PR #19, opened 2018-02-03
Test ethernomin

zyzek:


contracts/EtherNomin.sol
```diff
@@ -368,6 +354,28 @@ contract EtherNomin is ERC20FeeToken {
...
+    function canSelfDestruct()
+        public
+        view
+        returns (bool)
+    {
+        // Not being in liquidation implies the timestamp is uint max, so it would roll over.
+        // We need to check whether we're in liquidation first.
+        if (isLiquidating()) {
+            bool totalPeriodElapsed = liquidationTimestamp + liquidationPeriod < now;
+            bool allTokensReturned = (liquidationTimestamp + 1 weeks < now) && (nominPool == totalSupply);
```
DominicRomanowski: is the 1 weeks necessary?

contracts/EtherNomin.sol
```diff
@@ -456,19 +481,23 @@ contract EtherNomin is ERC20FeeToken {
...
+        uint proceeds;
         if (isLiquidating()) {
-            require(fiatBalanceAllowStale() >= proceeds);
+            proceeds = saleProceedsEtherAllowStale(n);
         } else {
-            require(fiatBalance() >= proceeds);
+            proceeds = saleProceedsEther(n);
         }
+
+        require(this.balance >= proceeds);
```
DominicRomanowski: couldn't this be: if (this.balance >= proceeds) { proceeds = this.balance }

contracts/EtherNomin.sol
zyzek: Not strictly necessary, but it prevents instant liquidation -> selfdestruction by someone.

contracts/EtherNomin.sol
zyzek: No, we don't allow anyone to be shortchanged. The appropriate response in this case is for the foundation to recollateralise the contract.

PR #37, opened 2018-02-22
include sent time variable for the setPrice funciton

mjspain12:


mjspain12:
Wrong target repository.

PR #43, opened 2018-02-26
Sigp report revisions

mjspain12:


contracts/EtherNomin.sol
```diff
@@ -657,7 +657,7 @@ contract EtherNomin is ERC20FeeToken {
 
     event LiquidationExtended(uint extension);
 
-    event PoolFeeRateUpdated(uint newFeeRate);
+    event poolFeeRateUpdated(uint newFeeRate);
```
DominicRomanowski: Please change this back to uppercase

contracts/HavvenEscrow.sol
```diff
@@ -283,6 +284,7 @@ contract HavvenEscrow is Owned, SafeDecimalMath {
 
         vestingSchedules[account].push([time, quantity]);
         totalVestedBalance = safeAdd(totalVestedBalance, quantity);
+        require(totalVestedBalance <= havven.balanceOf(this));
```
DominicRomanowski: move this to the top of the function (under the other two requires)

PR #67, opened 2018-05-28
Fix rounding error

blargity:
Made the changes to avoid rounding errors, so that users always get what they request exactly.

- Transfer of 5 = deduct exactly 5 from my account, then recipient gets 5 - fee
- Transfer (sender pays) = deduct 5 + fee from my account, then recipient gets exactly 5.

### DIFF
--- a/contracts/ExternStateFeeToken.sol
+++ b/contracts/ExternStateFeeToken.sol
@@ -237,8 +237,7 @@ contract ExternStateFeeToken is Proxyable, SafeDecimalMath {
     {
         require(to != address(0));
 
-        // The fee is deducted from the amount sent
-        uint fee = transferFeeIncurred(value);
+        uint fee = safeSub(value, priceToSpend(value));
         uint amountReceived = safeSub(value, fee);
 
         return _internalTransfer(sender, to, amountReceived, fee);
@@ -254,7 +253,7 @@ contract ExternStateFeeToken is Proxyable, SafeDecimalMath {
         require(to != address(0));
 
         // The fee is deducted from the amount sent
-        uint fee = transferFeeIncurred(value);
+        uint fee = safeSub(value, priceToSpend(value));
         uint amountReceived = safeSub(value, fee);
 
         // Reduce the allowance by the amount we're transferring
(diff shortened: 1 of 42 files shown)


contracts/ExternStateFeeToken.sol
```diff
@@ -209,30 +209,40 @@ contract ExternStateFeeToken is Proxyable, SafeDecimalMath {
...
+    /**
+     * @notice ERC20 friendly transfer function.
+     */
+    function _transfer_byProxy(address sender, address to, uint value)
+        internal
+        returns (bool)
+    {
+        require(to != address(0));
+
+        uint fee = safeSub(value, priceToSpend(value));
```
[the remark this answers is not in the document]
zyzek: Also, we have
```
fee = value - priceToSpend(value)
amountReceived = value - fee
```
Substituting the first into the second, I think we can instead do:
```
uint amountReceived = priceToSpend(value)
uint fee = safeSub(value, fee);
```
Which I find a bit clearer. It also saves us a single `safeSub`, although I suppose this is a bit of a micro-optimisation.
Unless there's a case where `priceToSpend(value) > value`, the two should be equivalent, but that looks mathematically impossible to me.

contracts/ExternStateFeeToken.sol
```diff
@@ -242,21 +252,48 @@ contract ExternStateFeeToken is Proxyable, SafeDecimalMath {
     {
         require(to != address(0));
 
-        // The fee is deducted from the sender's balance, in addition to
-        // the transferred quantity.
+        // The fee is deducted from the amount sent
+        uint fee = safeSub(value, priceToSpend(value));
```
zyzek: See above comments in `_transfer_byProxy`.

contracts/ExternStateFeeToken.sol
```diff
@@ -209,54 +210,85 @@ contract ExternStateFeeToken is Proxyable, SafeDecimalMath {
...
 
+    /**
+     * @notice ERC20 friendly transfer function.
+     */
+    function _transfer_byProxy(address sender, address to, uint value)
+        internal
+        returns (bool)
+    {
+        uint amountReceived = priceToSpend(value);
+        uint fee = safeSub(value, fee);
```
intmainvoid: Pretty sure this should be
```
uint fee = safeSub(value, amountReceived)
```

PR #68, opened 2018-05-30
Feature/issuance controller

blargity:
Adds:
 - Pausable: An interface other contracts can inherit from to receive a paused state, and the ability to be toggled by the owner. Includes a modifier to make it easy to check in the child contract.
 - Issuance Controller: Holds an amount of nUSD and HAV. Users are able to exchange ETH for nUSD and nUSD for HAV at fixed exchange rates which will be pushed into the contract by our existing oracle setup.


contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+date:       2018-05-20
+
+checked:    
+approved:   
+
+-----------------------------------------------------------------
+MODULE DESCRIPTION
+-----------------------------------------------------------------
+
+Issuance controler contract. The issuance controller provides
```
zyzek: `controler` -> `controller`

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+FILE INFORMATION
+-----------------------------------------------------------------
+
+file:       IssuanceController.sol
+version:    0.1
+author:     Kevin Brown
+
+date:       2018-05-20
+
+checked:
```
zyzek: I think we can removed this checked/approved stuff until we have an actual formal procedure for it.

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+checked:    
+approved:   
+
+-----------------------------------------------------------------
+MODULE DESCRIPTION
+-----------------------------------------------------------------
+
+Issuance controler contract. The issuance controller provides
+a way for users to acquire Nomins (Nomin.sol) by paying ETH
+and a way for users to acquire Havens (Havven.sol) by paying
```
zyzek: `Havens` -> `havvens`, `Nomins` -> `nomins`

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+contract IssuanceController is SafeDecimalMath, SelfDestructible, Pausable {
+
+    /* ========== STATE VARIABLES ========== */
+    Havven public havven;
+    Nomin public nomin;
+
+    // Address where the ether raised is transfered to
+    address public fundsWallet;
+
+    /* The address of the oracle which pushes the havven price to this contract */
```
zyzek: The oracle pushes both havven and ether prices.

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+    uint public priceStalePeriod = 3 hours;
+
+    /* The time the prices were last updated */
+    uint public lastPriceUpdateTime;
+    /* The USD price of havvens written in UNIT */
+    uint public usdToHavPrice;
+    /* The USD price of ETH written in UNIT */
+    uint public usdToEthPrice;
+    
+    uint public conversionFee;
```
zyzek: `conversionFee` seems to be unused. Do we not want a conversion fee?

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+        payable
+    {
+        exchangeForNomins();
+    } 
+
+    
+    /**
+     * @notice Exchange ETH to nUSD.
+     */
+    function exchangeForNomins()
```
zyzek: This may be worth renaming to `exchangeEtherForNomins` or something to make the trading pair clearer in the interface.

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+        priceStalePeriod = _time;
+    }
+
+    /* ========== MUTATIVE FUNCTIONS ========== */
+    /**
+     * @notice Access point for the oracle to update the prices of havvens / eth.
+     */
+    function updatePrices(uint newEthPrice, uint newHavvenPrice, uint timeSent)
+        external
+        onlyOracle  /* Should be callable only by the oracle. */
```
zyzek: Comment seems redundant.

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+    } 
+
+    
+    /**
+     * @notice Exchange ETH to nUSD.
+     */
+    function exchangeForNomins()
+        public 
+        payable
+        pricesNotStale // We can only do this when the prices haven't gone stale
```
zyzek: The comments on `pricesNotStale` and `notPaused` seem redundant.

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+        notPaused // And if the contract is paused we can't do this action either
+        returns (uint) // Returns the number of Nomins (nUSD) received
+    {
+        // How many Nomins are available for us to sell?
+        uint availableNomins = nomin.balanceOf(this);
+        uint requestedToPurchase = safeMul_dec(msg.value, usdToEthPrice);
```
zyzek: Do you think it's worth pointing out that this `safeMul_dec` works only because we chose 18 decimal places?

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+
+        return requestedToPurchase;
+    }
+
+    /**
+     * @notice Exchange nUSD for Havvens
+     */
+    function exchangeForHavvens(uint amount)
+        external
+        pricesNotStale // We can only do this when the prices haven't gone stale
```
zyzek: Same as above in re the modifier comments.

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+        // just watch ERC20 events from the Nomin contract filtered to our
+        // address.
+
+        return requestedToPurchase;
+    }
+
+    /**
+     * @notice Exchange nUSD for Havvens
+     */
+    function exchangeForHavvens(uint amount)
```
zyzek: Consider renaming this to something like `exchangeNominsForHavvens` or similar for clarity on the exchange pair being traded.

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+        pricesNotStale // We can only do this when the prices haven't gone stale
+        notPaused // And if the contract is paused we can't do this action either
+        returns (uint) // Returns the number of Havvens (HAV) received
+    {
+        // Does the sender have enough nUSD to request this exchange?
+        require(amount <= nomin.balanceOf(msg.sender));
+
+        // How many Havvens are they going to be receiving?
+        // Calculate the amount of Nomins we will receive after the transfer (minus fees)
+        uint amountReceived = safeDiv(nomin.amountReceived(amount), usdToHavPrice);
```
zyzek: Should this not be a safeDiv_dec?

contracts/Pausable.sol
```diff
@@ -0,0 +1,83 @@
...
+FILE INFORMATION
+-----------------------------------------------------------------
+
+file:       Pausable.sol
+version:    1.0
+author:     Kevin Brown
+
+date:       2018-05-22
+
+checked:
```
zyzek: No need for these checked and approved things.

contracts/Pausable.sol
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * @dev Constructor
+     * @param _owner The account which controls this contract.
+     */
+    constructor(address _owner)
+        Owned(_owner)
+        public
+    {
+        paused = false;
```
zyzek: These zero initialisations are unnecessary

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,306 @@
...
+    uint constant ORACLE_FUTURE_LIMIT = 10 minutes;
+
+    /* How long will the contract assume the price of any asset is correct */
+    uint public priceStalePeriod = 3 hours;
+
+    /* The time the prices were last updated */
+    uint public lastPriceUpdateTime;
+    /* The USD price of havvens written in UNIT */
+    uint public usdToHavPrice;
+    /* The USD price of ETH written in UNIT */
```
DominicRomanowski: Should be denominated, rather than written

DominicRomanowski:
I feel it would be worth adding a function to send nomins/havvens from the contract address itself, in case accidents occur...

contracts/IssuanceController.sol
blargity: Pretty sure we don't and this is reptilian DNA of when we thought there was an additional fee. I'll remove it.

contracts/IssuanceController.sol
blargity: Yup, adding.

tests/test_Pausable.py
```diff
@@ -0,0 +1,114 @@
...
+    def test_setPauseToggles(self):
+        initialPausedState = self.pausable.paused()
+        self.pausable.setPaused(self.contractOwner, not initialPausedState)
+        currentlyPausedState = self.pausable.paused()
+        self.assertEqual(not initialPausedState, currentlyPausedState)
+    
+    def test_setPauseIgnoresSame(self):
+        initialPausedState = self.pausable.paused()
+        initialLastPauseTime = self.pausable.lastPauseTime()
+        self.pausable.setPaused(self.contractOwner, initialPausedState)
```
zyzek: It would also be possible to check that the pause event was not emitted by checking if the logs list in the transaction is empty.

tests/test_Pausable.py
```diff
@@ -0,0 +1,114 @@
...
+    def test_setPauseUpdatesTime(self):
+        self.pausable.setPaused(self.contractOwner, True)
+        initialLastPauseTime = self.pausable.lastPauseTime()
+        sleep(1) # Wait one second so the lastPauseTime is different to now
+        self.pausable.setPaused(self.contractOwner, False)
+        afterFalseLastPausedTime = self.pausable.lastPauseTime()
+        self.assertEqual(initialLastPauseTime, afterFalseLastPausedTime)
+        self.pausable.setPaused(self.contractOwner, True)
+        finalLastPausedTime = self.pausable.lastPauseTime()
+        self.assertNotEqual(finalLastPausedTime, afterFalseLastPausedTime)
```
zyzek: Could check that the time was actually set to the correct value by also asserting that `finalLastPausedTime` is equal to `block_time(tx.blockNumber)`, where `tx` is the return value of `self.pausable.setPaused`, and `block_time` exists in one of the `utils` libraries.

tests/test_Pausable.py
```diff
@@ -0,0 +1,114 @@
...
+    def test_notPausedModifier(self):
+        originalValue = self.pausable.getSomeValue()
+        newValue = originalValue + 123
+        self.pausable.setSomeValue(self.contractOwner, newValue)
+        updatedValue = self.pausable.getSomeValue()
+        self.assertEqual(newValue, updatedValue)
+        self.pausable.setPaused(self.contractOwner, True)
+        self.assertReverts(self.pausable.setSomeValue, self.contractOwner, originalValue)
+        updatedValue = self.pausable.getSomeValue()
+        self.assertNotEqual(originalValue, updatedValue)
```
zyzek: Could also check that unpausing allows the function to work again afterwards.

contracts/IssuanceController.sol
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    /**
+     * @notice Set the stale period on the updated price variables
+     */
+    function setPriceStalePeriod(uint _time)
+        external
+        onlyOwner 
+    {
+        priceStalePeriod = _time;
```
zyzek: No `PriceStalePeriodUpdated` event?

zyzek:
I wonder if it's worth also adding a couple of getters for querying the havven and nomin balance of the contract to save people messing around with multiple ABIs in order to get this information.

tests/test_IssuanceController.py
```diff
@@ -0,0 +1,380 @@
...
+        cls.issuanceControllerEventDict = cls.event_maps['IssuanceController']
+
+    def test_constructor(self):
+        self.assertEqual(self.issuanceController.owner(), self.contractOwner)
+        self.assertEqual(self.issuanceController.fundsWallet(), self.fundsWallet)
+        self.assertEqual(self.issuanceController.oracle(), self.oracleAddress)
+        self.assertEqual(self.issuanceController.usdToEthPrice(), self.usdToEthPrice)
+        self.assertEqual(self.issuanceController.usdToHavPrice(), self.usdToHavPrice)
+        self.assertEqual(self.issuanceController.havven(), self.havven_contract.address)
+        self.assertEqual(self.issuanceController.nomin(), self.nomin_contract.address)
```
zyzek: Missing checks in `test_constructor` that the stale period and `lastUpdateTime` are initialised reasonably.

tests/test_IssuanceController.py
```diff
@@ -0,0 +1,432 @@
...
+    def test_updatePrices(self):
+        newEthPrice = self.usdToEthPrice + 100
+        newHavPrice = self.usdToHavPrice + 100
+        timeSent = block_time()
+        self.issuanceController.updatePrices(self.oracleAddress, newEthPrice, newHavPrice, timeSent)
+        self.assertEqual(self.issuanceController.usdToEthPrice(), newEthPrice)
+        self.assertEqual(self.issuanceController.usdToHavPrice(), newHavPrice)
+        self.assertEqual(self.issuanceController.lastPriceUpdateTime(), timeSent)
+
+    def test_updatePricesTooEarly(self):
```
zyzek: This test might be slightly more robust if it sent a price update before the reverted attempt, instead of relying on the test setup to do it.

tests/test_IssuanceController.py
```diff
@@ -0,0 +1,432 @@
...
+        timeSent = block_time() - 120
+        self.assertReverts(self.issuanceController.updatePrices, self.oracleAddress, self.usdToEthPrice, self.usdToHavPrice, timeSent)
+
+    def test_updatePricesTooLate(self):
+        ORACLE_FUTURE_LIMIT = 10 * 60
+        timeSent = block_time() + ORACLE_FUTURE_LIMIT + 60
```
zyzek: This should really be measured against `lastPriceUpdateTime`, rather than the block time. 
I might also be inclined to add a public getter for `ORACLE_FUTURE_LIMIT` to a `TestableIssuanceController`, and then use the timeSent as one second further than that, which eliminates a bit of the slack here.

tests/test_IssuanceController.py
```diff
@@ -0,0 +1,432 @@
...
+    def test_cannotTransferIfPriceStale(self):
+        amount = 10 * UNIT
+        nominsBalance = (amount * self.usdToEthPrice) // UNIT
+        base = self.nomin.amountReceived(nominsBalance)
+        self.issuanceController.setPriceStalePeriod(self.contractOwner, 1)
+        timeSent = block_time()
+        self.issuanceController.updatePrices(self.oracleAddress, self.usdToEthPrice, self.usdToHavPrice, timeSent)
+        exchanger = self.participantAddresses[0]
+        startingFundsWalletEthBalance = get_eth_balance(self.fundsWallet)
+        sleep(2) # Wait so the lastPriceUpdateTime is different to now
```
zyzek: Use the `fast_forward` function in utils instead of sleeping, which slows down the tests.

tests/test_IssuanceController.py
```diff
@@ -0,0 +1,432 @@
...
+        self.issuanceController.setPaused(self.contractOwner, True)
+        exchanger = self.participantAddresses[0]
+        startingFundsWalletEthBalance = get_eth_balance(self.fundsWallet)
+
+        # Set up the contract so it contains some nomins for folks to convert Ether for
+        self.nomin.giveNomins(MASTER, self.issuanceControllerContract.address, nominsBalance)
+        self.assertEqual(self.nomin.balanceOf(self.issuanceControllerContract.address), nominsBalance)
+
+        # Attmpt transfer
+        self.assertReverts(self.issuanceController.exchangeEtherForNomins, exchanger, amount)
```
zyzek: This should also test that the `exchangeNominsForHavvens` function doesn't operate when paused.

tests/test_IssuanceController.py
```diff
@@ -0,0 +1,432 @@
...
+        feesToPayInNomins = nominsToSend - baseToSend
+        exchanger = self.participantAddresses[0]
+        startingFundsWalletEthBalance = get_eth_balance(self.fundsWallet)
+
+        # Set up the contract so it contains some nomins for folks to convert Ether for
+        self.nomin.giveNomins(MASTER, self.issuanceControllerContract.address, nominsBalance)
+        self.assertEqual(self.nomin.balanceOf(self.issuanceControllerContract.address), nominsBalance)
+
+        # Transfer the amount to the receiver
+        txr = self.issuanceController.exchangeEtherForNomins(exchanger, ethToSend)
```
zyzek: Looks like the actual return value (the number of nomins received) of this function call isn't being checked. It should be equal to `baseToSend`.

tests/test_IssuanceController.py
```diff
@@ -0,0 +1,432 @@
...
+        # Ensure the result of the transfer is correct.
+        endingFundsWalletEthBalance = get_eth_balance(self.fundsWallet)
+        self.assertEqual(self.nomin.balanceOf(self.issuanceControllerContract.address), nominsBalance)
+        self.assertEqual(self.nomin.balanceOf(exchanger), 0)
+        self.assertEqual(self.nomin.feePool(), 0)
+        self.assertEqual(startingFundsWalletEthBalance, endingFundsWalletEthBalance)
+
+    # Withdraw havvens tests
+
+    def test_withdrawHavvens(self):
```
zyzek: Maybe worth testing that the transaction reverts if the user attempts to withdraw too many havvens (rather than just failing silently). Also partial withdrawals.

And the same thing for `test_withdrawNomins`.

tests/test_Pausable.py
blargity: Done!

PR #74, opened 2018-06-22
small changes to havven to add basic functions and cleanup

DominicRomanowski:
- added two arrays to constructor, to set issuers/nominsIssued for current issuers
- within transfers changed checking availableHavvens(sender) to new function transferableHavvens(sender)
- changed withdrawFee() to be external, from public
- changed name updateIssuanceData to computeIssuanceData
- changed remainingIssuableNomins to use safeSub
- added collateral function: havven.balanceOf(addr) + escrow.balanceOf(addr)
- renamed lockedHavvens to issuanceDraft, as it can go over the collateral of an address
- added lockedHavvens function, which caps the issuanceDraft up to the collateral of an address
- added transferableHavvens function which returns the number of havvens a user can transfer, considering the amount they have locked
- changed availableHavvens to use the collateral function rather than computing it
- fees now stored in a static address rather than in the underlying contract itself.
- constructor of Nomin now takes a total supply and a TokenState contract.

TODO: add a test for the constructor adding issuers

### DIFF
--- a/contracts/FeeToken.sol
+++ b/contracts/FeeToken.sol
@@ -63,10 +63,10 @@ contract FeeToken is ExternStateToken {
      * @param _feeAuthority The address which has the authority to withdraw fees from the accumulated pool.
      * @param _owner The owner of this contract.
      */
-    constructor(address _proxy, string _name, string _symbol, uint _totalSupply,
+    constructor(address _proxy, TokenState _tokenState, string _name, string _symbol, uint _totalSupply,
                 uint _transferFeeRate, address _feeAuthority, address _owner)
-        ExternStateToken(_proxy, _name, _symbol, _totalSupply,
-                         new TokenState(_owner, address(this)),
+        ExternStateToken(_proxy, _tokenState,
+                         _name, _symbol, _totalSupply,
                          _owner)
         public
     {
--- a/contracts/FeeToken.sol
+++ b/contracts/FeeToken.sol
@@ -49,6 +49,8 @@ contract FeeToken is ExternStateToken {
     uint constant MAX_TRANSFER_FEE_RATE = UNIT / 10;
     /* The address with the authority to distribute fees. */
     address public feeAuthority;
+    /* The address that fees will be pooled in. */
+    address constant FEE_ADDRESS = 0xfeefeefeefeefeefeefeefeefeefeefeefeefeef;
 
 
     /* ========== CONSTRUCTOR ========== */
@@ -158,7 +160,7 @@ contract FeeToken is ExternStateToken {
         view
         returns (uint)
     {
-        return tokenState.balanceOf(address(this));
+        return tokenState.balanceOf(FEE_ADDRESS);
     }
 
     /* ========== MUTATIVE FUNCTIONS ========== */
@@ -178,11 +180,11 @@ contract FeeToken is ExternStateToken {
         /* Insufficient balance will be handled by the safe subtraction. */
         tokenState.setBalanceOf(from, safeSub(tokenState.balanceOf(from), safeAdd(amount, fee)));
         tokenState.setBalanceOf(to, safeAdd(tokenState.balanceOf(to), amount));
-        tokenState.setBalanceOf(address(this), safeAdd(tokenState.balanceOf(address(this)), fee));
+        tokenState.setBalanceOf(FEE_ADDRESS, safeAdd(tokenState.balanceOf(FEE_ADDRESS), fee));
 
         /* Emit events for both the transfer itself and the fee. */
         emitTransfer(from, to, amount);
-        emitTransfer(from, address(this), fee);
+        emitTransfer(from, FEE_ADDRESS, fee);
 
         return true;
     }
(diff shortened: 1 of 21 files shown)


contracts/Havven.sol
```diff
@@ -609,41 +616,90 @@ contract Havven is ExternStateToken {
...
+
+    /**
+     * @notice The number of havvens that are free to be transferred by an account.
+     */
+    function transferableHavvens(address account)
+        public
+        view
+        returns (uint)
+    {
+        uint draft = issuanceDraft(account);
```
zyzek: These three cases should be commented to reflect their meaning.

contracts/Havven.sol
```diff
@@ -210,7 +210,7 @@ contract Havven is ExternStateToken {
      * If the provided address is 0x0, then a fresh one will be constructed with the contract owning all tokens.
      * @param _owner The owner of this contract.
      */
-    constructor(address _proxy, TokenState _tokenState, address _owner, address _oracle, uint _price)
+    constructor(address _proxy, TokenState _tokenState, address _owner, address _oracle, uint _price, address[] _issuers, uint[] _nominsIssued)
         ExternStateToken(_proxy, TOKEN_NAME, TOKEN_SYMBOL, HAVVEN_SUPPLY, _tokenState, _owner)
         /* Owned is initialised in ExternStateToken */
```
zyzek: Tbh I think this line is implicit from the @param. We can probably remove it.

contracts/Havven.sol
```diff
@@ -376,7 +383,7 @@ contract Havven is ExternStateToken {
          * their havvens are escrowed, however the transfer would then
```
zyzek: This comment belongs in the docstring of the `transferableHavvens` function.

contracts/Havven.sol
```diff
@@ -609,41 +616,90 @@ contract Havven is ExternStateToken {
...
+        uint bal = tokenState.balanceOf(account);
+        if (escrow != address(0)) {
+            bal = safeAdd(bal, escrow.balanceOf(account));
         }
+        return bal;
     }
 
     /**
-     * @notice Havvens that are locked, which can exceed the user's total balance + escrowed
+     * @notice Locked havvens based on nominsIssued, which can exceed the user's free and escrowed havvens.
```
zyzek: I think I'd rewrite this comment as something like "The collateral that would be locked by issuance, which can exceed the account's actual collateral."

contracts/Havven.sol
```diff
@@ -609,41 +616,90 @@ contract Havven is ExternStateToken {
...
+        uint debt = issuanceDraft(account);
+        uint collat = collateral(account);
+        if (debt > collat) {
+            return collat;
+        }
+        return debt;
     }
 
     /**
      * @notice Havvens that are not locked, available for issuance
```
zyzek: "Havvens that are not locked" -> "Collateral that is not locked". Also the sentence should end in a full stop.

I still think we should consider renaming this function to `unlockedCollateral` for the symmetry. Or maybe `lockedCollateral` should be `unavailableCollateral`?

contracts/Havven.sol
```diff
@@ -609,41 +616,90 @@ contract Havven is ExternStateToken {
...
+        uint issued = nominsIssued[account];
+        if (issued == 0) {
             return 0;
         }
-        return USDtoHAV(safeDiv_dec(nominsIssued[account], issuanceRatio));
+        return USDtoHAV(safeDiv_dec(issued, issuanceRatio));
+    }
+
+    /**
+     * @notice Locked Havvens that are capped at the user's free and escrowed havvens.
```
zyzek: I prefer instead something like "capped at the account's total collateral."

contracts/Havven.sol
```diff
@@ -609,41 +616,90 @@ contract Havven is ExternStateToken {
...
         view
         returns (uint)
     {
         uint locked = lockedHavvens(account);
-        uint bal = tokenState.balanceOf(account);
-        if (escrow != address(0)) {
-            bal += escrow.balanceOf(account);
-        }
-        if (locked > bal) {
+        uint bal = collateral(account);
```
zyzek: `bal` -> `collat`

contracts/FeeToken.sol
```diff
@@ -49,6 +49,8 @@ contract FeeToken is ExternStateToken {
     uint constant MAX_TRANSFER_FEE_RATE = UNIT / 10;
     /* The address with the authority to distribute fees. */
     address public feeAuthority;
+    /* The address that fees will be pooled in. */
+    address constant FEE_ADDRESS = 0xfeefeefeefeefeefeefeefeefeefeefeefeefeef;
```
paulhauner: I can see `Nomin.sol:163` still freezes by sending to `this` and the comments declare it is a "transfer ... to the fee pool". Shouldn't it send to `FEE_ADDRESS` now?

paulhauner: This `FEE_ADDRESS` concept seems to remove the ability to reclaim tokens sent to the `Nomin` contract address (it was previously "inherited" from the fee pool residing at its address). You could do it via arbitrary state modification still, just a thought.

zyzek: Sending to the Nomin contract address is disallowed inside the transfer functions, so that's still impossible. There's much less likelihood that people send nomins accidentally to a totally distinct address, which they would have to seek out manually.

Reclamation was only ever possible by withdrawing fees if you had issued some nomins, so I'm not sure that anything is different on that front.

PR #75, opened 2018-07-10
Added Nomin Airdropper contract

blargity:
Including unit tests.

### DIFF
--- a/contracts/NominAirdropper.sol
+++ b/contracts/NominAirdropper.sol
@@ -1,0 +1,80 @@
+/*
+-----------------------------------------------------------------
+FILE INFORMATION
+-----------------------------------------------------------------
+
+file:       NominAirdropper.sol
+version:    1.0
+author:     Kevin Brown
+
+date:       2018-07-09
+
+-----------------------------------------------------------------
+MODULE DESCRIPTION
+-----------------------------------------------------------------
+
+This contract was adapted for use by the Havven project for the
+airdropper contract that OmiseGO created here:
+https://github.com/omisego/airdrop/blob/master/contracts/Airdropper.sol
+
+It exists to save gas costs per transaction that'd otherwise be
+incurred running airdrops individually.
+
+Original license below.
+
+-----------------------------------------------------------------
+
+Copyright 2017 OmiseGO Pte Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+pragma solidity 0.4.24;
+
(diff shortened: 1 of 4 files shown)


contracts/NominAirdropper.sol
```diff
@@ -0,0 +1,80 @@
...
+
+        // Loop through each destination and perform the transfer.
+        uint256 i = 0;
+        
+        while (i < destinations.length) {
+            Nomin(tokenAddress).transferSenderPaysFee(destinations[i], values[i]);
+            i += 1;
+        }
+
+        return i;
```
intmainvoid: I think there should be a test case written for this - or this line discarded. AFAICT there is no way to return a value from a function that writes to the block chain

contracts/NominAirdropper.sol
```diff
@@ -0,0 +1,80 @@
...
+    {}
+
+    /**
+     * @notice Multisend airdrops tokens to an array of destinations.
+     * @dev The fee rate is in decimal format, with UNIT being the value of 100%.
+     */
+    function multisend(address tokenAddress, address[] destinations, uint256[] values)
+        external
+        onlyOwner
+        returns (uint256)
```
intmainvoid: I think there should be a test case written for this - or this line discarded. AFAICT there is no way to return a value from a function that writes to the block chain

tests/test_NominAirdropper.py
```diff
@@ -0,0 +1,99 @@
...
+    HavvenTestCase, ZERO_ADDRESS,
+)
+from tests.contract_interfaces.nomin_interface import PublicNominInterface
+from tests.contract_interfaces.havven_interface import HavvenInterface
+from tests.contract_interfaces.court_interface import FakeCourtInterface
+from tests.contract_interfaces.nomin_airdropper_interface import NominAirdropperInterface
+
+
+def setUpModule():
+    print("Testing Nomin...")
```
intmainvoid: Maybe this should be "`Testing NominAirdropper`"?

tests/test_NominAirdropper.py
```diff
@@ -0,0 +1,99 @@
...
+    
+    def test_assertIfUsedByNonOwner(self):
+        # Only owner should be able to use.
+        self.assertReverts(self.airdropper.multisend, DUMMY, self.nomin_contract.address, [DUMMY, DUMMY], [50 * UNIT, 100 * UNIT])
+
+    def test_correctlyAirdrops(self):
+        # Sending multiple transactions should result in sender paying fees, and correct amount being received.
+        self.airdropper.multisend(MASTER, self.nomin_contract.address, [DUMMY, DUMMY], [50 * UNIT, 100 * UNIT])
+
+        self.assertEqual(self.nomin.balanceOf(DUMMY), 150 * UNIT)
```
intmainvoid: Suggest adding test case where:
- an airdrop is conducted to addresses that are not the same

Although the following could be out of scope, it would probably be wise to test the following as well: 
- a large number of addresses (and values) are passed to the contract. The size of the array should attempt to be a little larger than that expected to be used in production (ie. > 50).
- The array of addresses contains a single bad address

blargity:
Didn't add the rest of the test cases requested given that:

- We're the only ones who are going to run this, and we can control the input precisely.
- The large number thing is going to be handled as much as it can be by solidity.
- Bad addresses will revert the transaction. This is expected, and we can control the input.

PR #77, opened 2018-07-23
Issuance updates

thekevinbrown:
Adds ETH -> HAV conversion in the issuance controller.

### DIFF
--- a/contracts/ExternStateToken.sol
+++ b/contracts/ExternStateToken.sol
@@ -30,12 +30,12 @@ import "contracts/SafeDecimalMath.sol";
 import "contracts/SelfDestructible.sol";
 import "contracts/TokenState.sol";
 import "contracts/Proxyable.sol";
-
+import "contracts/ReentrancyPreventer.sol";
 
 /**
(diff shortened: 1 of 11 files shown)


paulhauner:
Hi all,
  I'm leaving a few comments here as requested by Matthew M. from Havven. We performed a brief review of the changes introduced in this PR without automated testing.

We did not detect any security vulnerabilities during this review.

Nice work on implementing fixed-price functions, I think this was a good approach.

During the process we noticed the following things and raise them purely as informational:

1. The following functions are annotated as `public` when they could be `external` (this would save some gas and be more explicit):
   - exchangeEtherForNominsAtRate
   - exchangeEtherForHavvensAtRate
   - exchangeNominsForHavvensAtRate
   - nominsReceivedForEther
2. The fixed-rate "wrapper" functions implement the `pricesNotStale` and `notPaused` modifiers also implemented by the "wrapped" variable-rate exchange functions. This would result in slightly increased gas usage for trades which do not revert and slightly decreased gas usage for trades which do revert (they would revert early, spending less gas).

Happy to hear any comments or questions you might have :)

PR #78, opened 2018-08-02
Added descriptions to all of our require() statements

blargity:
Solidity now has a feature where we can supply an explanation for why we're reverting a transaction. This PR adds descriptions to all of our require statements across the code base.


blargity:
Note, gas increase in deploying Havven contract is 800,853 for the space to hold the strings. This brings our total gas required to deploy Havven to 5,248,205, which is well within the current block gas limit. The actual contract pushing the boundaries is PublicHavven, which is there just for testing, so while it's a fair bit more gas, it's only on deployment, and it will give us much more descriptive error messages at some point in the future, which is why I'm just allowing the testing framework to use a higher block gas limit instead of trying to cut down on the size of the strings.

PR #79, opened 2018-08-14
Issuance v3

thekevinbrown:
This version of the issuance controller with FIFO sales of user-sent nomins. 

Still requires extensive unit testing.


thekevinbrown:
I'm going to close this as it's stale now.

PR #81, opened 2018-10-30
Depot audit remediation

hav-noms:


contracts/Depot.sol
```diff
@@ -502,13 +526,17 @@ contract Depot is SafeDecimalMath, SelfDestructible, Pausable {
         external
         onlyNomin
     {
+        require(minimumDepositAmount <= amount, "Amount sent is less than the minimumDepositAmount");
```
thekevinbrown: This will result in deposits less than the minimum deposit amount getting ignored by the FIFO queue, but also not getting refunded to the user, as the result of the `tokenFallback` call is ignored by the token contracts. I would suggest instead of a `require` here we should transfer back the remainder at least.

contracts/Depot.sol
```diff
@@ -502,13 +526,17 @@ contract Depot is SafeDecimalMath, SelfDestructible, Pausable {
...
+
         // Ok, thanks for the deposit, let's queue it up.
         deposits[depositEndIndex] = nominDeposit({ user: from, amount: amount });
         // Walk our index forward as well.
         depositEndIndex = safeAdd(depositEndIndex, 1);
 
         // And add it to our total.
         totalSellableDeposits = safeAdd(totalSellableDeposits, amount);
+
+        emit NominDeposit(from, amount);
```
thekevinbrown: Do we need a NominDeposit event? What are we planning to do with these? If we do want to have them we should add them to the tests to insist that they happen.

test/Depot.js
```diff
@@ -198,6 +198,115 @@ contract('Depot', async function(accounts) {
...
+			// We need the owner to issue nomins
+			await havven.issueMaxNomins({ from: owner });
+			// Set up the depositor with an amount of nomins to deposit.
+			await nomin.transferSenderPaysFee(depositor, nominsBalance.toString(), { from: owner });
+		});
+
+		it('if the deposit nomin amount is a zero amount', async function() {
+			const nominsToDeposit = toUnit('0');;
+
+			await assert.revert(
```
thekevinbrown: I suspect this is reverting because of logic in the token contract, not in depot. We should see what the revert message is here, and I'd suggest trying to do something like a 100 wei deposit rather than a 0 deposit.

test/Depot.js
```diff
@@ -198,6 +198,115 @@ contract('Depot', async function(accounts) {
...
+
+		beforeEach(async function() {
+			// We need the owner to issue nomins
+			await havven.issueMaxNomins({ from: owner });
+			// Set up the depositor with an amount of nomins to deposit.
+			await nomin.transferSenderPaysFee(depositor, nominsBalance.toString(), { from: owner });
+		});
+
+		it('if the deposit nomin amount is a zero amount', async function() {
+			const nominsToDeposit = toUnit('0');;
```
thekevinbrown: Extra semicolon

thekevinbrown: (Side note, if you have installed the prettier plugin for VSCode it should fix these things up for you.)

test/Depot.js
```diff
@@ -198,6 +198,115 @@ contract('Depot', async function(accounts) {
...
+				)
+			);
+			const depositorBalance = await nomin.balanceOf(depositor);
+			assert.bnEqual(depositorBalance, nominsToDeposit);
+		});
+
+		it('if the deposit nomin of 10 amount is less than the minimumDepositAmount', async function() {
+			const nominsToDeposit = toUnit('10');
+
+			await assert.revert(
```
thekevinbrown: Again I think the tokenFallback reverting should not cause the parent call to revert. What's the revert message here?

test/Depot.js
```diff
@@ -198,6 +198,115 @@ contract('Depot', async function(accounts) {
...
+
+	describe('should accept nomin deposits', async function() {
+		const nominsBalance = toUnit('100');
+		const depositor = address1;
+
+		beforeEach(async function() {
+			// We need the owner to issue nomins
+			await havven.issueMaxNomins({ from: owner });
+			// Set up the depositor with an amount of nomins to deposit.
+			await nomin.transferSenderPaysFee(depositor, nominsBalance.toString(), { from: owner });
```
thekevinbrown: You shouldn't actually need the `.toString()` on `nominsBalance` here I don't think.

test/Depot.js
```diff
@@ -198,6 +198,115 @@ contract('Depot', async function(accounts) {
...
 
+	describe('should not accept nomin deposits and revert', async function() {
+		const nominsBalance = toUnit('100');
+		const depositor = address1;
+
+		beforeEach(async function() {
+			// We need the owner to issue nomins
+			await havven.issueMaxNomins({ from: owner });
+			// Set up the depositor with an amount of nomins to deposit.
+			await nomin.transferSenderPaysFee(depositor, nominsBalance.toString(), { from: owner });
```
thekevinbrown: You shouldn't actually need the `.toString()` on the `nominsBalance` here I don't think.

test/Depot.js
```diff
@@ -198,6 +198,115 @@ contract('Depot', async function(accounts) {
...
+				nominsToDeposit,
+				{
+					from: depositor,
+				}
+			)
+
+			const depotNominBalanceCurrent = await nomin.balanceOf(
+				depot.address
+			);
+			assert.bnEqual(depotNominBalanceCurrent, nominsToDeposit);
```
thekevinbrown: I think it'd be good to assert that the deposit is in the queue as well.

test/Depot.js
```diff
@@ -198,6 +198,115 @@ contract('Depot', async function(accounts) {
...
```
thekevinbrown: Same here. (Assert that it's in the queue)

contracts/Depot.sol
hav-noms: This will charge them nUSD tx fees on sending it back and forth. Is there no way to make this TX fail and revert?

contracts/Depot.sol
hav-noms: For the new Depot deposits and TX, screens in Mintr https://projects.invisionapp.com/share/TRN8FKUJEBU#/screens/329396605
Yes we'll add the tests.

test/Depot.js
hav-noms: All sorted

test/Depot.js
hav-noms: removed

contracts/Depot.sol
```diff
@@ -201,6 +203,20 @@ contract Depot is SafeDecimalMath, SelfDestructible, Pausable {
         emit PriceStalePeriodUpdated(priceStalePeriod);
     }
 
+    /**
+     * @notice Set the minimum deposit amount required to depoist nUSD into the FIFO queue
+     * @param _amount The new new minimum number of nUSD required to deposit
+     */
+    function setMinimumDepositAmount(uint _amount)
```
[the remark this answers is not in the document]
clementbalestrat: Done.

contracts/Depot.sol
clementbalestrat: @thekevinbrown we had to refactor `TokenFallbackCaller` so we use `preventReentrancy` only if it's a contract. Otherwise calling `nomin.transfer()` from `tokenFallback()` doesn't work.

test/Depot.js
hav-noms: Fixed

test/Depot.js
hav-noms: We fixed these tests to refund the depositor and are not testing a revert but the correct balances

contracts/Depot.sol
clementbalestrat: Done.

test/Depot.js
hav-noms: done

test/Depot.js
hav-noms: done

contracts/Depot.sol
```diff
@@ -201,6 +203,20 @@ contract Depot is SafeDecimalMath, SelfDestructible, Pausable {
...
+        external
+        onlyOwner
+    {
+        require(_amount > 0, "Minimum deposit amount must be greater than 0");
```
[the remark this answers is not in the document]
clementbalestrat: @thekevinbrown what about setting it to 0?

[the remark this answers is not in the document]
hav-noms: https://kovan.etherscan.io/tx/0x118ca794e488d76d30a0220ad21f94879422e2383ae4d28115030188f876bf45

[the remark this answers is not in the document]
hav-noms: removing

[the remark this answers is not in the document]
thekevinbrown: Yup, in that case it'll fail, you're correct.

But I still don't understand why we'd want to insist on 1 wei as a minimum. Why couldn't you set it to 0?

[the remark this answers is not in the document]
clementbalestrat: Yes good point, this is why we decided to remove it with Clint.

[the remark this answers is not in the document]
thekevinbrown: Or to put it another way, why is 1 wei a good minimum?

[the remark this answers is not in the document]
hav-noms: sorry not 1 wei, 1 nUSD

contracts/Depot.sol
```diff
@@ -404,7 +430,7 @@ contract Depot is SafeDecimalMath, SelfDestructible, Pausable {
         // And send them the Havvens.
         havven.transfer(msg.sender, havvensToSend);
 
-        emit Exchange("nUSD", nominAmount, "HAV", havvensToSend);
+        emit Exchange("nUSD", nominAmount, "HAV", havvensToSend, 1);
```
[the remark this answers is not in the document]
hav-noms: fixed

contracts/Depot.sol
```diff
@@ -269,8 +293,15 @@ contract Depot is SafeDecimalMath, SelfDestructible, Pausable {
                     deposit.amount = safeSub(deposit.amount, remainingToFulfill);
                     totalSellableDeposits = safeSub(totalSellableDeposits, remainingToFulfill);
 
-                    // Transfer the ETH to the depositor.
-                    deposit.user.transfer(safeDiv_dec(remainingToFulfill, usdToEthPrice));
+                    // Transfer the ETH to the depositor. Send is used instead of transfer
+                    // so a non payable contract wont block the FIFO queue on a failed
```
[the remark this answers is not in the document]
hav-noms: fixed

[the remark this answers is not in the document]
hav-noms: require removed

contracts/Depot.sol
```diff
@@ -501,14 +546,27 @@ contract Depot is SafeDecimalMath, SelfDestructible, Pausable {
...
-        depositEndIndex = safeAdd(depositEndIndex, 1);
-
-        // And add it to our total.
-        totalSellableDeposits = safeAdd(totalSellableDeposits, amount);
+        // A minimum deposit amount is designed to protect purchasers from over paying
+        // gas for fullfilling multiple small nomin deposits
+        if (amount < minimumDepositAmount) {
+            // We cant fail/revert the transaction or send the nomins back in a reentrant call.
+            // So we will keep your nomins balance seperate from the FIFO queue so you can withdraw them
+            smallDeposits[from] += safeAdd(smallDeposits[from], amount);
```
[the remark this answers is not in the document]
hav-noms: fixed

PR #113, opened 2019-02-20
Synthetix exchange event

jacko125:


utils/shouldFail.js
```diff
@@ -0,0 +1,30 @@
+const should = require('chai').should();
```
justinjmoses: Qu: Where is this now used? I can't see references to it anymore?

PR #118, opened 2019-03-06
HBN-762 A better deploy script

justinjmoses:
* The script has been moved into the folder `publish`, with instructions here: https://github.com/Synthetixio/synthetix/tree/HBN-762/publish
* It's split into three parts `build`, `deploy` and `verify`. Each step is run separately, but requires the previous one have been run.
* The most important file to consider is [config.json](https://github.com/Synthetixio/synthetix/blob/HBN-762/publish/deployed/rinkeby/config.json) - this needs to be changed before running to reflect which contracts to deploy. It's found in the `deploymentPath` input you pass the function. 
* It also does away with `.` namespacing and just keeps contract names their full names (e.g. `ProxysUSD`) and keeps track of the name of the underlying contract (e.g. `Proxy`). 
* Finally, the `deploy` step has been rewritten to allow for resuming from within broken deployments - it does this by mutating the [config.json](https://github.com/Synthetixio/synthetix/blob/HBN-762/publish/deployed/rinkeby/config.json) input as it deploys each contract. 

Here's me re-running all steps on `RINKEBY` - notice it skips redeployment as it isn't necessary.

here-goes


publish/index.js
```diff
@@ -0,0 +1,736 @@
...
+				// TODO - add these to the JSON file for the deployment
+				const deployedAt = new Date(result.data.result[0].timeStamp * 1000);
+				const deployedTxn = `https://${network}.etherscan.io/tx/${result.data.result[0].hash}`;
+
+				console.log(gray(` - Deployed at ${deployedAt}, see ${deployedTxn}`));
+
+				// Grab the last 50 characters of the compiled bytecode
+				const compiledBytecode = compiled[name].evm.bytecode.object.slice(-50);
+
+				const pattern = new RegExp(`${compiledBytecode}(.*)$`);
```
justinjmoses: Note: the existing code used `(.+)$` but this failed for `SafeDecimalMath` that didn't have anything after the `compiledBytecode` before the end of input and would fail. Hence I changed the `+` to `*`.

publish/index.js
```diff
@@ -0,0 +1,736 @@
...
+						.setTarget(feePool.options.address)
+						.send(deployer.sendParameters());
+				}
+			}
+
+			const synthetixState = await deployContract({
+				name: 'SynthetixState',
+				args: [account, account],
+			});
+			const proxySynthetix = await deployContract({ name: 'ProxySynthetix', args: [account] });
```
justinjmoses: ⚠️❓Should we ever be redeploying this on `mainnet`? Shouldn't this always be marked to never deploy on mainnet? Is that the same for all of the existing proxy contracts?

hav-noms: No. We should be just adding the address and not configure as deploy

hav-noms:
Can you add a flag or another command to transform and output to this json to push into the website
https://github.com/Synthetixio/synthetix-website-static-cms/blob/master/content/pages/tokens.json

justinjmoses:
@hav-noms I added a step `generate-token-list`. I didn't have it output the `pageTitle` or `pageDescription`, just the list of tokens. I decided that hard-coding the title or description into this repo seemed like a bad idea, so there would be a tiny manual step to take the token JSON output and insert it into the existing `tokens.json` file before uploading. 

See https://github.com/Synthetixio/synthetix/tree/HBN-762/publish#generate-token-file

PR #120, opened 2019-03-13
Removing tests from language detection

justinjmoses:
This should remove tests from showing up as language detection, and highlight Solidity as the majority language not JavaScript.

### DIFF
--- a/publish/index.js
+++ b/publish/index.js
@@ -245,8 +245,13 @@ program
 				}
 				const { address } = deployedContract.options;
 
-				// in case we've already been verified, keep info from then
-				const { timestamp, txn } = deployment[name] || {};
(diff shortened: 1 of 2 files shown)


publish/index.js
```diff
@@ -245,8 +245,13 @@ program
 				}
 				const { address } = deployedContract.options;
 
-				// in case we've already been verified, keep info from then
-				const { timestamp, txn } = deployment[name] || {};
+				let timestamp = new Date();
```
justinjmoses: This was a fix I added to ensure that when doing a deploy alongside exisiting contracts, it doesn't remove their `timestamp` and `txn` entries from the `deployment.json` file

PR #125, opened 2019-04-15
HBN-722 Upgrading to the latest truffle & web3

justinjmoses:
I tried this branch again and now with truffle `5.0.12` and web3 `1.0.0-beta.52`, the issues with the tests were resolved (issues with BN translation in an earlier version of web3).

The major change is that we must invoke any overloaded methods using `.methods['name(param1,...)']` - this includes all `transfer`, `transferFrom`, `transferSenderPaysFee` and `transferFromSenderPaysFee` methods, to denote which ones include the additional `bytes data` param.


jacko125:
Justin, I just ran the test suite locally and it looks fine. 

What was the fix to update to the new version of truffle / web3 ?

package.json
```diff
@@ -53,23 +54,22 @@
 		"eslint-plugin-prettier": "^2.6.2",
 		"eslint-plugin-promise": "^4.0.1",
 		"eslint-plugin-standard": "^4.0.0",
-		"ganache-cli": "^6.1.8",
+		"ganache-cli": "^6.3.0",
```
jacko125: "ganache-cli": "^6.4.1", using this in inflationarySupply branch already

PR #127, opened 2019-04-18
HBN-834 Adding the inverted synth functionality, along with new synths

justinjmoses:
* New crypto synths `sETH`, `sBNB`
* New inverted crypto synths `iBTC`, `iETH`, `iBNB`
* Added more content in `synths.json` detailing underlying asset for all synths plus the inverted amounts 
* Changes to `ExchangeRates.sol` to do the inverted math, and to track the mapping of inverted prices. I opted to keep the logic in here rather than detached state as it's only tracking the handful of inverted synths we added (called invertedPrices) and there's an array of these now as well so we can iterate over the list if we ever need to.
* Minor change to `Synthetix.sol` to check `rateNotFrozen` of `destinationCurrency` before allowing an exchange
* Linting improvements and fixes.
* Adding `eth-gas-reporter` to test output
* Cleaning up dependencies
* Incorporating fixes to the publish script from the 757 branch - no longer checking `mainnet` but rather checking if owner privilege before attempting action


publish/deployed/kovan/config.json
```diff
@@ -6,13 +6,13 @@
...
 	"ExchangeRates": {
-		"deploy": false
+		"deploy": true
 	},
 	"FeePool": {
 		"deploy": false
 	},
 	"Synthetix": {
-		"deploy": false
+		"deploy": true
```
hav-noms: @justinjmoses @kaiynne just a reminder a redeploy of Synthetix requires all existing 19 synths to be added again. I'll dig up the old run sheet to reuse with these new synths.

publish/index.js
```diff
@@ -328,16 +329,16 @@ program
 					account,
 					account,
 					account,
-					web3.utils.toWei('0.0015', 'ether'),
-					web3.utils.toWei('0.0015', 'ether'),
+					web3.utils.toWei('0.0015'),
```
hav-noms: We gotta make sure we set this to '0.0', '0.003' for MAINNET release of Staking Rewards. We need to redo all tests to support no transfer fees and higher exchange fees still

justinjmoses: I'll change it now - great catch!

publish/deployed/kovan/config.json
justinjmoses: via calling `addSynth` right? We actually do this in the publish script already - https://github.com/Synthetixio/synthetix/pull/127/files/069ae78b593319e07ee34a151d7b22c4859d9c03#diff-11230b41d0a9d05b6d97799af010d7d4R551.

publish/index.js
justinjmoses: Your comment here raises the general point that whenever we change things in `mainnet` via owner calls  - like fees or inverted sythns - we need to update the publish script to track these.

PR #141, opened 2019-05-09
HBN-863 Remove sRUB, sNZD, sPLN, sCAD, sSGD, sBRL

hav-noms:
Owner can call Synthetix.removeSynth() on most but not sRUB, sPLN because of dust. 

@justinjmoses is this all required on the scripts to remove a synth?


justinjmoses:
No the script doesn't have the idea of removing anything. 

We'd either a) have to add a new flag called `remove` for `config.json` on each contract to remove or b) a new part of the script to do removals, and make it only for synths, and then remove them there. Since it's a very specific action - removing synths - I'm inclined to do it in b).

justinjmoses:
I updated this PR with a few things:

1. Broke out the publish script into modules
1. Added the `remove-synths` command to the publish script
1. Reverted the `config.json` and `synths.json` files back to their original state in Kovan

So you run this to remove them in KOVAN: `node publish remove-synths -n kovan -d publish/deployed/kovan -g 5 -s sRUB, sNZD, sCAD, sSGD, sBRL`

justinjmoses:
I updated this with the fix to actually work 😳 and ran it on `sCAD` on Kovan and committed the auto changes to that environment's JSON file

PR #147, opened 2019-05-21
Feepool upgrade

hav-noms:


justinjmoses:
Ah why did you delete the lock file? You can use `npm run clean-install` to refresh everything?

PR #172, opened 2019-06-19
Sip reduce fee periods

jacko125:
WIP branch for reduce fee periods


jacko125:
WIP branch for reduce fee periods

PR #184, opened 2019-08-02
set exchangeRate to run at 500 optimizer runs

jacko125:


publish/src/solidity.js
```diff
@@ -61,8 +62,8 @@ module.exports = {
 			)
 		);
 
-		const warnings = output.errors.filter(e => e.severity === 'warning');
```
jacko125: output.errors could actually be `undefined` and throw error here if no check

PR #189, opened 2019-08-08
allow building exchangeRates contract at higher runs

jacko125:


publish/src/solidity.js
```diff
@@ -49,6 +49,7 @@ module.exports = {
 					settings: {
 						optimizer: {
 							enabled: true,
+							runs: runs,
```
justinjmoses: [np] this can just be `runs` using ES6 syntax

PR #220, opened 2019-09-04
Hbn 988 check proxy target for only synthetix fee pool modifier

jacko125:


publish/src/commands/deploy.js
```diff
@@ -765,42 +765,163 @@ const deploy = async ({
...
+		let proxyForSynth, proxyERC20ForSynth;
+		if (currencyKey === 'sUSD' || currencyKey === 'sETH') {
+			proxyForSynth = await deployContract({
+				name: `Proxy${currencyKey}`,
+				source: 'Proxy',
+				args: [account],
+				force: addNewSynths,
+			});
+
+			proxyERC20ForSynth = await deployContract({
```
justinjmoses: Qu: don't you think we should call them all `ProxyERC20${currencyKey}` (the new ones) and have the old ones use `Proxy${ currencyKey}` ? That way we'll keep the pattern where the prefix of the contract name is the source.

publish/src/commands/deploy.js
```diff
@@ -765,42 +765,163 @@ const deploy = async ({
...
-				`Synth ${currencyKey}`,
-				currencyKey,
-				account,
-				w3utils.asciiToHex(currencyKey),
-			].concat(additionalConstructorArgsMap[subclass] || []),
-			force: addNewSynths,
-		});
+
+		let synth;
+		if (currencyKey === 'sETH' && network !== 'local') {
```
justinjmoses: There's quite a bit duped here, I think it can be better handled with one statement that determines whether or not to use the `synthetix` or `synthetixProxy` and the `feePool` or `proxyFeePool`

publish/src/commands/deploy.js
```diff
@@ -843,101 +966,202 @@ const deploy = async ({
...
+					);
+					await synth.methods
+						.setIntegrationProxy(proxyERC20ForSynth.options.address)
+						.send(deployer.sendParameters());
 				} else {
 					appendOwnerAction({
-						key: `Synthetix.addSynth(Synth${currencyKey})`,
-						target: synthetixAddress,
-						action: `addSynth(${synthAddress})`,
+						key: `Synth${currencyKey}.setIntegrationProxy(proxyERC20${currencyKey})`,
```
justinjmoses: [np] `Proxy` should be capitalized

publish/src/commands/deploy.js
```diff
@@ -843,101 +966,202 @@ const deploy = async ({
...
+								key: `Synth${currencyKey}.setFeePool(FeePool)`,
+								target: synthAddress,
+								action: `setFeePool(${feePoolAddress})`,
+							});
+						}
+					}
+				}
+			}
+		} else {
+			//  New Synth.sol with currencyKey - bytes32, with feePoolProxy and synthetixProxy
```
justinjmoses: Again there's a lot duplicated here - I think you can de-dupe without too much effort. Keep in mind as well that we want to cauterize the lines with the exceptional code that we'll remove once we migrate to the new sETH and sUSD proxies.

publish/src/commands/deploy.js
hav-noms: Thats true, I did say to keep it Proxy because ERC20 everywhere is so nasty to read. Really this one is the new proxy so maybe we just call it Proxy.sol?

justinjmoses: Hmmm, yeah we need to start cleaning this stuff up - it's becoming quite difficult to manage the different proxies. We need to keep Proxy.sol for proxies to non-ERC20 stuff, like FeePool. 

How about we rename `ProxyERC20.sol` to `ProxyToken.sol`, then we can have:
- `ProxyTokenSynthetix`
- `ProxyTokensUSD`
...
- `ProxyFeePool` 
... 

The only complication is that `Proxyable.sol` has both `proxy` and `integrationProxy`, where the latter is for our `ProxyToken` implementation, which is non-obvious.

publish/src/commands/deploy.js
jacko125: @justinjmoses We might have a situation where I have to keep the transferFee logic in Synthetix and FeePool as sETH synth contract relies on these functions.  So stubbing synthInitiatedFeePayment

```    /**
     * @notice Function that allows synth contract to delegate sending fee to the fee Pool.
     * @dev Only the synth contract can call this function.
     * @param from The address fee is coming from.
     * @param sourceCurrencyKey source currency fee from.
     * @param sourceAmount The amount, specified in UNIT of source currency.
     * @return Boolean that indicates whether the transfer succeeded or failed.
     */
    function synthInitiatedFeePayment(
        address from,
        bytes4 sourceCurrencyKey,
        uint sourceAmount
    )
        external
        returns (bool)
    {
        return true;
    }```

PR #251, opened 2019-09-23
sip-18

hav-noms:
https://sips.synthetix.io/sips/sip-18


publish/src/commands/deploy.js
```diff
@@ -858,6 +863,16 @@ const deploy = async ({
 		});
 	}
 
+	// ensure Depot has sUSD synth address setup correctly
+	await runStep({
```
justinjmoses: 👍

PR #258, opened 2019-09-25
update package-lock.json and ganache cli set to 6.4.1, bump cache on circleci

jacko125:


jacko125:
This should fix up issue we discussed during standup.

PR #262, opened 2019-09-26
Arb rewarder

hav-noms:


publish/deployed/kovan/config.json
```diff
@@ -1,4 +1,7 @@
 {
+	"ArbRewarder": {
+		"deploy": true
```
justinjmoses: @jacko125 can you also add in the kovan address in the `deployment.json`?

PR #271, opened 2019-09-27
merge develop arbrewarder -> alpha

jacko125:


publish/deployed/kovan/config.json
```diff
@@ -1,4 +1,7 @@
 {
+	"ArbRewarder": {
+		"deploy": true
```
justinjmoses: why `true`? isn't it already deployed? Can you change this to `false` and add the relevant info to the `deployment.json` file for `kovan`?

jacko125: Fixed this to be false, I don't think we will deploy the arbRewarder contract to kovan testnet as the Uniswap testnet is not on kovan.

publish/src/commands/deploy.js
```diff
@@ -948,6 +948,33 @@ const deploy = async ({
...
+		contract: 'ArbRewarder',
+		target: arbRewarder,
+		read: 'synthetix_rates',
+		expected: input => input === exchangeRates.options.address,
+		write: 'setExchangeRates',
+		writeArg: exchangeRates.options.address,
+	});
+
+	// Ensure synthetix ProxyERC20 on arbRewarder set
+	await runStep({
```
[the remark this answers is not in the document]
jacko125: @hav-noms  Yep, added. Should be good to only check on mainnet with expected addresses that we put in (These hopefully won't change often but in PR).

publish/src/commands/deploy.js
```diff
@@ -117,6 +117,16 @@ const deploy = async ({
...
 
+	// load accounts used by local ganache in keys.json
+	const users = Object.entries(
+		JSON.parse(fs.readFileSync(path.join(__dirname, '..', '..', '..', 'keys.json'))).private_keys
+	).map(([pub, pri]) => ({
+		public: pub,
+		private: `0x${pri}`,
+	}));
+
+	privateKey = users[0].private;
```
justinjmoses: This file shouldn't need to know anything about the local ganache keyfile. The way I worked around this was having the privateKey passed in as an argument above (that's why it only invokes `privateKey = envPrivateKey` when `!local`). https://github.com/Synthetixio/synthetix/blob/master/int-test/index.js#L58

justinjmoses: Long story short, you should remove these lines above you added

PR #276, opened 2019-10-01
Hbn 1025 chainlink max gas price

jacko125:
This is a stub to show how Synthetix.exchange() can use a gasPrice oracle from any oracle source (Tests and deploy script isn't update yet).


justinjmoses:
This is no longer required.

PR #291, opened 2019-10-21
Staking pool - First Iteration

### LINKED ISSUE #192: SNX Staking Pool
Smart Contract Bounty. 
- Send your SNX to delegate all management. Mint, burn, collect fees/rewards
- Pool Admin sets % of all fees / rewards as management fee
- Internal mapping to assign the fees / rewards based on SNX users contributed to be withdrawn
- Open source so anyone can deploy and manage a pool and be a pool admin and market their staking pools

JGcarv:
Hello!

This is a PR that resolves #192 .

The Staking Pool contract is rather simple. Users can deposit SNX and earn liquidity tokens, which is a regular ERC20 that can be transferred, and manager can execute a few functions, like issue and exchange synths, claim fees, and vest rewards. 

Some particulars details:
* I'm making use of proxies to make deployment cheaper, but decided not to use the provided `Proxy` and `Proxyable` implementations because a lt of the functionality they provide wouldn't be that useful and would increase costs.
 
* The manager fees are only being charged on claimedFees, because that's the current way to add value to the pool. It's easy to implement a joining fee you think it's adequate.

* Since fees from the feePool can't be claimed on SNX there's two different approaches for dealing with this: 
1. Either using the depot to trade Synths for SNX 
This approach allows for a simpler code as well simpler withdrawing mechanics, but amount earned from fees will be smaller due to the exchange rate from depot.
2. keep a sUSD balance that users will receive together with their SNX when withdrawing. 
In this scenario the code is a bit more comp[lex, but the pool does not need to deal with the exchange costs.

In this PR I'm using the depot, because it's a bit cleaner, but a implementation using the second alternative can be found here: https://gist.github.com/JGcarv/84056424d7a5052e0aabe33591ebb7ea  

Let me know which approach do you think it's best. Maybe we could offer the two options for the factory and let managers decide.

There's a few thing that I'll continue working on:
1. Improve the withdrawing liquidation steps
2. Add natspec
3. Increase code coverage

Also, branched from a very old commit, so let me know if you want me to rebase this branch.


k06a:
Why not to subclass from ERC20 and to use proper internal API?

JGcarv:
I guess I didn't noticed that the system already had OpenZepplin as a dependency, so I kind of didn't use it. I'll update it today

JGcarv:
Hello @k06a .

I've tried making the recommended changes, but I realized I would have problems with the proxy.

Since I'm using a contract for storage and a contract for implementation, I would have to import the ERC20 on the `StakingPoolStorage`, which means I would be also import it on the `StakingPoolProxy`.

That would double the size of `stakingPoolProxy`, and it would increase the gas costs of deploying it and that seems like a trade off that's not worth it.

What do you think?

justinjmoses:
Hi @JGcarv would you mind changing the base to `community/develop` and merging the latest with that?

justinjmoses:
@JGcarv 
> I'm making use of proxies to make deployment cheaper, but decided not to use the provided Proxy and Proxyable implementations because a lt of the functionality they provide wouldn't be that useful and would increase costs.

Proxies are primarily for upgrading contracts without having users to update their addresses. Is that your intent here? 

> I've tried making the recommended changes, but I realized I would have problems with the proxy.

> Since I'm using a contract for storage and a contract for implementation, I would have to import the ERC20 on the StakingPoolStorage, which means I would be also import it on the StakingPoolProxy.

The proxy doesn't need any special methods - ours don't - users invoke the Proxy using the ABI for the underlying and it works as expected. E.g. for our `ERC20` proxy of `sETH` - https://contracts.synthetix.io/ProxyERC20#readContract - you'll see the ABI methods available are just for the proxy - yet if you invoke that contract address with the [Synth ABI](https://raw.githubusercontent.com/Synthetixio/synthetix-js/master/lib/abis/mainnet/ProxyERC20.json)

You could use our `ProxyERC20` -  https://github.com/Synthetixio/synthetix/blob/master/contracts/ProxyERC20.sol

JGcarv:
> Proxies are primarily for upgrading contracts without having users to update their addresses. Is that your intent here?

Not really, I'm using proxies for much cheaper deployments on chain. Since we have a factory, and users should be able to deploy a staking pool from it, it would be quite expensive to handle the whole pool bytecode. So the solution is to use a smaller contract, which contain only the storage layout and a fallback function to delegate all calls to a target reference implementation.

As a result from this, both the `proxy` and the `target` contracts must have the same storage layout, otherwise we could read/write from different slots. That's the purpose of the stakingPoolStorage contract, which is the first inheritance both on the proxy and the target. Since the implementation of `ERC20.sol` defines storage variables(`balances`, `_totalSupply`, etc), I need to import this file on `stakingPoolStorage`, which would increase the size/cost for the proxy.

> You could use our ProxyERC20

That's a good point. While using this doesn't solve the size issue, in fact it makes it worse, it does make me wonder if you think it's necessary for the proxy contracts to have the ERC20 methods when read by tools like Etherscan. How will users will interact with this contract? Will there be a front-end of sorts, or they'll have to rely of metamask/etherscan/etc?
    
> would you mind changing the base to community/develop and merging the latest with that?

No problem, I'll do that

.vscode/settings.json
```diff
@@ -1,11 +0,0 @@
-{
```
justinjmoses: would you mind not deleting this and the other vscode config?

justinjmoses:
> That's a good point. While using this doesn't solve the size issue, in fact it makes it worse, it does make me wonder if you think it's necessary for the proxy contracts to have the ERC20 methods when read by tools like Etherscan. 

Actually `ProxyERC20` was developed for integrations - Uniswap and Kyber specifically did not play nicely with our `Proxy.sol` - they expected top level functions to exist on the Proxy which didn't (they were passed through) and hence we needed `ProxyERC20`. cc @hav-noms for more detail here.  

> How will users will interact with this contract? Will there be a front-end of sorts, or they'll have to rely of metamask/etherscan/etc?  

@hav-noms - what's the plan for the staking pool front-end?

.vscode/settings.json
JGcarv: Sure, I have know idea what happened to be there

PR #297, opened 2019-10-29
SIP 21: Double exchange fee rate on swing trades

hav-noms:
Just a WIP for review


package.json
```diff
@@ -7,7 +7,7 @@
 	"scripts": {
 		"clean-install": "rm -rf build && rm ./package-lock.json && rm -rf node_modules/* && npm install",
 		"coverage": "solidity-coverage",
-		"ganache": "ganache-cli -e 1000000000000 -l 0x7a1200 -t 2018-03-13T00:00:00",
```
hav-noms: was 8M now 9.5M

contracts/Synthetix.sol
```diff
@@ -183,33 +183,33 @@ contract Synthetix is ExternStateToken {
     }
     // ========== SETTERS ========== */
 
-    function setFeePool(IFeePool _feePool)
```
hav-noms: TEMP: Out of gas on deployment. Need to increase ganache to 9.5M

contracts/Synthetix.sol
```diff
@@ -542,7 +468,16 @@ contract Synthetix is ExternStateToken {
         uint fee = 0;
 
         if (chargeFee) {
-            amountReceived = feePool.amountReceivedFromExchange(destinationAmount);
+            // Get the exchange fee rate
+            uint exchangeFeeRate = feePool.exchangeFeeRate();
+            // Is this a swing trade? long to short or sort to long.
```
jacko125: @hav-noms hey Clinton, I was thinking, in the scenario where a trader makes an exchange from `iTRX` to `sUSD` to close out the position and then -> `sUSD` into `iBTC` to go from one inverse to another, they would have the exchange fee doubled on both moves, making inverse synths cost double the fee to always get into and out. 

I thought we discussed if the sourceKey or destKey was `sUSD`, we would keep exchange fee same as it would require the trader to move first into `sUSD` and then to another inverse synth which will cost double the fee to execute such a move anyway.

justinjmoses: Makes sense to me. We'll need those test cases as well.

PR #302, opened 2019-11-05
SIP25 - decouple gasPriceLimit oracle and remove exchange.updateLock(…

### LINKED ISSUE #301: SIP 25 - decouple max gas price limit oracle
SIP 25 - decouple max gas price limit oracle. 

https://sips.synthetix.io/sips/sip-25

jacko125:
closes #301

### DIFF
--- a/publish/src/commands/deploy.js
+++ b/publish/src/commands/deploy.js
@@ -58,6 +58,7 @@ const deploy = async ({
 	buildPath = DEFAULTS.buildPath,
 	deploymentPath,
 	oracleExrates,
+	oracleGasLimit,
 	oracleDepot,
 	privateKey,
 	yes,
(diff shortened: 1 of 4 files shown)


justinjmoses:
Can you also add setting it to the deploy script? We should set it initially and then in successive deploys use the existing version (as we do with the oracle address)

jacko125:
> Can you also add setting it to the deploy script? We should set it initially and then in successive deploys use the existing version (as we do with the oracle address)

@justinjmoses  Got it, currently breaking int-tests once PR was raised, I will need to be set in the deployment script so that its ready.

publish/src/commands/deploy.js
```diff
@@ -536,13 +543,23 @@ const deploy = async ({
...
 
+	// setup gasLimitOracle on Synthetix
 	// setup exchange gasPriceLimit on Synthetix
 	const gasPriceLimit = w3utils.toWei('35', 'gwei');
 	if (network === 'local') {
 		await runStep({
 			contract: 'Synthetix',
 			target: synthetix,
-			account: oracleExrates,
+			read: 'gasLimitOracle',
```
justinjmoses: @jacko125 this is only for local tests - we still need a check to make sure `gasLimitOracle` is set in all other networks (so if we deploy a new Synthetix in the future, this field gets set)

PR #313, opened 2019-11-07
Adding in view to read fee rate for exchange

justinjmoses:
This adds the following `view` to `Synthetix`: `feeRateForExchange(bytes32 source, bytes32 dest)`. This allows dApps and, more importantly, other contracts to determine what the fee would be a given `exchange`.

It adds gas to the `Synthetix` contract deployment, but only around `4k`: from `6,742,579` to `6,746,386` (as you can [see here in gas checks](https://github.com/Synthetixio/synthetix/runs/293117793))

### DIFF
--- a/contracts/Synthetix.sol
+++ b/contracts/Synthetix.sol
@@ -288,6 +288,30 @@ contract Synthetix is ExternStateToken {
         return availableSynths.length;
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


jacko125:
Looks good, agreed with Clinton this can be refactored into exchangeRates or feePool where exchange fee rate stored.

PR #314, opened 2019-11-09
Speeding up unit tests

justinjmoses:
At attempt to run tests in parallel on CircleCI to see how much time savings we can get.


justinjmoses:
Looks like it gets the build down to about 9 mins from 16/17. https://circleci.com/workflow-run/a2bb5797-6f4c-43e2-8092-6b4d2c400e2f. Better but still far from great.

k06a:
This also could help: https://github.com/trufflesuite/truffle/pull/2540

justinjmoses:
Thanks @k06a - that would help locally for sure - though i'd be curious to see how well ganache handles multiple contract tests all deploying to it simultaneously. 

I'm actually an advocate of moving away from `truffle` altogether and simply rely on `mocha-parallel-tests` and `ethers` - we don't use truffle for deployment and it's just another dependency that adds little value to our testing.

cgewecke:
@justinjmoses 

Another slow thing is `eth-gas-reporter` - it makes a quite a few extra calls to collect data. If you only need the report occasionally maybe there's a way to toggle it on and off in CI by inspecting the commit message? Might be worth a quick check with 'spec' reporter to see how much speed-up you get without it. 

I see this mocha config quite often.
```
reporter: process.env.GAS_REPORTER ? 'eth-gas-reporter' : 'spec'
```

cf: [eth-gas-reporter 171][1]

[1]: https://github.com/cgewecke/eth-gas-reporter/issues/171

cgewecke:
Ah now that I think about it will probably mess CodeChecks CI up unless data gets saved on merge..

justinjmoses:
Thanks @cgewecke - yeah the results are indeed strange - though they have worked in the past. Do you know if the gas diff computed via CodeChecks is against the last run on `master` or is it for the base branch of the PR `develop`?

cgewecke:
@justinjmoses Codechecks has always seemed a little weird here because both Travis and CircleCI are running. In CircleCI I think CC sometimes [needs some config help][1] identifying the target branch if it's not master or dev... 

Travis should correctly identify the target without help as long as push builds are enabled (default). If the data's gone off track it might take a couple merged PRs to establish correct baselines again.

One solution might be to use only Travis for gas checks and have that build be [fast_finish / allowed_failures][2]. Then you could look at gas as necessary but Circle would be the 'real' CI. Idk.

[1]: https://github.com/codechecks/docs/blob/master/configuration.md#settings
[2]: https://blog.travis-ci.com/2013-11-27-fast-finishing-builds

k06a:
It was merged: https://github.com/trufflesuite/truffle/pull/2540

justinjmoses:
@k06a looks good, though I suspect we'd have issues running this if all test specs are sharing the same ganache deployment, no?

k06a:
If you use before() and beforeEach() correctly it should work correctly. But hope in this case ganache is not the slowest point 😁

justinjmoses:
The tests share the same deployment information no? How would that not cause issues?

justinjmoses:
Closed in favor of #433 433

PR #324, opened 2019-11-12
gas opt issue synths

hav-noms:


contracts/Synthetix.sol
```diff
@@ -541,7 +541,7 @@ contract Synthetix is ExternStateToken {
      * @param currencyKey The currency to register synths in, for example sUSD or sAUD
      * @param amount The amount of synths to register with a base of UNIT
      */
-    function _addToDebtRegister(bytes32 currencyKey, uint amount)
+    function _addToDebtRegister(bytes32 currencyKey, uint amount, uint existingDebt)
```
jacko125: @hav-noms  Good, saving another call to get debtBalanceOf when its loaded before.

contracts/Synthetix.sol
```diff
@@ -835,6 +839,25 @@ contract Synthetix is ExternStateToken {
         }
     }
 
+    function _remainingIssuableSynthsinXDR(address issuer)
```
jacko125: @hav-noms  Could we refactor `remainingIssuableSynths(issuer, currencyKey)` to return the (alreadyIssued & remainingMaxIssuableAmount) based on the currencyKey entered. This should work with other functions that currently need remainingIssuableSynths

```    /**
     * @notice The remaining synths an issuer can issue against their total synthetix balance.
     * @param issuer The account that intends to issue
     * @param currencyKey The currency to price issuable value in
     */
    function remainingIssuableSynths(address issuer, bytes32 currencyKey)
        public
        view
        // Don't need to check for synth existing or stale rates because maxIssuableSynths will do it for us.
        returns (uint)
    {
        uint alreadyIssued = debtBalanceOf(issuer, currencyKey);
        uint maxIssuable = maxIssuableSynths(issuer, currencyKey);

        if (alreadyIssued >= maxIssuable) {
            maxIssuable = 0;
        } else {
            maxIssuable  = maxIssuable.sub(alreadyIssued);
        }

return (
maxIssuable, 
alreadyIssued
   )
    }```

Comment (hav-noms) [reply contracts/Synthetix.sol]: OK I'll try it out. Thanks

File: contracts/Synthetix.sol [hunk 3]
```diff
@@ -621,14 +616,34 @@ contract Synthetix is ExternStateToken {
         external
         optionalProxy
     {
+        (uint maxIssuable, uint debtXDR) = _getMaxIssuableInCurrency(currencyKey);
```
jacko125: @hav-noms  Would be clearer if `(uint maxIssuable, uint existingDebtInXDR)` when we pass this value onwards to _internalIssueSynths

PR #327, opened 2019-11-14
Inverse pricing refactoring

justinjmoses:
- Added two parameters to `ExchangeRates.setInversePricing()` to handle the edge case of upgrading `ExchangeRates` while one or more inverted rates are frozen. This allows the owner to set the inverted parameters on the new contract to match those of the existing.
- Upgraded all currency keys to use `toBytes32` which explicitly adds right padding when required. 
- Removed `preferredCurrency` tests from `SynthetixState` as it's no longer used
- Added better documentation to `rateOrInverted`
- Minor changes to `ExchangeRates.removeInversePricing()` to only `emit` the event when a rate is found, and to revert if not.
- removed `ExchangeRates.priceUpdateLock`
- Refactored `int-test` to use `mocha` instead of just a plain script
- Refactored and moved tests from unit, int, testnet into `test-contracts`, `test-publish` and `test-testnet` respectively. All tests live under the `test` folder now. 
- Moving old python tests and scripts to `legacy` folder for the time being
- Added tests to `publish` to test redeployment of `ExchangeRates` properly persisting frozen synths when necessary


package.json
```diff
@@ -11,10 +11,10 @@
 		"ganache:int": "npm run ganache -- --acctKeys keys.json",
 		"lint": "solhint \"contracts/**/*.sol\" && eslint \"**/*.js\"",
 		"lint:fix": "eslint --fix \"**/*.js\"",
-		"test": "concurrently --kill-others --success first \"npm run ganache > /dev/null\" \"truffle compile && wait-port 8545 && truffle test\"",
```
justinjmoses: the `truffle compile` here was redundant - it will always run before `truffle test`

.circleci/config.yml
```diff
@@ -1,52 +1,116 @@
...
+      - run: npm run test:publish
+      - run: cat test/publish/test.log
+
+  test-kovan:
+    <<: *env_defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run: npm run test:testnet -- --network kovan
```
jacko125: Whats the strategy for failure recovery here ?

contracts/ExchangeRates.sol
```diff
@@ -374,11 +385,13 @@ contract ExchangeRates is SelfDestructible {
...
-                break;
+                // Track the event
+                emit InversePriceConfigured(currencyKey, 0, 0, 0);
+
+                return;
             }
         }
-
-        emit InversePriceConfigured(currencyKey, 0, 0, 0);
+        require(false, "No inverse price found");
```
jacko125: Maybe this could be more explicit in checking that the currencyKey exists in rates / invertedKeys and revert.

index.js
```diff
@@ -2,6 +2,12 @@
 
 const fs = require('fs');
 const path = require('path');
+const w3utils = require('web3-utils');
+
+/**
+ * Converts a string into a hex representation of bytes32, with right padding
+ */
+const toBytes32 = key => w3utils.rightPad(w3utils.asciiToHex(key), 64);
```
jacko125: @justinjmoses  Maybe we suggest a PR to web3.js to include this into their library

.circleci/config.yml
justinjmoses: Well failure means lost testnet ETH, SNX and sUSD, so it wouldn't be a tragedy. Though perhaps one mitigation here is to use much lower values of SNX and sUSD (basically dust) so it's not a meaningful amount.

jacko125: Think that is good idea!

index.js
justinjmoses: hmmm - i think they are trying to move away from magically doing stuff like this. they prefer we be explicit, which makes sense.

PR #328, opened 2019-11-17
Audit remediations

jacko125:


contracts/FeePool.sol
```diff
@@ -417,7 +410,9 @@ contract FeePool is Proxyable, SelfDestructible, LimitedSetup {
         public
         optionalProxy_onlyOwner
         onlyDuringSetup
-    {
+    {        
+        require (startingDebtIndex <= synthetixState.debtLedgerLength(), "Cannot import bad data");
```
justinjmoses: Should this be `<`? The `startingDebtIndex` will always be less than the ledger length?

jacko125: Correct, ensures that the feePeriods we import into feePool are valid, as `startingDebtIndex` will always be the debt ledger length or greater.   `startingDebtIndex` is set on a feePeriod when we are closing a period and a new period reads the debt ledger length at the time.

PR #329, opened 2019-11-18
Fixing flaky rewards test

justinjmoses:
This PR is to address one test that would intermittently fail on CircleCI only (could not repro locally or on Travis). Most of the test flakes were failing with:

```
[1]      AssertionError: Number is too large to be close (Delta between actual and expected is 576923076923076923076931, but variance was only 10): expected false to be truthy
[1]       at Function.assertBNClose [as bnClose] (utils/testUtils.js:293:9)
[1]       at Context.it (test/RewardsIntegrationTests.js:733:11)
[1]       at process._tickCallback (internal/process/next_tick.js:68:7)
```

but occasionally one would fail with:

```
[1]      Error: Returned error: VM Exception while processing transaction: revert No supply is mintable -- Reason given: No supply is mintable.
[1]      at PromiEvent (node_modules/truffle/build/webpack:/packages/contract/lib/promievent.js:6:1)
[1]       at TruffleContract.mint (node_modules/truffle/build/webpack:/packages/contract/lib/execute.js:158:1)
[1]       at Context.it (test/contracts/RewardsIntegrationTests.js:712:20)
[1]       at process._tickCallback (internal/process/next_tick.js:68:7)
```

Which suggested the issue was that the fee period wasn't correctly completing. 

I tried various solutions via trial and error (most involving mining extra blocks), and the working solution involves adding a small additional number of seconds to fast forward before closing the fee period, suggesting the problem is that the fee period isn't ready to be closed when we attempt it. This fix, while ungraceful, prevents false negatives in tests for now. Longer term, investigating why this is happening on CircleCI would be worthwhile.

### DIFF
--- a/test/RewardsIntegrationTests.js
+++ b/test/RewardsIntegrationTests.js
@@ -22,6 +22,8 @@ const getInstance = getContractInstance(web3);
 contract('Rewards Integration Tests', async accounts => {
 	// Updates rates with defaults so they're not stale.
 	const updateRatesWithDefaults = async () => {
+		await mineBlock();
+
 		const timestamp = await currentTime();
 
 		await exchangeRates.updateRates(
--- a/test/RewardsIntegrationTests.js
+++ b/test/RewardsIntegrationTests.js
@@ -33,8 +33,7 @@ contract('Rewards Integration Tests', async accounts => {
 		);
 	};
 
-	const closeFeePeriodAndFastForward = async () => {
-		// console.log('closeFeePeriodAndFastForward');
+	const fastForwardAndCloseFeePeriod = async () => {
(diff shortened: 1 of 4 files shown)


test/RewardsIntegrationTests.js
```diff
@@ -189,7 +191,7 @@ contract('Rewards Integration Tests', async accounts => {
 
 		it('should allocate the 3 accounts a third of the rewards for 1 period', async () => {
 			// Close Fee Period
-			await closeFeePeriodAndFastForward();
+			await fastForwardAndCloseFeePeriod();
```
jacko125: We should rename this on the feePool tests too

PR #331, opened 2019-11-19
Index synth dedupe

justinjmoses:


publish/src/commands/deploy.js
```diff
@@ -938,7 +938,7 @@ const deploy = async ({
 			});
 		}
 
-		if (proxyFeePool) {
+		if (proxyFeePool && synth) {
```
justinjmoses: Unrelated fix - was causing issues when not deploying new synths.

test/publish/index.js
```diff
@@ -383,6 +383,8 @@ describe('publish scripts', function() {
 
 												fs.writeFileSync(configJSONPath, JSON.stringify(configForExrates));
 
+												this.timeout(60000);
```
justinjmoses: Give more time to the deployer if a bunch of new contracts are being deployed

PR #336, opened 2019-11-20
Synthetix v2.13.1 Arcturus Release into ropsten

justinjmoses:


justinjmoses:
Note: this failed due to an issue with the rinkeby testnet script not correctly picking up on the nonce to use `Error: Returned error: replacement transaction underpriced` - seems like it continued on the the SNX transfer before the ETH transfer was mined. It's flaky so we need to look at fixing it.

justinjmoses:
Yeah the second time it failed with `nonce too low`, so it's definitely struggling on testnet. Let me be explicit with the nonce

PR #340, opened 2019-11-26
Only issue burn susd

justinjmoses:


test/contracts/FeePool.js
```diff
@@ -683,9 +683,9 @@ contract('FeePool', async accounts => {
 		assert.bnEqual(await sAUDContract.balanceOf(owner), oldSynthBalance.add(feesAvailable[0]));
 	});
 
-	it('should revert when a user tries to double claim their fees', async () => {
+	it('should revert when a user tries to double claim th eir fees', async () => {
```
hav-noms: their

hav-noms:
LGTM. Awaiting feeClaim updates

migrations/2_deploy_synthetix_system.js
```diff
@@ -316,12 +316,12 @@ module.exports = async function(deployer, network, accounts) {
 
 	// Initial prices
 	const { timestamp } = await web3.eth.getBlock('latest');
-
+	// XDR: 1 USD
```
justinjmoses: I updated the comments here to reflect what we're actually doing (even though setting the XDR rate explicitly isn't kosher).

jacko125: https://github.com/Synthetixio/synthetix/blob/only-issue-burn-susd/contracts/ExchangeRates.sol#L203-L206

This might just override the XDR rate being sent in ?

test/contracts/FeePool.js
```diff
@@ -621,13 +620,13 @@ contract('FeePool', async accounts => {
...
 
-		await feePool.claimFees(sUSD, { from: account1 });
+		await feePool.claimFees({ from: account1 });
 
 		assert.bnClose(feesAvailableOwner[0], totalFees.div(web3.utils.toBN('2')), '8');
 		assert.bnClose(feesAvailableAcc1[0], totalFees.div(web3.utils.toBN('2')), '8');
 	});
 
-	it('should allow a user to claim their fees in sAUD', async () => {
+	it('should allow a user to claim their fees in sAUD after some exchanging', async () => {
```
hav-noms: The test name should just say sUSD not sAUD right? after some exchanging sounds like the test would do the exchange but it doesn't. Maybe the test should be named "should allow a user to claim 50% of the fees as they hold 50% of the debt"

migrations/2_deploy_synthetix_system.js
justinjmoses: Yes it does

test/contracts/FeePool.js
justinjmoses: the test does do the the exchange? See lines 647-648

PR #347, opened 2019-12-04
Integration with Chainlink: Phase 1 - SIP-32

justinjmoses:
Addresses https://sips.synthetix.io/sips/sip-32


contracts/ExchangeRates.sol
```diff
@@ -489,34 +543,6 @@ contract ExchangeRates is SelfDestructible {
         return (_localRates, anyRateStale);
     }
 
-    /**
```
justinjmoses: These are redundant and don't seem to be used anywhere anymore, so removing them.

package.json
```diff
@@ -11,7 +11,7 @@
 		"ganache:int": "npm run ganache -- --acctKeys keys.json",
 		"lint": "solhint \"contracts/**/*.sol\" && eslint \"**/*.js\"",
 		"lint:fix": "eslint --fix \"**/*.js\"",
-		"test": "concurrently --kill-others --success first \"npm run ganache > /dev/null\" \"wait-port 8545 && truffle test test/contracts/*\"",
+		"test": "concurrently --kill-others --success first \"npm run ganache > /dev/null\" \"wait-port 8545 && truffle compile && truffle test ${TRUFFLE_TEST:=test/contracts/*}\"",
```
justinjmoses: As discussed, calling `compile` first will create the `build` folder if not already created and store the compiled contracts -without this truffle test will always recompile everything in memory :(

test/contracts/DelegateApprovals.js
```diff
@@ -1,5 +1,7 @@
 const DelegateApprovals = artifacts.require('DelegateApprovals');
 
+require('.'); // import common test scaffolding
```
justinjmoses: This allows all the tests to be run in isolation

test/contracts/Synthetix.js
```diff
@@ -705,10 +707,9 @@ contract('Synthetix', async accounts => {
 			synthetix.methods['transfer(address,uint256)'](account2, value, { from: account1 })
 		);
 		await assert.revert(
-			synthetix.methods['transfer(address,uint256,bytes)'](account2, value, data),
-			{
+			synthetix.methods['transfer(address,uint256,bytes)'](account2, value, data, {
```
justinjmoses: The update to `assert.revert` to include an optional `reason` found this bug.

contracts/ExchangeRates.sol
```diff
@@ -396,25 +408,67 @@ contract ExchangeRates is SelfDestructible {
...
+        }
+    }
+
+    /**
+     * @notice Add a pricing aggregator for the given key. Note: existing aggregators may be overridden.
+     * @param currencyKey THe currency key to add an aggregator for
+     */
+    function addAggregator(bytes32 currencyKey, address aggregatorAddress) external onlyOwner {
+        AggregatorInterface aggregator = AggregatorInterface(aggregatorAddress);
+        require(aggregator.latestTimestamp() >= 0, "Given Aggregator is invalid");
```
justinjmoses: Note: if the address is not a valid `AggregatorInterface`, this will revert with no reason given.

contracts/ExchangeRates.sol
```diff
@@ -123,11 +127,19 @@ contract ExchangeRates is SelfDestructible {
     }
 
     function rates(bytes32 code) public view returns(uint256) {
-        return uint256(_rates[code].rate);
+        if (aggregators[code] != address(0)) {
+            return uint(aggregators[code].latestAnswer() * 1e10);
```
jacko125: @justinjmoses  Maybe could use `SafeDecimalMath.unit()` here in place of 1e10

contracts/ExchangeRates.sol
```diff
@@ -396,25 +408,67 @@ contract ExchangeRates is SelfDestructible {
...
 
-                return;
+                return true;
             }
         }
+        return false;
     }
+    /**
+     * @notice Remove a pricing aggregator for the given key
+     * @param currencyKey THe currency key to remove an aggregator for
```
hav-noms: T**H**e, and the comments above

contracts/ExchangeRates.sol
justinjmoses: You mean `1e18`? Chainlink's aggregators use 10 decimal places for their values, so `unit` wouldn't work here.

PR #356, opened 2019-12-10
Release supply schedule to Kovan 

jacko125:


contracts/Math.sol
```diff
@@ -0,0 +1,44 @@
...
+    * Calculates to 18 digits of precision with SafeDecimalMath.unit()
+    */
+    function powDecimal(uint x, uint n)
+        internal
+        pure
+        returns (uint)
+    {
+        // https://mpark.github.io/programming/2014/08/18/exponentiation-by-squaring/
+        
+        // double exp(double x, int n) {
```
justinjmoses: [np] You don't need to comment the whole algorithm here - I think just the link above is sufficient

contracts/SupplySchedule.sol
```diff
@@ -1,250 +1,273 @@
...
------------------------------------------------------------------
-
-file:       SupplySchedule.sol
-version:    1.0
-author:     Jackson Chan
-            Clinton Ennis
-date:       2019-03-01
-
 -----------------------------------------------------------------
 MODULE DESCRIPTION
```
justinjmoses: Let's remove this description entirely and rely on the new docs.

contracts/Math.sol
```diff
@@ -0,0 +1,44 @@
+pragma solidity 0.4.25;
+
+import "./SafeDecimalMath.sol";
+
+contract Math {
```
justinjmoses: We need to add tests for this contract.

justinjmoses: And why not use a `library`?

contracts/SupplySchedule.sol
```diff
@@ -1,250 +1,273 @@
...
+    // Time of the last inflation supply mint event
     uint public lastMintEvent;
 
-    Synthetix public synthetix;
+    // Counter for number of weeks since the start of supply inflation
+    uint public weekCounter;
 
-    uint constant SECONDS_IN_YEAR = 60 * 60 * 24 * 365;
+    // The number of SNX rewarded to the caller of Synthetix.mint()
+    uint public minterReward = 50 * SafeDecimalMath.unit();
```
justinjmoses: I still think this is too high. It should be 10 or less.

test/contracts/RewardsIntegrationTests.js
```diff
@@ -699,67 +699,64 @@ contract('Rewards Integration Tests', async accounts => {
 			// await logFeePeriods();
 
 			// Account2 should have 67% of the minted rewards
-			const account2Escrow3 = await rewardEscrow.getVestingScheduleEntry(account2, 2); // Account2's 3rd escrow entry
+			// const account2Escrow3 = await rewardEscrow.getVestingScheduleEntry(account2, 2); // Account2's 3rd escrow entry
```
justinjmoses: Are there any assertions left here? If not, then why `mint()` above?

test/contracts/Synthetix.js
```diff
@@ -2322,87 +2323,98 @@ contract('Synthetix', async accounts => {
...
+		// These tests are using values modeled from https://sips.synthetix.io/sips/sip-23
+		// https://docs.google.com/spreadsheets/d/1a5r9aFP5bh6wGG4-HIW2MWPf4yMthZvesZOurnG-v_8/edit?ts=5deef2a7#gid=0
+		const INITIAL_WEEKLY_SUPPLY = 75e6 / 52;
 
 		const DAY = 86400;
 		const WEEK = 604800;
-		const YEAR = 31536000;
 
-		const YEAR_TWO_START = 1551830400;
+		const INFLATION_START_DATE = 1551830400; // 2019-03-06T00:00:00+00:00
```
justinjmoses: This matches what's in ganache on startup right? This seems very brittle - better to grab the first timestamp from ganache after startup

test/contracts/Synthetix.js
```diff
@@ -2413,114 +2425,145 @@ contract('Synthetix', async accounts => {
...
+			assert.bnEqual(newTotalSupply, existingTotalSupply.add(expectedSupplyToMint));
+			assert.bnEqual(newTotalSupply, expectedTotalSupply);
+			assert.bnEqual(mintableSupply, expectedSupplyToMint);
 		});
 
-		it('should revert when trying to mint again within the 7 days period', async () => {
-			// Issue
-			const weeklyIssuance = divideDecimal(secondYearSupply, 52 / 2);
-			const expectedSupplyToMint = weeklyIssuance;
+		// it('should be able to mint again after another 7 days period', async () => {
```
justinjmoses: If we want to keep these around, let's change them to `xit` to skip for now, otherwise just remove them completely.

contracts/Math.sol
jacko125: Comments removed

contracts/Math.sol
jacko125: Updated Math.sol to be used as a library. Next step will be writing testing contract to test library.

test/contracts/SupplySchedule.js
```diff
@@ -1,86 +1,83 @@
...
+	powerToDecimal,
+	ZERO_ADDRESS,
+} = require('../utils/testUtils');
 const BN = require('bn.js');
 
-contract('SupplySchedule', async accounts => {
-	const DAY = 86400;
-	const WEEK = 604800;
-	const YEAR = 31536000;
+contract.only('SupplySchedule', async accounts => {
```
justinjmoses: Please remove

test/contracts/Synthetix.js
hav-noms: these have been restored

test/contracts/RewardsIntegrationTests.js
hav-noms: Yeah deferring fixing these again. Have restored the commented out parts

contracts/SupplySchedule.sol
jacko125: Complete

jacko125:
> Math still needs some tests.

@justinjmoses  Math library tests completed and added PublicMaths contract for testing

PR #362, opened 2019-12-13
sip-30 Deprecate erc223 

hav-noms:
https://sips.synthetix.io/sips/sip-30


contracts/Depot.sol
```diff
@@ -45,7 +45,7 @@ contract Depot is SelfDestructible, Pausable {
     using SafeDecimalMath for uint;
 
     /* ========== STATE VARIABLES ========== */
-    ISynthetix public synthetix;
+    address public snxProxy;
```
hav-noms: use snxProxy here instead to remove call from update script

contracts/Depot.sol
```diff
@@ -420,7 +420,7 @@ contract Depot is SelfDestructible, Pausable {
         fundsWallet.transfer(msg.value);
 
         // And send them the SNX.
-        synthetix.transfer(msg.sender, synthetixToSend);
+        IERC20(snxProxy).transfer(msg.sender, synthetixToSend);
```
hav-noms: Don't need ISynthetix.sol its only doing a snx.transfer

contracts/Depot.sol
```diff
@@ -549,43 +549,27 @@ contract Depot is SelfDestructible, Pausable {
...
-     *         if they'd like. You can equally just transfer synths to this contract and it will work
-     *         exactly the same way but with one less call (and therefore cheaper transaction fees)
+     * @notice depositSynths: Allows users to deposit synths via the approve / transferFrom workflow     
      * @param amount The amount of sUSD you wish to deposit (must have been approved first)
      */
     function depositSynths(uint amount)
         external
     {
-        // Grab the amount of synths
+        // Grab the amount of synths. Will fail if not approved first
```
hav-noms: mintr must now call depositSynths()

contracts/Synthetix.sol
```diff
@@ -276,68 +275,38 @@ contract Synthetix is ExternStateToken {
...
-     */
-    function transfer(address to, uint value, bytes data)
         public
         optionalProxy
         returns (bool)
     {
-        // Ensure they're not trying to exceed their locked amount
-        require(value <= transferableSynthetix(messageSender), "Insufficient balance");
+        // Ensure they're not trying to exceed their staked SNX amount
+        require(value <= transferableSynthetix(messageSender), "Cannot transfer staked or escrowed SNX");
```
hav-noms: Message is longer that 32 bytes but its worth while message. hopefull will reduce support. I've save bytecode space in other messages and requires

contracts/Depot.sol
jacko125: Looks good, dapp would need to let users to `approve()` first.

contracts/Synth.sol
```diff
@@ -129,37 +107,15 @@ contract Synth is ExternStateToken {
         public
         optionalProxy
         returns (bool)
-    {
-        require(from != 0xfeefeefeefeefeefeefeefeefeefeefeefeefeef, "The fee address is not allowed");
```
jacko125: Good removing this require as per Anton's comments

publish/deployed/kovan/deployment.json
```diff
@@ -29,11 +29,11 @@
...
 		"ExchangeRates": {
 			"name": "ExchangeRates",
-			"address": "0x1ec58775e6169e475cCfC10168e5c945c6e56a4b",
+			"address": "0xE15f59e3806787a1Ef621a3595511DF7999E97D2",
 			"source": "ExchangeRates",
-			"link": "https://kovan.etherscan.io/address/0x1ec58775e6169e475cCfC10168e5c945c6e56a4b",
-			"timestamp": "2019-12-13T04:58:00.000Z",
-			"txn": "https://kovan.etherscan.io/tx/0x0a63d2c1ba51e62fe7fed403f55f693477a175c13be83e88330816fcc2aaa18e",
+			"link": "https://kovan.etherscan.io/address/0xE15f59e3806787a1Ef621a3595511DF7999E97D2",
+			"timestamp": "2019-12-10T04:51:08.000Z",
```
jacko125: Need to get the new ExchangeRates for Kovan

PR #366, opened 2019-12-13
add Synthetix header

jacko125:


publish/src/solidity.js
```diff
@@ -6,6 +6,40 @@ const solidifier = require('solidifier');
 const solc = require('solc');
 const { COMPILED_FOLDER } = require('./constants');
 
+const addSynthetixHeader = ({ content, contract }) => {
```
justinjmoses: [np] I'd opt for a separate little module for this as it's pretty distracting in this file

PR #367, opened 2019-12-15
Add synthetix header to flattened contracts

jacko125:


publish/src/solidity-header.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+module.exports = {
+	addSynthetixHeader({ content, contract }) {
```
justinjmoses: [np] Now this should be `addSolidityHeader` or even just `insertCodeHeader`

PR #374, opened 2019-12-18
Fixes for deploy

justinjmoses:


publish/src/check-aggregator-prices.js
```diff
@@ -0,0 +1,68 @@
...
+						address: synth.aggregator,
+						apikey: process.env.ETHERSCAN_KEY,
+					},
+				});
+				abi = JSON.parse(result);
+			}
+
+			const liveAggregator = new web3.eth.Contract(abi, synth.aggregator);
+
+			const [first, second] = await Promise.all([
```
jacko125: could name them as Aggregator / exchangeRates ?

PR #381, opened 2020-01-21
Release 2.19.x kovan

jacko125:


contracts/FeePool.sol
```diff
@@ -71,7 +71,7 @@ contract FeePool is Proxyable, SelfDestructible, LimitedSetup {
     uint public constant MAX_FEE_PERIOD_DURATION = 60 days;
 
     // Users are unable to claim fees if their collateralisation ratio drifts out of target treshold
-    uint public targetThreshold = (10 * SafeDecimalMath.unit()) / 100;
+    uint public targetThreshold = (1 * SafeDecimalMath.unit()) / 100;
```
justinjmoses: But this isn't the version deployed in kovan is it?

jacko125: Correct, This isn't deployed into kovan yet. I think it will be the plan to have it go in with XDR removal today. 

I think we can hold off merging in and combine with XDR changes first then.

justinjmoses: Please do and then you can merge

PR #385, opened 2020-01-29
Sip 35 skinny ether collateral

hav-noms:


contracts/Depot.sol
```diff
@@ -433,7 +433,7 @@ contract Depot is SelfDestructible, Pausable {
      * @param guaranteedEtherRate The ether exchange rate which must be honored or the call will revert.
      * @param guaranteedSynthetixRate The synthetix exchange rate which must be honored or the call will revert.
      */
-    function exchangeEtherForSynthetixAtRate(uint guaranteedEtherRate, uint guaranteedSynthetixRate)
```
justinjmoses: Why are we changing this name - are we redeploying it?

contracts/Synth.sol
```diff
@@ -118,21 +127,61 @@ contract Synth is ExternStateToken {
...
+    {   
+        uint amountInUSD;
+
+        // sUSD can be transferred to FEE_ADDRESS directly
+        if (currencyKey == "sUSD") {
+            amountInUSD = value;
+            super._internalTransfer(messageSender, to, value);
+        } else {
+            // else exchange synth into sUSD and send to FEE_ADDRESS
+            ISynthetix(synthetixProxy).synthInitiatedExchange(messageSender, currencyKey, value, "sUSD", FEE_ADDRESS);
```
justinjmoses: note: this is going away with SIP-38 and will need to use `exchange`

test/contracts/EtherCollateral.js
```diff
@@ -0,0 +1,399 @@
...
+	getEthBalance,
+	toUnit,
+	// multiplyDecimal,
+	// divideDecimal,
+	ZERO_ADDRESS,
+} = require('../utils/testUtils');
+
+const { toBytes32 } = require('../../.');
+
+contract.only('EtherCollateral', async accounts => {
```
justinjmoses: don't forget to remove this guy

justinjmoses: alternatively you can use `TRUFFLE_TEST=test/contracts/EtherCollateral.js npm test` to do this via the CLI

contracts/Synth.sol
hav-noms: Thanks noted will need to change this on the merge @jacko125

contracts/Depot.sol
hav-noms: Yes in case the Depot is drained of all of its sUSD it's getting ready for a redeploy. This change was in response to questions from @oleiba in discord where it seems ambiguous what synthetix is when returned whether it was synths when its actually the SNX token.

PR #390, opened 2020-02-09
SIP-37 Fee reclamation rebates

justinjmoses:
Todo
- [x] `Issuer.burnSynths` tests wait for waiting period when required
- [x] `ExchangeState` tests
- [x] `AddressResolver` tests
- [x] `ExchangeRate` tests for historical rates
- [x] `Synth.transferableSynths` tests
- [x] Doc updates


contracts/ExchangeRates.sol
```diff
@@ -81,50 +83,27 @@ contract ExchangeRates is SelfDestructible {
...
-
-    function _setRate(bytes32 code, uint256 rate, uint256 time) internal {
-        _rates[code] = RateAndUpdatedTime({rate: uint216(rate), time: uint40(time)});
+    function setRateStalePeriod(uint _time) external onlyOwner {
+        rateStalePeriod = _time;
+        emit RateStalePeriodUpdated(rateStalePeriod);
     }
 
-    /* ========== SETTERS ========== */
+    /* ========== MUTITATIVE FUNCTIONS ========== */
```
jacko125: @justinjmoses  small spelling here - MUTITATIVE -> MUTATIVE

justinjmoses: 👍

contracts/Exchanger.sol
```diff
@@ -131,26 +173,126 @@ contract Exchanger is MixinResolver {
...
+
+    function appendExchange(address account, bytes32 src, uint amount, bytes32 dest, uint amountReceived) internal {
+        IExchangeRates exRates = exchangeRates();
+        uint roundIdForSrc = exRates.getCurrentRoundId(src);
+        uint roundIdForDest = exRates.getCurrentRoundId(dest);
+        exchangeState().appendExchangeEntry(account, src, amount, dest, amountReceived, now, roundIdForSrc, roundIdForDest);
+    }
+
+    function getRoundIdsAtPeriodEnd(address account, bytes32 currencyKey, uint index) internal view returns (uint, uint) {
+        (bytes32 src, , bytes32 dest, , uint timestamp, uint roundIdForSrc, uint roundIdForDest) = exchangeState()
```
jacko125: @justinjmoses  `getRoundIdsAtPeriodEnd()` is used internally by  `settlementOwing()` only.  Could optimise for gas by not having to load the exchangeEntry from `exchangeState` again in https://github.com/Synthetixio/synthetix/blob/sip-37-fee-reclamation-rebates/contracts/Exchanger.sol#L276-L277 and reuse the state by passing it through to  `getRoundIdsAtPeriodEnd()`.

justinjmoses: Ah I believe the trouble I had was that `settlementOwing()` kept hitting the `Stack too deep` error with the number of local variables. I suppose we could break it out even more though into additional functions?

contracts/ExchangeRates.sol
```diff
@@ -395,8 +232,83 @@ contract ExchangeRates is SelfDestructible {
...
+        uint startingRoundId,
+        uint startingTimestamp,
+        uint timediff
+    ) external view returns (uint) {
+        uint roundId = startingRoundId;
+        uint nextTimestamp = 0;
+        while (true) {
+            (, nextTimestamp) = getRateAndTimestampAtRound(currencyKey, roundId + 1);
+            // if there's no new round, then the previous roundId was the latest
+            if (nextTimestamp == 0 || nextTimestamp > startingTimestamp + timediff) {
```
jacko125: @justinjmoses  If we want to optimise the search for the `lastRoundID` that matches the criteria and not continuously loop until finding an matching entry:
1. timestamp >  (trade startingTimeStamp + 3 minutes)
2. OR nextTimestamp == 0 

Instead of iterating from `roundID + 1` until we find the roundID needed , using binary search algorithm would reduce the cost to O(log n) , relying on the fact that roundID is sorted and incremental to split size of values to search (roundID - lastestRoundID)   

Chainlink aggregator should have a function to return latestRound
`function latestRound() public view returns (uint256) {
        return roundId;
    }`

and `exchangeRates` would need to store a mapping of the latestRoundID for each currencyKey.

jacko125: Alternatively, possible to check if currentRoundID equals the trade roundID already (in cases where no rates updated after 'n' minutes.

```
// if latest currentRound equals trade round 
if (getCurrentRoundId(currencyKey) == roundId) {
      return roundId;
 }

// loop through roundIDs 
```

justinjmoses: Hmm - binary search is definitely a better algorithm, but given the fact that the timeDiff is measured in minutes and oracle updates are at absolute most once per minute, don't you think it's overkill?

As for `getCurrentRoundId()` - yeah that could be a good optimization to add.

jacko125: Thats definitely true, I realised that in the worse case for our case it'll only be a few prices to search within the 3 minutes and it'll be overkill, or even no updates within 3 minutes for less volatile assets.

PR #391, opened 2020-02-10
Moves emit call to be after state mutation

evgenyboxer:
After consulting with @hav-noms I have moved the emit call on `acceptOwnership` to come after the state mutation.


evgenyboxer:
closing this after some further research, it is actually ok. will add some comments around this function.

PR #392, opened 2020-02-10
update to call notifyRewardAmount

hav-noms:


contracts/RewardsDistributionRecipient.sol
```diff
@@ -0,0 +1,19 @@
...
+
+import "./Owned.sol";
+
+
+contract RewardsDistributionRecipient is Owned {
+    address rewardsDistribution;
+
+    function notifyRewardAmount(uint256 reward) external;
+
+    modifier onlyRewardsDistribution() {
```
justinjmoses: Note: This is what I meant by `State.sol` being badly named - we could use it here for this use-case, it's just not *state* as such, so it doesn't make sense that this extend "State".

jacko125: @justinjmoses  It should be named `Associated.sol` which `state.sol` essentially does

hav-noms: @justinjmoses yeah I agree. This could also implement the address resolver for the modifier. 
There are a lot of places though it does not make sense to lookup the address each time. (I would ague all of them) maybe we could mix in a local resolver which the central resolver gets updated by the owner and has the authority to push the new addresses out to each consumer.

PR #395, opened 2020-02-13
update multiCollateralSynths to use address.resolver. Add multiCollat…

jacko125:
…eralKey to synth

### DIFF
--- a/contracts/MultiCollateralSynth.sol
+++ b/contracts/MultiCollateralSynth.sol
@@ -41,7 +41,7 @@ contract MultiCollateralSynth is Synth {
     /* ========== VIEWS ======================= */
 
     function multiCollateral() internal view returns (address) {
-        require(resolver.getAddress(multiCollateralKey) != address(0), "Resolver is missing MultiCollateral address");
+        require(resolver.getAddress(multiCollateralKey) != address(0), "Resolver is missing multiCollateral address");
         return resolver.getAddress(multiCollateralKey);
     }
 
--- a/contracts/MultiCollateralSynth.sol
+++ b/contracts/MultiCollateralSynth.sol
@@ -22,6 +22,7 @@ import "./Synth.sol";
 
 contract MultiCollateralSynth is Synth {
     /* ========== CONSTRUCTOR ========== */
+    bytes32 public multiCollateralKey;
 
     constructor(
         address _proxy,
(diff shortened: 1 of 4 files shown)


contracts/MultiCollateralSynth.sol
```diff
@@ -22,6 +22,7 @@ import "./Synth.sol";
 
 contract MultiCollateralSynth is Synth {
     /* ========== CONSTRUCTOR ========== */
+    bytes32 public multiCollateralKey;
```
justinjmoses: Qu: worth a setter for this if we need to update it or is that overkill?

contracts/MultiCollateralSynth.sol
```diff
@@ -31,14 +32,17 @@ contract MultiCollateralSynth is Synth {
...
+    ) public Synth(_proxy, _tokenState, _tokenName, _tokenSymbol, _owner, _currencyKey, _totalSupply, _resolver) {
+        multiCollateralKey = _multiCollateralKey;
+    }
 
     /* ========== VIEWS ======================= */
 
     function multiCollateral() internal view returns (address) {
-        require(resolver.getAddress("MultiCollateral") != address(0), "Resolver is missing MultiCollateral address");
-        return resolver.getAddress("MultiCollateral");
+        require(resolver.getAddress(multiCollateralKey) != address(0), "Resolver is missing MultiCollateral address");
```
justinjmoses: [np] Instead of capitalizing "MultiCollateral", I'd camelCase it to indicate that it's a variable

migrations/2_deploy_synthetix_system.js
```diff
@@ -307,18 +307,35 @@ module.exports = async function(deployer, network, accounts) {
...
+				`Synth ${currencyKey}`,
+				currencyKey,
+				owner,
+				toBytes32(currencyKey),
+				web3.utils.toWei('0'),
+				resolver.address,
+				toBytes32('EtherCollateral'),
+				{ from: deployerAccount }
+			);
+		} else {
```
justinjmoses: copy pasta. 

image

contracts/MultiCollateralSynth.sol
jacko125: @justinjmoses  isn't `multiCollateralKey` camel case ?

contracts/MultiCollateralSynth.sol
jacko125: I'd say its overkill given we can re-deploy synths behind proxy, 

If say `collateralkey` changes on address resolver , synthetix.sol would likely be upgraded too.

PR #396, opened 2020-02-13
update ETHCollateral to use address resolver

hav-noms:


test/contracts/EtherCollateral.js
```diff
@@ -176,20 +177,21 @@ contract('EtherCollateral', async accounts => {
...
-				}
-			);
-
-			assert.equal(await instance.synthProxy(), synthProxy.address);
-			assert.equal(await instance.sUSDProxy(), sUSDContract.address);
-			assert.equal(await instance.depot(), depot.address);
+			assert.equal(await instance.resolver(), addressResolver.address);
+		});
+
+		it('should access its dependencies via the address resolver', async () => {
```
justinjmoses: [np] not sure this is really worth it - you're just checking the address resolver is correctly setup in the migration script.

hav-noms: Yeah which is the same as before. Testing we got what we needed in the constructor. I guess it might help other failing tests that something might not be right in the deployment.

PR #400, opened 2020-02-15
SIP-37 Additional test coverage

justinjmoses:
- Primarily added new ABI checker that only expected mutative functions exist from the contract
- Making sure all restricted mutative functions are indeed restrictive

### DIFF
--- a/test/contracts/Synth.js
+++ b/test/contracts/Synth.js
@@ -114,7 +114,7 @@ contract('Synth', async accounts => {
 				});
 			});
 		});
-		describe('when non-multiCollateral tries to burn', () => {
+		describe('when non-internal tries to burn', () => {
 			it('then it fails', async () => {
 				await onlyGivenAddressCanInvoke({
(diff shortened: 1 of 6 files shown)


test/contracts/Synth.js
```diff
@@ -83,6 +87,45 @@ contract('Synth', async accounts => {
...
+			it('then it fails', async () => {
+				await onlyGivenAddressCanInvoke({
+					fnc: sUSDContract.issue,
+					args: [account1, toUnit('1')],
+					accounts,
+					reason: 'Only Synthetix, FeePool, Exchanger or Issuer contracts allowed',
+				});
+			});
+		});
+		describe('when non-multiCollateral tries to burn', () => {
```
jacko125: @justinjmoses  This test might be duplicated as testing synth.js here but refers non-multiCollateral

justinjmoses: It's just the name. Fixing

PR #403, opened 2020-02-17
Remove Synthetix.effectiveValue in favor of ExchangeRates.effectiveValue

justinjmoses:
cc @clementbalestrat @evgenyboxer - please make sure the dApps no longer reference `Synthetix.effectiveValue()` but rather invoke the same function on `ExchangeRates`

### DIFF
--- a/contracts/Exchanger.sol
+++ b/contracts/Exchanger.sol
@@ -201,8 +201,10 @@ contract Exchanger is MixinResolver {
         // // Issue their new synths
         _synthetix.synths(destinationCurrencyKey).issue(destinationAddress, amountReceived);
 
-        // Remit any fees if required
-        remitFeeIfAny(_exRates, _synthetix, fee, destinationCurrencyKey);
+        // Remit the fee if required
+        if (fee > 0) {
+            remitFee(_exRates, _synthetix, fee, destinationCurrencyKey);
+        }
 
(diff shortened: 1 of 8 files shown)


contracts/Exchanger.sol
```diff
@@ -200,13 +201,8 @@ contract Exchanger is MixinResolver {
...
 
-        // Remit the fee in sUSDs
-        if (fee > 0) {
-            uint usdFeeAmount = _synthetix.effectiveValue(destinationCurrencyKey, fee, sUSD);
-            _synthetix.synths(sUSD).issue(feePool().FEE_ADDRESS(), usdFeeAmount);
-            // Tell the fee pool about this.
-            feePool().recordFeePaid(usdFeeAmount);
-        }
+        // Remit any fees if required
+        remitFeeIfAny(_exRates, _synthetix, fee, destinationCurrencyKey);
```
jacko125: @justinjmoses  For this refactoring, 

would it be better to check if any fees before calling internal func? 

```
if (fee > 0) { 
remitFee(IExchangeRates _exRates, ISynthetix _synthetix, uint fee, bytes32 currencyKey
}
```

justinjmoses: I guess it's cheaper that calling another function yah? OK I'll fix.

PR #409, opened 2020-02-18
all audit resolutions and tests passing

hav-noms:


test/contracts/EtherCollateral.js
```diff
@@ -564,14 +564,14 @@ contract('EtherCollateral', async accounts => {
 							const openLoanIDsByAccount = await etherCollateral.openLoanIDsByAccount(address1);
 							assert.bnEqual(openLoanIDsByAccount[0], loan2ID);
 						});
-						xit('Loan Closed event emits the fees charged', async () => {
+						xit('LoanClosed event emits the fees charged', async () => {
```
justinjmoses: is this still pending?

hav-noms: Yes I still need to finalize getting these pending tests done. The interest calcs are well tested in other tests but on the event there is a small delta between the local offchain calculation and the final interest charged which is due to the time blockchain time lag. I could query the contract how much the interest was on the loan which is going to be exactly the same value but if there was an error in the contract calculations we wouldn't know. I was wondering whether to just round the test to 4 decimal places.

PR #410, opened 2020-02-19
add owner only function to burn any amount of synth at FEE_ADDRESS

jacko125:
Owner in feePool to burn FEE_ADDRESS synths


jacko125:
transfers of synths to FEE_ADDRESS will be treated as a fee payment with next release. function not needed and replaced with Synths burning when transfer to 0x address.

PR #426, opened 2020-02-23
New json tests

justinjmoses:
- Created a new test target `test:deployments` which tests the following
    - all synths in synths.json are added to `Synthetix`, and inverses are set as expected
    - all `AddressResolver` addresses are set
    - all ABIs from Etherscan are set (verified) and match expected ones
- Also fixed the testnet deployment JSON files where `ProxysUSD` had source incorrectly set to `Proxy` not `ProxyERC20`
- And reset the inverse pricing on the inverted synths in all testnets to match mainnet

cc @clementbalestrat - this will replace all those tests in synthetix-js checking contract addresses that don't belong there


migrations/2_deploy_synthetix_system.js
```diff
@@ -450,7 +449,6 @@ module.exports = async function(deployer, network, accounts) {
 			feePoolEternalStorage.address,
 			feePoolState.address,
 			issuer.address,
-			etherCollateral.address, // MultiCollateral for Synth uses EtherCollateral
```
jacko125: @justinjmoses  Is the idea that etherCollateral won't be updated any time soon?

hav-noms: It's in the list up higher as EtherCollateral not MultiCollateral but yeah sETH needs this address in the resolver

justinjmoses: Recall that SynthsETH is given `EtherCollateral` as a constructor argument. This was leftover.

jacko125: Correct, so no other contracts directly looks up `MultiCollateral` key https://github.com/Synthetixio/synthetix/blob/master/contracts/MultiCollateralSynth.sol#L27-L40

justinjmoses: Correct, so `MultiCollateral` is leftover that needs to be removed

PR #431, opened 2020-02-25
Remove XDR post release

hav-noms:


contracts/SynthetixState.sol
```diff
@@ -163,28 +160,28 @@ contract SynthetixState is State, LimitedSetup {
         emit IssuanceRatioUpdated(_issuanceRatio);
     }
 
-    /**
```
justinjmoses: We really need a way to have the latest code in synthetix yet also have the existing deployed code referenced in there for safekeeping. I need to think on a good way to do this.

hav-noms: yeah I wanted to get rid of the XDR and the extra compiler warnings

PR #432, opened 2020-02-25
gas optimisation for issuing and burning synths

jacko125:


.circleci/config.yml
```diff
@@ -3,6 +3,7 @@ env_defaults: &env_defaults
   working_directory: ~/repo
   docker:
     - image: circleci/node:10.16
+  resource_class: xlarge
```
justinjmoses: Merge in my latest changes to `develop` and you can remove this

test/utils/testUtils.js
```diff
@@ -447,4 +461,14 @@ module.exports = {
 
 	getEthBalance,
 	bytesToString,
+
+	sUSD,
```
justinjmoses: I'm not sure it really makes sense to add all these here to the output - it doesn't save us much to import it rather than just mapping them thru `toBytes32`. Maybe just export `defaultCurrencyKeys` and not the others?

jacko125: It's cleaning up the places where we do the `toBytes32` in all the different test scripts into a single place to use. It'll be less readible to use the `defaultCurrencyKeys`  in the updateRates field.

PR #433, opened 2020-02-25
Run Contract tests in parallel in CI

justinjmoses:
- Speeds up contract tests by splitting into 4 parallel tasks inside the `test_contracts` job
- Moves gas checks to travis (as each parallel gas check in CircleCI would override the others)


justinjmoses:
To get it faster, I need to configure truffle to use a JSON reporter (instead of the `eth-gas-reporter`), then get it to output JSON with all the truffle output removed, then have these results uploaded to CircleCI via https://circleci.com/docs/2.0/configuration-reference/#store_test_results

PR #434, opened 2020-02-26
Resolver gas improvements

justinjmoses:
https://sips.synthetix.io/sips/sip-46


test/contracts/FeePool.js
```diff
@@ -1385,11 +1385,13 @@ contract('FeePool', async accounts => {
 
 			// overwrite the Synthetix address in the resolver so we can issue
 			await addressResolver.importAddresses(['Synthetix'].map(toBytes32), [owner], { from: owner });
+			await XDRContract.setResolverAndSyncCache(addressResolver.address, { from: owner });
```
hav-noms: I was going to say this can be ripped out but we reverted removing XDRs. Shall we put that commit back in now to remove XDRs?

justinjmoses: Looks like you did so thanks :pray: - I just merged those changes in

contracts/EtherCollateral.sol
```diff
@@ -85,8 +85,21 @@ contract EtherCollateral is Owned, Pausable, ReentrancyGuard, MixinResolver {
     // Account Open Loan Counter
     mapping(address => uint256) public accountOpenLoanCounter;
 
+    /* ========== ADDRESS RESOLVER CONFIGURATION ========== */
```
jacko125: EtherCollateral contract would be hard to upgrade to use the new addressCache pattern (until the 3 month trial ends)

PR #436, opened 2020-02-27
Improvements to deploy script for AddressResolver

justinjmoses:
This tidies up the deploy script to better handle AddressResolver updates:

- If AddressResolver has not changed, then `importAddresses` is only called for any new contracts deployed (otherwise it will be called with all)
- If any contract with `setResolver()` does not have the current (or new) AddressResolver, invoke `setResolver()` with it. (Allowing for new AddressResolver deployments).

### DIFF
--- a/publish/src/commands/deploy.js
+++ b/publish/src/commands/deploy.js
@@ -1096,54 +1096,83 @@ const deploy = async ({
 	// -------------------------
 
 	if (addressResolver) {
-		await runStep({
-			gasLimit: 500e3, // higher gas required
-			contract: `AddressResolver`,
-			target: addressResolver,
-			write: 'importAddresses',
-			writeArg: [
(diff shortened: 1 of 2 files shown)


publish/src/commands/deploy.js
```diff
@@ -1096,54 +1096,83 @@ const deploy = async ({
...
+					addressesNotInResolver.map(({ name }) => toBytes32(name)),
+					addressesNotInResolver.map(({ address }) => address),
 				],
-			],
-		});
+			});
+		}
+
+		// Now for all targets that have a setResolver, we need to ensure the resolver is set
+		for (const [contract, target] of Object.entries(deployer.deployedContracts)) {
```
jacko125: good check that addressResolver is set correctly here

PR #438, opened 2020-02-28
add helper function canBurnSynths() 

jacko125:
add helper function canBurnSynths()


contracts/Issuer.sol
```diff
@@ -42,6 +42,12 @@ contract Issuer is MixinResolver {
         return IssuanceEternalStorage(resolver.requireAndGetAddress("IssuanceEternalStorage", "Missing IssuanceEternalStorage address"));
     }
 
+    /* ========== VIEWS ========== */
+
+    function canBurnSynths(address account) public view returns (bool) {
```
jacko125: @clementbalestrat Clem this function should help the dapp check if burnSynths is open for user.

justinjmoses: This can be `external` and also go in `IIssuer.sol`

contracts/Issuer.sol
```diff
@@ -114,7 +120,7 @@ contract Issuer is MixinResolver {
         external
         onlySynthetix
     {
-        require(now >= lastIssueEvent(from).add(minimumStakeTime), "Minimum stake time not reached");
+        require(canBurnSynths(from), "Minimum stake time not reached");
```
justinjmoses: it's a little more gas to use this as a `public` function rather than an internal

contracts/Issuer.sol
jacko125: Done. public so that the dapp can use the function.

contracts/Issuer.sol
```diff
@@ -281,5 +287,5 @@ contract Issuer is MixinResolver {
 
     /* ========== EVENTS ========== */
 
-    event minimumStakeTimeUpdated(uint minimumStakeTime);
+    event MinimumStakeTimeUpdated(uint minimumStakeTime);
```
jacko125: comment from Kyle (iosiro)

PR #447, opened 2020-03-04
Sips 10 & 14 improve delegation powers

hav-noms:
https://sips.synthetix.io/sips/sip-10
https://sips.synthetix.io/sips/sip-14


contracts/interfaces/IDelegateApprovals.sol
```diff
@@ -0,0 +1,14 @@
+pragma solidity 0.4.25;
+
+
+interface IDelegateApprovals {
+    function rateForCurrency(bytes32 currencyKey) external view returns (uint);
```
jacko125: maybe dont need this one here

hav-noms: oh whoops... how'd that get in there

contracts/Synthetix.sol
```diff
@@ -269,22 +269,6 @@ contract Synthetix is ExternStateToken, MixinResolver {
         return _transferFrom_byProxy(messageSender, from, to, value);
     }
 
-    function issueSynths(uint amount) external optionalProxy {
```
justinjmoses: Right yeah it makes sense to start moving these out now.

hav-noms: @justinjmoses @jacko125  I'm not 100% certain this is a good way to go just yet. I am concerned about composability. 
Before integration could just rely on the synthetix proxy to call issue /burn on now the requirement would be to integrate the address resolver so they can always be pointed to the current issuer.  Maybe I'm overthinking it but if the docs explain this well enough could avoid people hard coding the. Issuer address only to have it deprecated when we upgrade it. Also, nobody has integrated with Issue and burn yet but delegation should open that up.

@justinjmoses does this also mean we remove synthetix.exchange() for Exchanger.exchange()?

justinjmoses: Discussed offline. Summary: leave it all in `Synthetix` for now and put together a SIP to gather all these user functions into one kind of forwarder contract that will also emit these events.

contracts/Exchanger.sol
```diff
@@ -26,6 +27,7 @@ contract Exchanger is MixinResolver {
     bytes32 private constant CONTRACT_EXRATES = "ExchangeRates";
     bytes32 private constant CONTRACT_SYNTHETIX = "Synthetix";
     bytes32 private constant CONTRACT_FEEPOOL = "FeePool";
+    bytes32 private constant CONTRACT_DELEGATEAPPROVALS = "DelegateApprovals";
 
     bytes32[24] private addressesToCache = [CONTRACT_EXCHANGESTATE, CONTRACT_EXRATES, CONTRACT_SYNTHETIX, CONTRACT_FEEPOOL];
```
justinjmoses: it needs to be added in here as well

contracts/Issuer.sol
```diff
@@ -28,6 +29,7 @@ contract Issuer is MixinResolver {
     bytes32 private constant CONTRACT_EXCHANGER = "Exchanger";
     bytes32 private constant CONTRACT_SYNTHETIXSTATE = "SynthetixState";
     bytes32 private constant CONTRACT_FEEPOOL = "FeePool";
+    bytes32 private constant CONTRACT_DELEGATEAPPROVALS = "DelegateApprovals";
     bytes32 private constant CONTRACT_ISSUANCEETERNALSTORAGE = "IssuanceEternalStorage";
 
     bytes32[24] private addressesToCache = [
```
justinjmoses: ibid

contracts/Synthetix.sol
```diff
@@ -273,18 +273,34 @@ contract Synthetix is ExternStateToken, MixinResolver {
...
 
+    function issueMaxSynthsOnBehalf(address issueForAddress) external optionalProxy {
+        return issuer().issueMaxSynthsOnBehalf(issueForAddress, messageSender);
+    }
+
     function burnSynths(uint amount) external optionalProxy {
         return issuer().burnSynths(messageSender, amount);
     }
 
+    function burnSynthsOnBehalf(address burnForAddress, uint amount) external {
```
justinjmoses: not optionalProxy?

contracts/Issuer.sol
```diff
@@ -150,9 +167,18 @@ contract Issuer is MixinResolver {
         _internalBurnSynths(from, debtToRemoveAfterSettlement, existingDebt, totalSystemValue);
     }
 
+    function burnSynthsToTargetOnBehalf(address burnForAddress) external {
```
justinjmoses: onlySynthetix and missing an arg?

contracts/Exchanger.sol
hav-noms: for exchange sip-14

contracts/Synthetix.sol
hav-noms: Yes good catch thanks

justinjmoses:
@Clinton regarding the test flakes - I ran this 100 separate times locally and couldn't reproduce it. That being said, calculating the interest rate down to the second and expecting ganache's timestamps to match your expected probably isn't the best approach - can you instead request the time before and after the fastForward and use that amount of time to calculate the expected interest?

jacko125:
@justinjmoses  Makes sense to use the time before and after from ganache to calculate expected interest instead of relying up to the second. I can update those tests.

test/contracts/DelegateApprovals.js
```diff
@@ -1,9 +1,11 @@
 const DelegateApprovals = artifacts.require('DelegateApprovals');
+const { onlyGivenAddressCanInvoke } = require('../utils/setupUtils');
+const { toBytes32 } = require('../../.');
```
justinjmoses: @hav-noms can you add the only mutative function expectation checker to this test suite?

test/contracts/Issuer.js
```diff
@@ -1531,4 +1538,107 @@ contract('Issuer (via Synthetix)', async accounts => {
...
+				await assert.revert(synthetix.burnSynthsToTargetOnBehalf(authoriser, { from: delegate }));
+			});
+		});
+
+		it('should approveIssueOnBehalf for account1', async () => {
+			const authoriser = account1;
+			const delegate = account2;
+
+			await delegateApprovals.approveIssueOnBehalf(delegate, { from: authoriser });
+			const result = await delegateApprovals.canIssueFor(authoriser, delegate);
```
justinjmoses: this is the only test for canIssueFor - we need to test the failure case as well

test/contracts/Exchanger.js
```diff
@@ -100,7 +100,13 @@ contract('Exchanger (via Synthetix)', async accounts => {
...
 			abi: exchanger.abi,
 			ignoreParents: ['MixinResolver'],
-			expected: ['settle', 'setExchangeEnabled', 'setWaitingPeriodSecs', 'exchange'],
+			expected: [
+				'settle',
+				'setExchangeEnabled',
+				'setWaitingPeriodSecs',
+				'exchange',
+				'exchangeOnBehalf',
+			],
```
justinjmoses: we need a test for both the success and failure cases of `exchangeOnBehalf`

test/contracts/Issuer.js
justinjmoses: right ok i see the reverts on the failure cases for issue and burn above

test/contracts/Exchanger.js
jacko125: Correct, been added now.

PR #448, opened 2020-03-05
SIP-47 Ensure only exchanges with settlements adjust amounts down

justinjmoses:
This fix prevents the known side-effect of SIP-37 whereby exchanges with invalid amounts are succeeding with 0 amounts. Instead they should `revert` unless a settlement was required, in which case they should continue to pass (and thereby emptying out the settlement queue).

E.g. https://etherscan.io/tx/0x9a92ca42aa216b1cb406e1b09c50062b90ec2bfb5f10e584cc19978c5eece4a4 

For Exchanger.exchange, Issuer.burnSynths, Synth.transferAndSettle, Synth.transferFromAndSettle we will only reduce the amount down to the user's balance if there was any settlement. This means invalid amounts, instead of quietly succeeding as 0, will revert as they used to

### DIFF
--- a/contracts/Exchanger.sol
+++ b/contracts/Exchanger.sol
@@ -173,12 +173,12 @@ contract Exchanger is MixinResolver {
 
         uint sourceAmountAfterSettlement = sourceAmount;
 
-        // when the are entries that were settled
+        // when settlement was required
         if (numEntriesSettled > 0) {
             // ensure the sourceAmount takes this into account
             sourceAmountAfterSettlement = calculateAmountAfterSettlement(from, sourceCurrencyKey, sourceAmount, refunded);
 
-            // If, after settlement the user has no balance  left (highly unlikely), then return to prevent
+            // If, after settlement the user has no balance left (highly unlikely), then return to prevent
             // emitting events of 0 and don't revert so as to ensure the settlement queue is emptied
             if (sourceAmountAfterSettlement == 0) {
                 return 0;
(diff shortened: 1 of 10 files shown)


justinjmoses:
SIP-47

contracts/Exchanger.sol
```diff
@@ -169,20 +169,29 @@ contract Exchanger is MixinResolver {
...
 
-        (, uint refunded) = _internalSettle(from, sourceCurrencyKey);
+        (, uint refunded, uint numEntriesSettled) = _internalSettle(from, sourceCurrencyKey);
 
-        ISynthetix _synthetix = synthetix();
-        IExchangeRates _exRates = exchangeRates();
+        uint sourceAmountAfterSettlement = sourceAmount;
 
-        uint sourceAmountAfterSettlement = calculateAmountAfterSettlement(from, sourceCurrencyKey, sourceAmount, refunded);
+        // when the are entries that were settled
```
jacko125: comment typo ? `when there are entries that were settled`

contracts/Issuer.sol
```diff
@@ -138,14 +138,18 @@ contract Issuer is MixinResolver {
...
 
         // How much debt do they have?
         (uint existingDebt, uint totalSystemValue) = synthetix().debtBalanceOfAndTotalDebt(from, sUSD);
 
         require(existingDebt > 0, "No debt to forgive");
 
-        uint debtToRemoveAfterSettlement = exchanger().calculateAmountAfterSettlement(from, sUSD, amount, refunded);
+        uint debtToRemoveAfterSettlement = amount;
+
+        if (numEntriesSettled > 0) {
```
jacko125: Interesting, this check would save some gas instead of always calling `exchanger().calculateAmountAfterSettlement`  when burning.

PR #449, opened 2020-03-06
Allow all views to return when rates are stale

justinjmoses:
This will allow all contract views to still return with no reverts, even when rates are stale. The onus is now on the mutative functions to prevent actions during stale rates rather than breaking the entire system for any apps reading state.

**Note:** This requires updating `EtherCollateral` which won't be upgraded until the 3m trial ends.


justinjmoses:
Closing in favor of a draft PR

PR #450, opened 2020-03-06
Adding new header with ascii art and doc links

justinjmoses:
Ascii art via: http://patorjk.com/software/taag/#p=display&h=1&f=Small%20Slant&t=Synthetix

Note: the doubled `\` are necessary to escape them in text

Adds a heading like the below for a contract with an associated Proxy

```javascript
/*

⚠⚠⚠ WARNING WARNING WARNING ⚠⚠⚠

This is a TARGET contract - DO NOT CONNECT TO IT DIRECTLY IN YOUR CONTRACTS or DAPPS!

This contract has associated PROXY that MUST be used for all integrations - this TARGET will be REPLACED in an upcoming Synthetix release.
The proxy for this contract can be found here:

https://contracts.synthetix.io/ProxyFeePool

*/
/*
   ____            __   __        __   _
  / __/__ __ ___  / /_ / /  ___  / /_ (_)__ __
 _\ \ / // // _ \/ __// _ \/ -_)/ __// / \ \ /
/___/ \_, //_//_/\__//_//_/\__/ \__//_/ /_\_\
     /___/

* Synthetix: FeePool.sol
*
* Latest source (may be newer): https://github.com/Synthetixio/synthetix/blob/master/contracts/FeePool.sol
* Docs: https://docs.synthetix.io/contracts/FeePool
*
* Contract Dependencies: 
*	- EternalStorage
*	- ExternStateToken
*	- LimitedSetup
*	- MixinResolver
*	- Owned
*	- Proxyable
*	- SelfDestructible
*	- State
* Libraries: 
*	- Math
*	- SafeDecimalMath
*	- SafeMath
*
*/
```

### DIFF
--- a/contracts/AddressResolver.sol
+++ b/contracts/AddressResolver.sol
@@ -3,6 +3,7 @@ pragma solidity 0.4.25;
 import "./Owned.sol";
 
 
+// https://docs.synthetix.io/contracts/AddressResolver
 contract AddressResolver is Owned {
     mapping(bytes32 => address) public repository;
 
--- a/contracts/ArbRewarder.sol
+++ b/contracts/ArbRewarder.sol
@@ -7,6 +7,7 @@ import "./interfaces/IERC20.sol";
 import "./interfaces/IExchangeRates.sol";
 
 
+// https://docs.synthetix.io/contracts/ArbRewarder
 contract ArbRewarder is SelfDestructible, Pausable {
     using SafeMath for uint;
     using SafeDecimalMath for uint;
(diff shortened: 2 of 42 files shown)


hav-noms:
This sample looks good but one thing I really want to add is a massive warning on underlying Targets e.g. FeePool to not connect / use this address and we could even read in and print the proxy address they should connect to or use the AddressResolver with a link to a how to integrate Synthetix link in the docs.

jacko125:
Agree with Clinton, think what prompted this change was to help others know that for `Proxyable` contracts they should point to the proxy or proxyERC20 instead of underlying contract.

justinjmoses:
OK - I added the feature to the verifier - that's where we actually know which contract's source we are verifying. Note however that Etherscan tends to reuse source code for ABI verification, so perhaps I shouldn't link to the proxy contract in case it's the wrong one for the synth?

PR #451, opened 2020-03-06
SIP-48 Allow read all views even when rates stale

justinjmoses:
https://sips.synthetix.io/sips/sip-48

This will allow all contract views to still return with no reverts, even when rates are stale. The onus is now on the mutative functions to prevent actions during stale rates rather than breaking the entire system for any apps reading state.


hav-noms:
@justinjmoses as you mentioned its probably a safe place to also prevent issuing while the system is down. There are a few places with access to issuing now 
        bool isSynthetix 
        bool isFeePool 
        bool isExchanger 
        bool isIssuer 
including ETHCollateral.

contracts/Issuer.sol
```diff
@@ -119,7 +122,10 @@ contract Issuer is MixinResolver {
         _internalIssueSynths(from, maxIssuable, existingDebt, totalSystemDebt);
     }
 
-    function _internalIssueSynths(address from, uint amount, uint existingDebt, uint totalSystemDebt) internal {
+    function _internalIssueSynths(address from, uint amount, uint existingDebt, uint totalSystemDebt)
+        internal
+        snxRateNotStale
```
jacko125: @justinjmoses Checking snxRateNotStale for issue / burning would need to take into consideration that `all synths` need update to date rates too as `totalIssuedSynths` is used to get the `totalSystemDebt`

contracts/Synthetix.sol
```diff
@@ -127,8 +127,7 @@ contract Synthetix is ExternStateToken, MixinResolver {
         uint total = 0;
         uint currencyRate = exRates.rateForCurrency(currencyKey);
 
-        (uint[] memory rates, bool anyRateStale) = exRates.ratesAndStaleForCurrencies(availableCurrencyKeys());
-        require(!anyRateStale, "Rates are stale");
+        (uint[] memory rates, ) = exRates.ratesAndStaleForCurrencies(availableCurrencyKeys());
```
jacko125: @justinjmoses  Note that we will need to stop issue / burning of debt another way as we won't have a non-stale view of  `totalIssuedSynths` value once we strip out the  `require(!anyRateStale, "Rates are stale");`  here.  

Issuing and burning debt relies on knowing the total debt which is also the totalIssuedSynths

hav-noms:
LGTM. 
RE: EtherColatteral I think it safest to alter Synth to require !anyRateIsState to allow issuing and burning. You may be able to remove the check from the issuer then rather than do it twice at the Issuer and the Synth

contracts/Synthetix.sol
```diff
@@ -447,14 +447,8 @@ contract Synthetix is ExternStateToken, MixinResolver {
...
      * in this calculation.
-     * @notice SNX rate not stale is checked within debtBalanceOf
      */
-    function transferableSynthetix(address account)
-        public
-        view
-        rateNotStale("SNX") // SNX is not a synth so is not checked in totalIssuedSynths
-        returns (uint)
-    {
+    function transferableSynthetix(address account) public view returns (uint) {
```
jacko125: note -> transferableSynthetix() should check `debtBalanceOf` is not stale.

contracts/Synthetix.sol
```diff
@@ -267,7 +270,7 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         address from,
         address to,
         uint value
-    ) public optionalProxy returns (bool) {
+    ) public optionalProxy snxRateNotStale returns (bool) {
         systemStatus().requireSystemActive();
```
maxsam4, line 275: Do the `issueSynths` and `issueSynthsOnBehalf` require staleness checks?

contracts/Exchanger.sol
```diff
@@ -309,12 +313,12 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
             uint numEntriesSettled
         )
     {
-        // Note: this function can be called by anyone on behalf of anyone else
-
         systemStatus().requireExchangeActive();
 
         systemStatus().requireSynthActive(currencyKey);
 
+        require(!exchangeRates().rateIsStale(currencyKey), "Rate stale or not found");
```
jacko125, line 320: Given that `waitingPeriod` is 3mins and settlement looks at historical prices / or any price updates within `n` minutes, not sure if we need to check rate is stale or not here ? 

Could cause issue like MKR where we have to remove a synth / stop pushing rates but then people can't settle it anymore.

contracts/Synthetix.sol
```diff
@@ -134,8 +134,7 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         uint total = 0;
         uint currencyRate = exRates.rateForCurrency(currencyKey);
 
-        (uint[] memory rates, bool anyRateStale) = exRates.ratesAndStaleForCurrencies(availableCurrencyKeys());
-        require(!anyRateStale, "Rates are stale");
+        uint[] memory rates = exRates.ratesForCurrencies(availableCurrencyKeys());
```
jacko125, line 137: @justinjmoses  Note - we need to be careful here that if any rate is stale that `issuing` and `burning` any sUSD or debt should not be allowed as the totalIssuedSynth value needs to be not stale.

contracts/Synthetix.sol
```diff
@@ -134,8 +134,7 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         uint total = 0;
         uint currencyRate = exRates.rateForCurrency(currencyKey);
```
jacko125, line 135: `rateForCurrency` here would've thrown when `SNX` or the price the debt is priced in is stale.

contracts/Synthetix.sol
justinjmoses: yep - see my note in the desc of the PR. 

although `transfer(From)?` SNX - should it be prevented if rates are stale? It should as if anything is stale we can't figure out the correct size of debt pool, hence we can't figure out exactly how much SNX you need to have staked as collateral for your debt position

contracts/Exchanger.sol
justinjmoses: Really great pickup jacko. I'm removing this requirement.

justinjmoses: This will also fix that bug for old MKR holders who were purged and haven't yet settled.

contracts/Synthetix.sol
justinjmoses: They are above in `Issuer.sol` in the internal issue and burn functions

contracts/Synthetix.sol
```diff
@@ -118,6 +118,25 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
             IRewardsDistribution(requireAndGetAddress(CONTRACT_REWARDSDISTRIBUTION, "Missing RewardsDistribution address"));
     }
 
+    function _anySynthOrSNXRateIsStale() internal view returns (bool anyRateStale) {
+        bytes32[] memory currencyKeys = availableCurrencyKeys();
```
justinjmoses, line 122: @maxsam4 how expensive is doing this? As it's memory I don't imagine it's too costly, though perhaps there's a better way?

contracts/Synthetix.sol
```diff
@@ -118,6 +118,25 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+
+        bytes32[] memory currencyKeysWithSNX = new bytes32[](availableSynths.length + 1);
+
+        currencyKeys[availableSynths.length] = "SNX";
```
justinjmoses, line 126: how expensive is doing this? As it's memory I don't imagine it's too costly, though perhaps there's a better way?

contracts/Synthetix.sol
justinjmoses: Actually I changed my mind and added them here

contracts/Synthetix.sol
```diff
@@ -271,49 +293,49 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         return _transferFromByProxy(messageSender, from, to, value);
     }
 
-    function issueSynths(uint amount) external optionalProxy {
+    function issueSynths(uint amount) external noSynthOrSynthetixRateStale optionalProxy {
```
justinjmoses, line 296: Note: I decided for ease of reasoning to put these checks in here in `Synthetix` rather than underneath in `Issuer`. It's a bit more code this way, but no extra gas costs and feels like the tradeoff for readability is worthwhile. Thoughts?

contracts/FeePool.sol
```diff
@@ -319,7 +316,7 @@ contract FeePool is Owned, Proxyable, SelfDestructible, LimitedSetup, MixinResol
     /**
      * @notice Claim fees for last period when available or not already withdrawn.
      */
-    function claimFees() external optionalProxy returns (bool) {
+    function claimFees() external issuanceActive noSynthOrSynthetixRateStale optionalProxy returns (bool) {
         return _claimFees(messageSender);
```
jacko125, line 320: same modifiers are applied to `claimFees` and `claimOnBehalf` , if set at `_claimFees` ensures only one spot required to upgrade in future to sync both func checks.

contracts/Synthetix.sol
```diff
@@ -583,13 +518,34 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+        systemStatus().requireIssuanceActive();
+        _;
+    }
+
+    modifier noSynthOrSynthetixRateStale() {
+        require(!_anySynthOrSNXRateIsStale(), "A synth or SNX rate is stale");
+        _;
+    }
+
+    modifier exchangeActive(bytes32 src, bytes32 dest) {
```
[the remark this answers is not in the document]
jacko125: Synthetix contract getting close to full should consider this modifier as internal function.

contracts/FeePool.sol
justinjmoses: yes but this will surface clearly in the docs - seems a worthy tradeoff if we don't mind a slight impact to contract size here

contracts/Synthetix.sol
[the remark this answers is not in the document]
justinjmoses: Yeah but I'm trying to surface the requirements on our public functions so we have them clearly in the docs. If it's a trade-off between contract size limits and clearer code, I'd rather the latter and find ways to reduce the limit. 

One consideration is moving `issue*` / `burn*` to Issuer directly, and `exchange*` to exchanger, then third parties need to:

```solidity
IIssuer snxIssuer = snx.getAddress("Issuer");
snx.issueMaxSynths();
IExchanger snxExchanger = snx.getAddress("Exchanger");
snxExchanger.exchange(...);
```

contracts/FeePool.sol
```diff
@@ -807,6 +807,16 @@ contract FeePool is Owned, Proxyable, SelfDestructible, LimitedSetup, MixinResol
         _;
     }
 
+    modifier issuanceActive() {
+        systemStatus().requireIssuanceActive();
+        _;
+    }
+
+    modifier noSynthOrSynthetixRateStale() {
+        require(!synthetix().anySynthOrSNXRateIsStale(), "At least one synth or SNX rate is stale");
```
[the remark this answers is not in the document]
justinjmoses: Where are you seeing this? 

image

contracts/FeePool.sol
```diff
@@ -278,11 +277,9 @@ contract FeePool is Owned, Proxyable, SelfDestructible, LimitedSetup, MixinResol
     /**
      * @notice Close the current fee period and start a new one.
      */
-    function closeCurrentFeePeriod() external {
+    function closeCurrentFeePeriod() external issuanceActive {
```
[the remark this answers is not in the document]
justinjmoses: It was part of SIP-44 - can't close period while issuance is inactive - i'm just moving it up from a line of code into a modifier

contracts/FeePool.sol
justinjmoses: Yeah but I'm doing it to surface in the docs

jacko125: More of a concern for me is maintaining the code for `claimFees` and `claimFeesOnBehalf` overtime, if the spec for checks change for claiming, developer needs to remember to update both to match.

justinjmoses: That's why we have unit test coverage - we can refactor at will and rest assured it will be checked. Reduce human error by automating all the things!

https://media.giphy.com/media/d30rnjrGa9H1ceGc/giphy.gif

contracts/EtherCollateral.sol
```diff
@@ -447,6 +455,17 @@ contract EtherCollateral is Owned, Pausable, ReentrancyGuard, MixinResolver, IEt
...
     }
 
+    function exchangeRates() internal view returns (IExchangeRates) {
+        return IExchangeRates(requireAndGetAddress(CONTRACT_EXRATES, "Missing ExchangeRates address"));
+    }
+
+    /* ========== MODIFIERS ========== */
+
+    modifier sETHRateNotStale() {
+        require(!exchangeRates().rateIsStale("sETH"), "Cannot use while sETH rate is stale");
```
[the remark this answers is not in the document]
justinjmoses: Never heard of the guy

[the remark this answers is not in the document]
justinjmoses: https://github.com/Synthetixio/synthetix/pull/451/files#diff-733e929dd68f104f2b18041033bb427eR298
https://github.com/Synthetixio/synthetix/pull/451/files#diff-733e929dd68f104f2b18041033bb427eR304 
https://github.com/Synthetixio/synthetix/pull/451/files#diff-733e929dd68f104f2b18041033bb427eR310

😬😬😬

contracts/FeePool.sol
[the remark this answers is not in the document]
hav-noms: https://sips.synthetix.io/sips/sip-44 does not reason why closeCurrentFeePeriod() should not be called when issuanceActive. I'm trying to understand why closeCurrentFeePeriod() is blocked when !issuanceActive

contracts/FeePool.sol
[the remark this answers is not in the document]
hav-noms: Sorry 'maxed out' was a poor choice or wording. We'r mindful we're running out of room there and to keep it trim

contracts/FeePool.sol
hav-noms: My understanding was Kendrick could traverse the call stack surfacing all requires

contracts/FeePool.sol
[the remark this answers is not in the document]
hav-noms: Ah I forgot, this came to me later. We've been push FeePool to the deployment limit with optimize runs making it fatter but cheaper for users https://github.com/Synthetixio/synthetix/blob/master/publish/src/contract-overrides.js#L8

justinjmoses:
* Synthetix.exchange
     * develop: 470k
     * sip-48: 471k
     * this: 471k
* Synthetix.issue & burn
    * develop: 376k & 206k
    * sip-48: 406k, 242k
    * this: 437k & 269k
* FeePool.claimFees
    * develop: 290k
    * sip-48: 321k
    * this: 326k

justinjmoses:
develop
image

buidler.config.js
```diff
@@ -117,22 +117,34 @@ task('test:legacy', 'run the tests with legacy components')
 		await bre.run('test', taskArguments);
 	});
 
+const optimizeIfRequired = ({ bre, taskArguments: { optimizer } }) => {
```
justinjmoses, line 120: this is from: https://github.com/Synthetixio/synthetix/pull/533

buidler.config.js
```diff
@@ -165,6 +177,30 @@ task('compile')
 		}
 	});
 
+task('test')
```
justinjmoses, line 180: as above, please see #533

contracts/FeePool.sol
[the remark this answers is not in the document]
justinjmoses: It's true - I believe it was a decision at implementation - should we allow the closing of the fee period when issuance is disabled. And as you can see from L284 below, it is already in. Shall we remove it? We should discuss.

contracts/Proxy.sol
```diff
@@ -54,7 +54,8 @@ contract Proxy is Owned {
         }
     }
 
-    function() external payable {        
+    // solhint-disable no-complex-fallback
```
justinjmoses, line 57: From #533

justinjmoses:
Gas reports are back! 🤤🤤🤤🤤🤤🤤🤤 

Now we can see the impact of this PR against `develop`: https://github.com/Synthetixio/synthetix/pull/451/checks?check_run_id=717531407

contracts/EtherCollateral.sol
```diff
@@ -16,6 +16,7 @@ import "./interfaces/IFeePool.sol";
 import "./interfaces/ISynth.sol";
 import "./interfaces/IERC20.sol";
 import "./interfaces/IDepot.sol";
+import "./interfaces/IExchangeRates.sol";
```
justinjmoses, line 19: Here's where I'm doing something a bit atypical.

contracts/Issuer.sol
```diff
@@ -33,6 +37,11 @@ contract Issuer is Owned, MixinResolver, IIssuer {
 
     uint public minimumStakeTime = 24 hours; // default minimum waiting period after issuing synths
 
+    // Available Synths which can be used with the system
+    ISynth[] public availableSynths;
+    mapping(bytes32 => ISynth) public synths;
```
justinjmoses, line 42: I actually think this is best replaced in *Acrux* by `FlexibleContractStorage` (FCP), that way upgrades to Issuer will not require re-adding all Synths. That being said, I propose we leave this here in SIP-48 and upgrade it in the FCP SIP

justinjmoses: https://github.com/Synthetixio/SIPs/pull/148

contracts/Issuer.sol
```diff
@@ -93,17 +112,285 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+            // minus total issued synths from Ether Collateral from sETH.totalSupply()
+            if (excludeEtherCollateral && synth == "sETH") {
+                totalSynths = totalSynths.sub(etherCollateral().totalIssuedSynths());
+            }
+
+            uint synthValue = totalSynths.multiplyDecimalRound(rates[i]);
+            total = total.add(synthValue);
+        }
+
+        if (currencyRate == 0 && currencyKey == "SNX") {
```
zyzek, line 179: `currencyKey == "SNX"` already implies `currencyRate == 0` since the loop does not go as far as the SNX entry; there is no need to check the `currencyRate` here.

contracts/Issuer.sol
```diff
@@ -93,17 +112,285 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+    function _lastIssueEvent(address account) internal view returns (uint) {
         //  Get the timestamp of the last issue this account made
         return issuanceEternalStorage().getUIntValue(keccak256(abi.encodePacked(LAST_ISSUE_EVENT, account)));
     }
 
+    function _remainingIssuableSynths(address _issuer)
+        internal
+        view
+        returns (
+            // Don't need to check for synth existing or stale rates because maxIssuableSynths will do it for us.
```
zyzek, line 243: Comment is slightly off w.r.t. staleness and which function is checking what.

contracts/Issuer.sol
```diff
@@ -93,17 +112,285 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+        if (initialDebtOwnership == 0) return 0;
+
+        (debtBalance, , ) = _debtBalanceOfAndTotalDebt(_issuer, currencyKey);
+    }
+
+    function remainingIssuableSynths(address _issuer)
+        external
+        view
+        returns (
+            // Don't need to check for synth existing or stale rates because maxIssuableSynths will do it for us.
```
zyzek, line 351: Comment out of date.

contracts/Issuer.sol
```diff
@@ -93,17 +112,285 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+            uint maxIssuable,
+            uint alreadyIssued,
+            uint totalSystemDebt
+        )
+    {
+        (maxIssuable, alreadyIssued, totalSystemDebt, ) = _remainingIssuableSynths(_issuer);
+    }
+
+    function maxIssuableSynths(address _issuer) external view returns (uint) {
```
zyzek, line 360: This just duplicates the body of `_maxIssuableSynths`, should it not make a call to the internal function instead?

contracts/Issuer.sol
```diff
@@ -93,17 +112,285 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+
+    function transferableSynthetixAndAnyRateIsStale(address account, uint balance)
+        external
+        view
+        returns (uint transferable, bool anyRateIsStale)
+    {
+        // How many SNX do they have, excluding escrow?
+        // Note: We're excluding escrow here because we're interested in their transferable amount
+        // and escrowed SNX are not transferable.
+        // uint balance = synthetixERC20().balanceOf(account);
```
zyzek, line 376: I'm not sure if it's actually instructive to keep this `uint balance = synthetixERC20().balanceOf(account);` line around or not.

contracts/Issuer.sol
```diff
@@ -114,9 +401,54 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+
+        require(synths[currencyKey] == ISynth(0), "Synth already exists");
+        require(synthsByAddress[address(synth)] == bytes32(0), "Synth address already exists");
+
+        availableSynths.push(synth);
+        synths[currencyKey] = synth;
+        synthsByAddress[address(synth)] = currencyKey;
+    }
+
+    /**
```
zyzek, line 416: Kill the docstring.

contracts/Issuer.sol
```diff
@@ -114,9 +401,54 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+     * @notice Remove an associated Synth contract from the Synthetix system
+     * @dev Only the contract owner may call this.
+     */
+    function removeSynth(bytes32 currencyKey) external onlyOwner {
+        require(address(synths[currencyKey]) != address(0), "Synth does not exist");
+        require(IERC20(address(synths[currencyKey])).totalSupply() == 0, "Synth supply exists");
+        require(currencyKey != sUSD, "Cannot remove synth");
+
+        // Save the address we're removing for emitting the event at the end.
```
zyzek, line 425: No such event is emitted.

hav-noms:
LGTM lets get this in. Could you take a look at getting recovering coverage where it has dropped or even to 100

contracts/Issuer.sol
```diff
@@ -93,17 +112,285 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+            uint synthValue = totalSynths.multiplyDecimalRound(rates[i]);
+            total = total.add(synthValue);
+        }
+
+        if (currencyRate == 0 && currencyKey == "SNX") {
+            // if no rate while iterating through synths, then try SNX
+            currencyRate = rates[synthsAndSNX.length - 1];
+        } else if (currencyRate == 0) {
+            // and, in an edge case where the requested rate isn't a synth or SNX, then do the lookup
+            currencyRate = exchangeRates().rateForCurrency(currencyKey);
```
jacko125, line 184: In what scenario will the system allow an input for currencyKey in `_totalIssuedSynths` that is not a synths or SNX ? 

Might be better to return 0 in that case where user input is invalid / not supported, as `exchangeRates().rateForCurrency` would return a 0 anyway.

contracts/Issuer.sol
```diff
@@ -298,17 +650,19 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
 
         // What is the value of their SNX balance in sUSD?
-        uint collateralValue = exchangeRates().effectiveValue("SNX", collateral, sUSD);
+        uint collateralValue = exchangeRates().effectiveValue("SNX", collateralForAccount, sUSD);
 
         // What is their debt in sUSD?
-        (uint debtBalance, uint totalDebtIssued) = synthetix().debtBalanceOfAndTotalDebt(account, sUSD);
+        (uint debtBalance, uint totalDebtIssued, bool anyRateIsStale) = _debtBalanceOfAndTotalDebt(account, sUSD);
+
+        require(!anyRateIsStale, "A synth or SNX rate is stale");
```
jacko125, line 665: Can remove the rateNotStale() modifier in `synthetix.liquidateDelinquentAccount` if its checking here.

contracts/FeePool.sol
justinjmoses: Yeah it definitely does not surface requires nested in internal function calls.

contracts/Issuer.sol
```diff
@@ -93,17 +112,285 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+            currencyKeys[availableSynths.length] = "SNX";
+        }
 
-    function canBurnSynths(address account) public view returns (bool) {
-        return now >= lastIssueEvent(account).add(minimumStakeTime);
+        return currencyKeys;
     }
 
-    function lastIssueEvent(address account) public view returns (uint) {
+    function _anySynthOrSNXRateIsStale() internal view returns (bool anyRateStale) {
```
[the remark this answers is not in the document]
justinjmoses: Done.

contracts/Issuer.sol
```diff
@@ -93,17 +112,285 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+        uint total = 0;
+        uint currencyRate;
+
+        bytes32[] memory synthsAndSNX = _availableCurrencyKeysWithOptionalSNX(true);
+
+        (uint[] memory rates, bool anyRateStale) = exchangeRates().ratesAndStaleForCurrencies(synthsAndSNX);
+
+        for (uint i = 0; i < synthsAndSNX.length - 1; i++) {
+            // What's the total issued value of that synth in the destination currency?
+            // Note: We're not using exchangeRates().effectiveValue() because we don't want to go get the
```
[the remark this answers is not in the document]
justinjmoses: Adjusted

contracts/Issuer.sol
justinjmoses: I was just thinking the same thing after fixing the above comment :)

contracts/Issuer.sol
justinjmoses: It could be `ETH` - highly unlikely but possible. I'm reluctant to revert in views unless we strictly think this is a necessary restriction here?

contracts/Issuer.sol
justinjmoses: Fixed

contracts/Issuer.sol
```diff
@@ -93,17 +112,285 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+
+        if (alreadyIssued >= maxIssuable) {
+            maxIssuable = 0;
+        } else {
+            maxIssuable = maxIssuable.sub(alreadyIssued);
+        }
+    }
+
+    function _maxIssuableSynths(address _issuer) internal view returns (uint) {
+        // What is the value of their SNX balance in the destination currency?
```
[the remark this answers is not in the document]
justinjmoses: Fixed

contracts/Issuer.sol
justinjmoses: Fixed

contracts/Issuer.sol
justinjmoses: Yep good catch. Fixed.

contracts/Issuer.sol
justinjmoses: I think not. Removed.

contracts/Issuer.sol
justinjmoses: Done

contracts/Issuer.sol
```diff
@@ -114,9 +401,54 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+
+                break;
+            }
+        }
+
+        // And remove it from the synths mapping
+        delete synthsByAddress[address(synths[currencyKey])];
+        delete synths[currencyKey];
+
+        // Note: No event here as Synthetix contract exceeds max contract size
```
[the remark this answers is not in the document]
justinjmoses: Added.

contracts/Issuer.sol
justinjmoses: It is now.

contracts/Issuer.sol
justinjmoses: Yes I already did https://github.com/Synthetixio/synthetix/pull/451/files#diff-6f16722eb1b8986e839fa0ffd844709aL584

contracts/Synthetix.sol
```diff
@@ -581,10 +299,14 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
     }
 
     // totalSupply effectiveValue checks for rates stale
-    function liquidateDelinquentAccount(address account, uint susdAmount) external rateNotStale("SNX") optionalProxy returns (bool) {
+    function liquidateDelinquentAccount(address account, uint susdAmount) external optionalProxy returns (bool) {
         systemStatus().requireSystemActive();
```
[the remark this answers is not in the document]
justinjmoses: Good spot, done.

justinjmoses:
All comments addressed. RFAL.

contracts/FeePool.sol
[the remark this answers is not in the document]
justinjmoses: Will create a separate PR to make this system instead.

hav-noms:
LGTM resolved all my feedback

contracts/Issuer.sol
jacko125: Thanks!

PR #452, opened 2020-03-06
Upgrading solidity-coverage and using new plugin format

justinjmoses:
Fixes coverage reports (thanks to @cgewecke) and adds codecov.io reports via travis

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,4 +29,5 @@ keys.json
 .vscode/launch.json
 
 .DS_Store
-
+coverage.json
+coverage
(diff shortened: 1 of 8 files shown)


justinjmoses:
@cgewecke do you happen to have any ideas on the above error with the linker?

cgewecke:
@justinjmoses Will take a look...

cgewecke:
@justinjmoses #453 has some possible fixes....

justinjmoses:
You're a rockstar @cgewecke 🚀. Do you have an address for donations?

cgewecke:
@justinjmoses Oh that's really sweet of you - solidity-coverage 0.7.x was funded by an EF grant and getting it to run on some of the larger, more widely used projects was part of the initial proposal. It's received all the donations it needs, thanks though :)

PR #453, opened 2020-03-07
Fix coverage migrations

cgewecke:
[Edited]

**PR targets #452**

It looks like technically [deployer.link is async][1] - not sure why it's being weird while running in the plugin but awaiting the links seems to make it work.

Other changes were necessary as well...leaving some notes in the diff.

Please feel free to close if this contains changes you're not interested in - just opened to make sure it's possible to get it working.

[1]: https://github.com/trufflesuite/truffle/blob/develop/packages/deployer/src/actions/link.js#L3-L7

### DIFF
--- a/.solcover.js
+++ b/.solcover.js
@@ -1,3 +1,8 @@
 module.exports = {
-	skipFiles: ['test-helpers/PublicSafeDecimalMath.sol', 'test-helpers/PublicMath.sol'],
+  skipFiles: ['test-helpers/PublicSafeDecimalMath.sol', 'test-helpers/PublicMath.sol'],
+  client: require('ganache-cli'), // use ganache-cli version listed in dev deps
+  providerOptions: {
+    default_balance_ether: 10000000000000, // extra zero just in case (coverage consumes more gas)
+    time: new Date("2019-03-06T00:00:00"),
+  }
 };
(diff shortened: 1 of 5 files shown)


.solcover.js
```diff
@@ -1,3 +1,14 @@
...
+  client: require('ganache-cli'), // use ganache-cli version listed in dev deps
+  providerOptions: {
+    default_balance_ether: 10000000000000, // extra zero just in case (coverage consumes more gas)
+    time: new Date("2019-03-06T00:00:00"),
+    network_id: 55
+  },
+  mocha: {
+    grep: "@cov-skip", // Find everything with this tag
+    invert: true       // Run the grep's inverse set.
+  }
```
cgewecke: + Uses similar ganache settings to the one's defined in package.json for `test`
+ Runs on port 8545 / network_id 55 - was getting an error in Synthetix:mint about network "*" not being defined for `RewardEscrow` and this resolved it. Idk.
+ Defines a tag that lets mocha skip tests that check for exact account balances. Coverage distorts gas consumption and there were a couple `Depot` tests that failed because the numbers weren't right. (Skipping the test entirely might not be what you want though...)

migrations/2_deploy_synthetix_system.js
```diff
@@ -228,7 +229,7 @@ module.exports = async function(deployer, network, accounts) {
 		resolver.address,
 		{
 			from: deployerAccount,
-			gas: 8000000,
+			gas: block.gasLimit,
```
cgewecke: Coverage uses a high block gas limit. It instruments the contracts with additional Solidity statements so they're more expensive to deploy.

justinjmoses:
Wow - thanks so much Chris. This all looks solid. I'm going to merge it in and see

PR #455, opened 2020-03-10
Deprecate sip21

hav-noms:


contracts/Exchanger.sol
```diff
@@ -61,19 +61,7 @@ contract Exchanger is MixinResolver {
         // Get the base exchange fee rate
         uint exchangeFeeRate = feePool().exchangeFeeRate();
 
-        uint multiplier = 1;
```
justinjmoses: Qu: do we still need this function at all then? Or should we just always use it from `FeePool`?

test/contracts/Exchanger.js
```diff
@@ -1342,7 +1324,7 @@ contract('Exchanger (via Synthetix)', async accounts => {
 											from: iBTC,
 											to: sAUD,
 											toContract: sAUDContract,
-											exchangeFeeRateMultiplier: 2,
+											exchangeFeeRateMultiplier: 1,
```
justinjmoses: You can safely remove this prop

contracts/Exchanger.sol
hav-noms: We might need this for differential fees

justinjmoses: Gotcha

PR #459, opened 2020-03-16
update purge synths to batches and output encodeABI for protocolDAO

jacko125:


publish/src/commands/purge-synths.js
```diff
@@ -145,17 +148,29 @@ const purgeSynths = async ({
...
+			await performTransactionalStep({
+				account,
+				contract: `Synth${currencyKey}`,
+				target: Synth,
+				write: 'purge',
+				writeArg: [entries], // explicitly pass array of args so array not splat as params
+				gasLimit,
+				gasPrice,
+				etherscanLinkPrefix,
+				encodeABI: true,
```
justinjmoses: So `encodeABI` flag simply outputs the encoded data?

publish/src/util.js
```diff
@@ -197,6 +198,9 @@ const performTransactionalStep = async ({
 			appendOwnerAction(ownerAction);
 		}
 		return true;
+	} else if (encodeABI) {
```
justinjmoses: qu: shouldn't this be additional? i.e. shouldn't it be added to either the ownerActions or waiting for the user??

publish/src/commands/purge-synths.js
justinjmoses: Also, shouldn't this only be `true` when `network === 'mainnet'`?

jacko125: @justinjmoses  I think anything that requires the protocolDAO to action now. Yes will have a look at whether this can be appended to owner actions.

publish/src/util.js
jacko125: @justinjmoses  Good point, added to the ownerActions when its there. 

As for the Waiting for user , I want to make it grab the deployer wallet and once a user confirms it'll push to protocolDAO to stage it for other signers.

jacko125: Guess we hadn't thought through the deploy.js script to go to ProtocolDAO just yet.

publish/src/commands/purge-synths.js
```diff
@@ -143,7 +143,7 @@ const purgeSynths = async ({
 
 		if (Number(totalSupplyBefore) === 0) {
 			console.log(gray('Total supply is 0, exiting.'));
-			return;
+			continue;
```
justinjmoses: good catch

publish/src/commands/deploy.js
```diff
@@ -1013,10 +1014,10 @@ const deploy = async ({
 					console.log(gray(`Detected a new inverted synth for ${currencyKey}. Proceeding to add.`));
 					// Then a new inverted synth is being added (as there's no previous rate for it)
 					await setInversePricing({ freeze: false, freezeAtUpperLimit: false });
-				} else if (Number(totalSynthSupply) === 0) {
+				} else if (Number(totalSynthSupply) <= w3utils.toWei('10')) {
```
justinjmoses: is this really OK? 10 units of iBTC for example is not insignficant

publish/deployed/mainnet/synths.json
```diff
@@ -228,9 +228,9 @@
 		"desc": "Inverted Bitcoin",
 		"subclass": "PurgeableSynth",
 		"inverted": {
-			"entryPoint": 10600,
-			"upperLimit": 15900,
-			"lowerLimit": 5300
+			"entryPoint": 5275.00,
```
justinjmoses: [np] This should be separate from this PR

publish/src/util.js
```diff
@@ -183,10 +186,13 @@ const performTransactionalStep = async ({
 			etherscanLinkPrefix,
 		});
 
+		data = target.methods[write](...argumentsForWriteFunction).encodeABI();
```
justinjmoses: 👍 nice

publish/src/commands/deploy.js
jacko125: @justinjmoses  Agree on this point. Left in there yesterday from our run where iTRX (after purging had about 3500 tokes but 0 holders worth about $0.028 each) which was blocking. 

Are you happy if we move this number back to 0? 

A better check might be the totalSupply and also check the effectiveValue of that totalSupply here in the script?

publish/deployed/mainnet/synths.json
jacko125: Thanks for splitting this out to be released.

publish/src/commands/deploy.js
justinjmoses: > Are you happy if we move this number back to 0?

Yes. That will also fix the broken `test-publish` test

> A better check might be the totalSupply and also check the effectiveValue of that totalSupply here in the script?

You mean if the effectiveValue of the totalSupply in sUSD is below some threshold amount?

jacko125: yes

PR #465, opened 2020-03-20
Update synths.json

kaiynne:
Renaming inverted to inverse.


justinjmoses:
We will also need to PR these into `develop` separately.

justinjmoses:
Closed in favour of #467.

PR #466, opened 2020-03-20
Gnosis safe staging owner actions

jacko125:
Gnosis safe staging owner actions

TODO

- [x]  stage owner-actions.json as well to the protocol proxy. 

- [X] Check Gnosis safe API hasn't already got matching `to` and `data` and `!isExecuted`  of the preceding operation (otherwise confirm to skip).


publish/src/Gnosis-safe.abi.json
```diff
@@ -0,0 +1,585 @@
+[
```
justinjmoses: I really don't think we should be including this ABI in our repo. I think downloading this from Etherscan at deploy time is a reasonable approach. 

> As an aside, I have been thinking more and more of a better caching layer for ABI fetching. Seems silly to hit Etherscan's API for an immutable fetch. I've been thinking about creating an ABI service layer that uses Netlify for requests, GitHub as ABI storage, and Etherscan for the source of truth. Should be pretty simple to setup. Thoughts?

publish/src/commands/owner.js
```diff
@@ -48,22 +81,38 @@ const owner = async ({ network, newOwner, deploymentPath }) => {
...
+		gray(
+			'Skipping - Running through operations during deployment that couldnt complete as not owner.'
+		)
 	);
 
-	for (const [key, entry] of Object.entries(ownerActions)) {
-		const { action, link, complete } = entry;
-		if (complete) continue;
+	// new owner should be gnosis safe proxy address
+	const protocolDaoContract = getSafeInstance(web3, newOwner);
```
justinjmoses: I think you should consider moving the safe logic to the safe file entirely and keep this owner script clean

publish/src/commands/owner.js
```diff
@@ -48,22 +81,38 @@ const owner = async ({ network, newOwner, deploymentPath }) => {
...
+		yellow(`Using Protocol DAO Safe contract at ${protocolDaoContract.options.address} `)
+	);
 
-		await confirmOrEnd(
-			yellow('YOUR TASK: ') + `Invoke ${bgYellow(black(action))} (${key}) via ${cyan(link)}`
-		);
+	console.log(gray('Looking for contracts whose ownership we should accept'));
 
-		entry.complete = true;
-		fs.writeFileSync(ownerActionsFile, stringify(ownerActions));
```
justinjmoses: We should also remove the owner script entirely then - or else compile it during the deploy and remove it once all transactions are staged at the end of the deploy (to handle restarting deploys)

publish/src/safe-utils.js
```diff
@@ -12,6 +11,52 @@ const DELEGATE_CALL = 1;
...
+		outputs: [],
+		payable: false,
+		stateMutability: 'nonpayable',
+		type: 'function',
+	},
+	{
+		constant: true,
+		inputs: [],
+		name: 'nonce',
+		outputs: [{ internalType: 'uint256', name: '', type: 'uint256' }],
```
justinjmoses: all we need is `name` and `inputs` effectively

publish/src/commands/owner.js
```diff
@@ -81,46 +93,126 @@ const owner = async ({
...
+		});
 
-	// new owner should be gnosis safe proxy address
-	const protocolDaoContract = getSafeInstance(web3, newOwner);
-	// get protocolDAO nonce
-	const currentSafeNonce = await getSafeNonce(protocolDaoContract);
+		if (existingTx) {
+			console.log(
+				gray(
+					`Existing pending tx already submitted to gnosis safe - target address: ${target} and data: ${encodedData}`
```
justinjmoses: This logic should be moved over to the safe script - this is the owner script, safe-specific code should be elsewhere.

jacko125: Will have a look at refactoring more of this logic into safe-utils script.

PR #468, opened 2020-03-21
Deployment fixes for partial updates

justinjmoses:
* The current deploy script always updated a source's ABI even if not deployed. This is incorrect - it should only be updated when a deploy occurs. I reverted `Depot` and `EtherCollateral` in kovan's `deployment.json`
* Fixing this allowed me to continue with the deployment on kovan
* Had to explicitly add a few deps to the deploy script for `AddressResolver` for `Depot` and `EtherCollateral`  as these aren't being replaced on-chain yet we without an update we cannot lookup what they require.
* Ensuring `SynthetixEscrow` abi on `mainnet` `deployment.json` is actual ABI from old `HavvenEscrow`


publish/src/commands/deploy.js
```diff
@@ -325,10 +325,18 @@ const deploy = async ({
 			txn,
 			network,
 		};
-		deployment.sources[source] = {
-			bytecode: compiled[source].evm.bytecode.object,
-			abi: compiled[source].abi,
-		};
+		if (deployedContract.options.deployed) {
+			// track the new source and bytecode
+			deployment.sources[source] = {
```
jacko125: So this only updates the deployment.json abi / source bytecode if the contract was deployed ?

justinjmoses: correct, as it should

PR #469, opened 2020-03-22
Sip 44: System and synth disabling

justinjmoses:
### Four tiers of suspension
1. `System`
2. `Issuance` (also checks `System`)
3. `Exchange` (also checks `System`)
4. `Synth(bytes32)` (also checks `System`)

### Functions to restrict
- `Synthetix`
    - `issue*()`, `burn*()`
        - 👍require `Issuance` active
        - [x] Test
     - `transfer*()`
        - 👍require `System` active 
        - [x] Test
    - `settle()`
        - 👍require `Synth(currencyKey)` active
        - [x] Test
    - `exchange*()`
        - 👍require `Exchange` active
        - 👍require `Synth(src) && Synth(dest)` active
        - [x] Test
     - `mint()`
         - 👍require `Issuance` active
        - [x] Test
- `Synth`
    - `transfer*()`
         - 👍require `Synth(currencyKey)` active
        - [x] Test
- `Exchanger`
     - `settle()`
          - 👍require `Exchange` active
          - 👍require `Synth(currencyKey)` active
          - [x] Test
     - `exchange()`
        - 🚫 No check - used internally by `Synthetix.exchange()` (checked above), `Synth._transferToFeeAddress()` (which checks prior to invocation) and `PurgeableSynth.purge()` which can be run while system is suspended (as it's owner only)
- `EtherCollateral`
     - ` openLoan`
         - 👍require `Issuance` active
        - [x] Test
     - `closeLoan`
         - 👍require `Issuance` active
        - [x] Test
- `Depot`
     - 🚫 No check - will hit `Synth.transfer` or `Synthetix.transfer` checks
     - [x] Test
- `FeePool`
     - `claimFees*` 
         - 👍require `Issuance` active
         - [x] Test
     - `closeCurrentFeePeriod`
         - 👍require `Issuance` active
         - [x] Test

- [x] Remove `DappMaintenance` 
- [x] Remove `Exchanger.exchangeEnabled`


hav-noms:
@clementbalestrat & @evgenyboxer  I would like to deprecate DappMaintenance Contract and merge what functionality may be needed into this central system status contract. 
We should make sure the Dapps handle the events for 

- Synths being paused. Disable trading in the UI for those paused synths
- Exchange paused.
- Still need to keep prod down when system resumes for staging testing only keeping prod down.

contracts/SystemStatus.sol
```diff
@@ -0,0 +1,143 @@
...
+        emit IssuanceSuspended();
+    }
+
+    function resumeIssuance() external onlyOwner {
+        _requireAccessToResume(SECTION_ISSUANCE);
+        issuanceSuspended = false;
+        emit IssuanceResumed();
+    }
+
+    function suspendSynth(bytes32 currencyKey) external {
```
jacko125: @justinjmoses Could consider a bytes32 input of any `SuspensionCode` , ie,  `C01` for chainlink that is mapped by the dapps or other users on a table so we don't need the whole text on contract but a concise code that can be re-mapped for users. 

We wouldn't have to be describing many situations at this stage / or include the mapping or checks on the contract itself (ie as an ENUM). We can have a few generic codes that covers most things and when admin suspends a synth they can include the code.

contracts/Synth.sol
```diff
@@ -56,6 +64,8 @@ contract Synth is ExternStateToken, MixinResolver {
     function transfer(address to, uint value) public optionalProxy returns (bool) {
         _ensureCanTransfer(messageSender, value);
 
+        systemStatus().requireSynthActive(currencyKey);
```
jacko125: The check here for `requireSynthActive` could be moved into _ensureCanTransfer and cover `transfer()` + `transferFrom()`

justinjmoses: I'd considered that, then decided against it, but now I'm being swayed back 🤔

contracts/Exchanger.sol
```diff
@@ -269,6 +275,8 @@ contract Exchanger is MixinResolver {
     {
         // Note: this function can be called by anyone on behalf of anyone else
 
+        systemStatus().requireSynthActive(currencyKey);
```
jacko125: Note - Exchanging seems to be missing the requireSynthActive() check

justinjmoses: Jackson - it's in Synthetix: https://github.com/Synthetixio/synthetix/pull/469/files#diff-6f16722eb1b8986e839fa0ffd844709aR332

jacko125: I see it now,  had to expand the diff

contracts/Synthetix.sol
```diff
@@ -326,6 +357,8 @@ contract Synthetix is ExternStateToken, MixinResolver {
         optionalProxy
         returns (uint reclaimed, uint refunded, uint numEntriesSettled)
     {
+        systemStatus().requireSynthActive(currencyKey);
```
jacko125: @justinjmoses  This might be overlap with  the check in https://github.com/Synthetixio/synthetix/blob/sip-44-system-and-synth-disabling/contracts/Exchanger.sol#L278-L279

contracts/EtherCollateral.sol
```diff
@@ -334,6 +337,8 @@ contract EtherCollateral is Owned, Pausable, ReentrancyGuard, MixinResolver {
     // ========== PRIVATE FUNCTIONS ==========
 
     function _closeLoan(address account, uint256 loanID) private {
+        systemStatus().requireIssuanceActive();
+
```
jacko125: Should fix this bug here: https://github.com/Synthetixio/synthetix/blob/sip-44-system-and-synth-disabling/contracts/EtherCollateral.sol#L364

justinjmoses: That's a separate PR.

contracts/Synthetix.sol
justinjmoses: It does, but since `settle()` can be invoked here and directly in `Exchanger`, it's necessary to check both. We will need to migrate these to one area in the future.

contracts/SystemStatus.sol
```diff
@@ -0,0 +1,172 @@
...
+    function suspendSynth(bytes32 currencyKey, uint256 reason) external {
+        _requireAccessToSuspend(SECTION_SYNTH);
+        synthSuspension[currencyKey].suspended = true;
+        synthSuspension[currencyKey].reason = uint248(reason);
+        emit SynthSuspended(currencyKey, reason);
+    }
+
+    function resumeSynth(bytes32 currencyKey) external {
+        _requireAccessToResume(SECTION_SYNTH);
+        synthSuspension[currencyKey].suspended = false;
```
jacko125: @justinjmoses  You could delete the mapping here for `synthSuspension[currencyKey]`

test/contracts/PurgeableSynth.js
```diff
@@ -170,6 +171,15 @@ contract('PurgeableSynth', accounts => {
 					balanceBeforePurge = await iETHContract.balanceOf(account1);
 				});
 
+				describe('when the system is suspended', () => {
```
jacko125: good test here for purging

test/contracts/Synth.js
```diff
@@ -126,13 +127,80 @@ contract('Synth', async accounts => {
...
+					});
+					it('when transferFrom() is invoked, it works as expected', async () => {
+						await sUSDContract.transferFrom(owner, account1, amount, {
+							from: account1,
+						});
+					});
+				});
+			});
+		});
+		describe('when sETH is suspended', () => {
```
jacko125: @justinjmoses  Just a note on situation where sETH is suspended (issue with sETH price), we'd probably want to suspend > suspendIssuance / suspendSystem as well to stop etherCollateral from opening / closing loans as these won't be blocked by a `sETH` suspension.

PR #472, opened 2020-03-27
Deprecate ArbRewarder

hav-noms:


hav-noms:
> Please remove all traces from the `deployment.json` files as well
OK I need to go through each one and remove all the SNX before removing these. Including MAINNET which has ATM 680K SNX so I'll leave this here until this is done.

PR #474, opened 2020-03-28
Added a solidity test

maxsam4:
Just testing the waters (and CI) with a basic solidity test. Works on local.


justinjmoses:
Buidler doesn't support solidity unit tests and looks like it's not on their radar. Given how fast tests run now with buidler EVM I think it's reasonable to close this for now.

PR #475, opened 2020-03-29
Added script to catch function hash collisions

maxsam4:
The script will run as part of CI and will catch collisions between function selectors in proxy and target contracts.

The script currently runs on testnet and mainnet deployments. 

Ideally, it should also run on local deployments but I'm not sure how to generate the local deployment file.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,14 @@ jobs:
           at: .
       - run: npm run lint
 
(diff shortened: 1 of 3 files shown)


justinjmoses:
Apologies for not addressing this earlier @maxsam4. I am closing it for now as it's just not high enough priority for us to update it.

PR #483, opened 2020-03-30
Upgrade to Solidity 0.5.16

K-Ho:
Upgrade Betelegeuse release to Solidity 0.5.16

Not passing all tests (805 passing, 8 failing).

Tests are failing when calling Synthetix.mint(), within the beforeEach of RewardsIntegrationTests.js, as well as when testing mint in Synthetix.js. I believe the issue is from within _rewardsDistribution.distributeRewards(amountToDistribute);, when the Synthetix Proxy is called. It may be an issue of address payable -> address or vice-versa


contracts/AddressResolver.sol
```diff
@@ -1,16 +1,20 @@
...
 
 import "./Owned.sol";
 
 
 contract AddressResolver is Owned {
     mapping(bytes32 => address) public repository;
 
-    constructor(address _owner) public Owned(_owner) {}
+    constructor(address _owner) public Owned() {
+        require(_owner != address(0), "Owner address cannot be 0");
```
K-Ho: > Issue 1: can no longer have base constructor given twice warnings
> (i.e. ethereum/solidity#4199).
> This approach is one workaround - and at the very least we can dedupe this with a function in Owned.sol to perform these actions. An alternative is having only non-abstract classes invoke the Owner() constructor, and then perhaps having abstract contracts like MixinResolver fail if owner isn't set (by having the Owner() constructor called prior to the MixinResolver())

- @justinjmoses

K-Ho: Yep! Definitely cleaner to have an intializeOwner(address _owner) function inside of Owned.sol that gets called. -@K-Ho

contracts/EternalStorage.sol
```diff
@@ -96,7 +96,7 @@ contract EternalStorage is State {
         return BytesStorage[record];
     }
 
-    function setBytesValue(bytes32 record, bytes value) external onlyAssociatedContract {
+    function setBytesValue(bytes32 record, bytes calldata value) external onlyAssociatedContract {
```
K-Ho: > Qu: why do only these functions need this change?
-@justinjmoses

K-Ho: Think you answered below, but it's for external functions:

> Note that external functions require parameters with a data location of calldata

-@K-Ho

contracts/EtherCollateral.sol
```diff
@@ -16,9 +16,8 @@ contract EtherCollateral is Owned, Pausable, ReentrancyGuard, MixinResolver {
     using SafeDecimalMath for uint256;
 
     // ========== CONSTANTS ==========
-
-    uint256 constant ONE_THOUSAND = SafeDecimalMath.unit() * 1000;
-    uint256 constant ONE_HUNDRED = SafeDecimalMath.unit() * 100;
+    uint256 /*constant*/ ONE_THOUSAND = SafeDecimalMath.unit() * 1000;
+    uint256 /*constant*/ ONE_HUNDRED = SafeDecimalMath.unit() * 100;
```
K-Ho: > Qu: is this intentional?
-@justinjmoses

K-Ho: Yep, intentional:

> Values for constants that are not compile-time constants are disallowed.

https://solidity.readthedocs.io/en/v0.5.0/050-breaking-changes.html#variables

Either need to calculate these values beforehand or leave them as non-constants. Personally think going the route of leaving them as non-constants seems easiest.

-@K-ho

contracts/EtherCollateral.sol
```diff
@@ -255,9 +257,9 @@ contract EtherCollateral is Owned, Pausable, ReentrancyGuard, MixinResolver {
         totalFees = interest.add(_calculateMintingFee(synthLoan));
     }
 
-    function loanLifeSpan(address _account, uint256 _loanID) external view returns (uint256 loanLifeSpan) {
+    function loanLifeSpan(address _account, uint256 _loanID) external view returns (uint256 loanLifeSpanResult) {
```
K-Ho: > Qu: is this a new requirement - return value cannot clash with function name?
-@justinjmoses

K-Ho: Just checked again and realized this is not a requirement. It compiles and runs fine keeping the name as `loanLifeSpan`, but just gives a warning: `Warning: This declaration shadows an existing declaration.` Up to you guys if you want to rename or not to squash the warning :) -@k-Ho

contracts/ExchangeRates.sol
```diff
@@ -335,7 +335,7 @@ contract ExchangeRates is SelfDestructible {
     /**
      * @notice Retrieve the rates for a list of currencies
      */
-    function ratesForCurrencies(bytes32[] currencyKeys) external view returns (uint[]) {
+    function ratesForCurrencies(bytes32[] calldata currencyKeys) external view returns (uint[] memory) {
```
K-Ho: Right this is due to:

> Note that external functions require parameters with a data location of calldata.

(from https://solidity.readthedocs.io/en/v0.5.0/050-breaking-changes.html#explicitness-requirements)
-@justinjmoses

contracts/ExternStateToken.sol
```diff
@@ -156,14 +156,14 @@ contract ExternStateToken is SelfDestructible, Proxyable {
     bytes32 constant TRANSFER_SIG = keccak256("Transfer(address,address,uint256)");
 
     function emitTransfer(address from, address to, uint value) internal {
-        proxy._emit(abi.encode(value), 3, TRANSFER_SIG, bytes32(from), bytes32(to), 0);
+        proxy._emit(abi.encode(value), 3, TRANSFER_SIG, bytes32(uint256(uint160(from))), bytes32(uint256(uint160(to))), 0);
```
K-Ho: This error is: `TypeError: Explicit type conversion not allowed from "address" to "bytes32".`
Also trying to just do `bytes32(bytes20(from))` incorrectly formats the address.

contracts/FeePool.sol
```diff
@@ -25,7 +25,7 @@ contract FeePool is Proxyable, SelfDestructible, LimitedSetup, MixinResolver {
     uint public exchangeFeeRate;
 
     // Exchange fee may not exceed 10%.
-    uint public constant MAX_EXCHANGE_FEE_RATE = SafeDecimalMath.unit() / 10;
+    uint public /*constant*/ MAX_EXCHANGE_FEE_RATE = SafeDecimalMath.unit() / 10;
```
K-Ho: Same problem w/ constants that:

> Values for constants that are not compile-time constants are disallowed.

contracts/FeePool.sol
```diff
@@ -669,7 +669,10 @@ contract FeePool is Proxyable, SelfDestructible, LimitedSetup, MixinResolver {
         // If they don't have any debt ownership and they never minted, they don't have any fees.
         // User ownership can reduce to 0 if user burns all synths,
         // however they could have fees applicable for periods they had minted in before so we check debtEntryIndex.
-        if (debtEntryIndex == 0 && userOwnershipPercentage == 0) return;
+        if (debtEntryIndex == 0 && userOwnershipPercentage == 0) {
+            uint[2][FEE_PERIOD_LENGTH] memory nullResults;
+            return nullResults;
```
K-Ho: > Noted explicitness required.
-@justinjmoses

contracts/ExchangeRates.sol
```diff
@@ -1,11 +1,11 @@
-pragma solidity 0.4.25;
+pragma solidity ^0.5.16;
 
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "./SafeDecimalMath.sol";
 import "./SelfDestructible.sol";
 
 // AggregatorInterface from Chainlink represents a decentralized pricing network for a single currency key
-import "chainlink/contracts/interfaces/AggregatorInterface.sol";
+import "./chainlink/AggregatorInterface.sol";
```
K-Ho: > Instead of this, we can upgrade to @chainlink/contracts@0.0.3 and use https://github.com/smartcontractkit/chainlink/blob/develop/evm-contracts/src/v0.5/dev/AggregatorInterface.sol
-@justinjmoses

contracts/chainlink/AggregatorInterface.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity ^0.5.16;
```
K-Ho: > As above, we can use the one from Chainlink's npm module
-@justinjmoses

justinjmoses:
Ported to #499

PR #492, opened 2020-04-02
Migrating to buidler

justinjmoses:
- All contract tests ported to Buidler
- Removed migration script, replaced with setup utility to mock where possible
- Added support for coverage
- Ported our `testnet:local` and `test:publish` to use buidler-evm

test-contracts runs in under 6m, down from nearly an hour! 🙇 

> Note: the final missing piece is `buidler-gas-reporter` which a) isn't working with the latest buidler due to the limiting `peerDependency` and b) even when manually updated, produces empty reports. I'll look into this in a separate PR with a fork (@cgewecke)


justinjmoses:
@cgewecke running `npx @nomiclabs/buidler coverage` is failing with `Error: Returned error: Transaction gas limit is 1099511627775 and exceeds block gas limit of 9500000` even though I'm using the same `.solcover.js` file that works with truffle. I've tried adding `gasLimit: 9e12` to `providerOptions` but no luck. Any other thoughts?

buidler.config.js
```diff
@@ -0,0 +1,25 @@
...
+		artifacts: './build/artifacts',
+		cache: './build/cache',
+	},
+	networks: {
+		buidlerevm: {
+			// Initial date used by the following tests:
+			// - SupplySchedule
+			//
+			initialDate: '2019-03-06T00:00:00',
+			blockGasLimit: 0x1fffffffffffff,
```
cgewecke, line 22: @justinjmoses solidity-coverage doesn't work with buidlerevm yet. I think here you should add another network..like
```
coverage: {
  url: 'http://localhost:8555
}
```
... and then run `buidler coverage --network coverage` as the command.

cgewecke: EDIT - you're `.solcover.js` specifies port 8545, changing the network url above to reflect that:

https://github.com/Synthetixio/synthetix/blob/2dde2952befc0f215c8e9825f73d94bcc28582c0/.solcover.js#L2

justinjmoses: Thanks so much Chris! That fixes it. :pray:

alcuadrado: Thanks for giving support for your Buidler plugin, Chris 🙌

cgewecke: @justinjmoses Is the coverage difference here (e.g a drop) expected or does it seem like something's not right?

justinjmoses: Codecov is using the truffle coverage which has dropped as I've migrated tests to buidler, so you can ignore it. I checked out running `npm run coverage:unit` and the results are looking good. Once I move the rest of the tests over to `unit` from `contracts`, I'll update travis to use buidler coverage which will update codecov.

justinjmoses:
@cgewecke I'm getting `Exceeds block gas limit` with solidity-coverage on a number of our contracts that work fine when testing - see : https://travis-ci.org/github/Synthetixio/synthetix/jobs/672297483#L2100-L2148. I figure it's a simple config missing somewhere?

justinjmoses:
@cgewecke I have a PR to fix it: https://github.com/sc-forks/solidity-coverage/pull/494

test/contracts/Exchanger.js
```diff
@@ -140,11 +149,14 @@ contract('Exchanger (via Synthetix)', async accounts => {
 				describe('and 88 seconds elapses', () => {
 					// Note: timestamp accurancy can't be guaranteed, so provide a few seconds of buffer either way
 					beforeEach(async () => {
-						fastForward(88);
+						await fastForward(87);
```
hav-noms, line 152: was this supposed to be 87 or 88 as described?

test/contracts/RewardEscrow.js
```diff
@@ -281,11 +277,11 @@ contract('RewardEscrow', async accounts => {
 				await assert.revert(
 					rewardEscrow.appendVestingEntry(account1, toUnit('1'), { from: feePoolAccount })
 				);
-			});
+			}).timeout(60e3);
```
hav-noms, line 280: What affect will this have? will it fail the test if it times out or about this test?

test/contracts/RewardsIntegrationTests.js
```diff
@@ -53,49 +60,6 @@ contract('Rewards Integration Tests', async accounts => {
 		await updateRatesWithDefaults();
 	};
 
-	// const logFeePeriods = async () => {
```
hav-noms, line 56: I actually use these a every time I need to debug these tests

test/contracts/RewardsIntegrationTests.js
```diff
@@ -53,49 +60,6 @@ contract('Rewards Integration Tests', async accounts => {
...
-	// 				console.log(`  ${key}: ${period[key]}`);
-	// 			}
-	// 		}
-
-	// 		console.log();
-	// 	}
-	// 	console.log('------------------');
-	// };
-
-	// const logFeesByPeriod = async account => {
```
hav-noms, line 75: same with this

test/contracts/ProxyERC20.js
```diff
@@ -1,68 +1,79 @@
...
+	it('Must pass through to underlying via fallback function and emit on proxy', async () => {
+		const data = web3.eth.abi.encodeFunctionCall(
+			token.abi.find(({ name }) => name === 'somethingToBeProxied'),
+			['666', toBytes32('SNX')]
+		);
+		const txn = await proxyERC20.sendTransaction({ data, from: account3 });
+		// get rawLogs as logs not decoded because the truffle cannot decode the events from the
+		// underlying from the proxy invocation
+		const { topics } = txn.receipt.rawLogs[0];
+		assert.equal(topics[1], web3.eth.abi.encodeParameter('address', account3));
```
jacko125, line 54: good check here that messageSender set as expected

test/contracts/ProxyERC20.js
```diff
@@ -1,68 +1,79 @@
...
 
-	it('should tokenExchanger has ProxyERC20 set on deployment', async () => {
-		const _integrationProxyAddress = await tokenExchanger.integrationProxy();
-		assert.equal(proxyERC20.address, _integrationProxyAddress);
+	it('Must pass through to underlying via fallback function and emit on proxy', async () => {
+		const data = web3.eth.abi.encodeFunctionCall(
+			token.abi.find(({ name }) => name === 'somethingToBeProxied'),
+			['666', toBytes32('SNX')]
+		);
+		const txn = await proxyERC20.sendTransaction({ data, from: account3 });
```
jacko125, line 50: could use proxyThruTo from setupUtils here

test/contracts/Exchanger.js
justinjmoses: I was tinkering with it due to timing issues in the setup and forgot to put this back. Doing it now. 👍

test/contracts/RewardEscrow.js
justinjmoses: This allows individual tests to extend the default mocha timeout (which I believe is 20 secs): https://mochajs.org/#timeouts

I use the `it().timeout()` functionality as `this.timeout(...)` does not work with fat arrow functions due to the fact that they update `this` with the global context in the top-level `contract` or `describe` block.

test/contracts/RewardsIntegrationTests.js
justinjmoses: Once we have Solidity v5 you won't need these anymore! But sure, I'll put them back in for now

test/contracts/ProxyERC20.js
justinjmoses: 😊

test/contracts/ProxyERC20.js
justinjmoses: noted.

justinjmoses:
Additional improvements:

1. I removed the `web3` `artifacts` and `contracts` globals from `.eslintrc.js` and explicitly called out their usages in all tests
2. I added `use strict` headers to all tests
3. I took the `inflationStartTimestampInSecs` number and added it to the `synthetix` module and ensured everything used that number (rather than hardcoding it across the deployment and tests)
4. I made the snapshot and restores before and after tests explicit and used there where necessary. Please note that because some tests require the blocktime to start explicitly at `inflationStartTImestampInSecs` that *every* test that uses `fastForward` *must* also use snapshots to ensure that they don't pollute the EVM by upping the timestamp. @jacko125 this is a good reason to make the tests more robust and allow them to take any inflation start time rather than having to use the hard-coded amount. 
5. I explicitly added `chai.assert` to match the truffle assertions we were using
6. I ensured this `assert` was cloned and only that mutation be exported. 
7. I moved around test utilities and helpers to ensure only `test/utils/index.js` be the test utility file we share among different types of tests

PR #494, opened 2020-04-05
Owner script and getUsers function

justinjmoses:
## synthetix module
- adding `getUsers` function to synthetix for system users

## owner actions script
- adding `yes` flag to owner actions script
- adding default owner address parameter
- better error reporting when a non safe is connected

## nominate owner script
- adding default owner address parameter


jacko125:
@justinjmoses  Looks good these changes. I wanted to add the Market closure wallet to the `users` list is that ok if I push to the branch?

PR #499, opened 2020-04-18
Upgrade to Solidity 0.5.16

justinjmoses:
Here's the Solidity v5 upgrade branch. The lion's share of the effort was from @K-Ho :bow:; I cherry-picked his commits and massaged them into our `develop` branch.
 
## Notes

1. `MixinResolver`, `SelfDestructible` and `Pausable` are now all abstract. This is because they inherit from `Owned` but don't invoke the `Owned` constructor (this is as intended, so that we don't have the "Base constructor used twice issue"
2. All compiler warnings are now gone 😮 
3. I opted for the more permissive semver of solc `^0.516` (i.e. `0.5.x`) - happy to discuss being more restrictive. 

> Note: I've opted to keep this upgrade separate from testing Solidity v4 and v5 side-by-side, in order to keep this PR down in size. After merging, I'll work on that.


contracts/RewardsDistribution.sol
```diff
@@ -168,8 +175,12 @@ contract RewardsDistribution is Owned {
...
                 // If the contract implements RewardsDistributionRecipient.sol, inform it how many SNX its received.
                 bytes memory payload = abi.encodeWithSignature("notifyRewardAmount(uint256)", distributions[i].amount);
-                distributions[i].destination.call(payload);
-                // Note: we're ignoring the return value as it will fail for contracts that do not implement RewardsDistributionRecipient.sol
+
+                (bool success, ) = distributions[i].destination.call(payload);
+
+                if (!success) {
+                    // Note: we're ignoring the return value as it will fail for contracts that do not implement RewardsDistributionRecipient.sol
+                }
```
maxsam4, line 183: Is there any need of having this empty if statement?

contracts/State.sol
```diff
@@ -9,15 +9,20 @@ contract State is Owned {
...
         emit AssociatedContractUpdated(_associatedContract);
     }
 
     /* ========== SETTERS ========== */
 
     // Change the associated contract to a new address
-    function setAssociatedContract(address _associatedContract) external onlyOwner {
+    function setAssociatedContract(
+        address _associatedContract /* onlyOwner */
+    ) external {
```
maxsam4, line 25: `onlyOwner` not required here?

contracts/Synthetix.sol
```diff
@@ -59,18 +60,16 @@ contract Synthetix is ExternStateToken, MixinResolver {
...
-    /**
-     * @dev Constructor
-     * @param _proxy The main token address of the Proxy contract. This will be ProxyERC20.sol
-     * @param _tokenState Address of the external immutable contract containing token balances.
-     * @param _owner The owner of this contract.
-     * @param _totalSupply On upgrading set to reestablish the current total supply (This should be in SynthetixState if ever updated)
-     * @param _resolver The address of the Synthetix Address Resolver
-     */
-    constructor(address _proxy, TokenState _tokenState, address _owner, uint _totalSupply, address _resolver)
+    constructor(
```
maxsam4, line 63: natspec comments

contracts/test-helpers/MockExchanger.sol
```diff
@@ -34,10 +39,26 @@ contract MockExchanger {
     }
 
     function maxSecsLeftInWaitingPeriod(address account, bytes32 currencyKey) public view returns (uint) {
+        // silence compiler warnings
+        require(account != address(0), "Required account arg to suppress compiler warning");
+        require(currencyKey != bytes32(0), "Required currencyKey arg to suppress compiler warning");
```
maxsam4, line 44: Instead of adding extra code, the fn definition can be changed to 
`function maxSecsLeftInWaitingPeriod(address /* account */, bytes32 /* currencyKey */) public view returns (uint)`

Commenting out variable names suppresses the compiler warnings.

contracts/test-helpers/MockExchanger.sol
```diff
@@ -34,10 +39,26 @@ contract MockExchanger {
...
+        view
+        returns (
+            uint,
+            uint,
+            uint
+        )
+    {
+        // silence compiler warnings
+        require(account != address(0), "Required account arg to suppress compiler warning");
+        require(currencyKey != bytes32(0), "Required currencyKey arg to suppress compiler warning");
```
maxsam4, line 60: same as above

contracts/RewardsDistribution.sol
justinjmoses: It prevents the compiler warning for not doing anything with the result

contracts/State.sol
justinjmoses: Eagle eyes. 🙇 Going to write a test for that now.

contracts/FeePoolEternalStorage.sol
```diff
@@ -1,49 +1,20 @@
...
     constructor(address _owner, address _feePool) public EternalStorage(_owner, _feePool) LimitedSetup(6 weeks) {}
 
-    /**
-     * @notice Import data from FeePool.lastFeeWithdrawal
-     * @dev Only callable by the contract owner, and only for 6 weeks after deployment.
-     * @param accounts Array of addresses that have claimed
-     * @param feePeriodIDs Array feePeriodIDs with the accounts last claim
-     */
-    function importFeeWithdrawalData(address[] accounts, uint[] feePeriodIDs) external onlyOwner onlyDuringSetup {
+    function importFeeWithdrawalData(address[] calldata accounts, uint[] calldata feePeriodIDs)
```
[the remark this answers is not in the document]
justinjmoses: We're removing natspec altogether in favor of the docs. Natspec doesn't currently support state variables and internal functions which we want to include in our docs...

contracts/Synthetix.sol
justinjmoses: (ibid)

justinjmoses:
PR feedback addressed @maxsam4 - RFAL

contracts/IssuanceEternalStorage.sol
```diff
@@ -1,14 +1,11 @@
-pragma solidity 0.4.25;
+pragma solidity ^0.5.16;
 
 import "./EternalStorage.sol";
 
 
+// TODO: this contract is redundant and should be removed
```
hav-noms, line 6: Is this redundant? Can we remove it?

contracts/Owned.sol
```diff
@@ -1,14 +1,14 @@
...
 
 // https://docs.synthetix.io/contracts/Owned
 contract Owned {
     address public owner;
     address public nominatedOwner;
 
-    /**
-     * @dev Owned Constructor
-     */
+    // /**
```
hav-noms, line 9: might as well remove these natspec comments right?

contracts/Proxy.sol
```diff
@@ -20,10 +20,14 @@ contract Proxy is Owned {
         useDELEGATECALL = value;
```
hav-noms, line 20: I also think its time to remove the useDELEGATECALL path. We never used it and it should never be used.

contracts/Proxyable.sol
```diff
@@ -17,17 +17,20 @@ contract Proxyable is Owned {
      * optionalProxy modifiers, otherwise their invocations can use stale values. */
     address public messageSender;
```
hav-noms, line 18: Vote to remove the public accessor from messageSender. It used to be omitted and during an audit it was put back in from remediations however etherscan now always shows the last address that called the contract on the contract.... it was optically better hidden

contracts/PurgeableSynth.sol
```diff
@@ -70,6 +70,6 @@ contract PurgeableSynth is Synth {
     bytes32 private constant PURGED_SIG = keccak256("Purged(address,uint256)");
 
     function emitPurged(address account, uint value) internal {
```
hav-noms, line 72: @clementbalestrat @evgenyboxer from support issue today user asked where his iETH is that disappeared. We really need the exchange to show users they were purged out of a frozen iSynth and their inverse synth is now in sUSD

contracts/SelfDestructible.sol
```diff
@@ -1,31 +1,30 @@
...
+        emit SelfDestructBeneficiaryUpdated(owner);
     }
 
     /**
      * @notice Set the beneficiary address of this contract.
      * @dev Only the contract owner may call this. The provided beneficiary must be non-null.
      * @param _beneficiary The address to pay any eth contained in this contract to upon self-destruction.
      */
-    function setSelfDestructBeneficiary(address _beneficiary) external onlyOwner {
+    function setSelfDestructBeneficiary(address payable _beneficiary) external onlyOwner {
```
hav-noms, line 27: I know this is not part of this review but I would question if we event need a setSelfDestructBeneficiary anymore and our contracts would look optically better without setting a selfDestructBeneficiary address that gets all "funds" when blown up... It's only for ETH anyways and we have no contracts that need ETH sent to us that blow up.  We definitely shouldn't use this for ETH collateral either. Doesn't look good to let the owner use selfdestruct on a contract and get sent all the ETH

contracts/SupplySchedule.sol
```diff
@@ -26,10 +28,12 @@ contract SupplySchedule is Owned {
...
 
     // Address of the SynthetixProxy for the onlySynthetix modifier
-    address public synthetixProxy;
+    address payable public synthetixProxy;
 
     // Max SNX rewards for minter
-    uint public constant MAX_MINTER_REWARD = 200 * SafeDecimalMath.unit();
+    // TODO Uncomment constant
+    /*constant*/
+    uint public MAX_MINTER_REWARD = 200 * SafeDecimalMath.unit();
```
hav-noms, line 36: wanna do 1e18 * 200 like the others?

Also thats the TODO

contracts/Synthetix.sol
hav-noms: rip

contracts/SynthetixState.sol
```diff
@@ -1,13 +1,14 @@
...
+import "./State.sol";
 import "./LimitedSetup.sol";
 import "./SafeDecimalMath.sol";
-import "./State.sol";
+import "./Synthetix.sol";
 
 
 // https://docs.synthetix.io/contracts/SynthetixState
-contract SynthetixState is State, LimitedSetup {
+contract SynthetixState is Owned, State, LimitedSetup {
```
hav-noms, line 11: I much prefer how Owned has been lifted out

contracts/SynthetixState.sol
```diff
@@ -38,20 +39,18 @@ contract SynthetixState is State, LimitedSetup {
     // may not be issued against a given value of SNX.
     uint public issuanceRatio = SafeDecimalMath.unit() / 5;
     // No more synths may be issued than the value of SNX backing them.
-    uint constant MAX_ISSUANCE_RATIO = SafeDecimalMath.unit();
+    // TODO Uncomment constant
```
hav-noms, line 42: What is this?

contracts/PurgeableSynth.sol
clementbalestrat: Agreed. We need Zoe to have a thought about it.

contracts/ExternStateToken.sol
```diff
@@ -123,19 +128,30 @@ contract ExternStateToken is SelfDestructible, Proxyable {
     }
 
     /* ========== EVENTS ========== */
+    function addressToBytes32(address input) internal pure returns (bytes32) {
```
jacko125, line 131: Another way of approaching conversions for address type could be a library for converting  address to bytes32 instead of the internal function that gets inherited by the underlying contracts (Synth.sol and Synthetix.sol). 

```
contract ExternStateToken is Owned, SelfDestructible, Proxyable {
    using SafeMath for uint;
    using addressConversion for address;

    ....
    proxy._emit(abi.encode(value), 3, TRANSFER_SIG, from.toBytes32(), to.toBytes32(), 0);
}

library addressConversions {
   function toBytes32(address input) internal pure returns (bytes32) {
      return bytes32(uint256(uint160(input)));
   }
}
```

contracts/FeePool.sol
```diff
@@ -830,7 +844,7 @@ contract FeePool is Proxyable, SelfDestructible, LimitedSetup, MixinResolver {
             abi.encode(debtRatio, debtEntryIndex, feePeriodStartingDebtIndex),
             2,
             ISSUANCEDEBTRATIOENTRY_SIG,
-            bytes32(account),
+            bytes32(uint256(uint160(account))),
```
jacko125, line 847: This could be re-using the addressConversion library to convert address to Bytes32

contracts/Owned.sol
jacko125, line 9: Should this be removed / uncommented ?

contracts/PurgeableSynth.sol
```diff
@@ -70,6 +70,6 @@ contract PurgeableSynth is Synth {
...
-        proxy._emit(abi.encode(value), 2, PURGED_SIG, bytes32(account), 0, 0);
+        proxy._emit(abi.encode(value), 2, PURGED_SIG, bytes32(uint256(uint160(account))), 0, 0);
```
jacko125, line 73: note conversion here could reuse addressToBytes32() or possible library

contracts/RewardsDistribution.sol
```diff
@@ -168,8 +175,12 @@ contract RewardsDistribution is Owned {
 
                 // If the contract implements RewardsDistributionRecipient.sol, inform it how many SNX its received.
                 bytes memory payload = abi.encodeWithSignature("notifyRewardAmount(uint256)", distributions[i].amount);
-                distributions[i].destination.call(payload);
-                // Note: we're ignoring the return value as it will fail for contracts that do not implement RewardsDistributionRecipient.sol
+
+                (bool success, ) = distributions[i].destination.call(payload);
```
jacko125, line 179: @justinjmoses  If we're ignoring the return value / result of `.call()` then perhaps don't need to set `(bool success, )` at all and have the empty if statement to handle the compiler warning.

contracts/SupplySchedule.sol
jacko125, line 36: This should be set as constant - can remove the compile time constant error by setting `200 * 1e18`  instead of using `200 * SafeDecimalMath.unit();`

contracts/SynthetixState.sol
```diff
@@ -38,20 +39,18 @@ contract SynthetixState is State, LimitedSetup {
...
+    /*constant*/
+    uint MAX_ISSUANCE_RATIO = SafeDecimalMath.unit();
```
jacko125, line 44: Can update this to `uint constant MAX_ISSUANCE_RATIO = 1e18;`

contracts/IssuanceEternalStorage.sol
justinjmoses: It is redundant, but removing it requires updating our test setup script, deploy script and our deployment.json files. As for naming, both it and DelegateApprovals should use the [Base][Instance] ordering. E.g. `ProxyFeePool`, `TokenStatesUSD`. 
For EternalStorage instances, we should have `EternalStateIssuer` and `EternalStateDelegateApprovals`. This makes it much easier to reason about naming.

contracts/Owned.sol
justinjmoses: Yes, but let me do that in a follow up PR where I ensure everything is in the docs that need to be.

contracts/Proxy.sol
justinjmoses: Good point. Again, I'd like to do it in a follow up PR

contracts/Proxyable.sol
justinjmoses: Seconded. Though this should be in its own PR.

contracts/SupplySchedule.sol
justinjmoses: Yeah missed this, good spot.

contracts/SelfDestructible.sol
justinjmoses: I think that's a good point , worth a discussion (though separate PR).

contracts/SynthetixState.sol
justinjmoses: Mistake. Good catch

contracts/ExternStateToken.sol
justinjmoses: Yeah I had considered that as FeePool needs this function as well - though is it worth managing a new library just for this?

contracts/PurgeableSynth.sol
justinjmoses: Good catch - missed that one

contracts/RewardsDistribution.sol
justinjmoses: Not sure how that would work? Can you provide a suggestion?

justinjmoses:
RFAL

contracts/ExternStateToken.sol
jacko125: With v5 solidity it makes sense to have a library for address conversion as theres also the possible need to do conversions between `address` and `address payable` which can be included in the library , ie currently in supplySchedule. 

benefits would be reducing risks of inline conversions syntax error or copy-pasting `addressToBytes32()` into contracts that need it.

contracts/RewardsDistribution.sol
hav-noms: @jacko125 you mean just have `distributions[i].destination.call(payload);` and delete the rest?

jacko125: Yes, just the `distributions[i].destination.call(payload);` and delete the rest.

contracts/PurgeableSynth.sol
jacko125: @justinjmoses  Any update on this ?

contracts/RewardsDistribution.sol
justinjmoses: But then there's a compiler warning. How else to ignore that?

jacko125: Feels like the extra code to suppress the compiler warning is unnecessary.

justinjmoses: It's the only warning we have and for a few extra bytes of space it seems a manageable compromise?

PR #502, opened 2020-04-23
Adding legacy functionality to contract tests

justinjmoses:
This PR adds legacy testing capabilities to our contract tests. 

What it does is for all tests that rely on some contract that's not mocked, it will attempt to use the legacy source. 

For instance, `test/contracts/ExternStateToken.js` will test `contracts/ExternStateToken.sol` using `legacy/contracts/ProxyERC20.sol` and `legacy/contracts/TokenState.sol`.

### npm run test:legacy 

1. First it compiles all the contracts under `legacy/contracts` using solc `v0.4.25` (see `legacy/buidler.legacy.js`)
2. Then it copies these artifacts into the main artifact folder (from the same config above)
3. Then it runs a custom buidler task called `test:legacy` which:
     1. Enables a `legacy` flag in the BRE (builder runtime environment) and then starts the standard task
     2. Then it proceeds to kick off the regular `test` task (which compiles the latest sources from `contracts` and runs the test)

### npm run test

1. A function `linkWithLegacySupport` is exposed on the BRE. It figures out how to link the correct library (only `SafeDecimalMath` for now) depending on whether or not the artifact being linked is legacy or now. 
2. The truffle `contract('SomeContract', async cb => {... })` method is overridden to inject a `describe` block stating whether or not we are running in legacy mode
    1. If running in legacy mode, `artifacts.require` is overridden to require a legacy version of the artifact - if any - unless the artifact name matches the one being tested (i.e. not for `'SomeContract'` using the above example)

## Usage

It is run via `npm run test:legacy`

Optionally, run with the DEBUG flag to see which contracts use legacy sources

`DEBUG=1 npm run test:legacy`

Optionally, pass a collection of testfiles:

`npm run test:legacy -- test/contracts/ExternStateToken.js`.

> Notes: 
> 1. Coverage is run without legacy support (which makes sense)
> 2. In legacy mode, the test source itself is never the legacy file. E.g. when testing `FeePoolState` it will never use the legacy version of `FeePoolState`. 
> 3. All
-----

E.g running `DEBUG=1 npm run test:legacy -- test/contracts/FeePool.js`

image
image
image


test/contracts/ExternStateToken.js
```diff
@@ -16,146 +15,159 @@ contract('ExternStateToken', async accounts => {
...
-			'Some Token',
-			'TOKEN',
-			toUnit('1000'),
-			'18',
-			owner,
-			{
-				from: deployerAccount,
-			}
-		);
+	const ProxyERC20 = artifacts.require(`ProxyERC20`, { useLegacy: true });
```
justinjmoses, line 19: @hav-noms @jacko125 this is how you write a test to work in both regular mode and legacy mode. When run as `npm run test:legacy` - the legacy contracts will be loaded here instead

PR #503, opened 2020-04-25
Adding gas back in

justinjmoses:
Final TODO

- [x] use a flag on `test` instead of an environment variable


justinjmoses:
Ugh - if `buidler-gas-reporter` is used breaks all the buidler tests as you mentioned @maxsam4. Annoying. I'll have to only use the reporter in a separate config file and run that separately :/

cc @cgewecke @alcuadrado

alcuadrado:
Sorry for the inconvenience, we'll soon expose some APIs that `solidity-coverage` and `eth-gas-reporter` need to work with Buidler EVM.

cgewecke:
@alcuadrado @justinjmoses 

Not sure but this might be a problem at buidler-gas-reporter. If BuidlerEVM is running as an independent rpc node, the reporter's connection should work.

Also, the amount of time the gas job is taking seems crazy. We need to wrap data collection around the Buidler provider instead of making additional sync calls to fetch it from the chain and get that way down.

justinjmoses:
Hmm definitely getting further: https://travis-ci.org/github/Synthetixio/synthetix/jobs/690133769

@cgewecke are you able to take a quick look? 🙏

cgewecke:
@justinjmoses Will see if I can fix this weekend and get the speed improvements working as well. This shouldn't take 1/2 hr vs. a few minutes...

justinjmoses:
Any luck @cgewecke ?

cgewecke:
@justinjmoses #532 hopefully resolves.

justinjmoses:
Ah thank you @cgewecke :pray: Definitely a huge improvement, but I noticed that the gas reports are lacking a lot of coverage that they used to have. Here's the current one: https://github.com/Synthetixio/synthetix/pull/532/checks?check_run_id=708120597, and here's a prior one from March: https://github.com/Synthetixio/synthetix/runs/490408916

The reports we really need to see are the user-facing functions like `FeePool.claimFees()`, `Synthetix.exchange.*()`, `Synthetix.issue.*()` and `Synthetix.burn()`, and they are definitely in the tests... It looks like the gas reports are partial - I wonder if they are being overridden in successive specs?

cgewecke:
@justinjmoses Oh yes, that definitely looks like a bug. Will try to patch today.

justinjmoses:
Any luck @cgewecke ? Gas usage is a hot-topic these days in our community... 😬

cgewecke:
@justinjmoses Sorry I didn't get back to this yesterday. *Promise* I will look today.

cgewecke:
The diff will hopefully start showing up again in later PRs when this merges and there's a commit hash to compare the measurement to.

The "real time" currency price info is blank because CoinmarketCap now requires an API key to access its data. You should can enable it by setting a free tier key you [get from them here][1] as the value of a `coinmarketcap` option in the reporter buidler config.

Are there methods missing?

[1]: https://coinmarketcap.com/api/pricing/

justinjmoses:
Thanks so much @cgewecke :pray: :pray: :pray:

justinjmoses:
> The diff will hopefully start showing up again in later PRs when this merges and there's a commit hash to compare the measurement to.
Yep - that's what I figured

justinjmoses:
> Are there methods missing?

No it all looks solid. I'm going to tinker some more with it and throw it into our open PRs 😍

PR #505, opened 2020-04-27
[WIP] Added slither to CI

maxsam4:
It might take a couple of tries to get the CI config right but you can run slither locally with this dockerfile
```
FROM trailofbits/eth-security-toolbox
USER ethsec
ENV DEBIAN_FRONTEND noninteractive

RUN sudo apt update
RUN sudo apt install wget curl gnupg software-properties-common apt-utils git-core unzip python2.7 -y

RUN git clone https://github.com/Synthetixio/synthetix.git
WORKDIR /home/ethsec/synthetix

RUN git checkout develop
RUN npm i
RUN echo '{"buidler_cache_directory": "build/cache"}' >> slither.config.json
RUN slither .
```

PR #506, opened 2020-04-27
Added slither support

justinjmoses:
Replaces #505


justinjmoses:
@maxsam4 turns out we don't need a custom image to run `slither` locally (at least on a mac). I added a command to `package.json` to run slither via the trailofbits image.

PR #512, opened 2020-05-04
Improvements for third party integrations

justinjmoses:
- [x] AddressResolver.getSynth() functionality
- [x] IAddressResolver interface
- [x] Address Resolver forwarder


contracts/AddressResolver.sol
```diff
@@ -21,13 +23,20 @@ contract AddressResolver is Owned {
...
     }
 
-    function requireAndGetAddress(bytes32 name, string memory reason) public view returns (address) {
+    function requireAndGetAddress(bytes32 name, string calldata reason) external view returns (address) {
         address _foundAddress = repository[name];
         require(_foundAddress != address(0), reason);
         return _foundAddress;
     }
+
+    function getSynth(bytes32 key) external view returns (address) {
```
jacko125, line 36: This would be helpful !

contracts/Forwarder.sol
```diff
@@ -0,0 +1,33 @@
+pragma solidity ^0.5.16;
+
+import "./Owned.sol";
+
+
+// solhint-disable payable-fallback
+
+// https://docs.synthetix.io/contracts/Forwarder
+contract Forwarder is Owned {
```
justinjmoses, line 9: thoughts on this name? could also be `ProxyForwarder`, `ProxyReader`, `Reader`, ? (Keep in mind we'll use this a a prefix, so it'll look like `ProxyForwarderAddressResolver` when used for the resolver, say)

contracts/AddressResolver.sol
```diff
@@ -21,13 +28,19 @@ contract AddressResolver is Owned {
...
```
maxsam4, line 41: If there is no collision between synth and repository contract names, I'd love to see a new common function that first tries to load `repository[name]` and falls back to `synthetix.synths(name)` if the repository contract doesn't exist.

We'll then have to maintain an invariant that synth names do not collide with repository contract names but I think it will make life easier for external devs.

contracts/Forwarder.sol
maxsam4: `ProxyForwarder` seems redundant. I am ok with `Forwarder`. Can also consider names like `StaticProxy`, `ReadOnlyProxy`, `ReadProxy`.

We can also use something like `Broker` or `Mediator` but they will make more sense if used as a suffix. i.e. `AddressResolverBroker`.

contracts/AddressResolver.sol
```diff
@@ -2,16 +2,23 @@ pragma solidity ^0.5.16;
 
 import "./Owned.sol";
 
+import "./interfaces/IAddressResolver.sol";
+
+
+interface IHasSynths {
```
jacko125, line 8: why not `import interfaces/ISynthetix` here and `ISynthetix` rather than creating the once off `IHasSynth` interface ?

contracts/Forwarder.sol
jacko125: Forwarder / Broker (as a suffix) would be good names for this contract.

contracts/Forwarder.sol
```diff
@@ -0,0 +1,33 @@
...
+    address public target;
+
+    constructor(address _owner) public Owned(_owner) {}
+
+    function setTarget(address _target) external onlyOwner {
+        target = _target;
```
jacko125, line 15: possible to emit event on target update (helpful for integrators to watch this and be notified)

contracts/AddressResolver.sol
justinjmoses: Well once I've done the interface cleanup in #513 then I will (that's what kicked me off on that path)

contracts/AddressResolver.sol
justinjmoses: Yeah I'd considered that. So they do `getAddress("sETH")` rather than `getSynth("sETH")` right? It seemed clearer to have two separate functions - one to take contract names and one for synths. I always worry about creating too much special logic for external devs. Reasoning about it - contract names are different from synth keys, so it feels like two separate functions is clearer, no?

contracts/Forwarder.sol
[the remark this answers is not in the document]
justinjmoses: No this is designed to only be a reader - and limited as such.

contracts/Forwarder.sol
justinjmoses: Good catch

contracts/Forwarder.sol
justinjmoses: Right so we're stuck with prefixes right now so I think perhaps `ReadProxy` is clearest. I worry that `Forwarder` doesn't make it clear enough that it's any different from a regular Proxy

contracts/AddressResolver.sol
maxsam4: I'm fine with any really. Although, I'd don't completely agree on `contract names are different from synth keys` though. They can be the same thing if we just made sure we don't add collisions.

If you prefer only separate functions, I'm ok with that. People have to load different ABIs anyway. Unless you return the ABI as well from the smart contract :).

justinjmoses: Yes - they could be the same thing technically. I'm just not sure we *should* break that convention, as it makes it a bit harder to reason about IMO. 

Just as a reminder, my main focus here is for third party Solidity development, as in:
https://docs.synthetix.io/contracts/walkthrus/trading/#exchanging-in-solidity

justinjmoses: We can always add make the change to `AddressResolver` down the line if need be.

PR #517, opened 2020-05-08
More permissive pragmas for interfaces

justinjmoses:
This will allow third parties to include our interfaces in their projects and compile them easily.

Note: I had to work around chainlink's versioned interfaces, @thodges-gh any chance we could get permissible interfaces on the chainlink npm module in the future?

### DIFF
--- a/contracts/interfaces/IExchangeRates.sol
+++ b/contracts/interfaces/IExchangeRates.sol
@@ -1,27 +1,9 @@
 pragma solidity >=0.4.24;
 
 
-// via import "@chainlink/contracts-0.0.3/src/v0.5/dev/AggregatorInterface.sol";
-// included here to allow permissive compiler pragma
-interface AggregatorInterface {
-    function latestAnswer() external view returns (int256);
-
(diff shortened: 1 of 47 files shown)


contracts/interfaces/IExchangeRates.sol
```diff
@@ -1,11 +1,9 @@
-pragma solidity ^0.5.16;
-
-import "@chainlink/contracts-0.0.3/src/v0.5/dev/AggregatorInterface.sol";
+pragma solidity >=0.4.24;
 
 
 interface IExchangeRates {
     // Views
-    function aggregators(bytes32 currencyKey) external view returns (AggregatorInterface);
```
justinjmoses, line 8: simply use `address` as the interfaces from chainlink don't have permissive enough pragmas to be used here

jacko125:
Why not pragma `>=0.4.25` so `0.4.25` is minimum as what our current deployed contracts?

justinjmoses:
Trying to be as permissive to third party devs as possible as there are a number of teams still using 0.4.24 as far as I can see

PR #519, opened 2020-05-12
add list of suspension reasons

jacko125:
add list of suspension reasons


index.js
```diff
@@ -93,12 +93,24 @@ const getUsers = ({ network = 'mainnet', user } = {}) => {
 	return user ? users.find(({ name }) => name === user) : users;
 };
 
+const getSuspensionReason = code => {
```
justinjmoses, line 96: can you also add it to the binary and to the readme? :pray:

jacko125: done

jacko125:
@justinjmoses  Updated the API should be matching up with the other ones now.

PR #521, opened 2020-05-15
Sip 56 differencial fees

hav-noms:
https://sips.synthetix.io/sips/sip-56


contracts/Exchanger.sol
```diff
@@ -387,17 +379,32 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
         return timestamp.add(waitingPeriodSecs).sub(now);
     }
 
+    function getAmountsForExchange(
+        uint sourceAmount, 
+        bytes32 sourceCurrencyKey, 
+        bytes32 destinationCurrencyKey
+    ) public view returns (uint amountReceived, uint fee, uint exchangeFeeRate){
```
justinjmoses, line 386: public functions cost more gas, so we should start using a combination of an `internal` and an `external` function - we have the space now

test/contracts/FeePool.js
```diff
@@ -1484,4 +1432,131 @@ contract('FeePool', async accounts => {
...
+				args: [[sUSD], [toUnit('0.1')]],
+				accounts,
+				address: owner,
+				reason: 'Only the contract owner may perform this action',
+			});
+		});
+		it('when input array lengths dont match then revert ', async () => {
+			await assert.revert(
+				feePool.setExchangeFeeRateForSynths([sUSD, sAUD], [toUnit('0.1')], { from: owner })
+			);
```
justinjmoses, line 1452: please add revert reasons everywhere where possible

test/contracts/FeePool.js
```diff
@@ -1484,4 +1432,131 @@ contract('FeePool', async accounts => {
...
+		});
+		it('when owner sets an exchange fee rate larger than MAX_EXCHANGE_FEE_RATE then revert', async () => {
+			await assert.revert(
+				feePool.setExchangeFeeRateForSynths([sUSD], [toUnit('11')], {
+					from: owner,
+				})
```
justinjmoses, line 1458: ibid

test/contracts/FeePool.js
```diff
@@ -1484,4 +1432,131 @@ contract('FeePool', async accounts => {
...
+				const sUSDRate = await feePool.getExchangeFeeRateForSynth(sUSD);
+				assert.bnEqual(sUSDRate, fxBIPS);
+				const sBTCRate = await feePool.getExchangeFeeRateForSynth(sBTC);
+				assert.bnEqual(sBTCRate, cryptoBIPS);
+				const sETHRate = await feePool.getExchangeFeeRateForSynth(sETH);
+				assert.bnEqual(sETHRate, cryptoBIPS);
+			});
+		});
+
+		describe.only('Given synth exchange fee rates to update', async () => {
```
justinjmoses, line 1492: please try to remove these so we get full code coverage reports

contracts/FeePool.sol
```diff
@@ -405,6 +400,23 @@ contract FeePool is Owned, Proxyable, SelfDestructible, LimitedSetup, MixinResol
...
+        for (uint i = 0; i < synthKeys.length; i++) {              
+            require(exchangeFeeRates[i] <= MAX_EXCHANGE_FEE_RATE, "MAX_EXCHANGE_FEE_RATE exceeded");
+            feePoolEternalStorage().setUIntValue(
+                keccak256(abi.encodePacked(SYNTH_EXCHANGE_FEE_RATE, synthKeys[i])), 
+                exchangeFeeRates[i]
+            );                 
+        }        
+    }
+
+    function getExchangeFeeRateForSynth(bytes32 synthKey) public view returns (uint)
```
justinjmoses, line 416: `external `

contracts/FeePool.sol
```diff
@@ -405,6 +400,23 @@ contract FeePool is Owned, Proxyable, SelfDestructible, LimitedSetup, MixinResol
...
+    {
+        return feePoolEternalStorage().getUIntValue(keccak256(abi.encodePacked(SYNTH_EXCHANGE_FEE_RATE, synthKey)));
+    }
```
justinjmoses, line 419: Might be worth adding to a utility contract for the front-end that returns `exchangeRatesForSynths(bytes[]32 calldata synths) external view returns (bytes32[] memory)` cc @clementbalestrat

contracts/Exchanger.sol
```diff
@@ -387,17 +379,32 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
     function calculateExchangeAmountMinusFees(
-        bytes32 sourceCurrencyKey,
+        bytes32 /* sourceCurrencyKey */,
         bytes32 destinationCurrencyKey,
         uint destinationAmount
-    ) internal view returns (uint amountReceived, uint fee) {
+    ) internal view returns (uint amountReceived, uint fee, uint exchangeFeeRate) {
         // What's the fee on that currency that we should deduct?
-        amountReceived = destinationAmount;
+        amountReceived = destinationAmount; // TODO is this needed?
```
justinjmoses, line 403: you're right we can remove this

contracts/Exchanger.sol
justinjmoses: this should also be on `IExchanger` and @clementbalestrat will need to know about it for the dApps.

contracts/FeePool.sol
```diff
@@ -405,6 +400,23 @@ contract FeePool is Owned, Proxyable, SelfDestructible, LimitedSetup, MixinResol
             rewardsClaimed: rewardsClaimed
         });
     }
+    
+    function setExchangeFeeRateForSynths(bytes32[] calldata synthKeys, uint256[] calldata exchangeFeeRates) external onlyOwner
```
justinjmoses, line 404: Thinking about how best to manage these on-chain with the upcoming delegated migrator. Using in-memory dynamic arrays seems to be the most efficient:

```solidity
function arrayedDynamic() external {
        bytes32[] memory keys = new bytes32[](5); 
        uint256[] memory vals = new uint256[](5);
        
        keys[0] = "SNX";
        vals[0] = 1;
        
        keys[1] = "sUSD";
        vals[1] = 2;
        
        keys[2] = "sETH";
        vals[2] = 3;
        
        keys[3] = "sBNB";
        vals[3] = 4;
        
        keys[4] = "sBTC";
        vals[4] = 5;
        
        feePool().setExchangeFeeRateForSynths(keys, vals);
        
        // uses 26395 gas in a simple test
    }

 function simple() external {
        // presume feePool.setExchangeFeeRateForSynth(bytes32 key, uint val) external;

        feePool.setExchangeFeeRateForSynth("SNX", 1);
        feePool.setExchangeFeeRateForSynth("sUSD", 2);
        feePool.setExchangeFeeRateForSynth("sETH", 3);
        feePool.setExchangeFeeRateForSynth("sBNB", 4);
        feePool.setExchangeFeeRateForSynth("sBTC", 5);
        // uses 34930 gas and increases quickly for added synths
    }
    
    function arrayedStatic() external {
        // presume feePool.setExchangeFeeRateForSynths(bytes32[64] calldata keys, uint[64] calldata vals) external;

        bytes32[64] memory keys; 
        keys[0] = "SNX";
        keys[1] = "sUSD";
        keys[2] = "sETH";
        keys[3] = "sBNB";
        keys[4] = "sBTC";
        uint256[64] memory vals;
        vals[0] = 1;
        vals[1] = 2;
        vals[2] = 3;
        vals[3] = 4;
        vals[4] = 5;
        feePool().setExchangeFeeRateForSynths(keys, vals);
        
        // uses 35879 gas 
    }
    
```

cc @maxsam4

maxsam4: Solidity does not support array constants (not in v0.5 at least) so using dynamic memory arrays seems like our best bet.

If we really want to squeeze some juice, we can abi encode our data and store it as a constant and then do a low-level call. It will make the execution cost approximately 5% cheaper.

```
pragma solidity ^0.5.0;

contract Foo {
    
    function setExchangeFeeRateForSynths(bytes32[] calldata keys, uint256[] calldata vals) external {
        
    }

}

contract Bar {
    
    Foo feePool;
    
    bytes constant data = "0x28a1170d000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000005534e5800000000000000000000000000000000000000000000000000000000007355534400000000000000000000000000000000000000000000000000000000734554480000000000000000000000000000000000000000000000000000000073424e42000000000000000000000000000000000000000000000000000000007342544300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000500000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000300000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000005";
    
    constructor(Foo pool) public {
        feePool = pool;
    }
    
    
    function arrayedDynamic() external {

        bytes32[] memory keys = new bytes32[](5); 
        uint256[] memory vals = new uint256[](5);
        
        keys[0] = "SNX";
        vals[0] = 1;
        
        keys[1] = "sUSD";
        vals[1] = 2;
        
        keys[2] = "sETH";
        vals[2] = 3;
        
        keys[3] = "sBNB";
        vals[3] = 4;
        
        keys[4] = "sBTC";
        vals[4] = 5;
        
        feePool.setExchangeFeeRateForSynths(keys, vals);
        
        // 26221 total
        // 4949 execution cost
    }
    
    function arrayedDynamic2() external {
        
        address(feePool).call(data);
        
        // 25963 total
        // 4691 execution cost
    }
    
    // Helper function to encode data. Does not need to be part of production contract.
    function encode() public view returns (bytes memory) {
        bytes32[] memory keys = new bytes32[](5); 
        uint256[] memory vals = new uint256[](5);
        
        keys[0] = "SNX";
        vals[0] = 1;
        
        keys[1] = "sUSD";
        vals[1] = 2;
        
        keys[2] = "sETH";
        vals[2] = 3;
        
        keys[3] = "sBNB";
        vals[3] = 4;
        
        keys[4] = "sBTC";
        vals[4] = 5;
        
        return (
            abi.encodePacked( 
                feePool.setExchangeFeeRateForSynths.selector,
                abi.encode(keys, vals) 
            )
        );
    }
}
```

test/contracts/FeePool.js
hav-noms: Done

test/contracts/FeePool.js
hav-noms: Still a WIP. These will be gone when not in draft.

contracts/Exchanger.sol
hav-noms: Done

test/contracts/FeePool.js
hav-noms: done

contracts/FeePool.sol
hav-noms: FeePool is maxed out. Maybe we could add it to the exchange as a view function?

contracts/Exchanger.sol
hav-noms: gone

contracts/FeePool.sol
justinjmoses: We probably need to pull out the Fee claiming logic to `Claimer`....

contracts/Exchanger.sol
```diff
@@ -149,7 +137,7 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
             );
 
             // and deduct the fee from this amount
-            (uint amountShouldHaveReceived, ) = calculateExchangeAmountMinusFees(src, dest, destinationAmount);
+            (uint amountShouldHaveReceived, ,) = calculateExchangeAmounts(src, dest, destinationAmount);
```
jacko125, line 140: During settlement, for exchanges made previously the `exchangeFeeRate` could be different from the present rate and with differential fees this would change more frequently for pairs than the current flat 30bps exchange fee.   

The `exchangeEntry.exchangeFeeRate` is stored already but isn't used in settlements from the fee reclamations release. 

`calculateExchangeAmounts` for settlement could reuse the `exchangeEntry.exchangeFeeRate` stored to ensure fee reclamations are based on the exchange fee set at the time of trade.

hav-noms: Of course! Yes we talked about this. Thanks I'll make that change

contracts/FeePool.sol
```diff
@@ -405,6 +387,23 @@ contract FeePool is Owned, Proxyable, SelfDestructible, LimitedSetup, MixinResol
...
+    {        
+        require(synthKeys.length == exchangeFeeRates.length, "Array lengths dont match");
+        for (uint i = 0; i < synthKeys.length; i++) {
+            require(exchangeFeeRates[i] <= MAX_EXCHANGE_FEE_RATE, "MAX_EXCHANGE_FEE_RATE exceeded");
+            feePoolEternalStorage().setUIntValue(
```
jacko125, line 396: Good that we store the exchangeRate for each Synth inside the feePoolEternalStorage so won't have to set them each time FeePool is re-deployed. 

- Gas opt wise, as this would add 2 external reads to EternalStorage for each exchange tx, could feePool have a local cache like mixin resolver's cache, that loads all the synth rates from EternalStorage and updates when `setExchangeFeeRateForSynths` is invoked ?

hav-noms: That is a good idea. I would prefer to cache the prices in the Exchanger though so then it would not be cross contract each exchange

PR #526, opened 2020-05-21
Sip 56 differential fees

hav-noms:
New branch and PR after rebasing develop
https://sips.synthetix.io/sips/sip-56

- All fee calculation views moved from FeePool to Exchanger

- [x] calculateExchangeAmounts for settlement should reuse the exchangeEntry.exchangeFeeRate 
from @jacko125 
During settlement, for exchanges made previously the exchangeFeeRate could be different from the current rate and with differential fees this would change more frequently for pairs than the current flat 30bps exchange fee.

The exchangeEntry.exchangeFeeRate is stored already but isn't used in settlements from the fee reclamations SIP-37 release. For example, if fees go up for a particular pair, then it would be applied retrospectively on settlement and a trader might find themselves paying more.

calculateExchangeAmounts for settlement could reuse the exchangeEntry.exchangeFeeRate stored to ensure fee reclamations are based on the exchange fee set at the time of trade.
- [x] Test changing exchangeFeeRate after exchange then settle
- [ ] Should we add `exchanger.exchangeRatesForSynths(bytes[]32 calldata synths) external view returns (bytes32[] memory)`


justinjmoses:
1. Yes, I think the logic for rates for exchange go into the `Exchanger` completely - there's no need for them to live in `FeePool` anymore. 

2. I think instead of reusing and overloading `FeePoolEternalStorage`, we look at creating a new version of `EternalStorage` called `SynthetixStorage` that allows setting from anything in the `AddressResolver`, but it uses a mapping of `bytes32` for the contract name storage itself.  (so prefixes the keys with the name of the contract that is invoking it (using a reverse mapping in `AddressResolver` from `address => bytes32`). This way we can start to ensure all contract-level configuration lives in one eternal storage contract going forward, so SCCPs can be controlled by controlling access to this. 

```solidity
pragma solidity ^0.5.16;

interface IAddressResolver {
    // need to add this reverse lookup to AddressResolver
    function getContractForAddress(address) external returns (bytes32);
}

contract SynthetixStorage {
    mapping(bytes32 => mapping(bytes32 => uint)) internal UIntStorage;
    
    IAddressResolver public resolver; 
    
    constructor(IAddressResolver _resolver) public {
        resolver = _resolver;
    }
    
    function setUIntValue(bytes32 contractName, bytes32 record, uint value) external onlyContractOrAccessControl(contractName) {
        UIntStorage[contractName][record] = value;
    }

   function setUIntValues(bytes32 contractName, bytes32[] calldata records, uint[] values) external onlyContractOrAccessControl(contractName) {
      // we could add these for each type for easy loading in
      // ...
   }

    modifier onlyContractOrAccessControl(bytes32 contractName) {
        bytes32 contractAtSender = resolver.getContractForAddress(msg.sender);
        
        if (contractAtSender == bytes32(0)) {
            // then check access control ... 
        } else {
           require(contractAtSender == contractName);
            
        }
        _;
    }
}
```

@hav-noms @jacko125 @maxsam4 thoughts?

contracts/Exchanger.sol
```diff
@@ -387,19 +371,25 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
-        // What's the fee on that currency that we should deduct?
-        amountReceived = destinationAmount;
-
-        // Get the exchange fee rate
-        uint exchangeFeeRate = feeRateForExchange(sourceCurrencyKey, destinationCurrencyKey);
-
+    function feeRateForExchange(        
+        bytes32 /* sourceCurrencyKey */, // API for source in case pricing model evolves to include source rate
+        bytes32 destinationCurrencyKey
+    ) public view returns (uint exchangeFeeRate){
```
justinjmoses, line 377: in the interests of gas savings, let's use `external` as much as possible with `internal` companions for those functions we also need to invoke internally

contracts/Exchanger.sol
```diff
@@ -387,19 +371,25 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
+        exchangeFeeRate = feePool().getExchangeFeeRateForSynth(destinationCurrencyKey);
+    }
+    
+    function getAmountsForExchange(
+        uint sourceAmount, 
+        bytes32 sourceCurrencyKey, 
+        bytes32 destinationCurrencyKey
+    ) public view returns (uint amountReceived, uint fee, uint exchangeFeeRate) {
```
justinjmoses, line 385: (ibid)

contracts/interfaces/IExchanger.sol
```diff
@@ -8,9 +8,7 @@ interface IExchanger {
         bytes32 currencyKey,
         uint amount,
         uint refunded
-    ) external view returns (uint amountAfterSettlement);
-
-    function feeRateForExchange(bytes32 sourceCurrencyKey, bytes32 destinationCurrencyKey) external view returns (uint);
```
justinjmoses, line 13: this needs to be brought back in

contracts/Exchanger.sol
hav-noms: done

contracts/Exchanger.sol
hav-noms: added internal _getAmountsForExchange

contracts/Exchanger.sol
hav-noms: added internal _feeRateForExchange

contracts/interfaces/IExchanger.sol
hav-noms: done

test/contracts/Exchanger.js
```diff
@@ -100,12 +100,21 @@ contract('Exchanger (via Synthetix)', async accounts => {
...
 			}
 		);
+
+		// set a 0.5% exchange fee rate (1/200)
+		exchangeFeeRate = toUnit('0.005');
+		await setExchangeFeeRateForSynths({
+			owner,
+			feePool,
+			synthKeys,
+			exchangeFeeRates: synthKeys.map(x => exchangeFeeRate),
```
justinjmoses, line 116: [np] no need to define `x`

test/contracts/Exchanger.js
```diff
@@ -281,46 +290,202 @@ contract('Exchanger (via Synthetix)', async accounts => {
...
+			exchangeFeeRate = await feePool.getExchangeFeeRateForSynth(sBTC);
+
 			const actualFeeRate = await exchanger.feeRateForExchange(sEUR, sBTC);
 			assert.bnEqual(actualFeeRate, exchangeFeeRate, 'Rate must be the exchange fee rate');
 		});
 		it('for two inverse synths, returns the regular exchange fee', async () => {
-			const actualFeeRate = await exchanger.feeRateForExchange(iBTC, toBytes32('iETH'));
+			exchangeFeeRate = await feePool.getExchangeFeeRateForSynth(iETH);
+
+			const actualFeeRate = await exchanger.feeRateForExchange(iBTC, iETH);
```
justinjmoses, line 303: either this logic needs moving to Exchanger or we need to move these tests to FeePool.

test/contracts/Exchanger.js
```diff
@@ -281,46 +290,202 @@ contract('Exchanger (via Synthetix)', async accounts => {
...
+					);
+					orgininalFee = fee;
+					orginalFeeRate = exchangeFeeRate;
+
+					await feePool.setExchangeFeeRateForSynths([sAUD], [multiplyDecimal(bipsFX, factor)], {
+						from: owner,
+					});
+				});
+				it('then return the fee tripled', async () => {
+					// eslint-disable-next-line no-unused-vars
```
justinjmoses, line 454: why disable here?

test/contracts/Exchanger.js
```diff
@@ -281,46 +290,202 @@ contract('Exchanger (via Synthetix)', async accounts => {
...
+					const { amountReceived, fee } = await exchanger.getAmountsForExchange(amount, sUSD, sAUD);
+					assert.bnEqual(fee, multiplyDecimal(orgininalFee, factor));
+				});
+				it('then return the feeRate tripled', async () => {
+					// eslint-disable-next-line no-unused-vars
```
justinjmoses, line 459: (ibid)

test/contracts/Exchanger.js
```diff
@@ -281,46 +290,202 @@ contract('Exchanger (via Synthetix)', async accounts => {
...
+				}
+			);
+		});
+		describe('when calling getAmountsForExchange', () => {
+			describe('and the destination is a crypto synth', () => {
+				let received;
+				let destinationFee;
+				let feeRate;
+				beforeEach(async () => {
+					await synthetix.exchange(sUSD, amountIssued, sBTC, { from: account1 });
```
justinjmoses, line 351: I think this section would be better off without the `exchange` - it's really not necessary to validate that the `getAmountsForExchange()` function works. The coverage for that part should go into wherever you are testing `exchange` itself.

test/contracts/Exchanger.js
```diff
@@ -281,46 +290,202 @@ contract('Exchanger (via Synthetix)', async accounts => {
...
+			});
+
+			describe('when tripling an exchange rate', () => {
+				const amount = toUnit('1000');
+				const factor = toUnit('3');
+
+				let orgininalFee;
+				let orginalFeeRate;
+				beforeEach(async () => {
+					// eslint-disable-next-line no-unused-vars
```
justinjmoses, line 440: please remove

test/contracts/Exchanger.js
```diff
@@ -396,8 +561,14 @@ contract('Exchanger (via Synthetix)', async accounts => {
...
 			describe('and the exchange fee rate is 1% for easier human consumption', () => {
 				beforeEach(async () => {
+					// Warning: this is mutating the global exchangeFeeRate for this test block and will be reset when out of scope
 					exchangeFeeRate = toUnit('0.01');
-					await setExchangeFee({ owner, feePool, exchangeFeeRate });
+					await setExchangeFeeRateForSynths({
+						owner,
+						feePool,
+						synthKeys,
+						exchangeFeeRates: synthKeys.map(x => exchangeFeeRate),
```
justinjmoses, line 570: please remove the `x`

test/contracts/Exchanger.js
```diff
@@ -396,8 +561,14 @@ contract('Exchanger (via Synthetix)', async accounts => {
 			});
 			describe('and the exchange fee rate is 1% for easier human consumption', () => {
 				beforeEach(async () => {
+					// Warning: this is mutating the global exchangeFeeRate for this test block and will be reset when out of scope
```
justinjmoses, line 564: Do you think this is a necessary comment? We are using `addSnapshotBeforeRestoreAfterEach()` above so it should be inferred...

test/contracts/Exchanger.js
```diff
@@ -979,6 +1170,32 @@ contract('Exchanger (via Synthetix)', async accounts => {
...
+											});
+											describe('when another minute passes and the exchange fee rate has increased', () => {
+												beforeEach(async () => {
+													await fastForward(60);
+													feePool.setExchangeFeeRateForSynths([sBTC], [toUnit('0.1')], {
+														from: owner,
+													});
+												});
+												describe('when settle() is invoked for sBTC', () => {
+													it('then it settles with a rebate using the exchange fee rate at time of trade', async () => {
```
justinjmoses, line 1194: good coverage 👍

test/contracts/Exchanger.js
justinjmoses: You are already covering it down here: https://github.com/Synthetixio/synthetix/pull/526/files#diff-4d03a8bb67a402e2360871c1aed458e0R1388

test/contracts/FeePool.js
```diff
@@ -109,6 +116,15 @@ contract('FeePool', async accounts => {
...
 		// Send a price update to guarantee we're not stale.
 		await updateRatesWithDefaults();
+
+		// set a 0.3% default exchange fee rate
+		const exchangeFeeRate = toUnit('0.003');
+		await setExchangeFeeRateForSynths({
+			owner,
+			feePool,
+			synthKeys,
+			exchangeFeeRates: synthKeys.map(x => exchangeFeeRate),
```
justinjmoses, line 126: no `x`

test/contracts/Issuer.js
```diff
@@ -1405,12 +1416,18 @@ contract('Issuer (via Synthetix)', async accounts => {
...
 				beforeEach(async () => {
 					amount = toUnit('1250');
 					await setExchangeWaitingPeriod({ owner, exchanger, secs: 60 });
 					// set the exchange fee to 0 to effectively ignore it
-					await setExchangeFee({ owner, feePool, exchangeFeeRate: '0' });
+					await setExchangeFeeRateForSynths({
+						owner,
+						feePool,
+						synthKeys,
+						exchangeFeeRates: synthKeys.map(x => exchangeFeeRate),
```
justinjmoses, line 1429: `x`

test/contracts/Issuer.js
```diff
@@ -22,6 +22,7 @@ const { toBytes32 } = require('../..');
 
 contract('Issuer (via Synthetix)', async accounts => {
 	const [sUSD, sAUD, sEUR, SNX] = ['sUSD', 'sAUD', 'sEUR', 'SNX'].map(toBytes32);
+	const synthKeys = [sUSD, sAUD, sEUR, SNX];
```
justinjmoses, line 25: SNX is a synth key?

justinjmoses: You can also pass them thru here: https://github.com/Synthetixio/synthetix/pull/526/files#diff-6a144895e3eabd852dbaf475d7e21bfbR62

test/contracts/RewardsIntegrationTests.js
```diff
@@ -55,6 +58,8 @@ contract('Rewards Integration Tests', async accounts => {
 		'ETH',
 	].map(toBytes32);
 
+	const synthKeys = [sUSD, sAUD, sEUR, sBTC, SNX, iBTC, sETH, ETH];
```
justinjmoses, line 61: is SNX a synth key?

test/contracts/RewardsIntegrationTests.js
```diff
@@ -160,6 +174,13 @@ contract('Rewards Integration Tests', async accounts => {
 		}));
 
 		MINTER_SNX_REWARD = await supplySchedule.minterReward();
+
+		await setExchangeFeeRateForSynths({
+			owner,
+			feePool,
+			synthKeys,
+			exchangeFeeRates: synthKeys.map(x => exchangeFeeRate),
```
justinjmoses, line 182: `x`

test/contracts/Synth.js
```diff
@@ -53,6 +53,7 @@ contract('Synth', async accounts => {
 				'Synth',
 				'ExchangeRates',
 				'FeePool',
+				'FeePoolEternalStorage',
```
justinjmoses, line 56: ?

test/publish/index.js
```diff
@@ -477,12 +479,15 @@ describe('publish scripts', function() {
 								assert.strictEqual(web3.utils.fromWei(balance), '5000', 'Balance should match');
 							});
 							it('and their sBTC balance is 1000 - the fee', async () => {
-								const expected = await callMethodWithRetry(
-									FeePool.methods.amountReceivedFromExchange(web3.utils.toWei('1000'))
+								// const expected = await callMethodWithRetry(
```
justinjmoses, line 482: please remove

test/publish/index.js
```diff
@@ -538,12 +543,19 @@ describe('publish scripts', function() {
 											const balance = await callMethodWithRetry(
 												sUSDContract.methods.balanceOf(accounts.first.public)
 											);
-											const sUSDGainedFromPurge = await callMethodWithRetry(
-												FeePool.methods.amountReceivedFromExchange(sBTCBalanceAfterExchange)
+											// const sUSDGainedFromPurge = await callMethodWithRetry(
```
justinjmoses, line 546: please remove

hav-noms: done

test/publish/index.js
hav-noms: done

test/contracts/Synth.js
hav-noms: Test `when transferring it to FEE_ADDRESS it should exchange into sUSD first before sending:` fails with 
     Error: Transaction reverted: function returned an unexpected amount of data
without it

test/contracts/RewardsIntegrationTests.js
hav-noms: nope. good catch

hav-noms: removed

test/contracts/Exchanger.js
hav-noms: I disagree. feeRateForExchange purpose is for exchange rules. This is where we had doubling fee rate logic for inverse trades its also there incase we have future sips specifying more exotic exchange calculations that include the source. Its logic for the exchanger where the reading of the rates will be refactored to SynthetixEternalStorage. We're planning to move away from the FeePool not add to it.

contracts/FeePool.sol
```diff
@@ -835,11 +810,11 @@ contract FeePool is Owned, Proxyable, SelfDestructible, LimitedSetup, MixinResol
...
     }
 
-    event ExchangeFeeUpdated(uint newFeeRate);
-    bytes32 private constant EXCHANGEFEEUPDATED_SIG = keccak256("ExchangeFeeUpdated(uint256)");
+    event SynthExchangeFeeUpdated(bytes32 synthKey, uint newExchangeFeeRate);
+    bytes32 private constant SYNTHEXCHANGEFEEUPDATED_SIG = keccak256("SynthExchangeFeeUpdated(bytes32,uint256)");
 
-    function emitExchangeFeeUpdated(uint newFeeRate) internal {
-        proxy._emit(abi.encode(newFeeRate), 1, EXCHANGEFEEUPDATED_SIG, 0, 0, 0);
+    function emitExchangeFeeUpdated(bytes32 synthKey, uint newExchangeFeeRate) internal {
```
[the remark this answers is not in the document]
hav-noms: You're right It doesn't. It needs to move to SynthetixEternalStorage. This is emitting off we're it currently gets emitted its the best place for now since the address wont change on upgrades. unlike exchanger. Unless we move it to emit off Synthetix

test/contracts/Synth.js
justinjmoses: Ah ok. So the problem is that `FeePool` is mocking `FeePoolEternalStorage` and you need it to use the real fee pool. You can solve it this way or you can solve it by moving `FeePoolEternalStorage` from `mocks` to `deps` in `setup.js` for `FeePool`

justinjmoses: It's hard to say what's better frankly...

contracts/FeePool.sol
```diff
@@ -404,6 +386,24 @@ contract FeePool is Owned, Proxyable, SelfDestructible, LimitedSetup, MixinResol
             rewardsClaimed: rewardsClaimed
         });
     }
+    
+    function setExchangeFeeRateForSynths(bytes32[] calldata synthKeys, uint256[] calldata exchangeFeeRates) external onlyOwner
```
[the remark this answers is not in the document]
hav-noms: setExchangeFeeRateForSynths is not the one thats wasteful as its an onlyOwner call to set the rates.
getExchangeFeeRateForSynths is and should be moved to exchanger calling SynthetixEternalStorage instead.

contracts/Exchanger.sol
```diff
@@ -148,8 +136,8 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
                 destRoundIdAtPeriodEnd
             );
 
-            // and deduct the fee from this amount
-            (uint amountShouldHaveReceived, ) = calculateExchangeAmountMinusFees(src, dest, destinationAmount);
+            // and deduct the fee from this amount using the exchangeFeeRate from storage
+            uint amountShouldHaveReceived = destinationAmount.multiplyDecimal(SafeDecimalMath.unit().sub(exchangeFeeRate));
```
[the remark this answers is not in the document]
hav-noms: Gotcha, I'll split it out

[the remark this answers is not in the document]
hav-noms: It's now `uint amountShouldHaveReceived = _getAmountReceivedForExchange(destinationAmount, exchangeFeeRate);`

test/contracts/Exchanger.js
hav-noms: replaced with `() => {return exchangeFeeRate;}),`

test/contracts/RewardsIntegrationTests.js
hav-noms: replaced with `() => {return exchangeFeeRate;}),`

test/contracts/Exchanger.js
justinjmoses: If you aren’t the one to tackle this debt and requirements change, someone will have to know to migrate this logic over. Even though this is likely the way we are going in the next release, it’s extra overhead for someone to have to know that there’s this exception to the pattern that tests for a function are in the spec that matches the contract name.

justinjmoses: Though that being said, I believe you’re right to keep this as you’re asserting the view is correct in Exchanger. So I stand corrected.

test/contracts/Exchanger.js
justinjmoses: Actually just `() => exchangeFeeRate` will do

contracts/Exchanger.sol
```diff
@@ -257,18 +245,13 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
-            sourceCurrencyKey,
-            sourceAmountAfterSettlement,
-            destinationCurrencyKey
-        );
-
         uint fee;
+        uint exchangeFeeRate;
 
-        (amountReceived, fee) = calculateExchangeAmountMinusFees(
+        (amountReceived, fee, exchangeFeeRate) = _getAmountsForExchange(
```
jacko125, line 251: @hav-noms  `_getAmountsForExchange` would that be more clearer if named `_getAmountsForExchangeMinusFees` so that we know amountReceived is less fees ?

contracts/Exchanger.sol
```diff
@@ -297,7 +280,8 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
             sourceCurrencyKey,
             sourceAmountAfterSettlement,
             destinationCurrencyKey,
-            amountReceived
+            amountReceived,
+            exchangeFeeRate
```
jacko125, line 284: Good appending the exFeeRate here instead of it being loaded again in appendExchange()

contracts/Exchanger.sol
```diff
@@ -387,33 +371,61 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
         uint amount,
         bytes32 dest,
-        uint amountReceived
+        uint amountReceived,
+        uint exchangeFeeRate
     ) internal {
         IExchangeRates exRates = exchangeRates();
         uint roundIdForSrc = exRates.getCurrentRoundId(src);
-        uint roundIdForDest = exRates.getCurrentRoundId(dest);
-        uint exchangeFeeRate = feePool().exchangeFeeRate();
```
jacko125, line 416: Good, saves one external call here

test/contracts/Exchanger.js
```diff
@@ -23,16 +23,19 @@ const { toBytes32 } = require('../..');
...
 		'sAUD',
 		'sEUR',
 		'SNX',
 		'sBTC',
 		'iBTC',
 		'sETH',
+		'iETH',
 	].map(toBytes32);
 
+	const synthKeys = [sUSD, sAUD, sEUR, SNX, sBTC, iBTC, sETH, iETH];
```
jacko125, line 37: SNX might not be a synthKey needed to set an exchangeFee for in this test

hav-noms: fixed thanks

test/contracts/Exchanger.js
hav-noms: done

contracts/Exchanger.sol
justinjmoses: ++

contracts/FeePool.sol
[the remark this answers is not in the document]
justinjmoses: yep - by "these" I meant the pair of them, but I take your point that it's really the getter that's the problem

contracts/FeePool.sol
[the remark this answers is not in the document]
justinjmoses: The thing is, if we change where we emit from between deployments, we have to code that difference into the subgraphs. But we can manage with this, *c'est la vie*.

test/contracts/Synth.js
justinjmoses: Could you please add a comment here about this and above in Exchanger so that when we move the fee logic into Exchanger completely we'll know to remove those entries? It's not easy to follow the trail of dependencies so keeping the list as trim as possible is the best bet.

contracts/interfaces/IExchanger.sol
```diff
@@ -23,14 +21,25 @@ interface IExchanger {
...
+
     // Mutative functions
     function exchange(
         address from,
         bytes32 sourceCurrencyKey,
         uint sourceAmount,
         bytes32 destinationCurrencyKey,
         address destinationAddress
-    ) external returns (uint amountReceived);
+    ) external returns (uint amountReceived);
```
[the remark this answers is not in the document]
hav-noms: done

justinjmoses:
@hav-noms I merged in `develop` so we get gas reports and fixed up some minor issues from the PR, including formatting

PR #528, opened 2020-05-21
Adding contract size output to compile task for buidler

justinjmoses:
Ensure the buidler tests uses optimized sources when requested, and output compiled sizes.

Run via `npx buidler compile --showsize --optimizer`

Added both opts to circleci, so it will fail if the contract is oversize (as using `--optimizer` will disable `allowUnlimitedContractSize`)

### Opts

* `--showsize` output file sizes
* `--optimizer` enable the optimizer to run with `200` runs

sizes-buidler

### DIFF
--- a/buidler.config.js
+++ b/buidler.config.js
@@ -16,6 +16,10 @@ const {
 	constants: { inflationStartTimestampInSecs, AST_FILENAME, AST_FOLDER, BUILD_FOLDER },
 } = require('.');
 
+const {
+	DEFAULTS: { optimizerRuns },
+} = require('./publish/src/commands/build');
+
 const log = (...text) => console.log(gray(...['└─> [DEBUG]'].concat(text)));
 
 const GAS_PRICE = 20e9; // 20 GWEI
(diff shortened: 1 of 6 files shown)


buidler.config.js
```diff
@@ -107,6 +114,54 @@ task('test:legacy', 'run the tests with legacy components')
...
+	.addFlag('showsize', 'Show size of compiled contracts')
+	.addFlag('optimizer', 'Compile with the optimizer')
+	.setAction(async (taskArguments, bre, runSuper) => {
+		if (taskArguments.optimizer) {
+			// Use optimizer (slower) but simulates real contract size limits and gas usage
+			// Note: does not consider actual deployed optimization runs from
+			// publish/src/contract-overrides.js
+			console.log(gray('Adding optimizer, runs', yellow(optimizerRuns)));
+			bre.config.solc.optimizer = { enabled: true, runs: optimizerRuns };
+			bre.config.networks.buidlerevm.allowUnlimitedContractSize = false;
```
jacko125, line 127: this would be good to set so we don't start building on Synthetix contract and not even aware it's over-size

justinjmoses:
It's available in CircleCI reports as such:

image

PR #531, opened 2020-05-23
Moving issuance views to issuer (for SIP-48)

justinjmoses:
Here is a proposal for how to reduce size of Synthetix, by moving issuance-specific views over and keeping.

This is a PR into the SIP-48 branch of #451.

These are the contract size of `develop` (not rates stale)
image

And these are the contract size of this:
image

I'm going to add average gas prices for issuance actions here shortly to compare

### DIFF
--- a/contracts/Issuer.sol
+++ b/contracts/Issuer.sol
@@ -8,13 +8,17 @@ import "./interfaces/IIssuer.sol";
 // Libraries
 import "./SafeDecimalMath.sol";
 
-// Inheritance
+// Internal references
 import "./IssuanceEternalStorage.sol";
 import "./interfaces/ISynthetix.sol";
 import "./interfaces/IFeePool.sol";
 import "./interfaces/ISynthetixState.sol";
 import "./interfaces/IExchanger.sol";
 import "./interfaces/IDelegateApprovals.sol";
+import "./IssuanceEternalStorage.sol";
+import "./interfaces/IExchangeRates.sol";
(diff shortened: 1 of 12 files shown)


justinjmoses:
Hmmm - I tried a few manual reports
* Synthetix.exchange
     * develop: 470k
     * sip-48: 471k
     * this: 471k
* Synthetix.issue & burn
    * develop: 376k & 206k
    * sip-48: 406k, 204k
    * this: 437k & 234k

contracts/Issuer.sol
```diff
@@ -8,13 +8,17 @@ import "./interfaces/IIssuer.sol";
...
 
-// Inheritance
+// Internal references
 import "./IssuanceEternalStorage.sol";
 import "./interfaces/ISynthetix.sol";
 import "./interfaces/IFeePool.sol";
 import "./interfaces/ISynthetixState.sol";
 import "./interfaces/IExchanger.sol";
 import "./interfaces/IDelegateApprovals.sol";
+import "./IssuanceEternalStorage.sol";
```
krebernisak, line 18: Hi! Just browsing the repo to get a feel for it, and I figured I should make my first comment :)

I noticed a duplicated import here. It is ignored by the compiler, but you should probably remove it.

```suggestion
```

justinjmoses: Yep good spot, thanks!

justinjmoses:
Latest improvements

image

PR #532, opened 2020-05-26
Update buidler-gas-reporter to 0.1.4-beta.2

cgewecke:
**PR targets #503**

A fix for #503. Uses a beta version of buidler-gas-reporter that wraps BuidlerEVM's provider to get the additional data it needs. 

+ Still a little slower than normal... (~ 12 minutes).  
+ Looks like this would need to merge to verify the correctness of the codechecks diff 

**NB:** this is a temporary patch. Patricio notes in [this buidler-gas-reporter PR comment][1] that the provider logic is being refactored at Buidler. The plugin version installed here will need an update when he publishes his new work. 

(Leaving explanatory notes in the diff, please just ping if you'd like changes or clarification.)

[1]: https://github.com/cgewecke/buidler-gas-reporter/pull/18#issuecomment-633753475


buidler.config.js
```diff
@@ -188,4 +189,10 @@ module.exports = {
 			baseNetworkConfig
 		),
 	},
+	gasReporter: {
+		enabled: process.env.GAS ? true : false,
+		showTimeSpent: true,
+		currency: 'USD',
+		outputFile: 'test-gas-used.log',
+	}
```
cgewecke, line 197: Deleted the special gas config and moved everything here. The reporter is toggled `on` by an environment variable set in the npm script for `test:gas`. Otherwise off.

package.json
```diff
@@ -16,7 +16,7 @@
 		"slither": "pip3 install --user slither-analyzer && slither .",
 		"prepublishOnly": "npm run generate-asts",
 		"test": "buidler test",
-		"test:gas": "concurrently --kill-others --success first \"npx buidler node --config buidler.gas.js > /dev/null\" \"wait-port 8545 && npx buidler test --network localhost --config buidler.gas.js\" || cat test-gas-used.log",
+		"test:gas": "GAS=1 buidler test || cat test-gas-used.log",
```
cgewecke, line 19: No longer needs an independent node.

PR #529, opened 2020-05-22
Improve test Coverage 

hav-noms:
Remove deprecated code
1. deprecate Proxy.setUseDELEGATECALL 
2. deprecate SynthetixState.setPreferredCurrency

Added tests for improved unit test coverage

PR #533, opened 2020-05-27
Adding --gasprofile to tests

justinjmoses:
Provides a new flag `--gasprofile` to `buidler test` that only runs a handful of tests and outputs their gas usage. (Once #503 is merged, this can simply use that to show gas and this just becomes a small filter of tests to run to profile gas)

image

### DIFF
--- a/contracts/Proxy.sol
+++ b/contracts/Proxy.sol
@@ -54,7 +54,8 @@ contract Proxy is Owned {
         }
     }
 
-    function() external payable {        
+    // solhint-disable no-complex-fallback
+    function() external payable {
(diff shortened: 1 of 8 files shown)


contracts/Proxy.sol
```diff
@@ -54,7 +54,8 @@ contract Proxy is Owned {
         }
     }
 
-    function() external payable {        
+    // solhint-disable no-complex-fallback
```
justinjmoses, line 57: FYI @hav-noms this was introduced after your coverage PR #529 but it's not obvious why. I guess it was being missed somehow and now we reduced the logic it realizes it is complex? 🤷

PR #536, opened 2020-05-29
Sip 15 liquidation mechanism

hav-noms:
https://sips.synthetix.io/sips/sip-15


contracts/Liquidations.sol
```diff
@@ -0,0 +1,232 @@
...
+import "./EternalStorage.sol";
+
+// Inheritance
+import "./interfaces/ISynthetix.sol";
+import "./interfaces/ISynthetixState.sol";
+import "./interfaces/IIssuer.sol";
+
+
+// https://docs.synthetix.io/contracts/Liquidations
+// contract Liquidations is Owned, MixinResolver {
```
justinjmoses, line 21: don't forget to remove

publish/src/commands/deploy.js
```diff
@@ -472,6 +472,28 @@ const deploy = async ({
 		});
 	}
 
+	const liquidations = await deployContract({
+		name: 'Liquidations',
+		args: [account, addressOf(resolverAddress)],
+	});
+
+	const eternalStorageLiquidations = await deployContract({
+		name: 'EternalStorageLiquidations',
```
justinjmoses, line 481: Thanks for naming this correctly :pray:

jacko125: How about the key in addressResolver? 

`bytes32 private constant CONTRACT_LIQUIDATIONETNERALSTORAGE = "LiquidationEternalStorage"; `

publish/src/commands/deploy.js
```diff
@@ -472,6 +472,28 @@ const deploy = async ({
...
+		source: 'EternalStorage',
+		args: [account, addressOf(liquidations)],
+	});
+
+	if (liquidations && eternalStorageLiquidations) {
+		await runStep({
+			contract: 'EternalStorage',
```
justinjmoses, line 488: this needs to be the name `EternalStorageLiquidations`

contracts/interfaces/ILiquidations.sol
```diff
@@ -0,0 +1,24 @@
...
+    // Mutative Functions
+    function flagAccountForLiquidation(address account) external;
+
+    // Restricted: used internally to Synthetix
+    function removeAccountInLiquidation(address account) external;
+
+    function checkAndRemoveAccountInLiquidation(address account) external;
+
+    // owner only
+    function setLiquidationDelay(uint time) external;
```
justinjmoses, line 17: You probably don't need these here in the public interface

contracts/Liquidations.sol
```diff
@@ -0,0 +1,232 @@
...
+    struct LiquidationEntry {
+        uint deadline;
+    }
+
+    bytes32 private constant sUSD = "sUSD";
+
+    /* ========== ADDRESS RESOLVER CONFIGURATION ========== */
+
+    bytes32 private constant CONTRACT_SYNTHETIX = "Synthetix";
+    bytes32 private constant CONTRACT_LIQUIDATIONETNERALSTORAGE = "EternalStorageLiquidations";
```
justinjmoses, line 35: [np] can you follow the naming convention `CONTRACT_<label>`?

contracts/Liquidations.sol
```diff
@@ -0,0 +1,232 @@
...
+
+
+// https://docs.synthetix.io/contracts/Liquidations
+// contract Liquidations is Owned, MixinResolver {
+contract Liquidations is Owned, MixinResolver, ILiquidations {
+    using SafeMath for uint;
+    using SafeDecimalMath for uint;
+
+    struct LiquidationEntry {
+        uint deadline;
```
justinjmoses, line 27: are you planning on adding more to this?

hav-noms: we should remove the struct

contracts/Liquidations.sol
hav-noms: We're going with EternalStorageLiquidations right?

contracts/Liquidations.sol
```diff
@@ -0,0 +1,232 @@
...
+    // Add internal viewer for synthetix / issuer contract to check _OpenForLiqudation(collateralRatio)
+
+    // get liquidationEntry for account
+    // returns deadline = 0 when not set
+    function _getLiquidationEntryForAccount(address account) internal view returns (LiquidationEntry memory _liquidation) {
+        _liquidation.deadline = liquidationEternalStorage().getUIntValue(_getKey(LIQUIDATION_DEADLINE, account));
+    }
+
+    function _getKey(bytes32 _scope, address _account) internal pure returns (bytes32) {
+        return keccak256(abi.encodePacked(_scope, _account));
```
justinjmoses, line 111: perhaps assume `LIQUIDATION_DEADLINE` is the scope? unless you're planning to add other storage keys?

contracts/Liquidations.sol
```diff
@@ -0,0 +1,232 @@
...
+        liquidationDelay = time;
+
+        // emit event
+        emit LiquidationDelayUpdated(time);
+    }
+
+    // Collateral ratio is higher when less collateral backing debt
+    // Upper bound is 1.0 (100%)
+    function setLiquidationRatio(uint _liquidationRatio) external onlyOwner {
+        require(_liquidationRatio < MAX_LIQUIDATION_RATIO, "ratio > MAX_LIQUIDATION_RATIO");
```
justinjmoses, line 125: [np] the require reason should say `>=` - same as the ones below

contracts/Liquidations.sol
```diff
@@ -0,0 +1,232 @@
...
+        // emit event
+        emit LiquidationPenaltyUpdated(penalty);
+    }
+
+    /* ========== MUTATIVE FUNCTIONS ========== */
+    function flagAccountForLiquidation(address account) external {
+        LiquidationEntry memory liquidation = _getLiquidationEntryForAccount(account);
+
+        // Don't set liquidation if account flagged already
+        if (liquidation.deadline > 0) return;
```
justinjmoses, line 153: worth reverting here you think?

contracts/Liquidations.sol
```diff
@@ -0,0 +1,232 @@
...
+import "./MixinResolver.sol";
+import "./interfaces/ILiquidations.sol";
+
+// Libraries
+import "./SafeDecimalMath.sol";
+
+// Internal references
+import "./EternalStorage.sol";
+
+// Inheritance
```
justinjmoses, line 14: internal references?

publish/src/commands/deploy.js
justinjmoses: The key should match what's used in the deploy script, and it's `EternalStorageLiquidations` so that's what should be here as well...

justinjmoses: That way the code below will know that `EternalStorageLiquidations` is required in the `AddressResolver` and will add it

publish/src/commands/deploy.js
hav-noms: done

contracts/Liquidations.sol
justinjmoses: Yep. It's the right pattern to use - source then implementation - like `Proxy:Synthetix` or `Synth:sUSD`

contracts/Liquidations.sol
jacko125: We removed `isFlagged` from the struct to just use deadline is set by checking `deadline > 0` (something there). 

Will refactor the struct out, but possible to add back in if we have to store the person who flags the liquidation.

contracts/Liquidations.sol
hav-noms: done

contracts/interfaces/ILiquidations.sol
hav-noms: removed

contracts/Liquidations.sol
hav-noms: fixed

contracts/Exchanger.sol
```diff
@@ -163,6 +163,20 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
 
+    function hasWaitingPeriodOrSettlementOwing(address account, bytes32 currencyKey) external view returns (bool) {
+        if (maxSecsLeftInWaitingPeriod(account, currencyKey) > 0) {
+            return true;
+        }
+
+        (uint reclaimAmount, uint rebateAmount, ) = settlementOwing(account, currencyKey);
+
+        if (reclaimAmount > 0 || rebateAmount > 0) {
+            return true;
```
justinjmoses, line 174: Even if there's nothing owing, there could still be settlement entries needed to be cleared out. You could instead check for `entries == 0` if you wanted to be sure there's no settlement entries pending. Otherwise, if you just want to ensure they can burn what they have, then I'd suggest checking `reclaimAmount > 0` because a rebate would be an `issue` instead.

contracts/Liquidations.sol
```diff
@@ -0,0 +1,236 @@
...
+import "./SafeDecimalMath.sol";
+
+// Internal references
+import "./EternalStorage.sol";
+import "./interfaces/ISynthetix.sol";
+import "./interfaces/ISynthetixState.sol";
+import "./interfaces/IExchangeRates.sol";
+import "./interfaces/IIssuer.sol";
+
+import "@nomiclabs/buidler/console.sol";
```
justinjmoses, line 18: Perhaps just wrap a big TEMP around this... or we could also make it a lint failure though we'd need to write a custom solhint plugin for this...

contracts/Synthetix.sol
```diff
@@ -581,6 +609,52 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         return true;
     }
 
+    // totalSupply effectiveValue checks for rates stale
```
justinjmoses, line 612: This has implications on SIP-48 #451

contracts/Liquidations.sol
hav-noms: lint failure sounds good

contracts/Liquidations.sol
hav-noms: for the internal View typing for the AddressResolver

contracts/Exchanger.sol
jacko125: Updated to only check `reclaimAmount > 0`

contracts/Synthetix.sol
jacko125: Might be able to move the implementation of liquidateDelinquentAccount into `issuer` along with SIP-48

contracts/Issuer.sol
```diff
@@ -240,6 +275,64 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+    ) external onlySynthetix returns (uint totalRedeemed, uint amountToLiquidate) {
+        // Ensure waitingPeriod and sUSD balance is settled as burning impacts the size of debt pool
+        require(!exchanger().hasWaitingPeriodOrSettlementOwing(liquidator, sUSD), "sUSD needs to be settled");
+        ILiquidations _liquidations = liquidations();
+
+        // Check account is liquidation open
+        require(_liquidations.isOpenForLiquidation(account), "Account not open for liquidation");
+
+        // require liquidator has enough sUSD
+        require(IERC20(address(synthetix().synths(sUSD))).balanceOf(liquidator) >= susdAmount, "Not enough sUSD");
```
justinjmoses, line 299: you could probably save an extra step here by having `SynthsUSD` added as a dep directly... though perhaps worth trying with `--gas` and compare the diff?

contracts/Issuer.sol
```diff
@@ -240,6 +275,64 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+
+        uint liquidationPenalty = _liquidations.liquidationPenalty();
+
+        // What is the value of their SNX balance in sUSD?
+        uint collateralValue = exchangeRates().effectiveValue("SNX", synthetix().collateral(account), sUSD);
+
+        // What is their debt in sUSD?
+        (uint debtBalance, uint totalDebtIssued) = synthetix().debtBalanceOfAndTotalDebt(account, sUSD);
```
justinjmoses, line 307: This will be an internal call once I'm done with SIP-48 (#451)

contracts/Issuer.sol
```diff
@@ -189,7 +204,26 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+        uint existingDebt,
+        uint totalDebtIssued
+    ) internal {
+        // liquidation requires sUSD to be already settled / not in waiting period
+
+        // Remove liquidated debt from the ledger
+        _removeFromDebtRegister(burnForAddress, amount, existingDebt, totalDebtIssued);
+
+        // synth.burn does a safe subtraction on balance (so it will revert if there are not enough synths).
+        synthetix().synths(sUSD).burn(liquidator, amount);
```
justinjmoses, line 223: (see comment below)

contracts/Liquidations.sol
```diff
@@ -0,0 +1,272 @@
...
+    /* ========== MODIFIERS ========== */
+    modifier onlySynthetixOrIssuer() {
+        bool isSynthetix = msg.sender == address(synthetix());
+        bool isIssuer = msg.sender == address(issuer());
+
+        require(isSynthetix || isIssuer, "Liquidations: Only the synthetix or Issuer contract can perform this action");
+        _;
+    }
+
+    modifier rateNotStale(bytes32 currencyKey) {
```
justinjmoses, line 260: is this used?

contracts/Liquidations.sol
```diff
@@ -0,0 +1,272 @@
...
+
+    function synthetixState() internal view returns (ISynthetixState) {
+        return ISynthetixState(requireAndGetAddress(CONTRACT_SYNTHETIXSTATE, "Missing SynthetixState address"));
+    }
+
+    function issuer() internal view returns (IIssuer) {
+        return IIssuer(requireAndGetAddress(CONTRACT_ISSUER, "Missing Issuer address"));
+    }
+
+    function exchangeRates() internal view returns (IExchangeRates) {
```
justinjmoses, line 79: Is this still used?

contracts/Issuer.sol
jacko125: SIP-48 would move all the synths mapping into Issuer, that would be better as well when it just access it internally. 

What did you mean by added as a dep directly as well?

contracts/Liquidations.sol
jacko125: good pickup, this was added as per commit `2a9f779a50bc7246aaa6950c7c6c76f5c31992d4` to add snx stale check to flag liquidations.

Flagging account for Liquidation should check SNX is not stale. This should be a test case in SIP-15 and tested.


As per SIP-44, these mutative actions should add 
- `synthetix.liquidateDelinquentAccount` check for `systemStatus().requireSystemActive()`
- `liquidations.flagAccountForLiquidation` check for `systemStatus().requireSystemActive()`

contracts/Liquidations.sol
jacko125: yep

hav-noms: looks like its not. We can pull these out now

contracts/Issuer.sol
jacko125: A few other places where it calls synthetix() will also be an internal call once SIP-48 is done.

contracts/Issuer.sol
justinjmoses: I meant you could have SynthsUSD as a direct dependency in Issuer, but you're right, probably easier to just wait for SIP-48

contracts/Liquidations.sol
justinjmoses: (I unresolved this)

contracts/Exchanger.sol
```diff
@@ -163,6 +163,20 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
+        }
+
+        (uint reclaimAmount, , ) = settlementOwing(account, currencyKey);
+
+        if (reclaimAmount > 0) {
+            return true;
+        }
+
+        return false;
```
justinjmoses, line 177:
```suggestion
        return reclaimAmount > 0;
```

contracts/Liquidations.sol
jacko125: Updated to use it to check `rateNotStale("SNX")`

contracts/Exchanger.sol
jacko125: @justinjmoses  Updated this , could you check and resolve?

contracts/Liquidations.sol
```diff
@@ -0,0 +1,293 @@
...
+        require(time <= MAX_LIQUIDATION_DELAY, "Must be less than 1 month");
+        require(time >= MIN_LIQUIDATION_DELAY, "Must be greater than 1 day");
+
+        liquidationDelay = time;
+
+        emit LiquidationDelayUpdated(time);
+    }
+
+    // Accounts Collateral/Issuance ratio is higher when there is less collateral backing their debt
+    // Upper bound liquidationRatio is 1 + penalty (100% + 10% = 110%)
```
zyzek, line 183: This doesn't seem to be enforced? (if it is to be enforced then I suppose to would need to be enforced in `setLiquidationPenalty` as well)

contracts/Liquidations.sol
```diff
@@ -0,0 +1,293 @@
...
+
+    // totalIssuedSynths checks synths for staleness
+    // check snx rate is not stale
+    function flagAccountForLiquidation(address account) rateNotStale("SNX") external {
+        systemStatus().requireSystemActive();
+
+        LiquidationEntry memory liquidation = _getLiquidationEntryForAccount(account);
+        require(liquidation.deadline == 0, "Account already flagged for liquidation");
+
+        uint accountsIssuanceRatio = synthetix().collateralisationRatio(account);
```
zyzek, line 209: The naming is a bit confusing: the variable `accountsIssuanceRatio` should probably be `accountsCollateralisationRatio` to match the function call, and as the issuance ratio is the global target that does not differ from account to account.

contracts/Liquidations.sol
```diff
@@ -0,0 +1,293 @@
...
+        eternalStorageLiquidations().deleteUIntValue(_getKey(LIQUIDATION_DEADLINE, _account));
+        // delete liquidation caller
+        eternalStorageLiquidations().deleteAddressValue(_getKey(LIQUIDATION_CALLER, _account));
+
+        emit AccountRemovedFromLiqudation(_account, now);
+    }
+
+    // Returns the inverse view of the issuanceRatio to collateralRatio (1/v)
+    function inverseRatio(uint value) internal pure returns (uint) {
+        return (SafeDecimalMath.unit().divideDecimalRound(value));
```
zyzek, line 269: This function looks like an inlining candidate: it's used only in one place. Also redundant parens.

contracts/Liquidations.sol
```diff
@@ -0,0 +1,293 @@
...
+     * V = Collateral
+     * P = liquidation penalty
+     * Calculates amount of synths = (t * D - V) / (t - (1 + P))
+     */
+    function calculateAmountToFixCollateral(
+        uint debtBalance,
+        uint collateral
+    ) external view returns (uint) {
+        // What is target ratio ?
+        uint target = SafeDecimalMath.unit().divideDecimal(synthetixState().issuanceRatio());
```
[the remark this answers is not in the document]
zyzek: Although if no refactor is desired, at least `target` could be defined with the `inverseRatio` function. Otherwise that function looks like a candidate for inlining as it's only used in one place.

I'd probably also remove the comment, which is a little uninformative.

contracts/Liquidations.sol
hav-noms: thanks @zyzek will inline it. I though twe might be using it more than once

hav-noms: fixed

contracts/Liquidations.sol
hav-noms: fixed

contracts/Liquidations.sol
jacko125: We should add inverseRatio view back in if the intention was for it to help anyone reading to quickly get the % of the issuanceRatio, ie 800% ?

contracts/Liquidations.sol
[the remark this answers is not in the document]
jacko125: @zyzek  Thanks Anton, I do like the substitution of  `target` with the issuance ratio `r` and simplifying the formula so that we don't have to calculate the target first.

contracts/Liquidations.sol
hav-noms: I thought we might use it more than once internally but its only in 

```
function liquidationCollateralRatio() external view returns (uint) {
        return SafeDecimalMath.unit().divideDecimalRound(liquidationRatio);
}

File: contracts/Liquidations.sol (hunk 10 above)
Comment (justinjmoses) [reply contracts/Liquidations.sol]: Right - it's just not "Inheritance" - so that should be removed

File: contracts/Exchanger.sol [hunk 24]
```diff
@@ -149,6 +149,12 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
         return (reclaimAmount, rebateAmount, numEntries);
     }
 
+    function hasWaitingPeriodOrSettlementOwing(address account, bytes32 currencyKey) external view returns (bool) {
+        (uint reclaimAmount, , ) = settlementOwing(account, currencyKey);
+
+        return reclaimAmount > 0 || maxSecsLeftInWaitingPeriod(account, currencyKey) > 0;
```
justinjmoses, line 155: [np] if they are in the waiting period then it's more gas intensive to also check settlement.

contracts/Issuer.sol
```diff
@@ -189,7 +204,28 @@ contract Issuer is Owned, MixinResolver, IIssuer {
             debtToRemoveAfterSettlement = exchanger().calculateAmountAfterSettlement(from, sUSD, amount, refunded);
         }
 
-        _internalBurnSynths(from, debtToRemoveAfterSettlement, existingDebt, totalSystemValue);
+        uint maxIssuable = synthetix().maxIssuableSynths(from);
```
justinjmoses, line 207: [np] You can move this to `_internalBurnSynths` to de-dupe and save a bit of contract space

contracts/Issuer.sol
```diff
@@ -240,6 +277,68 @@ contract Issuer is Owned, MixinResolver, IIssuer {
...
+    ) external onlySynthetix returns (uint totalRedeemed, uint amountToLiquidate) {
+        // Ensure waitingPeriod and sUSD balance is settled as burning impacts the size of debt pool
+        require(!exchanger().hasWaitingPeriodOrSettlementOwing(liquidator, sUSD), "sUSD needs to be settled");
+        ILiquidations _liquidations = liquidations();
+
+        // Check account is liquidation open
+        require(_liquidations.isOpenForLiquidation(account), "Account not open for liquidation");
+
+        // require liquidator has enough sUSD
+        require(IERC20(address(synthetix().synths(sUSD))).balanceOf(liquidator) >= susdAmount, "Not enough sUSD");
```
justinjmoses, line 301: This would actually fail during SafeMath subtract in `_burnSynthsForLiquidation`: https://github.com/Synthetixio/synthetix/pull/536/files#diff-0430106cb3fbbaf9c3bb9c8d08cb2b7aR225 - so technically not necessary to have this explicit check across two separate contract

contracts/Liquidations.sol
```diff
@@ -0,0 +1,284 @@
...
+        CONTRACT_EXRATES
+    ];
+
+    /* ========== CONSTANTS ========== */
+    uint public constant MAX_LIQUIDATION_RATIO = 1e18; // 100% issuance ratio
+
+    uint public constant MAX_LIQUIDATION_PENALTY = 1e18 / 4; // Max 25% liquidation penalty / bonus
+
+    uint public constant MAX_LIQUIDATION_DELAY = 2629743; // 1 Month
+    uint public constant MIN_LIQUIDATION_DELAY = 86400; // 1 day
```
justinjmoses, line 56: Why not just say `1 days`?

contracts/Synthetix.sol
```diff
@@ -646,4 +665,23 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+    bytes32 internal constant ACCOUNTLIQUIDATED_SIG = keccak256("AccountLiquidated(address,uint256,uint256,address)");
+
+    function emitAccountLiquidated(
+        address account,
+        uint256 snxRedeemed,
+        uint256 amountLiquidated,
+        address liquidator
+    ) internal {
+        proxy._emit(
+            abi.encode(amountLiquidated, snxRedeemed, liquidator),
```
justinjmoses, line 679: ⚠️⚠️⚠️ The order is off here - this should be `snxRedeemed` then `amountLiquidated`, and the test below doesn't catch this: https://github.com/Synthetixio/synthetix/pull/536/files#diff-5d866c3d835a60f39355e8fc3301539eR1029-R1032

contracts/interfaces/IExchanger.sol
```diff
@@ -21,10 +21,7 @@ interface IExchanger {
             uint numEntries
         );
 
-    function feeRateForExchange(bytes32 sourceCurrencyKey, bytes32 destinationCurrencyKey)
```
justinjmoses, line 24: Why is this being removed? I think it may have been a bad merge.

test/contracts/Issuer.js
```diff
@@ -180,6 +181,14 @@ contract('Issuer (via Synthetix)', async accounts => {
 				reason: 'Only the synthetix contract can perform this action',
 			});
 		});
+		it('liquidateDelinquentAccount() cannot be invoked directly by a user', async () => {
```
justinjmoses, line 184: I feel pretty strongly that test coverage belongs in the test spec that is targeting it. That means that tests for invoking liquidations be covered here, and also coverage for burning to target removing the flag, etc. That way, others can reason about where coverage exists and feel safe to mock external classes as much as possible.

test/contracts/Liquidations.js
```diff
@@ -0,0 +1,1227 @@
...
+} = require('./helpers');
+
+const { toBytes32 } = require('../..');
+
+const MockExchanger = artifacts.require('MockExchanger');
+
+contract('Liquidations', accounts => {
+	const [sUSD, SNX] = ['sUSD', 'SNX'].map(toBytes32);
+	const [, owner, oracle, account1, alice, bob, carol, david] = accounts;
+	const [hour, day, week, month] = [3600, 86400, 604800, 2629743];
```
justinjmoses, line 24: Same comment above with "month"

test/contracts/Liquidations.js
```diff
@@ -0,0 +1,1227 @@
...
+			SystemStatus: systemStatus,
+			FeePoolState: feePoolState,
+			Issuer: issuer,
+		} = await setupAllContracts({
+			accounts,
+			synths: ['sUSD'],
+			contracts: [
+				'AddressResolver',
+				'ExchangeRates',
+				'Exchanger', // required for Synthetix to check if exchanger().hasWaitingPeriodOrSettlementOwing
```
justinjmoses, line 57: [np] If that's the only reason, it could be mocked and you use `setup.mockGenericContractFnc` to ensure that the Exchanger mock returns true for `hasWaitingPeriodOrSettlementOwing`

test/contracts/Liquidations.js
```diff
@@ -0,0 +1,1227 @@
...
+				);
+
+				// now have Liquidations resync its cache
+				await liquidations.setResolverAndSyncCache(addressResolver.address, { from: owner });
+			});
+			it('removeAccountInLiquidation() can only be invoked by synthetix', async () => {
+				await onlyGivenAddressCanInvoke({
+					fnc: liquidations.removeAccountInLiquidation,
+					args: [alice],
+					address: owner, // TODO: is this supposed to be synthetix.address
```
justinjmoses, line 322: Why can't you change this to `synthetix.address`?

test/contracts/Liquidations.js
```diff
@@ -0,0 +1,1227 @@
...
+					accounts,
+					reason: 'Liquidations: Only the synthetix or Issuer contract can perform this action',
+				});
+			});
+			it('removeAccountInLiquidation() can only be invoked by issuer', async () => {
+				await onlyGivenAddressCanInvoke({
+					fnc: liquidations.removeAccountInLiquidation,
+					args: [alice],
+					address: owner, // TODO: is this supposed to be issuer.address
```
justinjmoses, line 331: as above but `issuer.address`?

test/contracts/Liquidations.js
```diff
@@ -0,0 +1,1227 @@
...
+	describe('when anyone calls liquidateDelinquentAccount on alice', () => {
+		let exchanger;
+		describe('then do liquidation checks', () => {
+			beforeEach(async () => {
+				exchanger = await MockExchanger.new(synthetix.address);
+				await addressResolver.importAddresses(['Exchanger'].map(toBytes32), [exchanger.address], {
+					from: owner,
+				});
+				// now have synthetix resync its cache
+				await synthetix.setResolverAndSyncCache(addressResolver.address, { from: owner });
```
justinjmoses, line 414: [np] you can do this and the one below in parallel with await `Promise.all([...])`

test/contracts/Liquidations.js
```diff
@@ -0,0 +1,1227 @@
...
+						});
+					});
+
+					describe('when Alice burnSynthsToTarget to fix her c-ratio ', () => {
+						let burnTransaction;
+						beforeEach(async () => {
+							await updateSNXPrice('1');
+							burnTransaction = await synthetix.burnSynthsToTarget({ from: alice });
+						});
+						xit('then AccountRemovedFromLiquidation event is emitted', async () => {
```
justinjmoses, line 661: why pending?

test/contracts/Liquidations.js
```diff
@@ -0,0 +1,1227 @@
...
+									// Record Alices state
+									aliceDebtBefore = await synthetix.debtBalanceOf(alice, sUSD);
+									aliceSNXBefore = await synthetix.collateral(alice);
+
+									// Record Carol State
+									carolSNXBefore = await synthetix.balanceOf(carol);
+
+									// Carol Liquidates Alice
+									for (let i = 0; i < 10; i++) {
+										await synthetix.liquidateDelinquentAccount(alice, sUSD5, {
```
justinjmoses, line 884: `await Promise.all([...])`

justinjmoses, line 884:
```suggestion
await Promise.all(Array(10).fill(0).map(() => synthetix.liquidateDelinquentAccount(alice, sUSD5, {from:carol}))
```

test/contracts/Liquidations.js
```diff
@@ -0,0 +1,1227 @@
...
+											from: carol,
+										});
+									}
```
justinjmoses, line 887:
```suggestion
await Promise.all(Array(10).fill(0).map(() => synthetix.liquidateDelinquentAccount(alice, sUSD5, {from:carol}))
```

test/contracts/Liquidations.js
```diff
@@ -0,0 +1,1227 @@
...
+
+									assert.bnEqual(await sUSDContract.balanceOf(carol), sUSD50);
+
+									// Record Alices state
+									aliceDebtBefore = await synthetix.debtBalanceOf(alice, sUSD);
+									aliceSNXBefore = await synthetix.collateral(alice);
+
+									// Record Carol State
+									carolSNXBefore = await synthetix.balanceOf(carol);
+
```
justinjmoses, line 942: This setup is shared with the section above, so you could group them together with the same setup code under another `describe` and dedupe the setup code in a shared `beforeEach`

test/contracts/Liquidations.js
```diff
@@ -0,0 +1,1227 @@
...
+							await updateSNXPrice('1');
+
+							aliceDebtBalance = await synthetix.debtBalanceOf(alice, sUSD);
+
+							burnTransaction = await synthetix.burnSynths(aliceDebtBalance, { from: alice });
+						});
+						it('then alice has no more debt', async () => {
+							assert.bnEqual(toUnit(0), await synthetix.debtBalanceOf(alice, sUSD));
+						});
+						xit('then AccountRemovedFromLiquidation event is emitted', async () => {
```
justinjmoses, line 739: (ibid)

contracts/Synthetix.sol
```diff
@@ -118,6 +120,10 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
             IRewardsDistribution(requireAndGetAddress(CONTRACT_REWARDSDISTRIBUTION, "Missing RewardsDistribution address"));
     }
 
+    function liquidations() internal view returns (ILiquidations) {
+        return ILiquidations(requireAndGetAddress(CONTRACT_LIQUIDATIONS, "Missing Liquidations address"));
+    }
+
```
justinjmoses, line 126: Looks like you aren't using this at all, it should be removed.

Interestingly I found this by looking through the code-coverage report:
https://codecov.io/gh/Synthetixio/synthetix/compare/20ede9e46a54618facd4c967553cc5ed7fb1448d...225de7c4005b03189b500ecd9b170dab4342f3f1/src/contracts/Synthetix.sol?before=contracts/Synthetix.sol#L123

contracts/Synthetix.sol
jacko125: @justinjmoses  I swapped the order around yesterday and must've not updated it. 

Re tests - the logs are not propagating through the tx.logs , this was something similar to what was happening in the ProxyERC20 tests.

contracts/Exchanger.sol
jacko125: will move the waiting period check to be done before.

contracts/interfaces/IExchanger.sol
jacko125: @hav-noms  Could you check whether this is supposed to be removed from the merging in of `develop` branch ?

contracts/Issuer.sol
jacko125: The require to check their sUSD balance is earlier here as it would save a lot more gas rather than letting `_burnSynthsForLiquidation` revert the transaction after loading all the debtbalance and states.

contracts/Liquidations.sol
jacko125: @zyzek  Correct, what we came across was this won't be strictly enforced if the `liquidationPenalty` was updated afterwards it could still break the limits. 

So then both  `liquidationRatio` and `liquidationPenalty` would need to checked and updated accordingly during an SCCP update.

contracts/Exchanger.sol
jacko125: Good [nit]pickup !

contracts/Issuer.sol
jacko125: If we check for `maxIssuable` in `_internalBurnSynths`, then for `burnSynthsToTarget` it would load the `maxIssuable` again.

contracts/Liquidations.sol
```diff
@@ -0,0 +1,284 @@
...
+        uint deadline = now.add(liquidationDelay);
+
+        _storeLiquidationEntry(account, deadline, msg.sender);
+
+        emit AccountFlaggedForLiquidation(account, deadline);
+    }
+
+    // Internal function to remove account from liquidations
+    // Does not check collateral ratio is fixed
+    function removeAccountInLiquidation(address account) external onlySynthetixOrIssuer {
```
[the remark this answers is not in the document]
jacko125: @justinjmoses  Updated, now only checking is Issuer.

contracts/Synthetix.sol
jacko125: Interesting, the gas coverage report is really helpful, this must've been left during the refactoring of `liquidateDelinquentAccount` to Issuer contract. 

Fixed now.

contracts/Liquidations.sol
```diff
@@ -0,0 +1,293 @@
...
+    uint public constant MIN_LIQUIDATION_DELAY = 86400; // 1 day
+
+    // Storage keys
+    bytes32 public constant LIQUIDATION_DEADLINE = "LiquidationDeadline";
+    bytes32 public constant LIQUIDATION_CALLER = "LiquidationCaller";
+
+    /* ========== STATE VARIABLES ========== */
+    uint public liquidationDelay = 2 weeks; // liquidation time delay after address flagged
+    uint public liquidationRatio = 1e18 / 2; // 0.5 issuance ratio when account can be flagged for liquidation
+    uint public liquidationPenalty = 1e18 / 10; // 10%
```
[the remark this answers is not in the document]
zyzek: Okay so I resolved this to my satisfaction now, I guess I missed the very obvious point that a user who liquidates an account must receive a higher value of SNX than the sUSD they burn to liquidate it. If the formula was
`S = (t * D - V) / (t - P)`, then the user would have to receive `S * P` worth of SNX rather than `S * (1 + P)` in order to make the maths work out that the target ratio is properly restored.

In other words, yes, changing the formula would be more versatile, but it would allow unprofitable liquidations. E.g. If `P` was 50%, then a liquidator burning 100 sUSD would receive back only 50 sUSD worth of SNX. Not sure if there are cases you would want to allow this, except perhaps the instance where an account was completely underwater (which the insurance fund is supposed to cover in the end).

test/contracts/Liquidations.js
jacko125: Don't have the synthetix contract PK to call from?

test/contracts/Liquidations.js
hav-noms: fixed

test/contracts/Liquidations.js
hav-noms: done

contracts/Liquidations.sol
hav-noms: done

test/contracts/Liquidations.js
jacko125: is there a way to make the invoke the call from `issuer.address` ?

test/contracts/Liquidations.js
jacko125: as discussed, this is where decodeLogs / getDecodedLogs would help

test/contracts/Liquidations.js
justinjmoses: What you've done is the best approximation - I'd just make that clear. The TODO is unnecessary.

PR #537, opened 2020-05-29
SIP-53 Binary Options

justinjmoses:
Accompanied by documentation at synthetixio/synthetix-docs#16, and a SIP at https://sips.synthetix.io/sips/sip-53.


contracts/BinaryOption.sol
```diff
@@ -0,0 +1,163 @@
+pragma solidity ^0.5.16;
+
+// Libraries
+import "./SafeDecimalMath.sol";
+
+// Internal references
+import "./BinaryOptionMarket.sol";
+
+contract BinaryOption {
```
justinjmoses, line 9: Is it an instance of `IERC20`? If so can you explicitly use the interface from openzeppelin?

justinjmoses: Additionally, can you expose and extend from an interface for `BinaryOption` for third party consumption?

contracts/BinaryOption.sol
```diff
@@ -0,0 +1,163 @@
...
+
+    constructor(address initialBidder, uint initialBid) public {
+        market = BinaryOptionMarket(msg.sender);
+        bidOf[initialBidder] = initialBid;
+        totalBids = initialBid;
+    }
+
+    /* ========== VIEWS ========== */
+
+    function price() public view returns (uint) {
```
justinjmoses, line 41: [np] I know you're using this internally as well but we've been preferring to have `external` and writing an internal function for any internal references to reduce gas.

contracts/BinaryOption.sol
```diff
@@ -0,0 +1,163 @@
...
+    function refund(address bidder, uint newRefund) external onlyMarket {
+        // The safe subtraction will catch refunds that are too large.
+        bidOf[bidder] = bidOf[bidder].sub(newRefund);
+        totalBids = totalBids.sub(newRefund);
+    }
+
+    // This must only be invoked after bidding.
+    function claim(address claimant) external onlyMarket returns (uint optionsClaimed) {
+        uint claimable = claimableBy(claimant);
+        // No options to claim? Nothing happens.
```
justinjmoses, line 75: Qu: should we revert here? Is it more instructive to a user?

contracts/BinaryOptionMarket.sol
```diff
@@ -0,0 +1,507 @@
...
+import "./SafeDecimalMath.sol";
+
+// Internal references
+import "./BinaryOptionMarketFactory.sol";
+import "./BinaryOption.sol";
+import "./interfaces/IExchangeRates.sol";
+import "./interfaces/ISynth.sol";
+import "./interfaces/IFeePool.sol";
+
+contract BinaryOptionMarket is Owned, MixinResolver {
```
justinjmoses, line 17: Could you provide an interface here also?

contracts/interfaces/IBinaryOptionMarket.sol
```diff
@@ -0,0 +1,87 @@
+pragma solidity ^0.5.16;
```
justinjmoses, line 1: could you please make this pragma permissive like the other interfaces we have?

contracts/interfaces/IBinaryOption.sol
```diff
@@ -0,0 +1,18 @@
+pragma solidity ^0.5.16;
+
+import "../interfaces/IBinaryOptionMarket.sol";
+import "../interfaces/IERC20.sol";
+
+contract IBinaryOption is IERC20 {
```
justinjmoses, line 6: I've been trying to avoid this approach - using a `contract` here. I since fixed all the interfaces to be `interface` and accept that some contracts implement multiple interfaces and cast them either as their interface or as `IERC20` where necessary. Happy to discuss the trade-offs, but I really would like to avoid this pattern going forward

test/contracts/BinaryOptionMarket.js
```diff
@@ -0,0 +1,1969 @@
...
+			initialShortBid,
+			{ from: initialBidder }
+		);
+
+		market = await BinaryOptionMarket.at(tx.logs[1].args.market);
+		const options = await market.options();
+		long = await BinaryOption.at(options.long);
+		short = await BinaryOption.at(options.short);
+
+		await sUSDSynth.approve(market.address, sUSDQty, { from: initialBidder });
```
justinjmoses, line 145: [np] you can do these in parallel with `await Promise.all`

test/contracts/BinaryOptionMarket.js
```diff
@@ -0,0 +1,1969 @@
...
+		});
+
+		const functions = [
+			['incrementTotalDeposited', []],
+			['decrementTotalDeposited', []],
+			['paused', [false]],
+		];
+
+		for (const f of functions) {
+			await mockGenericContractFnc({
```
justinjmoses, line 161: I'd do these calls in parallel

test/contracts/BinaryOptionMarketFactory.js
```diff
@@ -0,0 +1,1352 @@
...
+				'This action cannot be performed while the contract is paused'
+			);
+		});
+	});
+
+	describe('Market tracking', () => {
+		it('Multiple markets can exist simultaneously, and debt is tracked properly across them.', async () => {
+			const now = await currentTime();
+			const markets = await Promise.all(
+				[toUnit(1), toUnit(2), toUnit(3)].map(price =>
```
justinjmoses, line 708: can you share some of this setup with a `beforeEach` block to avoid duplication of setup in the `it` blocks?

test/contracts/setup.js
```diff
@@ -497,6 +511,10 @@ const setupAllContracts = async ({
 			],
 			deps: ['SystemStatus', 'FeePoolState', 'AddressResolver'],
 		},
+		{
+			contract: 'BinaryOptionMarketFactory',
```
justinjmoses, line 515: Are you also able to add this to the deploy script? https://github.com/Synthetixio/synthetix/blob/develop/publish/src/commands/deploy.js - we should also add something to the publish test to assert that it was deployed correctly: https://github.com/Synthetixio/synthetix/blob/develop/test/publish/index.js

test/contracts/setup.js
```diff
@@ -10,6 +10,8 @@ const {
 
 const SUPPLY_100M = web3.utils.toWei((1e8).toString()); // 100M
 
+const { toUnit } = require('../utils')();
```
justinjmoses, line 13: Not just easier to use `web3.utils.toWei` (if we destruct it to just `toWei`)?

contracts/BinaryOption.sol
zyzek: The market calls this function on the short and long instances and most of the time one of these will have no options claimable. But I have updated the `claimOptions` function in `BinaryOptionMarket` to revert in case the user has no claimable options at all.

I have also made the same change in `exerciseOptions`.

contracts/interfaces/IBinaryOptionMarket.sol
zyzek: Done

test/contracts/setup.js
zyzek: Done

contracts/BinaryOption.sol
zyzek: Done, and I've also updated other places this occurs.

contracts/BinaryOption.sol
zyzek: Done.

test/contracts/BinaryOptionMarket.js
zyzek: Done.

test/contracts/BinaryOptionMarket.js
zyzek: Done

test/contracts/setup.js
zyzek: I had to do a bit of refactoring to get the contract size down, but this has now been addressed.
It seems to deploy correctly on a local chain, but I suspect the way I've done it is not ideal: https://github.com/Synthetixio/synthetix/blob/8ae8111294e60e0dfcc611e856800236a5a044ad/publish/src/commands/deploy.js#L785

In particular:

* Is there a better place for these constants? This is a code smell.
* It looks to me like deployment to live chains involves a manual deployment and then adding a `deploy: false` entry for the new contracts in the `config.json` file for each chain. Is this correct, or is there something else that needs to be done for the testnets and mainnet configurations?
* We still need to choose some reasonable initial values for these parameters.

And a test still needs to be written.

contracts/interfaces/IBinaryOption.sol
zyzek: The new contracts now implement their interfaces, which are proper interfaces.

contracts/BinaryOptionMarketManager.sol
```diff
@@ -0,0 +1,391 @@
...
+            [fees.poolFee, fees.creatorFee, fees.refundFee]
+        );
+        market.setResolverAndSyncCache(resolver);
+        _addMarket(address(market));
+
+        // The debt can't be incremented in the new market's constructor because until construction is complete,
+        // the manager doesn't know its address in order to grant it permission.
+        uint initialDeposit = bids[0].add(bids[1]);
+        totalDeposited = totalDeposited.add(initialDeposit);
+        _sUSD().transferFrom(msg.sender, address(market), initialDeposit);
```
zyzek, line 280: I suppose to avoid requiring the user from having to `approve`, this (and `BinaryOptionMarket.bid`) could perform the transfer directly, though of course this would require modifying the `Synth` code.

contracts/BinaryOptionMarketManager.sol
```diff
@@ -0,0 +1,419 @@
...
+
+    function numMaturedMarkets() external view returns (uint) {
+        return _maturedMarkets.elements.length;
+    }
+
+    function maturedMarkets(uint index, uint pageSize) external view returns (address[] memory) {
+        return _maturedMarkets.getPage(index, pageSize);
+    }
+
+    function _isValidSynth(bytes32 oracleKey) internal view returns (bool) {
```
justinjmoses, line 148: This will allow any rate, even non-synths, like `SNX` and `ETH` - is that intentional?

contracts/BinaryOptionMarket.sol
```diff
@@ -0,0 +1,606 @@
...
+        }
+        return Phase.Expiry;
+    }
+
+    /* ---------- Market Resolution ---------- */
+
+    function _oraclePriceAndTimestamp() internal view returns (uint price, uint updatedAt) {
+        IExchangeRates exRates = _exchangeRates();
+        uint currentRoundId = exRates.getCurrentRoundId(oracleDetails.key);
+        return exRates.rateAndTimestampAtRound(oracleDetails.key, currentRoundId);
```
justinjmoses, line 174: For future notice, we should really expose `getRateAndUpdatedTime` as an external function on `ExchangeRates` rather than having to do this and paying extra gas.

contracts/BinaryOptionMarketManager.sol
zyzek: Yes, but happy to modify this given a reason.

zyzek: Fixed.

contracts/BinaryOptionMarket.sol
zyzek: #560

justinjmoses: OK then!

PR #544, opened 2020-06-04
Upgrading buidler to 1.3.7

justinjmoses:
Fixes invalid opcode checker via https://github.com/nomiclabs/buidler/pull/627


justinjmoses:
Did you happen to see this @cgewecke?

cgewecke:
@justinjmoses Sorry, just ran across this now.

It should work if you blow out the package-lock.json and run `npm install` again. (It did for me on Node 12 anyway). Lock may have gotten corrupted somehow.

Atm it's depositing `eth-gas-reporter` *inside* `buidler-gas-reporter/node_modules` instead of flat packing it at the outer layer as it should. 

This means Buidler's mocha can't locate the reporter in the dep tree as it searches recursively outwards.

justinjmoses:
Good call @cgewecke - I updated to the latest `solidity-coverage` as well :pray:

justinjmoses:
Yeah @zyzek it was a mistake to include those files in this PR (I was switching branches and missed that addition). I've now removed it from this PR.

PR #551, opened 2020-06-10
refactor Token Wrapper out

hav-noms:
- [x] Refactor out the TokenWrapper
- [x] Implement IStakingRewards


contracts/StakingRewards.sol
```diff
@@ -98,25 +71,29 @@ contract StakingRewards is TokenWrapper, RewardsDistributionRecipient {
...
+    }
+
+    /* ========== MUTATIVE FUNCTIONS ========== */
+
+    function stake(uint256 amount) public nonReentrant updateReward(msg.sender) {
         require(amount > 0, "Cannot stake 0");
-        super.stake(amount);
+        _totalSupply = _totalSupply.add(amount);
+        _balances[msg.sender] = _balances[msg.sender].add(amount);
+        stakingToken.safeTransferFrom(msg.sender, address(this), amount);
```
justinjmoses, line 84: qu: is it worth trying to do `transferFromAndSettle` if possible?

justinjmoses: Nvm - having synths directly in here isn't the primary use-case.

contracts/StakingRewards.sol
```diff
@@ -5,79 +5,52 @@ import "openzeppelin-solidity-2.3.0/contracts/math/SafeMath.sol";
...
-        rewardPerTokenStored = rewardPerToken();
-        lastUpdateTime = lastTimeRewardApplicable();
-        if (account != address(0)) {
-            rewards[account] = earned(account);
-            userRewardPerTokenPaid[account] = rewardPerTokenStored;
-        }
-        _;
+    /* ========== VIEWS ========== */
+
+    function totalSupply() public view returns (uint256) {
```
justinjmoses, line 48: Can you make this and the others all `external`?

contracts/interfaces/IStakingRewards.sol
```diff
@@ -2,21 +2,22 @@ pragma solidity >=0.4.24;
 
 
 interface IStakingRewards {
+    // Views
```
justinjmoses, line 5: can you add `totalSupply()` and `balanceOf(...)` here as well? (If it was a full ERC20 token then I'd suggest not to as it should instead be cast to `IERC20` by third parties, but as it's partial I think it makes sense to include them, no?)

hav-noms: Agree

contracts/StakingRewards.sol
hav-noms: totalsupply and balance of are used internally also

hav-noms: Can only do it on stake and withdraw

contracts/interfaces/IStakingRewards.sol
hav-noms: RFAL

contracts/StakingRewards.sol
```diff
@@ -98,25 +71,29 @@ contract StakingRewards is TokenWrapper, RewardsDistributionRecipient {
         return balanceOf(account).mul(rewardPerToken().sub(userRewardPerTokenPaid[account])).div(1e18).add(rewards[account]);
     }
 
-    // stake visibility is public as overriding LPTokenWrapper's stake() function
-    function stake(uint256 amount) public updateReward(msg.sender) {
+    function getRewardForDuration() public view returns (uint256) {
```
jacko125, line 74: the helper view here can be external.

hav-noms: done

contracts/StakingRewards.sol
justinjmoses: but you can refactor those usages to  `_totalSupply` and `_balances` no?

contracts/interfaces/IStakingRewards.sol
justinjmoses: thanks

contracts/StakingRewards.sol
hav-noms: of course! done

hav-noms:
RFAL

PR #552, opened 2020-06-16
Add External Rewards Recovery function

hav-noms:
Just a quicky for the the ERC Recover for external rewards

### DIFF
--- a/contracts/StakingRewards.sol
+++ b/contracts/StakingRewards.sol
@@ -145,7 +145,7 @@ contract StakingRewards is TokenWrapper, RewardsDistributionRecipient {
     }
 
     // Added to support recovering LP Rewards from other systems to be distributed to holders
-    function recoverERC20(address tokenAddress, uint256 tokenAmount) public onlyOwner {
+    function recoverERC20(address tokenAddress, uint256 tokenAmount) external onlyOwner {
         // If it's SNX we have to query the token symbol to ensure its not a proxy or underlying
         bool isSNX = (keccak256(bytes("SNX")) == keccak256(bytes(ERC20Detailed(tokenAddress).symbol())));
         // Cannot recover the staking token or the rewards token
--- a/contracts/StakingRewards.sol
+++ b/contracts/StakingRewards.sol
@@ -59,11 +59,12 @@ contract StakingRewards is TokenWrapper, RewardsDistributionRecipient {
     event Staked(address indexed user, uint256 amount);
     event Withdrawn(address indexed user, uint256 amount);
     event RewardPaid(address indexed user, uint256 reward);
+    event Recovered(address indexed token, uint256 amount);
 
     constructor(
         address _owner,
         address _rewardsDistribution,
-        address _rewardsToken, 
+        address _rewardsToken,
         address _stakingToken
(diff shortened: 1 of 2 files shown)


contracts/StakingRewards.sol
```diff
@@ -141,4 +142,13 @@ contract StakingRewards is TokenWrapper, RewardsDistributionRecipient {
         periodFinish = block.timestamp.add(DURATION);
         emit RewardAdded(reward);
     }
+
+    // Added to support recovering LP Rewards from other systems to be distributed to holders
+    function recoverERC20(address tokenAddress, uint256 tokenAmount) public onlyOwner {
```
zyzek, line 147: If the eventual use case is for these tokens to then be transferred on to some other address, perhaps this function should take an argument for the recipient? Otherwise, looks good.

hav-noms: yeah the path here now is the protocolDAO owns it and can withdraw it then send it to whoever. I think its added security for now to have it pull to the multi-sig first.

hav-noms:
RFAL

test/contracts/StakingRewards.js
```diff
@@ -131,6 +144,56 @@ contract('StakingRewards', async accounts => {
...
 	});
 
+	describe('External Rewards Recovery', () => {
+		const amount = toUnit('5000');
+		beforeEach(async () => {
+			// Send ERC20 to StakingRewards Contract
+			await externalRewardsToken.transfer(stakingRewards.address, amount, { from: owner });
+			assert.bnEqual(await externalRewardsToken.balanceOf(stakingRewards.address), amount);
+		});
+		it('should revert if recovering staking token', async () => {
```
justinjmoses, line 154: can you also add a `onlyGivenAddressCanInvoke` test block?

contracts/StakingRewards.sol
```diff
@@ -141,4 +143,14 @@ contract StakingRewards is TokenWrapper, RewardsDistributionRecipient {
...
```
justinjmoses, line 148: Why not `external`?

hav-noms: fixed

test/contracts/StakingRewards.js
hav-noms: Done

hav-noms:
@justinjmoses RFAL and merge when approved

contracts/StakingRewards.sol
```diff
@@ -141,4 +143,14 @@ contract StakingRewards is TokenWrapper, RewardsDistributionRecipient {
         periodFinish = block.timestamp.add(DURATION);
         emit RewardAdded(reward);
     }
+
+    // Added to support recovering LP Rewards from other systems to be distributed to holders
+    function recoverERC20(address tokenAddress, uint256 tokenAmount) external onlyOwner {
+        // If it's SNX we have to query the token symbol to ensure its not a proxy or underlying
+        bool isSNX = (keccak256(bytes("SNX")) == keccak256(bytes(ERC20Detailed(tokenAddress).symbol())));
```
justinjmoses, line 150: you can probably try regular compare

hav-noms: discussed in core. we're going with keccak256

PR #553, opened 2020-06-17
Sip 58 emit exchange entry settlements

jacko125:


contracts/Exchanger.sol
```diff
@@ -156,6 +156,65 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
         return (reclaimAmount, rebateAmount, numEntries);
     }
 
+    // Internal function to emit events for each individual rebate and reclaim entry
+    function _settlementOwing(address account, bytes32 currencyKey)
```
justinjmoses, line 160: Why not instead have a settlementOwing internal function that takes a boolean - whether or not to emit events - then you don't have to dupe all this code?

contracts/Exchanger.sol
```diff
@@ -497,4 +567,37 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
         systemStatus().requireSynthActive(currencyKey);
         _;
     }
+
+    // ========== EVENTS ==========
+
+    event ExchangeEntryAppended(
```
justinjmoses, line 573: these events all need the `timestamp` of the original exchange, it was in the spec: https://sips.synthetix.io/sips/sip-58

This is what the dapps will use to try to connect the reclaim/rebates to the exchanges themselves cc @clementbalestrat

jacko125: @justinjmoses I realised that for events the better pattern is to use the block timestamp the event was emitted on instead of adding an extra timestamp to the data.

contracts/Exchanger.sol
jacko125: I'm consistently hitting stack too deep error when trying to return extra values the required values in the `settlementOwing` view function instead of a separate mutable function. 

When we try to return a list of reclaim / rebate amounts for each entry to be settled in `settlementOwing` there isn't room for the extra local variables.   

```
function settlementOwing(address account, bytes32 currencyKey)
        public
        view
        returns (
            uint reclaimAmount,
            uint rebateAmount,
            uint numEntries,
            bytes32[] memory srcs,
            uint[] memory reclaims,
            uint[] memory rebates
        )
```

The required info for emitting each entry settled as a combined event are: 
        
```
    event ExchangeEntrySettlement(
        bytes32 src,
        uint256 amount,
        bytes32 dest,
        uint256 reclaim,
        uint256 rebate,
        uint256 srcRoundIdAtPeriodEnd,
        uint256 destRoundIdAtPeriodEnd,
        uint256 exchangeTimestamp
    );
```

contracts/ExchangeState.sol
```diff
@@ -8,18 +8,7 @@ import "./interfaces/IExchangeState.sol";
...
-        bytes32 dest;
-        uint amountReceived;
-        uint exchangeFeeRate;
-        uint timestamp;
-        uint roundIdForSrc;
-        uint roundIdForDest;
-    }
-
-    mapping(address => mapping(bytes32 => ExchangeEntry[])) public exchanges;
+    mapping(address => mapping(bytes32 => IExchangeState.ExchangeEntry[])) public exchanges;
```
jacko125, line 11: @justinjmoses  If ExchangeState isn't redeployed it should be fine. The updated `IExchangeState` interface will be compiled when the new `Exchanger` contract is deployed.

contracts/Exchanger.sol
```diff
@@ -497,4 +583,39 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
+        address indexed account,
+        bytes32 src,
+        uint256 amount,
+        bytes32 dest,
+        uint256 amountReceived,
+        uint256 exchangeFeeRate,
+        uint256 roundIdForSrc,
+        uint256 roundIdForDest
```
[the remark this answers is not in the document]
jacko125: I think it's easy for the dapps / synthetix exchange to correlate `ExchangeEntryAppended` event using the block it was included in. @clementbalestrat  any thoughts on this?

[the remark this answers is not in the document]
clementbalestrat: Isn't it supposed to be the same?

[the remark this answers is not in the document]
justinjmoses: The question is do we include `timestamp` in the `ExchangeEntryAppended` event to match the one in `ExchangeEntrySettled` or does the dApp simply lookup the timestamp via an evm call to block.timestamp for `ExchangeEntryAppended`. I think more the former to be explicit, but we want your input @clementbalestrat

justinjmoses:
Updated. Please take a look

justinjmoses:
Gas reports look pretty good! Just a heads up to not merge this into develop yet - let's wait until after the release on Tuesday in case we need to run a hotfix through `develop`

test/contracts/Exchanger.js
```diff
@@ -666,16 +701,43 @@ contract('Exchanger (via Synthetix)', async accounts => {
...
+											args: [
+												account1,
+												sUSD,
+												amountOfSrcExchanged,
+												sEUR,
+												expectedSettlement.reclaimAmount,
+												new web3.utils.BN(0),
+												new web3.utils.BN(1),
+												new web3.utils.BN(3),
+												exchangeTime + 1,
```
justinjmoses, line 736: How do you know it's +1? Is it due to how buidler evm mines?

jacko125: Correct, buidler evm increments each txn / block by 1 second

PR #554, opened 2020-06-18
Use deployedBytecode instead of bytecode during build

K-Ho:
The contract size being logged is the size of `bytecode`, or `initcode`, while the size limitation of `24KB` only applies to the `deployedBytecode`.


justinjmoses:
Closed in favor of #557

PR #559, opened 2020-06-21
Add precommit format

justinjmoses:
This will fail the commit if the lint fails, but the formatted files are not added to the commit. 

This could be fixed by adding a `git add -u` but then that may unintentionally add files to the commit that aren't intended. This should really be amended to "format everything staged and stage those formats". Looking into options if anyone has any good suggestions?


zyzek (dismissed):

hav-noms (dismissed):
You asked us to review even tho still in Draft. Cant wait for this to speed up PR's with all those pesky np's!

package.json
```diff
@@ -55,6 +58,11 @@
 	"directories": {
 		"test": "test"
 	},
+	"husky": {
+		"hooks": {
+			"pre-commit": "npm run lint && npm run format"
```
jacko125, line 63: this would be super helpful

jacko125 (dismissed):

justinjmoses:
Note: I updated this to use `pretty-quick` to do the formatting as well

zyzek (dismissed):

jacko125 (approved):
LGTM, installed husky locally already

ajsantander (dismissed):
Awesome!

PR #562, opened 2020-06-25
Fix gas reports

justinjmoses:
It works by skipping tests contained within a block with `@gas-skip` in the description. 

Here are all the mocha options for reference we could be using: https://mochajs.org/api/mocha

### DIFF
--- a/buidler.config.js
+++ b/buidler.config.js
@@ -195,9 +195,11 @@ task('test')
 		if (gas) {
 			console.log(gray(`Enabling ${yellow('gas')} reports, tests will run slower`));
 			bre.config.gasReporter.enabled = true;
-			console.log(gray(`Ignoring test specs containing`, yellow('@gas-skip')));
-			bre.config.mocha.grep = '@gas-skip';
-			bre.config.mocha.invert = true;
+			if (!grep) {
+				console.log(gray(`Ignoring test specs containing`, yellow('@gas-skip')));
+				bre.config.mocha.grep = '@gas-skip';
+				bre.config.mocha.invert = true;
+			}
 		}
 
 		await runSuper(taskArguments);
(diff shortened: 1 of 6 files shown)


justinjmoses:
There's still work here to try to figure out what's going on, but I suspect it's to do with ganache.

This is an example of the errors picked up in gas mode only: https://travis-ci.org/github/Synthetixio/synthetix/jobs/701160060

If this section (https://github.com/Synthetixio/synthetix/blob/fix-gas-reports/test/contracts/BinaryOption.js#L139-L197) is run in isolation via `describe.only` and `@gas-skip` is removed, then the following is generated:

image

cc @cgwecke @alcuadrado

buidler.config.js
```diff
@@ -195,6 +195,9 @@ task('test')
 		if (gas) {
 			console.log(gray(`Enabling ${yellow('gas')} reports, tests will run slower`));
 			bre.config.gasReporter.enabled = true;
+			console.log(gray(`Ignoring test specs containing`, yellow('@gas-skip')));
+			bre.config.mocha.grep = '@gas-skip';
```
zyzek, line 199: This breaks grep support for gas-instrumented tests. Not the absolute worst thing (we can still use .only).

Approving, conditional on this being acceptable for now (since presumably we're going to come back and fix the gas tests for BinaryOption once we work out why it's doing what it's doing.).

justinjmoses: Good consideration. I will make it only change grep if one isn't already set. 

As for the fix, I'm hopeful that a future version of `buidler-gas-reporter` will be more compatible because I can't isolate it either (even after triple checking the `async/await` pairs)

zyzek: The resolution looks good to me.

Very mysterious case on this one, but I'm wondering if it's related to the fact that the contracts are dynamically-instantiated, since it appears to occur in the BinaryOption and BinaryOptionMarket suites, but not in the BinaryOptionMarketManager one.

justinjmoses: Yeah I was thinking something similar 🤔

PR #572, opened 2020-07-01
xlarge for circleci tests

jacko125:
Task to look at: 

Splitting the test-contracts to run in parallel as previously to speed up tests.


justinjmoses:
drops from 10ms to 8mins - so a 20% gain. not bad

PR #578, opened 2020-07-01
Examples: Adding historical lookup

justinjmoses:
This is a useful lookup state at block feature that is very handy for inspecting older state.

Todo in the future
- Should be able to determine from block number which address of the contract to use from `versions.json` file


justinjmoses:
Using this, I was able to determine that the sETH holder who didn't have enough for the settlement required prior Acrux was because after the waiting period of 1 hour at the time they had a rebate not a reclaim

image

justinjmoses:
And the waitingPeriodSecs: 
image

PR #583, opened 2020-07-02
Sip 65 Decentralized circuit breaker

justinjmoses:
### Answered Questions
1. What factor to use (keeping in mind @jacko125's point about inverse synth movement being higher around the limits) `3`
2. Is it worth tracking the last exchange prices publicly? ✅


contracts/Exchanger.sol
```diff
@@ -232,10 +248,14 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
         require(sourceCurrencyKey != destinationCurrencyKey, "Can't be same synth");
         require(sourceAmount > 0, "Zero amount");
 
-        bytes32[] memory synthKeys = new bytes32[](2);
-        synthKeys[0] = sourceCurrencyKey;
-        synthKeys[1] = destinationCurrencyKey;
-        require(!exchangeRates().anyRateIsStale(synthKeys), "Src/dest rate stale or not found");
+        // NOte: the below technique causes a stack too deep issue, either refactor the whole function or pay the slightly higher gas
```
justinjmoses, line 251: Note this - I will try again before making the PR final

index.js
```diff
@@ -173,8 +173,9 @@ const getSuspensionReasons = ({ code = undefined } = {}) => {
 	const suspensionReasonMap = {
 		1: 'System Upgrade',
 		2: 'Market Closure',
-		3: 'Circuit breaker',
-		99: 'Emergency',
+		55: 'Circuit Breaker (Phase one)', // https://sips.synthetix.io/SIPS/sip-55
```
justinjmoses, line 176: @jacko125 @hav-noms What do you think about this approach? Use SIP numbers for reason codes? I bumped Emergency way up to prevent any overlap with future SIPs, though we could use a low number like `10`. (Note: Market closure suspension doesn't have a SIP that I can see)

jacko125: good idea to use SIP code's as they can be easier to look up and check.

contracts/Exchanger.sol
```diff
@@ -381,7 +415,68 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
         return _internalSettle(from, currencyKey);
     }
 
+    function suspendSynthWithInvalidRate(bytes32 currencyKey) external {
+        systemStatus().requireSystemActive();
+        require(issuer().synths(currencyKey) != ISynth(0), "No such synth");
+        uint rate = exchangeRates().rateForCurrency(currencyKey);
```
jacko125, line 421: rate could be inlined as only used once

contracts/Exchanger.sol
```diff
@@ -381,7 +415,68 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
+            return _isDeviationAboveThreshold(lastRateFromExchange, currentRate);
+        }
+
+        // if no last exchange for this synth, then we need to look up last 3 rates (+1 for current rate)
+        (uint[] memory rates, ) = exchangeRates().ratesAndUpdatedTimeForCurrencyLastNRounds(currencyKey, 4);
+
+        // start at index 1 to ignore current rate
+        for (uint i = 1; i < rates.length; i++) {
+            // ignore any empty rates in the past (otherwise we will never be able to get validity)
+            if (rates[i] > 0 && _isDeviationAboveThreshold(rates[i], currentRate)) {
```
jacko125, line 458: In this loop, the `currentRate` is not updated to check whether the difference between each rates update is above deviation threshold, but compares the difference b/w the currentRate with each previous price.

justinjmoses: Actually this is how it is intended. From the SIP's test cases:

> Then the current rate will be within 1% of the previous three rates, and the exchange will continue, persisting the current rate

contracts/Exchanger.sol
```diff
@@ -171,11 +181,14 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
             // and deduct the fee from this amount using the exchangeFeeRate from storage
             uint amountShouldHaveReceived = _getAmountReceivedForExchange(destinationAmount, exchangeEntry.exchangeFeeRate);
 
-            if (exchangeEntry.amountReceived > amountShouldHaveReceived) {
+            // SIP-65 settlements where the amount at end of waiting period is beyond the threshold, then
+            // settle with no reclaim or rebate
+            bool isAboveThreshold = _isDeviationAboveThreshold(exchangeEntry.amountReceived, amountShouldHaveReceived);
+            if (!isAboveThreshold && exchangeEntry.amountReceived > amountShouldHaveReceived) {
```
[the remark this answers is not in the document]
justinjmoses: Yeah. I typically try to reduce cyclical complexity as a default, but I agree it's more readable your way. I will change it.

contracts/Exchanger.sol
jacko125: All good then 💯

PR #584, opened 2020-07-04
Removing redundant Synthetix reference from Synths

### LINKED ISSUE #579: Synths depend on the Synthetix contract even though they no longer use it.
Removing this AddressResolver dependency would mean that all the Synths do not need to be updated if only the Synthetix contract was updated.

https://github.com/Synthetixio/synthetix/blob/master/contracts/Synth.sol#L32

justinjmoses:
Closes #579


zyzek:
Looks good. Does this mean also that the Synth contract doesn't need Synthetix as a mock in setup.js?

https://github.com/Synthetixio/synthetix/blob/d67a779731186ff1760b7d2be727fa1b1b1fe2a3/test/contracts/setup.js#L475-L479

hav-noms:
LGTM anything else to add to this?

PR #585, opened 2020-07-04
SIP-44: Allow settlement even when exchanging is deactivated

justinjmoses:
Minor update to SIP-44: Helps performing settlement prior to a new release while preventing new exchanges from being added at the same time


contracts/Exchanger.sol
```diff
@@ -586,8 +586,6 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
     }
 
     modifier synthActive(bytes32 currencyKey) {
```
zyzek, line 588: Given that this is only used in one place now, why not just inline the modifier?

PR #592, opened 2020-07-14
Integrate utility

clementbalestrat:


contracts/BinaryOptionMarketData.sol
```diff
@@ -0,0 +1,114 @@
...
+
+    struct OraclePriceAndTimestamp {
+        uint price;
+        uint updatedAt;
+    }
+
+    // used for things that don't change over the lifetime of the contract
+    struct MarketParameters {
+        address creator;
+        IBinaryOptionMarket.Options options;
```
hav-noms, line 35: @zyzek I've refactored to use the Interfaces in the project it seems the opens struct is not accessible from the interface but was here https://github.com/Synthetixio/synthetix-utility/blob/master/contracts/BinaryOptionMarketData.sol#L89 Can you please fix this one

zyzek: Fixed the interfaces and also added some tests.

publish/src/commands/deploy.js
```diff
@@ -1083,6 +1083,23 @@ const deploy = async ({
 		args: [account, account, resolverAddress],
 	});
 
+	// ----------------
+	// SynthUtil setup
+	// ----------------
+	await deployContract({
```
justinjmoses, line 1089: we need to give it the address of the resolver proxy, not the resolver itself.

contracts/DappMaintenance.sol
```diff
@@ -0,0 +1,44 @@
...
+        isPausedMintr = isPaused;
+        emit MintrMaintenance(isPausedMintr);
+    }
+
+    function setMaintenanceModeSX(bool isPaused) external onlyOwner {
+        isPausedSX = isPaused;
+        emit SXMaintenance(isPausedSX);
+    }
+
+    event MintrMaintenance(bool isPaused);
```
justinjmoses, line 42: wanna name this something more generic like staking dApp?

contracts/BinaryOptionMarketData.sol
```diff
@@ -0,0 +1,113 @@
+pragma solidity ^0.5.16;
```
justinjmoses, line 1: what do you think about using a subfolder for these contracts - `views`, `view-helpers` or some such?

zyzek: I had the same thought. It might be time to have a bit more structure in the contracts folder altogether now that that directory is getting rather full.

publish/deployed/kovan/config.json
```diff
@@ -8,6 +8,9 @@
 	"BinaryOptionMarketManager": {
 		"deploy": false
 	},
+	"BinaryOptionMarketData": {
```
justinjmoses, line 11: so you don't actually need to do this - we typically will do this prior to the deploy

publish/deployed/local/config.json
```diff
@@ -14,6 +14,9 @@
 	"BinaryOptionMarketManager": {
 		"deploy": true
 	},
+	"BinaryOptionMarketData": {
```
justinjmoses, line 17: but you do need to do this 👍

PR #596, opened 2020-07-15
added transferOwnership functionality in Owned contract

### LINKED ISSUE #582: Owned.sol to include transferOwnership
Right now ownership transfers is a 2 step process
1. nominateNewOwner
2. acceptOwnership

Proposing to add not replace the functionality for `transferOwnership(address _newOwner)` which without nomination or acceptance will transfer the ownership direct to the given address.
This will speed up protocol upgrades when the automated scripts and potentially the delegated migrated can just assign the ownership without the 2 step process.

dev1644:
Resolves #582

### DIFF
--- a/contracts/Owned.sol
+++ b/contracts/Owned.sol
@@ -12,6 +12,12 @@ contract Owned {
         emit OwnerChanged(address(0), _owner);
     }
 
+    function transferOwnership(address newOwner) external onlyOwner {
+        require(newOwner != address(0), "Owner address cannot be 0");
+        emit OwnerChanged(owner, newOwner);
+        owner = newOwner;
+    }
+
     function nominateNewOwner(address _owner) external onlyOwner {
         nominatedOwner = _owner;
         emit OwnerNominated(_owner);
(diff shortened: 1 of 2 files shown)


justinjmoses:
@dev1644 Please remove the lockfile change from this PR.

dev1644:
Done @justinjmoses , Removed lock-file but can't run coverage as `Heap out of memory` error popping up. Root cause of the error is coming from Node runtime I searched for some solutions -:

- `export NODE_OPTIONS=--max_old_space_size=4096`, increasing maximum memory available at the run time I was able to run the coverage.

- `buidler --max-memory 4096 coverage --network coverage`, I also search through Builder's issues and found out [this](https://buidler.dev/troubleshooting/common-problems.html#out-of-memory-errors-when-compiling-large-projects). I tried to use this by modifying `coverage npm script` to `"coverage": "buidler --max-memory 4096 coverage --network coverage"`, but it was still breaking.

I also monitor the `node` runtime service executes after starting coverage tests, I noticed at the time compilation of coverage contracts the runtime was consuming 2.8 GB of memory, Default memory cap of `node runtime` is ~1.8 GB, so that might be causing the problem.

zyzek:
Node is allocated more memory for coverage now, so if you merge from `develop` this should be resolved.

https://github.com/Synthetixio/synthetix/blob/bf1c38fef3e6a82fffc07aed835a1049dea9cc3c/package.json#L11

justinjmoses:
Thanks @dev1644 however we need to discuss this issue more thoroughly before accepting any fixes to it (please see my comment on the issue). I am closing this for now but we can reopen in the future if need be.

PR #597, opened 2020-07-15
remove public accessor on messageSender from Proxyable

### LINKED ISSUE #580: Remove public accessor on Proxyable.messageSender
This used to be `address messageSender;`

but was updated to `address public messageSender;`

The impact is now messageSender is shown on etherscan showing the last called on the contract. It would be preferable to not show this.

dev1644:
Resolves #580


justinjmoses:
Sorry @dev1644 - even though this is an open issue, it isn't for general consumption. These kinds of changes require planning on our end and testing changes (for legacy support).

PR #612, opened 2020-07-23
Eth Collateral Trial 2

hav-noms:
- Fixes bug https://blog.synthetix.io/bug-disclosure/ 
- Implemented https://sips.synthetix.io/sccp/sccp-27
- Update Require to check sUSD balance in Depot https://github.com/Synthetixio/synthetix/compare/develop...eth-collat-tests?expand=1#diff-733e929dd68f104f2b18041033bb427eR382

Original SIP https://sips.synthetix.io/sips/sip-35

IOSIRO AUDIT https://gist.github.com/syncikin/2bec003f74c83f44a39c2a26afb2e0ac


contracts/interfaces/IDepot.sol
```diff
@@ -1,7 +1,7 @@
 pragma solidity >=0.4.24;
 
+contract IDepot {
```
justinjmoses, line 3: can we please put this back as an interface? those views can be external regardless

PR #613, opened 2020-07-23
Sip 64 flexible contract storage

justinjmoses:
- The creation of an abstract contract `ContractStorage` for future contract storage with `AddressResolver` access (replaces the old `State.sol` approach (see #640)). 
- Adds new `FlexibleStorage` contract for generic `uint`, `int`, `address`, `bytes32` and `bool` storage with `AddressResolver` support
- `FlexibleStorage` is non-permissioned - it cannot be updated and points to the ReadProxyAddressResolver to future-proof it
- Replaces `FeePoolEternalStorage` for synth exchange rates


contracts/FlexibleStorage.sol
```diff
@@ -0,0 +1,166 @@
...
+        bytes32 record,
+        address value
+    ) internal {
+        AddressStorage[_memoizeHash(contractName)][record] = value;
+        emit ValueSetAddress(contractName, record, value);
+    }
+
+    /* ========== VIEWS ========== */
+
+    function getUIntValue(bytes32 contractName, bytes32 record) external view returns (uint) {
```
ajsantander, line 57: Might be safe to revert if hashes[contractName] = 0x0 to avoid silent failures if a requested value does not exist, instead of returning 0.

ajsantander: To implement something like this I'd recommend something like:
```
    function getUIntValue(bytes32 contractName, bytes32 record) external view returns (uint) {
        return UIntStorage[_getHash(contractName)][record];
    }

    function _getHash(bytes32 contractName) internal view returns(bytes32 hash) {
		hash = hashes[contractName];
		require(hash != bytes32(0), "No hash found for the required contractName.");
    }
```

contracts/FlexibleStorage.sol
```diff
@@ -0,0 +1,166 @@
...
+// Internal References
+import "./interfaces/IAddressResolver.sol";
+
+
+// https://docs.synthetix.io/contracts/source/contracts/FlexibleStorage
+contract FlexibleStorage {
+    IAddressResolver public resolverProxy;
+
+    mapping(bytes32 => bytes32) public hashes;
+
```
ajsantander, line 12: Recommend adding a small comment here:
```
       /**
     Maps named properties by type, grouped by hashes.

     Eg.
     uint: mapping(hashes[contractName] => mapping(propertyName => uint))
     address: mapping(hashes[contractName] => mapping(propertyName => address))
     */
```

contracts/FlexibleStorage.sol
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /* ========== VIEWS ========== */
+
+    function getUIntValue(bytes32 contractName, bytes32 record) external view returns (uint) {
+        return UIntStorage[hashes[contractName]][record];
+    }
+
+    function getUIntValues(bytes32 contractName, bytes32[] calldata records) external view returns (uint[] memory) {
+        uint[] memory results = new uint[](records.length);
```
ajsantander, line 62: I have a feeling that these for loops could be avoided, thus producing a significant gas optimization. Let me look into it...

contracts/FlexibleStorage.sol
```diff
@@ -0,0 +1,166 @@
...
+        }
+        return results;
+    }
+
+    function getAddressValue(bytes32 contractName, bytes32 record) external view returns (address) {
+        return AddressStorage[hashes[contractName]][record];
+    }
+
+    function getAddressValues(bytes32 contractName, bytes32[] calldata records) external view returns (address[] memory) {
+        address[] memory results = new address[](records.length);
```
ajsantander, line 74: Minor gas optimization here if instead of using this pattern:
```
    function getValues(bytes32 register, bytes32[] calldata keys) external view returns (uint256[] memory) {
        uint256[] memory results = new uint256[](keys.length);
        
        for (uint i = 0; i < keys.length; i++) {
            results[i] = data[register][keys[i]];
        }
        
        return results;
    }
```
You declare the substructure outside of the loop, like this:
```
function getValues1(bytes32 register, bytes32[] calldata keys) external view returns (uint256[] memory) {
        uint256[] memory results = new uint256[](keys.length);
        
        mapping(bytes32 => uint256) storage subdata = data[register];
        for (uint i = 0; i < keys.length; i++) {
            results[i] = subdata[keys[i]];
        }
        
        return results;
    }
```

ajsantander: Player around with it in Remix and it shows ~200/300 gas optimization.

contracts/FlexibleStorage.sol
justinjmoses: We prefer to document over in docs rather than inline. We do make exceptions with complex code but I don't think this is that necessary.

contracts/FlexibleStorage.sol
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    function setAddressValue(
+        bytes32 contractName,
+        bytes32 record,
+        address value
+    ) external onlyContract(contractName) {
+        _setAddressValue(contractName, record, value);
+    }
+
```
ajsantander, line 115: I know that generics are not available in Solidity, but since everything in the EVM is ultimately stored as bytes32, couldn't we avoid duplicate code with something like:
```
bytes32 public value;
    
    function setUintValue(uint _value) public {
        setValue(bytes32(_value));
    }
    
    function setValue(bytes32 _value) public {
        value = _value;
    }
```

contracts/FlexibleStorage.sol
```diff
@@ -0,0 +1,166 @@
...
+        address[] calldata values
+    ) external onlyContract(contractName) {
+        require(records.length == values.length, "Input lengths must match");
+
+        for (uint i = 0; i < records.length; i++) {
+            _setAddressValue(contractName, records[i], values[i]);
+        }
+    }
+
+    function deleteAddressValue(bytes32 contractName, bytes32 record) external onlyContract(contractName) {
```
ajsantander, line 128: Also use `_getHash(...)` when deleting?

contracts/interfaces/IFlexibleStorage.sol
```diff
@@ -0,0 +1,22 @@
+pragma solidity >=0.4.24;
+
+
```
ajsantander, line 3: Add entries for `address` values?

ajsantander: Perhaps also declare FlexibleStorage as `contract FlexibleStorage is IFlexibleStorage` to ensure they're always in sync.

contracts/FlexibleStorage.sol
```diff
@@ -0,0 +1,166 @@
...
+        delete AddressStorage[hashes[contractName]][record];
+        emit ValueDeleted(contractName, record);
+    }
+
+    function migrateContractKey(
```
zyzek, line 133: Given that it is possible to migrate these keys by changing entries in the AddressResolver, I'm wondering if it is ultimately simpler and no less secure to simply allow either an owner or a designated admin address in the resolver to invoke this.

contracts/FlexibleStorage.sol
ajsantander: Continuing discussion in #622

contracts/FlexibleStorage.sol
ajsantander: Continuing discussion in #622

contracts/FlexibleStorage.sol
ajsantander: Experimenting with this. Will open a PR soon with a POC. On the meantime, here's a small experiment:
https://pastebin.com/gfPBT20e

contracts/FlexibleStorage.sol
ajsantander: Continuing discussion in #622

contracts/interfaces/IFlexibleStorage.sol
ajsantander: Continuing discussion in #622

contracts/FlexibleStorage.sol
justinjmoses: Yes but that would means adding the owner feature in here and we're working towards less and less permissioned systems.

contracts/FlexibleStorage.sol
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    /* ========== EVENTS ========== */
+
+    event ValueSetUInt(bytes32 contractName, bytes32 record, uint value);
+    event ValueSetInt(bytes32 contractName, bytes32 record, int value);
+    event ValueSetAddress(bytes32 contractName, bytes32 record, address value);
+    event ValueSetBool(bytes32 contractName, bytes32 record, bool value);
+    event ValueSetBytes32(bytes32 contractName, bytes32 record, bytes32 value);
+    event ValueDeleted(bytes32 contractName, bytes32 record);
```
justinjmoses, line 313: Let's include the delete events for each type and emit the value

contracts/FlexibleStorage.sol
```diff
@@ -0,0 +1,166 @@
...
+
+
+// https://docs.synthetix.io/contracts/source/contracts/FlexibleStorage
+contract FlexibleStorage {
+    IAddressResolver public resolverProxy;
+
+    mapping(bytes32 => bytes32) public hashes;
+
+    mapping(bytes32 => mapping(bytes32 => uint)) internal UIntStorage;
+    mapping(bytes32 => mapping(bytes32 => address)) internal AddressStorage;
```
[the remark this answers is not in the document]
justinjmoses: fixed

contracts/FlexibleStorage.sol
justinjmoses: Updated

zyzek:
lgtm

PR #616, opened 2020-07-24
System settings for sip 64

justinjmoses:
This is a preview of using SIP-64 for `SystemSettings` - a central place to manage SCCPs that would prevent the need to create migration scripts for SCCP changes in a post-DelegatedMigrator world.

## SCCP values (and on-chain checks for a unset value of `0`)

- [x] `SynthetixState.issuanceRatio`
    - [x] No funds at risk with `0` value - means that no one could issue.
- [x] `Exchanger.waitingPeriodSecs` (from SIP-37)
    - [x] `0` may be a valid state in the future - simply disables fee reclamation. No check required.
- [x] `Exchanger.priceDeviationThresholdFactor` (from SIP-65)
    - [x] No funds at rink with `0` value - means SIP-65 would simply be triggered and need to be deactivated
- [x] `FeePool.feePeriodDuration`
   - [x] `0` means fee periods could be closed immediately and new ones opened. This would be disruptive but no funds at risk if it were to occur
- [x] `FeePool.targetThreshold`
   - [x] If FeePool.targetThreshold isn't set and = 0, also safe as it means that there won't be a buffer between target Issuance ratio and when someone claims. Would get picked up pretty quickly by SNX claimers if they fail at 749% to claim.
- [x] `Liquidations.liquidationDelay`
   - [ ] @jacko125 to investigate
- [x] `Liquidations.liquidationRatio`
   - [ ] @jacko125 to investigate
- [x] `Liquidations.liquidationPenalty`
   - [x] Risks if liquidations.liquidationPenalty isn't set and = 0 is minimal as liquidations would only get no rewards in this case.
- [x] `ExchangeRates.rateStalePeriod` #638 
   - [x] No funds at risk with `0` value - simply means no one could issue or exchange
- [x] `Exchanger.exchangeFeeRate(bytes32 => uint)` (these need to be moved here so they can be changed with an SCCP): [changes are in this diff](https://github.com/Synthetixio/synthetix/compare/system-settings-for-sip-64~2...system-settings-for-sip-64)
   - [ ] Risk of stakers not receiving any exchange reward if value is `0`, however `0` could be a valid value in the future for some synths. Open question on whether or not to restrict a `0` rate on-chain
- [x] `Issuer.minimumStakeTime`

## Implications
- Higher gas costs as contracts like `Exchanger` need to go through `FlexibleStorage` to get values
- Defaults cannot be set on construction because `SystemSettings` requires the resolver have `FlexibleStorage` associated, and this won't be synced on construction. Thus, initial defaults need to be setup carefully in deploy scripts and test coverage needs to handle situations where no default has been set.

## Notes
- This approach uses `SystemSettings` which is the owner of these SCCP variables. The idea is that the pDAO would continue to be the owner of `SystemSettings` post-DelegatedMigrator, allowing the pDAO to modify SCCPs without needing to deploy a migration contract and wait for the grace period.
- The contracts that rely on these settings also reach into `FlexibleStorage` directly to read (thereby saving gas), and the `MixinSystemSettings` helps them do that by sharing the getters
- Look at how I tackle setting defaults in both `setup.js` and `deploy.js` - both are tested. The former in the regular testing framework, the latter in the publish tests


jacko125:
image

- `Synthetix.Exchange()`, `settle()` 12%+ and `Synth.transfer()` costs increased quite significant. 

- Possible to have `Issuer` and `Synths` read directly from `Flexible Storage`  instead of extra external call to `SystemSettings` and then the storage. Exchanges, Settlement and Synth transfers are high usage and affects many user's especially with the high gas cost environment we're going to continue having with Ethereum. 
   
- `SystemSettings`  can act as a setter for SCCP's and also for viewing externally as a single place to look at for SCCP values.

justinjmoses:
Yes thanks Jackson. Having `Issuer` and `Synths` read directly from `Flexible Storage` would help save a bit of gas but add to complexity. I will try see if I can come up with a good mid-ground.

justinjmoses:
Def cheaper to have contracts reach into `FlexibleStorage` directly: 

image

contracts/SystemSettingsLib.sol
```diff
@@ -0,0 +1,21 @@
+pragma solidity ^0.5.16;
+
+
+library SystemSettingsLib {
+    bytes32 internal constant CONTRACT_NAME = "SystemSettings";
+
+    bytes32 internal constant SETTING_WAITING_PERIOD_SECS = "waitingPeriodSecs";
+    bytes32 internal constant SETTING_PRICE_DEVIATION_THRESHOLD_FACTOR = "priceDeviationThresholdFactor";
+
+    function contractName() internal pure returns (bytes32) {
```
zyzek, line 10: Why not simply query the internal contants directly? Code size could be a consideration, but most of these seem to be used only very few times.

justinjmoses: As discussed offline, these getters are required sadly. The only other alternative is to make a base contract, which on second thought is probably better. Let me investigate

contracts/Exchanger.sol
```diff
@@ -146,6 +138,22 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
 
+    function getWaitingPeriodSecs() internal view returns (uint) {
+        return flexibleStorage().getUIntValue(SETTING_CONTRACT_NAME, SETTING_WAITING_PERIOD_SECS);
+    }
+
+    function getPriceDeviationThresholdFactor() internal view returns (uint) {
+        return flexibleStorage().getUIntValue(SETTING_CONTRACT_NAME, SETTING_PRICE_DEVIATION_THRESHOLD_FACTOR);
+    }
+
+    function waitingPeriodSecs() external view returns (uint) {
```
justinjmoses, line 149: So even though this view is available on `SystemSettings`, I think including external views on those contracts that need it is useful - it also allows us to add tests on these views that they have the right data injected

contracts/MixinSystemSettings.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity ^0.5.16;
+
+
+contract MixinSystemSettings {
```
justinjmoses, line 4: So theoretically this contract could also have helper functions as well, however if we add `FlexibleStorage` in here then we need to also include it in the list of contracts required from the resolver in the target contract mixing this in, and I'm not sure how to handle that gracefully. I suppose this could append FlexibleStorage to `addressesToCache` (if it was `internal` not `private`) as a function (like `Synth.appendToAddressCache()`) but cognizant about not adding too much to contract sizes. Any thoughts?

contracts/SystemSettings.sol
```diff
@@ -0,0 +1,195 @@
...
+    function setLiquidationPenalty(uint penalty) external onlyOwner {
+        require(penalty <= MAX_LIQUIDATION_PENALTY, "penalty > MAX_LIQUIDATION_PENALTY");
+
+        flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_LIQUIDATION_PENALTY, penalty);
+
+        emit LiquidationPenaltyUpdated(penalty);
+    }
+
+    // ========== EVENTS ==========
+    event WaitingPeriodSecsUpdated(uint waitingPeriodSecs);
```
zyzek, line 187: It might be worth considering either:

a) just having one unified event here like `SettingUpdated(bytes32 indexed setting, bytes data)`; or
b) Just watch for the type-specific updates in the flexible storage contract under the system status namespace.

Then clients can watch for a single event, rather than all the time having to pull down new ABIs as settings are added.

On a more general note, is it actually necessary to emit the storage updating events in flexible storage? If it's just an internal storage mechanism, I'm not sure it's necessary to have the events on the flexible storage unless we're intending to allow mirroring the entire flexible storage state in the subgraph or something. Especially if such updates will always have associated events in their own contexts like these ones here, anyway.

contracts/MixinSystemSettings.sol
zyzek: Could `MixinSystemSettings` itself be a `MixinResolver`? That shouldn't add too much size.

I assume unused internal helper functions are simply optimised out? Maybe that's not reasonable to assume.

contracts/Exchanger.sol
zyzek: This seems like the kind of thing where we should just produce test-specific contracts that publicly expose the things we want to test for, unless you mean specifically in the deployment tests?

contracts/Exchanger.sol
```diff
@@ -488,7 +486,9 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
         } else {
             factor = base.divideDecimal(comparison);
         }
-        return factor >= priceDeviationThresholdFactor;
+        uint _priceDeviationThresholdFactor = getPriceDeviationThresholdFactor();
```
zyzek, line 489: [np] Assigning to a local was not strictly necessary here

contracts/FeePool.sol
```diff
@@ -723,9 +729,9 @@ contract FeePool is Owned, Proxyable, SelfDestructible, LimitedSetup, MixinResol
      * @notice Calculate the collateral ratio before user is blocked from claiming.
      */
     function getPenaltyThresholdRatio() public view returns (uint) {
-        uint targetRatio = synthetixState().issuanceRatio();
+        uint targetRatio = getIssuanceRatio();
```
zyzek, line 732: [np] The local variable isn't strictly necessary

contracts/Exchanger.sol
jacko125: If these don't add much more gas to the contract executions and `Exchanger` size can fit, I think it's good having these convenience views here.

contracts/Liquidations.sol
```diff
@@ -35,35 +37,24 @@ contract Liquidations is Owned, MixinResolver, ILiquidations {
...
-
-    uint public constant MAX_LIQUIDATION_DELAY = 30 days;
-    uint public constant MIN_LIQUIDATION_DELAY = 1 days;
 
     // Storage keys
     bytes32 public constant LIQUIDATION_DEADLINE = "LiquidationDeadline";
     bytes32 public constant LIQUIDATION_CALLER = "LiquidationCaller";
 
-    /* ========== STATE VARIABLES ========== */
-    uint public liquidationDelay = 2 weeks; // liquidation time delay after address flagged
```
jacko125, line 63: Will want to check the liquidation defaults are set correctly if the settings contract is upgraded / Need to update this liquidationDelay to default `3` days as with SCCP.

contracts/interfaces/ISystemSettings.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity >=0.4.24;
+
+
+interface ISystemSettings {
```
jacko125, line 4: Did we want to add to the Interface all the other external views that have been added now?

publish/src/commands/deploy.js
```diff
@@ -1269,6 +1271,83 @@ const deploy = async ({
 		}
 	}
 
+	// again after all the resolvers have been set, then we can ensure the defaults of SystemSetting
+	// are set (requires FlexibleStorage to have been correctly configured)
+	if (systemSettings) {
```
jacko125, line 1276: @justinjmoses  I think we can add a double check at this stage in our deploy script during a release, for a `confirmation` and `double check` that the defaults are set correctly and up to date with the SCCP's ? So it reads all the on-chain defaults, spits it out, before allowing script to continue. 

As we saw with https://sips.synthetix.io/sccp/sccp-37 - Silver fee increase when these don't get updated in default can be missed on next release.

Another effective way we've dealt with this has been at the start of the release script to load these and display them at the start.

contracts/Exchanger.sol
justinjmoses: @zyzek In the case of these getters, it seems prudent to keep views on the contracts that need these props, as well as on `SystemSettings`. That said I added `issuanceRatio` to a number of contracts that needed it and it could be a confusing situation to anyone reading the contracts why these views have been duplicated, but honestly probably easy enough to keep them.

@jacko125 I think optimizing function names is overkill considering that the gas implications are in the 10s, not 1000s.

contracts/MixinSystemSettings.sol
justinjmoses: Actually there is a way to do it: https://github.com/Synthetixio/synthetix/pull/616/commits/7aa63f8c0e3341399ceda0b4ff325f7a299c746d 

It adds a bit more contract size onto each target with internal getters but prevents duplicating the logic everywhere for fetching props.

contracts/SystemSettings.sol
justinjmoses: I was just porting over the existing logic as-is from contracts, but now that you mention it, this could be a good step forward.

As for flexible storage - emitting the event is to keep it inline with `EternalStorage` and having a paper trail of events if we ever need to migrate data over to a new instance

contracts/Liquidations.sol
justinjmoses: For this it might be worth updating `flagAccountForLiquidation()` to ensure the values are not 0. Thoughts?

justinjmoses:
Latest results
image

jacko125:
- Risks if `liquidationPenalty` isn't set and = `0` is minimal as liquidations would only get no rewards in this case. 
- If `FeePool.targetThreshold` isn't set and = `0`, also safe as it means that there won't be a buffer between target Issuance ratio and when someone claims. Would get picked up pretty quickly by SNX claimers if they fail at 749% to claim.

PR #619, opened 2020-07-27
Experimental POC on deploying and testing in forked chains

ajsantander:
POC (not for merging)

This PR adds `--fork` options to some of the available commands/scripts:
* publish/deploy
* test:testnet

### publish/deploy
```
node publish/index deploy --network mainnet --deployment-path publish/deployed/mainnet --fork
```
Performs a "fake" mainnet deploy. Potential failures could be caught before the actual deploy. Would allow to run tests on the simulated deploy before doing the real deploy.

### test:testnet
```
yarn test:testnet --network mainnet --fork
```
Runs `test:testnet` on mainnet, without actually spending any funds, provided that the address in .env has SNX.

UPDATE: After considering this POC, we agreed that it is indeed a very interesting approach. We will open a new POC v2 with:
* Start the fork as a separate command, so that multiple commands like deploy, test, etc can be run on the same forked chain.
* Let commands detect when running on a fork and use that as a provider instead of firing up a new one.
* When forked, do not use .env, and instead pwn synthetixDAO account for full access and control.


publish/src/commands/deploy.js
```diff
@@ -241,6 +243,14 @@ const deploy = async ({
 		aggregatedPriceResults = padding + aggResults.join(padding);
 	}
```
ajsantander, line 245: This doesn't have to do with forking really, I just thought it'd be nice to display the Eth balance of the account that will be used on the deploy, and even abort if the account doesn't have enough balance to complete the deploy.

I'm just using an arbitrary 5 Eth for now.

publish/src/commands/deploy.js
```diff
@@ -241,6 +243,14 @@ const deploy = async ({
...
 
+	const balance = parseInt(
+		w3utils.fromWei(await deployer.web3.eth.getBalance(account), 'ether'),
+		10
+	);
+	if (balance < 5) {
+		throw new Error(`Deployer account's ETH balance is too low, ${balance}`);
+	}
+
```
ajsantander, line 253: This :point_up:  doesn't have to do with forking really, I just thought it'd be nice to display the Eth balance of the account that will be used on the deploy, and even abort if the account doesn't have enough balance to complete the deploy.

I'm just using an arbitrary 5 Eth for now.

test/testnet/index.js
```diff
@@ -144,6 +157,42 @@ program
...
+				sources['Synthetix'].abi,
+				targets['ProxyERC20'].address
+			);
+
+			const Depot = new web3.eth.Contract(sources['Depot'].abi, targets['Depot'].address);
+
+			let snxBalance = parseInt(
+				Web3.utils.fromWei(await Synthetix.methods.balanceOf(owner.address).call(), 'ether'),
+				10
+			);
```
ajsantander, line 171: Bringing the Synthetix contract up here, allows for an SNX balance calculation :point_up_2: 

If the SNX balance is lower than a certain threshold (arbitrary for now), SNX is purchased from the Depot contract. This wouldn't work in a fork of mainnet I think, but there should be an alternative mechanism for this case :point_down:

publish/src/Deployer.js
```diff
@@ -42,8 +43,15 @@ class Deployer {
...
 		this.contractDeploymentGasLimit = contractDeploymentGasLimit;
 
+		let provider;
+		if (fork) {
+			provider = getForkedChainProvider({ providerUrl, privateKey });
+		} else {
+			provider = new Web3.providers.HttpProvider(providerUrl);
+		}
+
 		// Configure Web3 so we can sign transactions and connect to the network.
```
ajsantander, line 53: Give provider to Deployer instead?

publish/src/commands/deploy.js
ajsantander: Dont fail, just warn.

test/testnet/index.js
```diff
@@ -71,7 +77,14 @@ program
 
 			let privateKey = envPrivateKey;
 
-			const web3 = new Web3(new Web3.providers.HttpProvider(providerUrl));
```
ajsantander, line 74: As above, have loadConnection provide a ready provider

ajsantander:
Closing this PR in favor of a POC v2. It's nice to have this one as reference as it is.

PR #626, opened 2020-07-29
New staking incentives

hav-noms:


jacko125:
@hav-noms  Deployments test failing when checking the `rewards.json` with whats on mainnet 

image

publish/deployed/mainnet/rewards.json
```diff
@@ -2,17 +2,17 @@
 	{
 		"name": "sETHUniswapV1",
 		"stakingToken": "0xe9cf7887b93150d4f2da7dfc6d502b216438f244",
-		"rewardsToken": "Synthetix"
+		"rewardsToken": "ProxyERC20"
```
justinjmoses, line 5: This is not required - it's not one of our labels, just something we use internally to check. It should stay as `Synthetix`. See https://github.com/Synthetixio/synthetix/blob/develop/test/deployments/index.js#L109-L124 for the checks

publish/deployed/mainnet/rewards.json
```diff
@@ -22,11 +22,26 @@
 	{
 		"name": "iETH",
 		"stakingToken": "ProxyiETH",
-		"rewardsToken": "Synthetix"
+		"rewardsToken": "ProxyERC20"
+	},
+	{
+		"name": "iETH2",
+		"stakingToken": "ProxyiETH",
+		"rewardsToken": "ProxyERC20"
```
justinjmoses, line 30: These should all be `Synthetix` (because the rewards token is not just within our protocol, it also includes `BAL` for instance`

publish/deployed/mainnet/rewards.json
hav-noms: @justinjmoses We cant. If you set it to Synthetix it deploys the staking contract with the underlying contract address. then on the next Synthetix release all SNX in the staking contract is unrecoverable. I experienced this on KOVAN already

publish/deployed/mainnet/rewards.json
hav-noms: @justinjmoses I agree but the Synthetix refers to the underlying contract not the Synthetix proxy

justinjmoses: Gocha OK so we need to either fix the rewards deploy script or fix the tests. Which do you think we should tackle?

hav-noms:
closing and re-opening at https://github.com/Synthetixio/synthetix/tree/develop-new-staking-incentives

PR #630, opened 2020-07-30
add % weights info to synth.json for sDEFI sCEX (#628)

jacko125:


justinjmoses:
The only thing I'd say is run a formatter on it - the trailing zeros would be removed

PR #637, opened 2020-07-31
Adds a new "publish/fork" command

ajsantander:
Expands on #619 "Experimental POC on deploying and testing in forked chains"

This allows you to run a new `fork` command, and thus keep a forked chain running on `localhost`. With this, the `deploy` and `test:testnet` commands can be run with the `--use-fork` option. When running either of these, the synthetixDAO address will be pwned, and will be used as the owner account instead of the one specified in `.env` via a private key.

Follow these steps to run the experiment:
1) Checkout this branch and run `yarn` or `npm install`
2) Run `node publish/index fork` to start a fork of mainnet on localhost.
3) Change some values to true in `publish/deployed/mainnet/config.json` to indicate which contracts you want to deploy.
4) Run `node publish/index deploy --network mainnet --deployment-path publish/deployed/mainnet --use-fork` to perform a fake deploy on mainnet. Notice that `deployments.json` is modified, but `owner-actions.json` is not, because of the pwning.
5) Run `npm run test:testnet -- --network mainnet --use-fork` to run tests on mainnet with zero costs.


publish/src/commands/fork.js
```diff
@@ -0,0 +1,51 @@
...
+	console.log('Forking:', network, 'at:', blockNumber);
+
+	// TODO: Validate incoming network?
+
+	const providerUrl = `https://${network}.infura.io/v3/${process.env.INFURA_PROJECT_ID}`;
+	const server = ganache.server({
+		fork: providerUrl,
+		gasLimit: 12e6,
+		unlocked_accounts: [
+			'0xeb3107117fead7de89cd14d463d340a2e6917769', // Synthetix protocolDAO
```
justinjmoses, line 16: You can fetch this by requiring `/index.js` and using `getUsers().owner`

package.json
```diff
@@ -88,11 +88,13 @@
...
 		"table": "^5.0.2",
 		"wait-port": "^0.2.2",
-		"web3": "1.2.6"
+		"web3": "^1.2.11"
 	},
 	"dependencies": {
 		"@chainlink/contracts-0.0.3": "npm:@chainlink/contracts@0.0.3",
 		"commander": "5.1.0",
+		"ganache-cli": "6.10.0-beta.2",
+		"ganache-core": "^2.11.0-beta.0",
```
ajsantander, line 97: These dependencies are not capricious changes. Unfortunately, they are necessary for ganache-cli to work correctly. Took me hours of testing to arrive to this conclusion :face_with_head_bandage:

publish/src/Deployer.js
```diff
@@ -45,8 +47,12 @@ class Deployer {
 		// Configure Web3 so we can sign transactions and connect to the network.
 		this.web3 = new Web3(new Web3.providers.HttpProvider(providerUrl));
 
-		this.web3.eth.accounts.wallet.add(privateKey);
-		this.web3.eth.defaultAccount = this.web3.eth.accounts.wallet[0].address;
+		if (useFork) {
+			this.web3.eth.defaultAccount = getUsers({ network, user: 'owner' }).address; // protocolDAO
```
ajsantander, line 51: pwned!

publish/src/commands/deploy.js
```diff
@@ -241,6 +244,16 @@ const deploy = async ({
...
+	const deployerBalance = parseInt(
+		w3utils.fromWei(await deployer.web3.eth.getBalance(account), 'ether'),
+		10
+	);
+	if (deployerBalance < 5) {
+		console.log(
+			yellow(`⚠ WARNING: Deployer account balance could be too low: ${deployerBalance} ETH`)
+		);
+	}
+
```
ajsantander, line 256: Just a nice to have :point_up_2: 
Warns if deployer is low on $$$

publish/src/commands/fork.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
```
ajsantander, line 1: This is the new fork command :point_down: 

It's basically launching ganache-server, which is ganache-cli without the commander (or whatever they use) cli interface. This allows us to fully control the thing from Javascript, without having to do some strange bash gymnastics.

publish/src/util.js
```diff
@@ -83,13 +83,13 @@ const loadAndCheckRequiredSources = ({ deploymentPath, network }) => {
...
 
-const loadConnections = ({ network }) => {
+const loadConnections = ({ network, useFork }) => {
 	if (network !== 'local' && !process.env.INFURA_PROJECT_ID) {
 		throw Error('Missing .env key of INFURA_PROJECT_ID. Please add and retry.');
 	}
 
 	const providerUrl =
-		network === 'local'
+		network === 'local' || useFork
```
ajsantander, line 92: This is interesting :point_up_2: 

`providerUrl` will be `localhost:8545` even if the network is mainnet, or ropsten when forking is active. So, it will still use the network's json files, but instead of going through e.g. an infura provider, it will go thru localhost.

package.json
justinjmoses: :100: 

Shouldn't they be `devDependencies` though?

publish/src/commands/fork.js
```diff
@@ -0,0 +1,55 @@
...
+
+const { getUsers } = require('../../../index.js');
+const ganache = require('ganache-core');
+
+const forkChain = async ({ network, blockNumber }) => {
+	// TODO: Remove or improve
+	console.log(`Forking ${network}...`);
+
+	// TODO: Validate incoming network?
```
justinjmoses, line 10: there's a function called `ensureNetwork` from `utils` you can use

publish/src/util.js
justinjmoses: Good note - can you add a comment about this in the code?

yarn.lock
```diff
@@ -0,0 +1,9820 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
justinjmoses, line 1: Yeah this is the issue with yarn - it adds lock files. We are all using `npm` - did you want to switch over? If not, you can probably run `yarn` with a config to use `package-lock.json`

justinjmoses: it's also causing travis to fail as it auto-detects yarn from the lock file

publish/src/commands/deploy.js
```diff
@@ -260,6 +273,7 @@ const deploy = async ({
 			? green('✅ YES\n\t\t\t\t') + newSynthsToAdd.join(', ')
 			: yellow('⚠ NO'),
 		'Deployer account:': account,
+		'Using a fork': useFork,
```
justinjmoses, line 276: Could you make this more prominent like the dry-run flag and put it up there next to it? Useful to sanity check before a deployment

PR #639, opened 2020-08-04
set bool for liquidation ratio and liquidation delay settings set

jacko125:
Possible idea on setting a flag when critical default settings are set

- FlexibleStorage stores bool values for new key 
- `MixinSystemSettings.sol` checks whether these defaults are set on the `FlexibleStorage` and reverts with reason if not set to protect un-intended scenario where liquidations can happen because of this. 
- Only delay and liquidation ratio is protected now.

- Extra cost of checking for liquidations should be rare as liquidations are not likely at this stage.    

- Add modifier to check `feePoolParams` set for `feePeriodDuration` to ensure we have a period set as having it at `0` will allow period to be closed immediately. 

- If feepool claiming target buffer threshold is not set, ie, `0` then it's safe to have no buffer in that case.  

- Add modifier for liquidation entry to check that the liquidation ratio and delay are set.


justinjmoses:
Thanks @jacko125. I fixed the unit tests pretty easily with https://github.com/Synthetixio/synthetix/pull/639/commits/7e6f400fba3a9828999127fbc172b3e520b8d7e4

However, the deployer fails cause when it tries to determine if the value is set:
image

We can fix this to make the deployer do a check of the prop being set, but do we create a function per setting to do that or make something generic and have the deploy script pass around the bytes32 records? 

In addition, I'm always hesitant about making getters revert (as you can imagine after SIP-48), so I wonder if this is the best way to solve it. I'm thinking the code should simply be more defensive everywhere when using these params rather than just assuming they have been set to a sane value.

jacko125:
@justinjmoses  Any thoughts on the best method to replace the whole `flexibleStorage` that the `SystemSettings` / `AddressResolver` is pointing to for the Liquidations and FeePool tests to check that the reverts are hitting when the defaults are not set in flexibleStorage?

contracts/MixinSystemSettings.sol
```diff
@@ -76,4 +79,24 @@ contract MixinSystemSettings is MixinResolver {
                 keccak256(abi.encodePacked(SETTING_EXCHANGE_FEE_RATE, currencyKey))
             );
     }
+
+    modifier liquidationParamsSet() {
```
justinjmoses, line 83: can't we just check against `0` - or do we need to explicitly know if it's been set?

justinjmoses:
Thanks @jacko125. Just adding a touch more coverage for the new `require`

PR #641, opened 2020-08-05
Upgrade to node v12

justinjmoses:


zyzek:
Should we also update the supported engine directive in package.json?

https://github.com/Synthetixio/synthetix/blob/4e2cdd1e1e16aaadadd296e79728a01414209cd6/package.json#L37

justinjmoses:
No so that directive let's installers of this as a dependency that it requires at least `nodejs v8`. As we don't have any code that's > nodejs 8 specifically, I'd suggest keeping this as low as possible.

PR #644, opened 2020-08-06
Removal of IssuanceEternalStorage in favor of existing FlexibleStorage

justinjmoses:


hav-noms:
This is great, gets rid of one of these Eternals. 

Will you import the existing previous 24 hour issuance records or will those stakers get a mint/burn relief?

justinjmoses:
A relief, but not much of one, only those who minted right before the
upgrade, and if it’s after a few period close then there’s no real gain.

PR #645, opened 2020-08-06
POC for moving synths to own storage contract

justinjmoses:


contracts/Issuer.sol
```diff
@@ -387,51 +371,18 @@ contract Issuer is Owned, MixinResolver, MixinSystemSettings, IIssuer {
         }
     }
 
+    function synths(bytes32 currencyKey) external view returns (ISynth) {
+        return availableSynths().synths(currencyKey);
+    }
+
     /* ========== MUTATIVE FUNCTIONS ========== */
 
     function addSynth(ISynth synth) external onlyOwner {
```
zyzek, line 380: As per above comments: Do these add/remove/list synth functions even belong in Issuer anymore if they can be accessed through the new contract?

contracts/AvailableSynths.sol
```diff
@@ -0,0 +1,109 @@
...
+        require(synthsByAddress[address(synth)] == bytes32(0), "Synth address already exists");
+
+        availableSynths.push(synth);
+        synths[currencyKey] = synth;
+        synthsByAddress[address(synth)] = currencyKey;
+
+        emit SynthAdded(currencyKey, address(synth));
+    }
+
+    function removeSynth(bytes32 contractName, bytes32 currencyKey) external onlyContract(contractName) {
```
zyzek, line 73: It seems a bit permissive to allow any contract in the resolver to add or remove synths from the system. It feels to me like the places these functions used to exist, i.e. in the issuer, are probably no longer needed and should just be moved here as owner actions. Given that they are the only ones requiring an owner on the issuer, the Owned dependency could be removed from the issuer and migrated to this contract, which seems more natural.

contracts/AvailableSynths.sol
```diff
@@ -0,0 +1,109 @@
...
+import "./ContractStorage.sol";
+import "./interfaces/IAvailableSynths.sol";
+
+// Internal references
+import "./interfaces/ISynth.sol";
+import "./interfaces/IERC20.sol";
+
+
+// https://docs.synthetix.io/contracts/source/contracts/AvailableSynths
+contract AvailableSynths is ContractStorage, IAvailableSynths {
```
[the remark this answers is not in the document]
justinjmoses: Yeah good point. I guess there could be value in just having something that's purely `State` for the address resolver. I'm gonna have more of think on this today.

contracts/Issuer.sol
justinjmoses: It may make more sense to have them here I suppose - the hope was to decouple state from a logic contract like Issuer.

contracts/AvailableSynths.sol
justinjmoses: `MixinResolver` requires an owner in order to update the cache, though theoretically we could migrate to using `ReadProxyAddressResolver` and have the `setResolverAndSyncCache` be simply a `resetCache` function that is called after a new address resolver is updated, but it's quite a bit of scope creep here.

contracts/Issuer.sol
```diff
@@ -162,7 +154,7 @@ contract Issuer is Owned, MixinResolver, MixinSystemSettings, IIssuer {
             if (synth == currencyKey) {
                 currencyRate = rates[i];
             }
-            uint totalSynths = IERC20(address(synths[synth])).totalSupply();
+            uint totalSynths = IERC20(address(_availableSynths.synths(synth))).totalSupply();
```
jacko125, line 157: Maybe `_totalIssuedSynths` can move into `availableSynths.sol`  instead of looping through all currencyKeys here and making 40+ external calls.

contracts/Issuer.sol
```diff
@@ -512,7 +463,7 @@ contract Issuer is Owned, MixinResolver, MixinSystemSettings, IIssuer {
         _setLastIssueEvent(from);
 
         // Create their synths
-        synths[sUSD].issue(from, amount);
+        availableSynths().synths(sUSD).issue(from, amount);
```
jacko125, line 466: Not sure about the further increase in gas costs for issuing and burning debt for all holders every week here compared to having an internal mapping of Synths inside `issuer.sol`.  

As discussed, the actual cost when `Issuer.sol` is re-deployed to `addSynths` back is negligible as deployer will be owner at this stage.

justinjmoses:
image

contracts/Issuer.sol
justinjmoses: Yeah I was thinking the same thing: https://github.com/Synthetixio/synthetix/pull/645/commits/98907f0955cabc54885d14c893e7f0d6eaa285f9

justinjmoses: (which is similar)

justinjmoses:
After the gas opt:

image

justinjmoses:
Closing this due to gas increase concerns.

PR #647, opened 2020-08-07
[DRAFT] Limit Orders

clementbalestrat:
Still a few tests to add


contracts/LimitOrders.sol
```diff
@@ -0,0 +1,130 @@
...
+import "./Pausable.sol";
+import "./Proxyable.sol";
+import "./SelfDestructible.sol";
+
+// Internal references
+import "./interfaces/ISynthetix.sol";
+import "./interfaces/IAddressResolver.sol";
+
+
+contract LimitOrders is Owned, Proxyable, SelfDestructible, Pausable {
```
justinjmoses, line 14: Do we want `Pausable`? If so we should use the constructor as well

contracts/LimitOrders.sol
```diff
@@ -0,0 +1,130 @@
...
+    /* ========== TYPES ========== */
+
+    struct LimitOrder {
```
justinjmoses, line 17: Let's create an interface `ILimitOrders` and this struct can live in there

contracts/LimitOrders.sol
```diff
@@ -0,0 +1,130 @@
...
+        uint256 weiDeposit;
+        uint256 executionFee;
+    }
+
+    /* ========== STATE VARIABLES ========== */
+
+    IAddressResolver public addressResolverProxy;
+
+    uint256 public orderCount;
+    bytes32 internal constant CONTRACT_SYNTHETIX = "Synthetix";
```
justinjmoses, line 32: If this has dependencies it needs *from* the AddressResolver, it needs to use `MixinResolver` - see other uses

contracts/LimitOrders.sol
```diff
@@ -0,0 +1,130 @@
...
+        uint256 minDestinationAmount;
+        uint256 weiDeposit;
+        uint256 executionFee;
+    }
+
+    /* ========== STATE VARIABLES ========== */
+
+    IAddressResolver public addressResolverProxy;
+
+    uint256 public orderCount;
```
justinjmoses, line 31: Is it `orderCount` or just `orderID`? It doesn't decrement when an order is cancelled so I think it's better as just an incrementing ID like Chainlink Aggregators use

contracts/ProxyLimitOrders.sol
```diff
@@ -0,0 +1,46 @@
+pragma solidity ^0.5.16;
+
+// Inheritance
+import "./Owned.sol";
+
+// Internal references
+import "./Proxyable.sol";
+
+
+contract ProxyLimitOrders is Owned {
```
justinjmoses, line 10: We don't need a new Proxy source for this - we can just deploy a new instance of the Proxy contract and use that, no?

Also, if we use `delegatecall` below then it will use this contract forevermore as storage - not something we currently do. If we want that approach we should create a new Proxy for this, but name it something generic

test/contracts/setup.js
```diff
@@ -179,6 +179,8 @@ const setupContract = async ({
 			toWei('0.02'), // refund fee
 		],
 		BinaryOptionMarketData: [],
+		ProxyLimitOrders: [owner],
```
justinjmoses, line 182: you also need to try call `ProxyLimitOrders.setTarget` after `LimitOrders` is created - that's what those `tryInvocationIfNotMocked` are used for

contracts/LimitOrders.sol
clementbalestrat: I think we do. I like the idea of pausing `newOrder` and `executeOrder` in case of any issue

contracts/LimitOrders.sol
```diff
@@ -0,0 +1,130 @@
...
+        uint destinationAmount = synthetix.exchangeOnBehalf(
+            order.submitter,
+            order.sourceCurrencyKey,
+            order.sourceAmount,
+            order.destinationCurrencyKey
+        );
+        require(destinationAmount >= order.minDestinationAmount, "target price not reached");
+        emit Execute(orderID, order.submitter, msg.sender);
+        gasUsed -= gasleft();
+        uint refund = ((gasUsed + 32231) * tx.gasprice) + order.executionFee; // magic number generated using tests
```
zyzek, line 111: I'm assuming this `32231` is supposed to cover the rest of the execution. The comment should describe what the magic number is instead of just saying that it is a magic number.

contracts/LimitOrders.sol
```diff
@@ -0,0 +1,130 @@
...
+        LimitOrder storage order = orders[orderID];
+        require(order.submitter == msg.sender, "Order already executed or cancelled");
+        msg.sender.transfer(order.weiDeposit);
+        delete orders[orderID];
+        emit Cancel(orderID);
+    }
+
+    function executeOrder(uint orderID) public notPaused {
+        ISynthetix synthetix = _synthetix();
+        uint gasUsed = gasleft();
```
zyzek, line 98: Might as well put `gasleft` before the call to `_synthetix()`

contracts/LimitOrders.sol
```diff
@@ -0,0 +1,130 @@
...
+            order.sourceCurrencyKey,
+            order.sourceAmount,
+            order.destinationCurrencyKey
+        );
+        require(destinationAmount >= order.minDestinationAmount, "target price not reached");
+        emit Execute(orderID, order.submitter, msg.sender);
+        gasUsed -= gasleft();
+        uint refund = ((gasUsed + 32231) * tx.gasprice) + order.executionFee; // magic number generated using tests
+        require(order.weiDeposit >= refund, "Insufficient weiDeposit");
+        order.submitter.transfer(order.weiDeposit - refund);
```
zyzek, line 113: The order should be deleted before the transfer is made. See the comment on the cancel function.

contracts/LimitOrders.sol
```diff
@@ -0,0 +1,130 @@
...
+        delete orders[orderID];
```
zyzek, line 114: Are gas refunds due to deleting the order accounted for in the magic number above?

contracts/LimitOrders.sol
```diff
@@ -0,0 +1,130 @@
...
+        require(orderID <= orderCount, "Order does not exist");
+        LimitOrder storage order = orders[orderID];
+        require(order.submitter != address(0), "Order already executed or cancelled");
+        uint destinationAmount = synthetix.exchangeOnBehalf(
+            order.submitter,
+            order.sourceCurrencyKey,
+            order.sourceAmount,
+            order.destinationCurrencyKey
+        );
+        require(destinationAmount >= order.minDestinationAmount, "target price not reached");
```
zyzek, line 108: If the price transiently moved into the right range, and then back out before the order could be filled, it will not trigger. Fixing this would of course require a much deeper integration of this component into the system so that you could perform a price look-back with an oracle round id. Probably out of scope for now, but just thought I'd mention it.

contracts/LimitOrders.sol
zyzek, line 113: I'm wondering given that the executor has every incentive to just use up all the gas they can to fill the transaction quickly, whether it might be cheaper in the end to just give the executor everything and save the call to refund the remaining dust. See previous comments on the gas logic.

contracts/LimitOrders.sol
zyzek, line 14: If we're adapting to the usual Synthetix proxy pattern, then along with inheriting Proxyable, all instances of `msg.sender` should be replaced with `messageSender`, and we need to encode the events to be passed to `Proxy._emit` as per the description here: https://docs.synthetix.io/contracts/source/contracts/proxy/#_emit . You can check, for example, Synthetix.sol for a previous implementation of the concept.

zyzek:
Needs more test cases.

contracts/LimitOrders.sol
clementbalestrat: Yeah right, I didn't think about that. I'm going to put it as it was initially: `latestID`

contracts/ProxyLimitOrders.sol
clementbalestrat: Agreed. Let's just use the existing one.

contracts/LimitOrdersState.sol
```diff
@@ -0,0 +1,113 @@
...
+        bytes32 sourceCurrencyKey;
+        uint256 sourceAmount;
+        bytes32 destinationCurrencyKey;
+        uint256 minDestinationAmount;
+        uint256 weiDeposit;
+        uint256 executionFee;
+    }
+
+    /* ========== STATE VARIABLES ========== */
+    IAddressResolver public addressResolverProxy;
```
justinjmoses, line 25: You don't need this - MixinResolver exposes you `resolver`

contracts/LimitOrdersState.sol
```diff
@@ -0,0 +1,113 @@
...
+    bytes32 internal constant CONTRACT_LIMITORDERS = "LimitOrders";
+    bytes32[24] private addressesToCache = [CONTRACT_LIMITORDERS];
+
+    uint256 public latestID;
+    mapping(uint256 => LimitOrder) public orders;
+
+    /* ========== CONSTRUCTOR ========== */
+
+    constructor(address _resolver) public MixinResolver(_resolver, addressesToCache) {
+        addressResolverProxy = IAddressResolver(_resolver);
```
justinjmoses, line 35: as above, not needed.

contracts/LimitOrdersState.sol
```diff
@@ -0,0 +1,113 @@
...
+
+// Internal references
+import "./interfaces/ILimitOrdersState.sol";
+import "./interfaces/IAddressResolver.sol";
+
+
+contract LimitOrdersState is MixinResolver {
+    /* ========== STRUCTS ========== */
+
+    struct LimitOrder {
```
justinjmoses, line 14: this can go into `ILimitOrdeersState`

contracts/LimitOrdersState.sol
```diff
@@ -0,0 +1,113 @@
...
+        orders[latestID] = LimitOrder(
+            _submitter,
+            _sourceCurrencyKey,
+            _sourceAmount,
+            _destinationCurrencyKey,
+            _minDestinationAmount,
+            _weiDeposit,
+            _executionFee
+        );
+        return latestID;
```
justinjmoses, line 94: You should probably add events on storage and deletion so we can track what's in here.

contracts/interfaces/ILimitOrdersState.sol
```diff
@@ -0,0 +1,44 @@
...
+    function deleteOrder(uint256 orderID, address submitter) external returns (uint);
+
+    // View functions
+    function getLatestID() external view returns (uint);
+
+    function getOrder(uint _orderID)
+        external
+        view
+        returns (
+            address,
```
justinjmoses, line 36: you can name these variables for readability

contracts/interfaces/ILimitOrdersState.sol
```diff
@@ -0,0 +1,44 @@
...
+        address submitter;
+        bytes32 sourceCurrencyKey;
+        uint256 sourceAmount;
+        bytes32 destinationCurrencyKey;
+        uint256 minDestinationAmount;
+        uint256 weiDeposit;
+        uint256 executionFee;
+    }
+
+    // Mutative functions
```
justinjmoses, line 16: typically we put the views first, then the mutative (or restricted) functions

publish/src/commands/deploy.js
```diff
@@ -1113,6 +1113,31 @@ const deploy = async ({
 		args: [account, resolverAddress],
 	});
 
+	// --------------------
+	// LimitOrders Setup
+	// --------------------
+	const proxyLimitOrders = await deployer.deployContract({
+		name: 'ProxyLimitOrders',
```
justinjmoses, line 1120: As the `name` is different from the `source`, you need to provide the `source` so the script knows what to use (see how `ReadProxyAddressResolver` achieves this)

publish/src/commands/deploy.js
```diff
@@ -1113,6 +1113,31 @@ const deploy = async ({
...
+	if (proxyLimitOrders && limitOrders) {
+		await runStep({
+			contract: 'ProxyLimitOrders',
+			target: proxyLimitOrders,
+			read: 'target',
+			expected: input => input === addressOf(limitOrders),
+			write: 'setTarget',
+			writeArg: addressOf(limitOrders),
+		});
+	}
```
justinjmoses, line 1139: what about `LimitOrderState`?

publish/src/commands/deploy.js
clementbalestrat: Yeah I haven't made the changes in that file yet

justinjmoses:
Closing this for now.

PR #648, opened 2020-08-07
SIP-63: Trading incentives

ajsantander:
Implementation of https://sips.synthetix.io/sips/sip-63 a.k.a. "Trading incentives"

Introduces a new TradingRewards contract, and connects it to the system via Exchanger. The connection can be turned on/off via SystemSettings.

Also adds new `exchangeWithTracking` and `exchangeOnBehalfWithTracking` functions both to Synthetix and Exchanger. These simply emit an event with a tracking code for tracking aggregator volumes.


contracts/TradingRewards.sol
```diff
@@ -0,0 +1,188 @@
...
+
+
+// TODO: Inherit RewardsDistributionRecipient, Pausable
+contract TradingRewards is ITradingRewards, ReentrancyGuard {
+    using SafeMath for uint;
+    using SafeERC20 for IERC20;
+
+    /* ========== STATE VARIABLES ========== */
+
+    uint _currentPeriodID;
```
justinjmoses, line 18: pls add state visibility to prevent warnings.

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,188 @@
...
+import "openzeppelin-solidity-2.3.0/contracts/math/SafeMath.sol";
+import "openzeppelin-solidity-2.3.0/contracts/token/ERC20/ERC20Detailed.sol";
+import "openzeppelin-solidity-2.3.0/contracts/token/ERC20/SafeERC20.sol";
+import "openzeppelin-solidity-2.3.0/contracts/utils/ReentrancyGuard.sol";
+
+import "./interfaces/ITradingRewards.sol";
+
+
+// TODO: Inherit RewardsDistributionRecipient, Pausable
+contract TradingRewards is ITradingRewards, ReentrancyGuard {
```
justinjmoses, line 12: What about Owned?

ajsantander: Pausable is Owned, so Owned is implicitly extended. What's your practice here, do you make these explicit? i.e. in this case yes, Owned would be added to the inheritance list.

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,181 @@
...
+
+    /* ========== STATE VARIABLES ========== */
+
+    uint private _currentPeriodID;
+    mapping(uint => Period) private _periods;
+
+    struct Period {
+        uint recordedFees;
+        uint totalRewards;
+        uint availableRewards;
```
jacko125, line 24: Period's ideally get  "finalised" / "closed" when `notifyRewards` is invoked by the Reward Distribution contract. There could be a flag here inside the `Period` struct to record that the period is finalised.

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,181 @@
...
+        _rewardsDistribution = newRewardsDistribution;
+    }
+
+    /* ========== RESTRICTED FUNCTIONS ========== */
+
+    function notifyRewardAmount(uint newRewards) external onlyRewardsDistribution {
+        uint currentBalance = _rewardsToken.balanceOf(address(this));
+        uint targetBalance = currentBalance.add(newRewards);
+        uint requiredAmount = targetBalance.sub(currentBalance);
+        if (requiredAmount > 0) {
```
jacko125, line 135: The pattern we have in `RewardsDistribution.sol` contract in SNX currently 

1. `transfers` the SNX reward to the recipients in the `distributeRewards` function first, 
2. then notify's the reward amount.

```for (uint i = 0; i < distributions.length; i++) {
            if (distributions[i].destination != address(0) || distributions[i].amount != 0) {
                remainder = remainder.sub(distributions[i].amount);

                // Transfer the SNX
                IERC20(synthetixProxy).transfer(distributions[i].destination, distributions[i].amount);

                // If the contract implements RewardsDistributionRecipient.sol, inform it how many SNX its received.

                // solhint-disable avoid-low-level-calls
                (bool success, ) = distributions[i].destination.call(payload);

                if (!success) {
                    // Note: we're ignoring the return value as it will fail for contracts that do not implement RewardsDistributionRecipient.sol
                }
            }
        }```

We  don't require the `_rewardsToken.safeTransferFrom(msg.sender, address(this), requiredAmount);` if that is pattern we can use.

File: contracts/TradingRewards.sol [hunk 5]
```diff
@@ -0,0 +1,181 @@
...
+            _rewardsToken.safeTransferFrom(msg.sender, address(this), requiredAmount);
+        }
+
+        _currentPeriodID = _currentPeriodID.add(1);
+
+        _periods[_currentPeriodID] = Period({totalRewards: newRewards, availableRewards: newRewards, recordedFees: 0});
```
jacko125, line 141: I think it's good here to track the `totalRewards` in a period and can use this to check whether a `Period` has been finalised and is claimable.

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,181 @@
...
+
+    function recoverTokens(address tokenAddress, uint amount) external onlyOwner {
+        require(tokenAddress != address(_rewardsToken), "Use other function for rewards");
+
+        IERC20(tokenAddress).safeTransfer(msg.sender, amount);
+
+        emit TokensRecovered(tokenAddress, amount);
+    }
+
+    function withdrawRewardsTokensFromCurrentPeriod(uint amount) external onlyOwner {
```
jacko125, line 154: The concept in the SIP should be that the current Period is not claimable as it hasn't been finalised until `notifyRewardAmount` is called to lock in the total fees paid during the period and also the amount of rewards to pay out for that period.

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,181 @@
...
+        uint maxRewards = participationRatio.mul(period.totalRewards);
+
+        uint alreadyClaimed = period.claimedRewardsForAccount[account];
+        availableRewards = maxRewards.sub(alreadyClaimed);
+    }
+
+    /* ========== MUTATIVE FUNCTIONS ========== */
+
+    // TODO: Implement onlyX modifier (onlyExchanger?)
+    function recordExchangeFee(uint amount, address account) external {
```
jacko125, line 88: recordExchangeFee should be protected as per TODO.

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,181 @@
...
+        return addr != address(0);
+    }
+
+    /* ========== VIEWS ========== */
+
+    function rewards(address account, uint periodID) external view returns (uint) {
+        return _calculateAvailableRewardsForAccountInPeriod(account, periodID);
+    }
+
+    function rewardsForPeriods(address account, uint[] calldata periodIDs) external view returns (uint totalRewards) {
```
jacko125, line 57: Another useful `view` for dapp's could be to return a breakdown as an array of the rewards for an `account` for each period from `uint[] calldata periodIDs` provided so the information could be presented in a table. 

We could query `mapping(uint => Period) private _periods` however that would be multiple calls via infura API to get all the data for a list of rewards someone has for each periods.

contracts/TradingRewards.sol
ajsantander: I see. Thanks! I thought it'd be simpler to just pull the needed amount, but if that's how its working elsewhere I'll change it.

contracts/TradingRewards.sol
ajsantander: Yeah, perhaps I should change the `withdrawRewardsTokensFromCurrentPeriod` name to `recoverRewardsTokens`.

My idea here is to disallow removing SNX from closed periods, but allow it for active periods since %'s aren't yet locked in. So, for example, the owner can send 10000 SNX for the next period, activate it, and then change his mind and withdraw 2000 SNX. However, as soon as the next period is activated and this one is claimable, the owner can no longer change his mind. Or, in other words, SNX can only be withdrawn from the current, not claimable period.

Does that make sense?

contracts/TradingRewards.sol
ajsantander: There's already a function for that: https://github.com/Synthetixio/synthetix/blob/79f0a8c5663ab1c7398404fd9863a7b165d9e50b/contracts/TradingRewards.sol#L57

Or did you mean something else?

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,241 @@
...
+        uint recordedFees;
+        uint totalRewards;
+        uint availableRewards;
+        mapping(address => uint) recordedFeesForAccount;
+        mapping(address => uint) claimedRewardsForAccount;
+    }
+
+    address private _rewardsDistribution;
+
+    IERC20 private _rewardsToken;
```
jacko125, line 39: `_rewardsToken` and `_rewardsDistribution` could be public

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,241 @@
...
+
+        _rewardsToken.safeTransfer(account, amountToClaim);
+
+        emit RewardsClaimed(amountToClaim, account, _currentPeriodID);
+    }
+
+    /* ========== RESTRICTED FUNCTIONS ========== */
+
+    // TODO: Implement onlyX modifier (onlyExchanger?)
+    function recordExchangeFeeForAccount(uint amount, address account) external {
```
jacko125, line 173: This would probably be `onlyExchanger` who will notify the trading rewards. 

You could inherit the `MixinResolver` if you want to use the `addressResolver` to keep track of the `Exchanger` contract and have it cached locally instead of having a separate setter for `Exchanger` address.

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,241 @@
...
+        require(_currentPeriodID > 0, "No period available");
```
jacko125, line 174: The require here might not be needed, when the contract is deployed the first `_currentPeriodID = 0` so the trading volume / fees can be recorded in there by default until `notifyRewards` is called.

contracts/TradingRewards.sol
jacko125: I mean one that would return a breakdown of each period (Like an array of each period's rewards amount) how much rewards there are, so not the `total sum`. 

This would help for dApp to display the rewards against each periodID

test/contracts/TradingRewards.js
```diff
@@ -0,0 +1,488 @@
...
+					recordedFees: 0,
+					totalRewards: 0,
+					availableRewards: 0,
+					recordedFeesForAccount: {},
+					claimedRewardsForAccount: {},
+				},
+			],
+		},
+
+		depositRewards: async function({ amount }) {
```
justinjmoses, line 45: [np] can use this syntax: `async depositRewards({ amount }) { ... }`

test/contracts/TradingRewards.js
```diff
@@ -0,0 +1,488 @@
...
+			this.data.rewardsBalance += amount;
+
+			token.transfer(rewards.address, toUnit(amount), { from: owner });
```
justinjmoses, line 48: you're not awaiting this?

test/contracts/TradingRewards.js
```diff
@@ -0,0 +1,488 @@
...
+					await rewards.getAvailableRewardsForAccountForPeriod(account, periodID)
+				);
+			}
+		});
+	};
+
+	// ---------------- TESTS ---------------- //
+
+	// TODO: why should I use this?
+	addSnapshotBeforeRestoreAfterEach();
```
justinjmoses, line 235: If you want to use a `before` to setup everything once and then ensure everything is reset using `evm_restore` before each test (for speed essentially)

test/contracts/TradingRewards.js
```diff
@@ -0,0 +1,488 @@
...
+		});
+
+		describe('when deploying the rewards contract', () => {
+			before('deploy rewards contract', async () => {
+				rewards = await TradingRewards.new(owner, token.address, rewardsDistribution, {
+					from: deployerAccount,
+				});
+			});
+
+			it('ensure only known functions are mutative', () => {
```
justinjmoses, line 262: [np] this is based on the ABI not the deployed instance, so it should live at the top level)

test/contracts/TradingRewards.js
justinjmoses: I'd say you can remove this and use `beforeEach` everywhere instead

contracts/Exchanger.sol
```diff
@@ -329,6 +352,73 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
+        );
+    }
+
+    function _exchangeWithTracking(
+        address from,
+        bytes32 sourceCurrencyKey,
+        uint sourceAmount,
+        bytes32 destinationCurrencyKey,
+        address destinationAddress,
+        bytes32 trackingCode
```
ajsantander, line 398: The parameter was not given to `_exchange(...)` directly, because it causes stack too deep errors.

contracts/Exchanger.sol
```diff
@@ -329,6 +352,72 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
+            exchangeForAddress,
+            sourceCurrencyKey,
+            sourceAmount,
+            destinationCurrencyKey,
+            exchangeForAddress,
+            trackingCode
+        );
+    }
+
+    function _exchangeWithTracking(
```
ajsantander, line 392: Wrapper function to avoid stack too deep errors that where caused when adding an additional parameter to `_exchange(...)`.

contracts/TradingRewards.sol
ajsantander: I usually avoid automatically generated getters, but if this is not the convention used in the codebase I could make a few of these public.

package.json
```diff
@@ -78,6 +78,7 @@
 		"eslint-plugin-prettier": "^2.6.2",
 		"eslint-plugin-promise": "^4.0.1",
 		"eslint-plugin-standard": "^4.0.0",
+		"lodash.clonedeep": "^4.5.0",
```
ajsantander, line 81: This is used in `TradingRewards.js` tests. It allows me to deep clone state when using snapshots.

contracts/Synthetix.sol
```diff
@@ -341,6 +376,37 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+        address indexed account,
+        bytes32 fromCurrencyKey,
+        uint256 fromAmount,
+        bytes32 toCurrencyKey,
+        uint256 toAmount,
+        address toAddress,
+        bytes32 trackingCode
+    );
+    bytes32 internal constant SYNTHEXCHANGEWITHTRACKING_SIG = keccak256(
+        "SynthExchangeWithTracking(address,bytes32,uint256,bytes32,uint256,address,bytes32)"
```
justinjmoses, line 389: I wonder if we should just keep this as `SynthExchange` and provide an extra param?

contracts/Exchanger.sol
```diff
@@ -329,6 +352,72 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
         );
     }
 
+    function exchangeWithTracking(
+        address from,
+        bytes32 sourceCurrencyKey,
+        uint sourceAmount,
+        bytes32 destinationCurrencyKey,
+        address destinationAddress,
```
justinjmoses, line 360: I was considering this before @jacko125 - is the `destinationAddress` sufficient for SIP-63 - it is right?

jacko125: @justinjmoses  @ajsantander  The `destinationAddress` would be sufficient for SIP-63, which is the `messageSender` being passed in currently from `Synthetix.exchange`. The only other case is the `Synthetix.exchangeOnBehalf` which passes in the `exchangeForAddress`

justinjmoses: Right but the SIP has `ISynthetix` and `IExchanger` also taking `address originator` - which is needed here as well right?

contracts/Synthetix.sol
jacko125: The graph should be able to handle that if it knows from which `block` to switch to the new ABI with extra param ?

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,312 @@
...
+
+        if (accountFees == 0) {
+            return 0;
+        }
+
+        uint participationRatio = accountFees.divideDecimal(period.recordedFees);
+        uint maxRewards = participationRatio.multiplyDecimal(period.totalRewards);
+
+        uint alreadyClaimed = period.claimedRewardsForAccount[account];
+        return maxRewards.sub(alreadyClaimed);
```
zyzek, line 154: Given that you can only accrue rewards in the current period, it seems that `alreadyClaimed` will always be zero or the entire balance.

We can simplify this logic by just checking if they've claimed at all -- as per my previous comment, simply set the account's recorded fees for the period to zero once they've claimed, since they can never increase again in claimable periods.

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    function _claimRewards(address account, uint periodID) internal {
+        Period storage period = _periods[_currentPeriodID];
+        require(period.isClaimable, "Period is not claimable");
+
+        uint amountToClaim = _calculateRewards(account, periodID);
+        require(amountToClaim > 0, "No rewards available");
+
+        period.claimedRewardsForAccount[account] = period.claimedRewardsForAccount[account].add(amountToClaim);
```
zyzek, line 180: From `_calculateRewards`, we have (abbreviating `period.claimedRewardsForAccount[account]` as `claimedRewards`):

```solidity
amountToClaim = maxRewards - claimedRewards
```

So here we're effectively setting

```solidity
claimedRewards += amountToClaim
```

which is equivalent to

```solidity
claimedRewards = maxRewards
```

So we could save a state lookup by just doing:

```solidity
amountToClaim = maxRewards - period.claimedRewardsForAccount[account]` 
period.claimedRewardsForAccount[account] = maxRewards
```

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,312 @@
...
+    function recoverEther(address payable recoverAddress) external onlyOwner {
+        require(recoverAddress != address(0), "Invalid recover address");
+
+        uint amount = address(this).balance;
+        recoverAddress.transfer(amount);
+
+        emit EtherRecovered(recoverAddress, amount);
+    }
+
+    function recoverTokens(address recoverAddress, address tokenAddress, uint amount) external onlyOwner {
```
zyzek, line 236: Did we want to disallow recovering the reward tokens?

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,312 @@
...
+        uint freeFromRewards = currentBalance.sub(_balanceLockedForRewards);
+        require(amount <= freeFromRewards, "Insufficient free rewards");
+
+        _rewardsToken.safeTransfer(recoverAddress, amount);
+
+        emit FreeRewardTokensRecovered(recoverAddress, amount);
+    }
+
+    // Warning: calling this on a period will effectively disable it.
+    function recoverAllLockedRewardTokensFromPeriod(address recoverAddress, uint periodID) external onlyOwner {
```
zyzek, line 265: I'm not sure about the optics of allowing incentive tokens to be recovered from any period. Maybe allow this only after enough periods have elapsed?

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,312 @@
...
+    function recordExchangeFeeForAccount(uint amount, address account) external onlyExchanger {
+        Period storage period = _periods[_currentPeriodID];
+
+        period.recordedFeesForAccount[account] = period.recordedFeesForAccount[account].add(amount);
+        period.recordedFees = period.recordedFees.add(amount);
+
+        emit ExchangeFeeRecorded(account, amount, _currentPeriodID);
+    }
+
+    function closeCurrentPeriodWithRewards(uint rewards) external onlyRewardsDistribution {
```
zyzek, line 206: Did we want to have a specified period duration?

contracts/Exchanger.sol
jacko125: Correct, the idea is that `originator` would override `destinationAddress` or `exchangeForAddress` if it is given.

contracts/TradingRewards.sol
ajsantander: A, gtocha. Ok, will do!

test/contracts/TradingRewards.js
ajsantander: Awesome. Thanks!

I ended up using my own snapshoting mechanism that also controls local js state to be compared with the on-chain state.

contracts/TradingRewards.sol
ajsantander: Reward tokens are recovered with either ` recoverFreeRewardTokens` or ` recoverAllLockedRewardTokensFromPeriod`.

Or are you talking about removing the ability to recover reward tokens (SNX) altogether?

contracts/TradingRewards.sol
ajsantander: Yeah, this is something that's not in the SIP and I am lightly suggesting. It's sort of an emergency function that allows the owner to withdraw locked SNX, no matter if it has been "promised" as rewards.

contracts/TradingRewards.sol
ajsantander: I don't think that there is any mention of the concept of time in the SIP, right @jacko125 ?

contracts/TradingRewards.sol
ajsantander: @jacko125 so which option should we go for?

Personally, I prefer keeping as much as possible in a contract private, and explicitly declaring getters. This way, whenever you se _myVar, you know its a variable declared in the contract scope.

contracts/Exchanger.sol
ajsantander: Ok, so to confirm and make sure that there is no confusion in this topic. This is the current state of the implementation, please bear with me in determining if it satisfies all the requirements.

New ABI functions were added, but not regarding any `originator` address (I'll refer to "originator" as the address for which we will be recording fee entries in TradingRewards). Interface changes were only made to accommodate the new `trackingCode` string, but not `originator`:

IExchanger
```
    // UNCHANGED
    function exchange(
        address from,
        bytes32 sourceCurrencyKey,
        uint sourceAmount,
        address destinationAddress
    ) external returns (uint amountReceived);

    // UNCHANGED
    function exchangeOnBehalf(
        address exchangeForAddress,
        address from,
        bytes32 sourceCurrencyKey,
        uint sourceAmount,
        bytes32 destinationCurrencyKey
    ) external returns (uint amountReceived);

    // NEW
    function exchangeWithTracking(
        address from,
        bytes32 sourceCurrencyKey,
        uint sourceAmount,
        address destinationAddress,
        bytes32 trackingCode                                 <-------------------- ADDED
    ) external returns (uint amountReceived);

    // NEW
    function exchangeOnBehalfWithTracking(
        address exchangeForAddress,
        address from,
        bytes32 sourceCurrencyKey,
        uint sourceAmount,
        bytes32 trackingCode                                 <-------------------- ADDED
    ) external returns (uint amountReceived);
```

IExchanger
```
    // UNCHANGED
    function exchange(
        bytes32 sourceCurrencyKey,
        uint sourceAmount,
        bytes32 destinationCurrencyKey
    ) external returns (uint amountReceived);

    // UNCHANGED
    function exchangeOnBehalf(
        address exchangeForAddress,
        bytes32 sourceCurrencyKey,
        uint sourceAmount,
        bytes32 destinationCurrencyKey
    ) external returns (uint amountReceived);

    // NEW
    function exchangeWithTracking(
        bytes32 sourceCurrencyKey,
        uint sourceAmount,
        bytes32 trackingCode                                 <-------------------- ADDED
    ) external returns (uint amountReceived);

    // NEW
    function exchangeOnBehalfWithTracking(
        address exchangeForAddress,
        bytes32 sourceCurrencyKey,
        uint sourceAmount,
        bytes32 trackingCode                                 <-------------------- ADDED
    ) external returns (uint amountReceived);
```

I did not see a need to change any interface to support the `originator` address since it will always be the `destinationAddress` passed on to `Exchanger._exchanger(...)` function.

Exchanger
```
    function _exchange(
        address from,
        bytes32 sourceCurrencyKey,
        uint sourceAmount,
        address destinationAddress
    ) internal returns (uint amountReceived) {
      ...
        // Record fee for trading rewards
        tradingRewards().recordExchangeFeeForAccount(fee, destinationAddress);
      ...
     }
```

This boils down to the following entry points ending up as the `originator`:
* Exchanger.exchange: destinationAddress
* Exchanger.exchangerOnBehalf: exchangeForAddress
* Exchanger.exchangeWithTracking: destinationAddress
* Exchanger.exchangerOnBehalfWithTracking: exchangeForAddress
* Synthetix.exchange: messageSender
* Synthetix.exchangeOnBehalf: exchangeForAddress
* Synthetix.exchangeWithTracking: messageSender
* Synthetix.exchangerOnBehalfWithTracking: exchangeForAddress

contracts/TradingRewards.sol
```diff
@@ -0,0 +1,312 @@
...
+    uint private _balanceLockedForRewards;
+    mapping(uint => Period) private _periods;
+
+    struct Period {
+        bool isClaimable;
+        uint recordedFees;
+        uint totalRewards;
+        uint availableRewards;
+        mapping(address => uint) recordedFeesForAccount;
+        mapping(address => uint) claimedRewardsForAccount; // TODO: Needed? a bool could be enough
```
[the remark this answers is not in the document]
ajsantander: Yeah, this was itching me a bit :)

I changed it to this:
```
    struct Period {
        bool isFinalized;
        uint recordedFees;
        uint totalRewards;
        uint availableRewards;
        mapping(address => uint) unaccountedFeesForAccount;
    }
```

When calculating rewards:
```
        uint accountFees = period.unaccountedFeesForAccount[account];

        if (accountFees == 0) {
            return 0;
        }

        uint participationRatio = accountFees.divideDecimal(period.recordedFees);
        return participationRatio.multiplyDecimal(period.totalRewards);
```

And when claiming rewards:
```
        uint amountToClaim = _calculateRewards(account, periodID);
        require(amountToClaim > 0, "No rewards available");

        period.unaccountedFeesForAccount[account] = 0;
```

contracts/Synthetix.sol
ajsantander: Cool. Checking with TheGraph on what happens to an event that changes signature...

ajsantander: So, I checked with The Graph and yes, as expected, we can overload events. It wouldn't change this code much though, line 379 above :point_up_2: would just become `event SynthExchange(`. We would still need separate `SYNTHEXCHANGEWITHTRACKING_SIG` and `emitSynthExchangeWithTracking(...)`. If we just replaced it with the new signature, the subgraphs wont be able to recognize the old signatures.

contracts/TradingRewards.sol
ajsantander: @justinjmoses what do you think about this? Is it overkill? Does it unnecessarily increase the attack/failure surface? Or do you like having an "emergency exit" like this?

contracts/TradingRewards.sol
jacko125: @ajsantander No not in the SIP, we decided to leave it open so it provides flexibility on how long the period can be , whether its 1 month or 1 week etc, and when the SNX rewards will be sent to fund it.

contracts/TradingRewards.sol
zyzek: Yeah, at least disallowing withdrawal of tokens which have been earmarked to pay out liquidity incentives.

contracts/Exchanger.sol
ajsantander: After discussion, I added the `address originator` parameter to the "withTracking" functions.

contracts/Exchanger.sol
```diff
@@ -320,13 +332,81 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
             exchangeForAddress
         );
+
+        _emitTrackingEvent(trackingCode);
+
+        _processTradingRewards(fee, originator);
+    }
+
+    function _emitTrackingEvent(bytes32 trackingCode) internal {
+        ISynthetixInternal(address(synthetix())).emitExchangeTracking(trackingCode);
```
justinjmoses, line 391: Yeah I like this as a separate event altogether.

contracts/Synthetix.sol
```diff
@@ -372,6 +410,13 @@ contract Synthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         );
     }
 
+    event ExchangeTracking(bytes32 trackingCode);
+    bytes32 internal constant EXCHANGE_TRACKING_SIG = keccak256("ExchangeTracking(bytes32)");
+
+    function emitExchangeTracking(bytes32 trackingCode) external onlyExchanger {
+        proxy._emit(abi.encode(trackingCode), 2, EXCHANGE_TRACKING_SIG, 0, 0, 0);
```
justinjmoses, line 417: Shouldn't we also emit more info about the exchange in case we want to know the volume?

contracts/Exchanger.sol
```diff
@@ -320,13 +332,81 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
+    }
+
+    function _processTradingRewards(uint fee, address originator) internal {
+        if (fee > 0 && originator != address(0)) {
+            address tradingRewardsAddress = address(tradingRewards());
+
+            // Record trading reward fees if TradingRewards contract address is not 0x1
+            bool isTradingRewardsActive = tradingRewardsAddress != address(1);
```
justinjmoses, line 399: not `address(0)`?

justinjmoses: Ah you're proposing to set it to `address(1)` by default... yeah I don't think it makes sense to start introducing a new pattern here. Couldn't `TradingRewards` simply expose a `view` that returns a `bool` whether or not it had an `SNX` balance and that way it would only run when `TradingRewards` was populated with `SNX`?

package.json
justinjmoses, line 81: you can also use `JSON.parse(JSON.stringify(...))` to clone deep without needing another thirdparty dep

ajsantander: That was my initial approach as well, but unfortunately BN's become just hex strings when decoded. To reconstruct them as proper objects, I need a deep copy library.

contracts/Exchanger.sol
```diff
@@ -391,7 +470,11 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
 
         // Remit the fee if required
         if (fee > 0) {
-            remitFee(fee, destinationCurrencyKey);
+            // Normalize fee to sUSD
+            // Note: fee is being reused to avoid stack too deep errors.
+            fee = exchangeRates().effectiveValue(destinationCurrencyKey, fee, sUSD);
+
+            remitFee(fee);
```
ajsantander, line 477: Inline fn contents?

contracts/Exchanger.sol
```diff
@@ -320,13 +337,91 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
+            destinationCurrencyKey,
+            exchangeForAddress
+        );
+
+        _emitTrackingEvent(trackingCode, amountReceived, destinationCurrencyKey);
+
+        _processTradingRewards(fee, originator);
+    }
+
+    function enableTradingRewardRecords(bool enabled) external onlyOwner {
```
justinjmoses, line 395: If we are going this route, the setter should be in `SystemSettings` and follow the same kind of pattern introduced in #616.

contracts/Exchanger.sol
```diff
@@ -320,13 +337,91 @@ contract Exchanger is Owned, MixinResolver, IExchanger {
...
+    function _processTradingRewards(uint fee, address originator) internal {
+        if (fee > 0 && originator != address(0)) {
+            bool isTradingRewardsEnabled = flexibleStorage().getBoolValue(CONTRACT_NAME, TRADING_REWARD_RECORDS_ENABLED);
+
+            if (isTradingRewardsEnabled) {
+                address tradingRewardsAddress = address(tradingRewards());
+
+                // Avoid reverting if the call fails
+                // solhint-disable-next-line
+                (bool success, ) = tradingRewardsAddress.call(
```
justinjmoses, line 418: do we still need this call then? seems excessive to also do a try/catch here effectively.

contracts/Exchanger.sol
```diff
@@ -133,6 +146,10 @@ contract Exchanger is Owned, MixinResolver, MixinSystemSettings, IExchanger {
         return getWaitingPeriodSecs();
     }
 
+    function tradingRewardsEnabled() external view returns (bool) {
+        return getTradingRewardsEnabled();
```
jacko125, line 150: good 👍

contracts/Exchanger.sol
```diff
@@ -290,7 +307,10 @@ contract Exchanger is Owned, MixinResolver, MixinSystemSettings, IExchanger {
         bytes32 destinationCurrencyKey,
         address destinationAddress
     ) external onlySynthetixorSynth returns (uint amountReceived) {
-        amountReceived = _exchange(from, sourceCurrencyKey, sourceAmount, destinationCurrencyKey, destinationAddress);
+        uint fee;
+        (amountReceived, fee) = _exchange(from, sourceCurrencyKey, sourceAmount, destinationCurrencyKey, destinationAddress);
+
+        _processTradingRewards(fee, destinationAddress);
```
jacko125, line 313: Just a note: a normal `exchange` via synthetix.exchange would input  `from` and `destinationAddress` as same address. SIP-63 should have this documented incase we need to show anyone who may be using a contract to trade, we can tell them that the `destinationAddress` is the one who gets the rewards.

jacko125 (approved):

test/contracts/EtherCollateral.js
```diff
@@ -27,6 +27,8 @@ contract('EtherCollateral', async accounts => {
 	const MONTH = 2629743;
 	const YEAR = 31536000;
 
+	const TEST_TIMEOUT = 160e3;
```
jacko125, line 30: Why increase test timeout here in Ethercollateral.js ?

contracts/Exchanger.sol
ajsantander: Sounds good @jacko125 
I've recently made an update to SIP-63. It would be great if you could add this in, since you're the most familiar with the possible usages of that interface.

test/contracts/EtherCollateral.js
ajsantander: As I mentioned earlier in a resolved conversation, I'm using a low end machine to run the test suites while I work, and the previous value was causing these runs to fail. I had to increase it a bit. Checked with Justin tho, seems fine.

contracts/Exchanger.sol
```diff
@@ -331,15 +411,15 @@ contract Exchanger is Owned, MixinResolver, MixinSystemSettings, IExchanger {
             // If, after settlement the user has no balance left (highly unlikely), then return to prevent
             // emitting events of 0 and don't revert so as to ensure the settlement queue is emptied
             if (sourceAmountAfterSettlement == 0) {
-                return 0;
+                return (0, 0);
```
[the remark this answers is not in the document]
justinjmoses: Discussed offline - it's not an attack vector

zyzek (approved):

PR #652, opened 2020-08-10
Fix verify from master

zyzek:


zyzek:
Closing in favour of releasing during Pollux.

PR #655, opened 2020-08-10
Improvements to Synthetix.Exchange gas costs 

jacko125:


justinjmoses:
image

contracts/Exchanger.sol
```diff
@@ -172,7 +172,7 @@ contract Exchanger is Owned, MixinResolver, MixinSystemSettings, IExchanger {
             IExchangeState.ExchangeEntry memory exchangeEntry = _getExchangeEntry(account, currencyKey, i);
```
jacko125, line 172: If `ExchangeState` could be upgraded, instead of making one external call per `_getExchangeEntry`, it would be better to have a function inside `ExchangeState` to return a list of ExchangeEntries given the required "indexes".

PR #656, opened 2020-08-10
Adding release tracking

justinjmoses:
1. Adds `publish/releases.json` to track upcoming releases
2. Adds `metadata` to `deployment.json : sources` for each new deployed source going forward
3. Ensures that `keccak256` of each source is added to `versions.json` on new deploy. 

> Note: the `keccak256` hash of the source is generated as part of the `solc` metadata and as such will differ if whitespace or comments in the source change. 

Having `publish/releases.json` helps with
a) tracking which contracts will be deployed in an upcoming release (no fat finger, also allows for automated fork testing of deploys from develop to mainnet)
b) historical list of previously released versions 
c) list of SIPs for programmatic setting of release notes


justinjmoses:
@ajsantander so we could use this in tandem with your #637 PR - this would track the list of contracts to enable in `config.json` and then we just run this process (build, fork deploy, testnet) in CircleCI on every PR change

publish/src/Deployer.js
```diff
@@ -187,10 +187,15 @@ class Deployer {
...
 		};
 		if (deployed) {
+			// remove the output from the metadata (don't dupe the ABI)
+			delete this.compiled[source].metadata.output;
+
 			// track the new source and bytecode
 			this.deployment.sources[source] = {
 				bytecode: this.compiled[source].evm.bytecode.object,
 				abi: this.compiled[source].abi,
+				source: Object.values(this.compiled[source].metadata.sources)[0],
```
justinjmoses, line 197: Note this adds the keccaked source and metadata to deployment. It does add size so we can potentially move them out to another file if need be, but the metadata can be useful for third party verification

PR #657, opened 2020-08-10
Webpack support

justinjmoses:
This adds a `browser.js` bundle with support for ABIs, addresses, users and synths.


package.json
```diff
@@ -69,15 +73,16 @@
...
 		"dotenv": "^6.1.0",
-		"eslint": "^5.5.0",
+		"eslint": "7.6.0",
 		"eslint-config-prettier": "^3.6.0",
 		"eslint-config-standard": "^12.0.0",
 		"eslint-plugin-havven": "^1.0.0",
-		"eslint-plugin-import": "^2.14.0",
-		"eslint-plugin-node": "^7.0.1",
+		"eslint-plugin-import": "2.22.0",
+		"eslint-plugin-node": "11.1.0",
```
justinjmoses, line 81: Upgrading this helped fix spurious `no-unpublished-require` errors from eslint as it couldn't parse the globs in `package.json` > `files`

justinjmoses:
image

index.js
```diff
@@ -222,18 +288,34 @@ const getSuspensionReasons = ({ code = undefined } = {}) => {
 	return code ? suspensionReasonMap[code] : suspensionReasonMap;
 };
 
+const wrap = ({ network, fs, path }) =>
```
jacko125, line 291: interesting to note that the wrapper sets in the `network` and `path` etc. Would you call it `wrap` or `wrapper` ?

PR #665, opened 2020-08-16
Minor gas optimization in claimRewards

maxsam4:
Reduces gas usage by around 10k in the `claimFees` function. It's lower than what I was aiming for but other optimizations would've required API breaking changes and a storage migration to a different format. It also simplifies the logic and makes it a bit easier to understand IMO.

Gas usage from tests:

|   | Min | Max | Avg |
|:-:|:-:|:-:|:-:|
| Before  | 261181 | 291201 | 270518 |
| After  | 251329 | 281341 | 260658 |


maxsam4:
Closing in favor of #670

PR #669, opened 2020-08-17
Sip 75 keeper isynths phase one

justinjmoses:
WIP

## TBD
- Removal of all centralized oracle code?
    - [ ] Yes
    - [ ] No

## TODO
- [ ] Coverage for new freeze txn and view functions
- [ ] Update of SIP-75 with minor specification changes (to `IExchangeRates`)

## Notes
- `ExchangeRates.RatesUpdated` no longer has inverse rate emitted but rather the companion long rate (to align it with `AnswerUpdated`).
- This change will require co-ordination with [synthetix-subgraph](https://github.com/Synthetixio/synthetix-subgraph/) to ensure that `AnswerUpdated` events on long synths cause a `RateUpdate` to the inverse (will require a reverse lookup on `ExchangeRates` - something like `currenciesUsingAggregator(address aggregator) external view returns (bytes32[] currencies)` to determine if a given aggregator update impacts an


contracts/ExchangeRates.sol
```diff
@@ -110,34 +99,40 @@ contract ExchangeRates is Owned, SelfDestructible, MixinResolver, MixinSystemSet
...
-        // rates already frozen in the current contract that need persisting across the upgrade
-        if (freeze) {
-            emit InversePriceFrozen(currencyKey);
+        if (freezeAtUpperLimit || freezeAtLowerLimit) {
+            // When indicating to freeze, we need to know the rate to freeze it at - either upper or lower
+            // this is useful in situations where ExchangeRates is updated and there are existing inverted
+            // rates already frozen in the current contract that need persisting across the upgrade
 
-            _setRate(currencyKey, freezeAtUpperLimit ? upperLimit : lowerLimit, now);
+            inversePricing[currencyKey].frozenAtUpperLimit = freezeAtUpperLimit;
```
jacko125, line 126: Could save storage gas if only either one `frozenAtUpperLimit` or `frozenAtLowerLimit` is set, not both as they can't both be `true`. 

In this scenario one would be false already and setting `false` again ontop.

jacko125, line 126: Is the intention here to reset both `frozenAtUpperLimit` and `frozenAtLowerLimit` to either false / true regardless?

contracts/ExchangeRates.sol
```diff
@@ -180,8 +176,47 @@ contract ExchangeRates is Owned, SelfDestructible, MixinResolver, MixinSystemSet
...
+        }
+    }
+
+    function currenciesUsingAggregator(address aggregator) external view returns (bytes32[] memory currencies) {
+        uint count = 0;
+        currencies = new bytes32[](aggregatorKeys.length);
+        for (uint i = 0; i < aggregatorKeys.length; i++) {
+            bytes32 currencyKey = aggregatorKeys[i];
+            if (address(aggregators[currencyKey]) == aggregator) {
+                currencies[count++] = currencyKey;
```
jacko125, line 215: @justinjmoses Is it going to be possible to have multiple currencies / prices based off the same `aggregator` address?

contracts/ExchangeRates.sol
```diff
@@ -489,17 +484,60 @@ contract ExchangeRates is Owned, SelfDestructible, MixinResolver, MixinSystemSet
         return false;
     }
 
+    function _rateOrInverted(bytes32 currencyKey, uint rate) internal view returns (uint newRate) {
+        // if an inverse mapping exists, adjust the price accordingly
+        InversePricing memory inverse = inversePricing[currencyKey];
```
jacko125, line 489: @justinjmoses  Actually for `_rateOrInverted` function, this view should be able to determine here whether the currency is an inverse synth or not and return the `rate` earlier. 

This would reduce the computation cost that happens in `_getInvertedRate` if the synth isn't inverse and reduce the cost.

jacko125 (changes requested):
Some points where gas efficiency can be improved for getting rates.

contracts/ExchangeRates.sol
justinjmoses: yes, so for example ETHUSD will support both `sETH` and `iETH`

contracts/ExchangeRates.sol
justinjmoses: When `setInversePricing` is called, it can be frozen upper, lower, or not at all. Keep in mind this function is used both a) when repricing a synth and b) populating a new `ExchangeRates` to replicate the state in an older one, hence the need, on occasion, to be able to set an inverse as frozen directly.

contracts/ExchangeRates.sol
justinjmoses: Done. Also some other cleanup added.

justinjmoses:
RFAL

publish/src/commands/deploy.js
```diff
@@ -1031,18 +1033,19 @@ const deploy = async ({
 				) {
 					if (oldExrates.options.address !== addressOf(exchangeRates)) {
 						const freezeAtUpperLimit = +w3utils.fromWei(currentRateForCurrency) === upperLimit;
```
ajsantander, line 1035: Consider that `Number.MAX_SAFE_INTEGER + 1 === Number.MAX_SAFE_INTEGER + 2` and `MAX_SAFE_INTEGER = 9007199254740991`, so might be a good idea to use big numbers here.

test/contracts/Exchanger.js
```diff
@@ -1829,7 +1829,7 @@ contract('Exchanger (via Synthetix)', async accounts => {
 				describe('when the iBTC synth is set with inverse pricing', () => {
 					const iBTCEntryPoint = toUnit(4000);
 					beforeEach(async () => {
-						exchangeRates.setInversePricing(
+						await exchangeRates.setInversePricing(
```
ajsantander, line 1832: :sweat_smile:

publish/releases.json
```diff
@@ -6,6 +6,7 @@
 			"minor": 27
 		},
 		"sources": [
+			"BinaryOptionMarketManager",
```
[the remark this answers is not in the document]
justinjmoses: Because of the change to the contract to align it with the update to `ExchangeRates`

contracts/ExchangeRates.sol
```diff
@@ -146,7 +141,8 @@ contract ExchangeRates is Owned, SelfDestructible, MixinResolver, MixinSystemSet
         inversePricing[currencyKey].entryPoint = 0;
         inversePricing[currencyKey].upperLimit = 0;
         inversePricing[currencyKey].lowerLimit = 0;
-        inversePricing[currencyKey].frozen = false;
+        inversePricing[currencyKey].frozenAtUpperLimit = false;
+        inversePricing[currencyKey].frozenAtLowerLimit = false;
```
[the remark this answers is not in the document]
justinjmoses: Yeah good catch. Unrelated code but optimizations always welcome 💯

contracts/ExchangeRates.sol
```diff
@@ -180,8 +176,47 @@ contract ExchangeRates is Owned, SelfDestructible, MixinResolver, MixinSystemSet
...
 
+    // SIP-75 Public keeper function to freeze a synth that is out of bounds
+    function freezeRate(bytes32 currencyKey) external {
+        InversePricing storage inverse = inversePricing[currencyKey];
+        require(inverse.entryPoint > 0, "Cannot freeze non-inverse rate");
+        require(!inverse.frozenAtUpperLimit && !inverse.frozenAtLowerLimit, "The rate is already frozen");
+
+        uint rate = _getRate(currencyKey);
+
+        if (rate == inverse.upperLimit || rate == inverse.lowerLimit) {
```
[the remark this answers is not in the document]
justinjmoses: Noted. I will adjust.

contracts/ExchangeRates.sol
```diff
@@ -180,8 +176,47 @@ contract ExchangeRates is Owned, SelfDestructible, MixinResolver, MixinSystemSet
...
     /* ========== VIEWS ========== */
 
+    // SIP-75 View to determine if freezeRate can be called safely
+    function canFreezeRate(bytes32 currencyKey) external view returns (bool) {
+        InversePricing memory inverse = inversePricing[currencyKey];
+        if (inverse.entryPoint == 0 || inverse.frozenAtUpperLimit || inverse.frozenAtLowerLimit) {
+            return false;
+        } else {
+            uint rate = _getRate(currencyKey);
+            return (rate == inverse.upperLimit || rate == inverse.lowerLimit);
```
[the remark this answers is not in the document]
justinjmoses: applied

contracts/ExchangeRates.sol
```diff
@@ -489,17 +484,59 @@ contract ExchangeRates is Owned, SelfDestructible, MixinResolver, MixinSystemSet
...
+        }
+
+        newRate = rate;
+
+        // These cases ensures that if a price has been frozen, it stays frozen even if it returns to the bounds
+        if (inverse.frozenAtUpperLimit) {
+            newRate = inverse.upperLimit;
+        } else if (inverse.frozenAtLowerLimit) {
+            newRate = inverse.lowerLimit;
+        } else if (inverse.entryPoint > 0) {
```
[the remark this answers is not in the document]
justinjmoses: Noted, fixing

contracts/ExchangeRates.sol
```diff
@@ -110,34 +99,40 @@ contract ExchangeRates is Owned, SelfDestructible, MixinResolver, MixinSystemSet
...
+            // rates already frozen in the current contract that need persisting across the upgrade
 
-            _setRate(currencyKey, freezeAtUpperLimit ? upperLimit : lowerLimit, now);
+            inversePricing[currencyKey].frozenAtUpperLimit = freezeAtUpperLimit;
+            inversePricing[currencyKey].frozenAtLowerLimit = freezeAtLowerLimit;
+            emit InversePriceFrozen(currencyKey, freezeAtUpperLimit ? upperLimit : lowerLimit, msg.sender);
+        } else {
+            // unfreeze if need be
+            inversePricing[currencyKey].frozenAtUpperLimit = false;
+            inversePricing[currencyKey].frozenAtLowerLimit = false;
```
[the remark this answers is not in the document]
justinjmoses: Nice catch. Fixing.

contracts/ExchangeRates.sol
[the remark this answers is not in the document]
justinjmoses: Actually then it will think `0` is a valid lowerLimit and it should not be. `0` will be returned for an inverse that has no underlying rate, and in which case it should not be frozen. I will add that condition as well

PR #672, opened 2020-08-19
Uniswap synths json list

jacko125:
- [ ]  Add ENS `setContentHash` functionality in script


publish/src/commands/persist-tokens.js
```diff
@@ -0,0 +1,131 @@
...
+const { confirmAction, ensureNetwork } = require('../util');
+const { gray, yellow } = require('chalk');
+
+const { schema } = require('@uniswap/token-lists');
+
+const Ajv = require('ajv');
+const ajv = new Ajv({ allErrors: true });
+
+const pinataSDK = require('@pinata/sdk');
+const pinata = pinataSDK(process.env.PINATA_KEY, process.env.PINATA_SECRET);
```
justinjmoses, line 14: I'd create this inside `uploadFileToIPFS` in case not having these in your .env file causes an issue

justinjmoses: (Did it myself)

index.js
```diff
@@ -32,6 +32,13 @@ const data = {
 
 const networks = ['local', 'kovan', 'rinkeby', 'ropsten', 'mainnet'];
 
+const networkToChainId = {
```
justinjmoses, line 35: @dvd-schwrtz I'm adding this into `synthetix` so you won't have to redefine it in synthetix-js

PR #673, opened 2020-08-20
Addind decode functionality to binary

justinjmoses:
Creates a new decode functionality for transactions that target the Synthetix ecosystem.

image

> Note: I've opted for `util.inspect` instead of `JSON.stringify` as the former can output ANSI colo(u)?rs. I suspect that's better in 99% of cases so I propose to make it a default

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const w3utils = require('web3-utils');
+const abiDecoder = require('abi-decoder');
 
 // load the data in explicitly (not programmatically) so webpack knows what to bundle
 const data = {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const w3utils = require('web3-utils');
+const abiDecoder = require('abi-decoder');
```
justinjmoses, line 4: In terms of browser bundle size, adding this contributes to a minimal addition (4.27MB to 4.32MB - i.e. 50KB). It could be done by looking up `signature` properties but this just seems fine for now.

jacko125 (approved):

PR #676, opened 2020-08-24
Some handy changes for the upcoming Pollux release

ajsantander:
#### Changes

* Adds a new `prepare-deploy` command that simply reads releases.json and uses that to flick contracts that need to be deployed in config.json.
* Allows the pwning of any address in the `fork` command. Used to be protocalDAO only, but now can also do oracle, etc.
* Fork command persists runs in a .db folder, which can be reset with a new `--reset` option.
* Updates the `import-fee-periods` script to use forking as well.
* Updates test:testnet to be able to work through invalid rates in a fork, and tests trading rewards being properly connected.

#### Forking workflow

These changes make the following workflow possible to simulate a release of Pollux in Kovan:
1. `node publish fork --network kovan --reset`
2. `node publish prepare-deploy --network kovan`
3. `node publish deploy --network kovan --use-fork --yes`
4. `node publish import-fee-periods --network kovan --use-fork --yes`
5. `node test/testnet --network kovan --use-fork --yes`


publish/src/commands/prepare-deploy.js
```diff
@@ -0,0 +1,69 @@
...
+
+	const deploymentPath = getDeploymentPathForNetwork(network);
+	ensureDeploymentPath(deploymentPath);
+
+	// Get config.js
+	const configFile = path.join(deploymentPath, CONFIG_FILENAME);
+	const config = JSON.parse(fs.readFileSync(configFile));
+
+	// Get releases.js
+	const releasesFile = path.join(deploymentPath, '../../', RELEASES_FILENAME);
```
justinjmoses, line 27: honestly I think this is better just as a top-level require as it has no variable path - as such, no need for `RELEASE_FILENAME` - just `require('../../releases.json')`

ajsantander: Fixed!

test/testnet/index.js
```diff
@@ -191,13 +193,55 @@ program
...
+				}
+
+				// Use these rates
+				const newRates = invalidCurrencyKeys.map(key => toWei('1', 'ether'));
+
+				// Set all invalid rates to 1:1
+				await exchangeRates.methods
+					.updateRates(invalidCurrencyKeys, newRates, await currentTime())
+					.send({ from: oracle, gas, gasPrice });
+			} else {
```
justinjmoses, line 241: [np] `else if`

ajsantander: Done

PR #678, opened 2020-08-25
Creating feeds.json and splitting out synths.json logic

justinjmoses:
1. Create a new top-level assets info JSON - `assets.json` with `asset`, `category`, `sign` and `desc`
2. Add a new JSON file per environment - a `feeds.json` for underlying assets. This allows some subset of assets to be tracked with aggregators for deployment, both those powering synths, as well as binary options markets and both `SNX` and `ETH` prices.  
3. On invoking `getSynths` for a network, the results will populate with asset data from `assets.json` and optionally a `feed` property populated from any `feeds.json` entry for that asset.


publish/deployed/local/feeds.json
```diff
@@ -0,0 +1,32 @@
+{
+	"SNX": {"asset": "SNX", "feed": "0x0", "standalone": true},
+	"ETH": {"asset": "ETH", "feed": "0x0", "standalone": true},
```
jacko125, line 3: Would COMP, KNC, LEND, REN also have `standalone: true` ? I guess its to signal that these are non-synths ?

justinjmoses: Good spot - had just fixed it already in my branch as well :)

jacko125 (approved):

index.js
```diff
@@ -174,52 +183,101 @@ const getAST = ({ source, path, fs, match = /^contracts\// } = {}) => {
...
 	}
 
+	const feeds = getFeeds({ network, path, fs, deploymentPath });
+
 	// copy all necessary index parameters from the longs to the corresponding shorts
 	return synths.map(synth => {
+		// mixin the asset details
+		synth = Object.assign({}, assets[synth.asset], synth);
+
+		if (feeds[synth.asset]) {
```
zyzek, line 237: Apart from sUSD, is there ever going to be a synth without a feed? Should a missing feed be an error?

publish/src/check-aggregator-prices.js
```diff
@@ -31,18 +37,18 @@ module.exports = async ({ network, providerUrl, synths, oldExrates }) => {
...
-			const liveAggregator = new web3.eth.Contract(abi, synth.aggregator);
+			const liveAggregator = new web3.eth.Contract(abi, feed);
 
 			const [aggAnswerRaw, exRatesAnswerRaw] = await Promise.all([
 				liveAggregator.methods.latestAnswer().call(),
-				oldExrates.methods.rateForCurrency(toBytes32(synth.name)).call(),
+				oldExrates.methods.rateForCurrency(toBytes32(currencyKey)).call(),
 			]);
 
 			const answer = (aggAnswerRaw / 1e8).toString();
```
zyzek, line 54: Hm, aren't these potentially BN's? Ordinary division feels a bit dicey to me for anything coming from the chain.
I assume this will need to be adapted for the V3 aggregator interface update and can wait until then, but just flagging it for now.

index.js
justinjmoses: Until SIP-81, there will be some on testnet that don't have a feed

publish/src/check-aggregator-prices.js
justinjmoses: This is just for a visual diff for the deploy script... It produces this output:

image

zyzek (approved):

PR #679, opened 2020-08-26
SIP-79 Gas tank

clementbalestrat:


contracts/GasTank.sol
```diff
@@ -0,0 +1,189 @@
...
+
+contract GasTank is Owned, MixinResolver, ReentrancyGuard {
+    /* ========== STATE VARIABLES ========== */
+
+    /* This value matches the required gas to execute the SpendGas function. It is added to the total gas spent
+		so keepers are fully refunded.
+     */
+    // TODO calculate this value with the tests
+    uint public constant PAYGAS_COST = 0;
+    mapping(address => bool) public allowance;
```
justinjmoses, line 24: Did you want to use FlexibleStorage for this? That way you won't lose state if you have to redeploy a new version?

justinjmoses: Also, `GasTankState` should be replaced with `FlexibleStorage`. Look at how we accomplish that using address keys in `Issuer`: https://github.com/Synthetixio/synthetix/blob/develop/contracts/Issuer.sol#L677-L681

contracts/SystemSettings.sol
```diff
@@ -38,6 +38,9 @@ contract SystemSettings is Owned, MixinResolver, MixinSystemSettings, ISystemSet
     // Minimum Stake time may not exceed 1 weeks.
     uint public constant MAX_MINIMUM_STAKE_TIME = 1 weeks;
 
+    // Max USD value of ether to be awarded to keepers
+    uint public constant MAX_KEEPER_FEE = 50;
```
clementbalestrat, line 42: This needs to be decided.

contracts/MixinSystemSettings.sol
```diff
@@ -21,6 +21,7 @@ contract MixinSystemSettings is MixinResolver {
     bytes32 internal constant SETTING_EXCHANGE_FEE_RATE = "exchangeFeeRate";
     bytes32 internal constant SETTING_MINIMUM_STAKE_TIME = "minimumStakeTime";
     bytes32 internal constant SETTING_AGGREGATOR_WARNING_FLAGS = "aggregatorWarningFlags";
+    bytes32 internal constant SETTING_KEEPER_FEE = "keeperFee";
```
justinjmoses, line 24: Is this the only keeper fee we'll have in the entire system? Perhaps it needs a more specific name?

clementbalestrat: hmmm I'm not sure. Happy to rename it `SETTING_GASTANK_KEEPER_FEE`

contracts/GasTank.sol
```diff
@@ -0,0 +1,189 @@
...
+        _depositEther(_account, msg.value);
+    }
+
+    function depositEther() external payable {
+        _depositEther(msg.sender, msg.value);
+    }
+
+    function _withdrawEther(address _account, uint _amount) internal {
+        require(_amount > 0, "Withdrawal amount must be greater than 0");
+        require(_gasTankState().balanceOf(_account) >= _amount, "Balance must be greater or equal to amount");
```
zyzek, line 135: We could just rely on the safe subtraction and skip the double check that the balance is sufficient here -- maybe not relevant pending any `FlexibleStorage` refactoring work.

contracts/GasTank.sol
```diff
@@ -0,0 +1,189 @@
...
+    function depositEtherOnBehalf(address _account) external payable {
+        require(_delegateApprovals().canManageGasTankFor(_account, msg.sender), "Not approved to act on behalf");
+        _depositEther(_account, msg.value);
+    }
+
+    function depositEther() external payable {
+        _depositEther(msg.sender, msg.value);
+    }
+
+    function _withdrawEther(address _account, uint _amount) internal {
```
zyzek, line 133: This probably also needs a `nonReentrant` modifier given that it's transferring to an unknown address.

zyzek, line 133: This is missing a parameter from the SIP, and I'm not sure the behaviour is quite right.
We need to be able to say that a given `msg.sender` is withdrawing from a certain account into a particular recipient address.
Note the difference in the signatures: the SIP suggests `function withdrawEtherOnBehalf(address account, address payable recipient, uint value) external`. Perhaps the `recipient` argument is not needed, and that the value should always be withdrawn to `msg.sender`, but I added it there for flexibility.

(I also point out a slight error in the SIP: it says a revert condition is `balanceOf(msg.sender) < value` though really it should be `balanceOf(account) < value`)

contracts/GasTank.sol
```diff
@@ -0,0 +1,189 @@
...
+        recipient.transfer(_amount);
+        emit EtherWithdrawn(msg.sender, recipient, _amount);
+    }
+
+    function withdrawEtherOnBehalf(address _recipient, uint _amount) external payable {
+        require(_delegateApprovals().canManageGasTankFor(_recipient, msg.sender), "Not approved to act on behalf");
+        _withdrawEther(_recipient, _amount);
+    }
+
+    function withdrawEther(address payable _recipient, uint _amount) external payable {
```
zyzek, line 147: This seems to allow anyone to withdraw from anyone's account.

contracts/GasTank.sol
```diff
@@ -0,0 +1,189 @@
...
+    function payGas(
+        address _spender,
+        address payable _recipient,
+        uint _gas
+    ) external nonReentrant returns (uint) {
+        require(isApprovedContract(msg.sender), "Contract is not approved");
+        require(_gasTankState().balanceOf(_spender) >= executionCost(_gas), "Spender balance is too low");
+        require(tx.gasprice >= _currentGasPrice(), "Gas price is too low");
+        require(tx.gasprice <= _gasTankState().maxGasPriceOf(_spender), "Spender gas price limit is reached");
+        uint etherSpent = executionCost(_gas);
```
zyzek, line 175: Assign to the variable before line 172 and reuse it in the `require` to save recomputation.

contracts/GasTank.sol
```diff
@@ -0,0 +1,189 @@
...
+
+    function currentEtherPrice() external view returns (uint currentEtherPriceWei) {
+        return _currentEtherPrice();
+    }
+
+    function _currentEtherPrice() internal view returns (uint currentGasPriceWei) {
+        return _exchangeRates().rateForCurrency("ETH");
+    }
+
+    function executionCost(uint _gas) internal view returns (uint etherCost) {
```
zyzek, line 100: I'm thinking this might be worth making public so that keepers can consult it when deciding whether to attempt to execute certain actions. If it should remain internal then add an underscore for consistency.

zyzek (changes requested):

PR #681, opened 2020-08-26
Adds a new CircleCI workflow with forking

ajsantander:
Uses the fork command's ability to persist the state of a fork in a .db folder to add a new workflow to the CI which automates the testing of a deployment + testnet tests.

Note that the workflow will only run when a commit touches the develop branch, which is achieved by attaching filters to each job in the new `fork` workflow. It has been tested without the filter: https://app.circleci.com/pipelines/github/Synthetixio/synthetix/3237/workflows/46774286-483a-4fdb-ba8b-ca5e7f3ace2f

In a future PR, I'd like to adapt this so that it runs a prod specific test suite instead of testnet tests. It'll be similar to this, but instead of running a command to test, it will run mocha with a bunch of isolated little tests, split into multiple files.


package.json
```diff
@@ -24,7 +24,14 @@
 		"test:etherscan": "node test/etherscan",
 		"test:local": "concurrently --kill-others --success first \"npx buidler node > /dev/null\" \"wait-port 8545 && node test/testnet --network local --yes\"",
 		"test:publish": "concurrently --kill-others --success first \"npx buidler node > /dev/null\" \"wait-port 8545 && mocha test/publish\"",
-		"test:testnet": "node test/testnet"
+		"test:testnet": "node test/testnet",
```
ajsantander, line 27: @justinjmoses is there a way to not have to do this? i.e. call node directly from CircleCI's config.yml.

publish/src/commands/deploy.js
```diff
@@ -1510,5 +1510,13 @@ module.exports = {
 				'Perform the deployment on a forked chain running on localhost (see fork command).',
 				false
 			)
-			.action(deploy),
+			.action(async (...args) => {
```
ajsantander, line 1513: Added this so that a failing deploy run causes the CI to fail.

package.json
justinjmoses: Yes you can simply do `node test/testnet` - it's a global like `npm` and `npx`

.circleci/config.yml
```diff
@@ -22,6 +22,59 @@ commands:
...
+            - build
+            - publish/deployed
+
+  fork_deploy:
+    description: Performs a deployment using a fork
+    parameters:
+      network:
+        type: string
+    steps:
+      - run:
```
justinjmoses, line 54: you can combine this and the wait_for_rpc into a CircleCI `command` and use that. See the `commands` at the top of the file for examples

package.json
```diff
@@ -24,7 +24,14 @@
...
+		"test:fork-start": "node publish fork --reset",
+		"test:fork-resume": "node publish fork",
```
justinjmoses, line 29: instead of all these here, I'd just invoke them via `node` in circle ci -seems excessive to add all these scripts to package.json

publish/src/commands/deploy.js
justinjmoses: Ah good catch.

.circleci/config.yml
```diff
@@ -169,6 +249,30 @@ jobs:
...
+      - prepare
+      - fork-prepare-mainnet:
+          requires:
+            - prepare
+          filters:
+            branches:
+              only: develop
+      - fork-deploy-mainnet:
+          requires:
+            - prepare
```
justinjmoses, line 263: this is a prior dependency, so you can remove it.

.circleci/config.yml
```diff
@@ -169,6 +249,30 @@ jobs:
 
 workflows:
   version: 2
+  fork:
```
justinjmoses, line 252: wanna add workflows here for kovan, rinkeby and ropsten as well?

.circleci/config.yml
```diff
@@ -22,6 +22,59 @@ commands:
...
+    description: Runs testnet tests on a fork
+    parameters:
+      network:
+        type: string
+    steps:
+      - run:
+          command: npm run test:fork-resume -- --network << parameters.network >>
+          background: true
+      - wait_for_rpc
+      - run: npm run test:fork-fees -- --network << parameters.network >>
```
justinjmoses, line 75: this is actually conditional on whether or not `FeePool` was re-deployed...

package.json
ajsantander: Lol duh. I don't know what drove me to the conclusion that this wasn't possible. Fixed.

.circleci/config.yml
ajsantander: Screenshot_20200827_180833
BAWM!

.circleci/config.yml
ajsantander: Yeah, I suggest that we keep it for now, and custom fit it as needed until prod testing is set up elsewhere. As it is now, its mainly a POC. A useful one.

.circleci/config.yml
justinjmoses: such a slick refactor :100:

justinjmoses: Test link to the above: https://app.circleci.com/pipelines/github/Synthetixio/synthetix/3284/workflows/0d11233d-27ae-468b-ad87-8863cb874b30

.circleci/config.yml
```diff
@@ -203,6 +304,105 @@ jobs:
 
 workflows:
   version: 2
+  fork:
+    jobs:
+      - prepare
```
justinjmoses, line 309: this should also be filtered, otherwise an additional prepare is added to every build on every commit: https://app.circleci.com/pipelines/github/Synthetixio/synthetix/3286/workflows/e861c5ae-ed5e-4a57-9511-b88a98f12ec0

.circleci/config.yml
```diff
@@ -224,15 +427,6 @@ workflows:
       - test-local:
           requires:
             - prepare
-      - pack:
```
justinjmoses, line 227: whoops - these seems to have been removed

PR #683, opened 2020-08-27
OVM Build Flag

K-Ho:
Adds a `--use-OVM` build flag that will run the build script using the Optimism `solc` fork for compilation. This will be helpful in ensuring contract sizes remain under `24KB` even with the overhead of our compilation process.


publish/src/commands/build.js
```diff
@@ -31,6 +31,7 @@ const build = async ({
 	testHelpers,
 	showWarnings,
 	showContractSize,
+	useOVM,
```
justinjmoses, line 34: Can you also add a console log in gray to log when the flag is enabled?

K-Ho: done!

justinjmoses:
Thanks @K-Ho - do you know if buidler supports passing through the solc compiler as well? might be good to have a flag in `buidler.config.js` to allow us to use your ovm-compiler when running tests.

K-Ho:
> Thanks @K-Ho - do you know if buidler supports passing through the solc compiler as well? might be good to have a flag in `buidler.config.js` to allow us to use your ovm-compiler when running tests.

Buidler doesn't support custom solc compilers. To get the tests running in the OVM, we need to import a new dependency and add two builder plugins - one that enables passing a path to a custom solc and another that sets the network provider to be an OVM provider.

justinjmoses:
OK thanks for that. Hopefully a newer version of buidler will make this a lot easier

PR #684, opened 2020-08-27
Migrating coverage and gas to circleci

justinjmoses:


jacko125 (approved):
great !

justinjmoses:
Coverage works great but the codechecks from the gas reports aren't being attached... the env vars are set. Hmmm not sure what's going on.

justinjmoses:
Fixed!

PR #693, opened 2020-08-28
Introduces production testing (via forking)

ajsantander:
#### Adds a new type of test: `.prod.js`

These tests have even less coverage than recently introduced `.spec.js` tests and much less than `.unit.js` tests, and are run against a forked network instead of a mocked system deployed locally. Coverage would be something like this:

```
              manual                ==> mainnet or testnet
          prod prod prod            ==> local forked chain, real Synthetix
    spec spec spec spec spec        ==> local chain, mocked Synthetix  ┐
unit unit unit unit unit unit unit  ==> local chain                    ┘ measured in actual coverage
```

To run the tests:
1. `npm run fork` (or `node publish fork --reset --network kovan`)
2. `npm run test:prod` (or `npx buidler test:prod --network localhost`)

The tests are located in `test/prod/`, separated from the rest of the tests to reduce clutter and avoiding globing or hand picking in the main test task.

Note that prod tests are very similar to spec tests. We thought about avoiding code duplication and actually making spec tests able to run on either environment, but we soon realized that this would involve some "surgical" refactoring on existing tests, vs proposing a way that is easy to move forward with.

Also, note how I am suggesting that we move test utils into dedicated folders, and that these utils move away from being monolithic bags of code and are split into very small and very specific files.


publish/src/solidity.js
```diff
@@ -65,6 +65,7 @@ module.exports = {
 
 	compile({ sources, runs, useOVM }) {
 		// Note: require this here as silent error is detected on require that impacts pretty-error
+		// eslint-disable-next-line node/no-missing-require
```
ajsantander, line 68: This is actually not part of my PR, but I had to change it in order for lint checks to pass.

index.js
```diff
@@ -46,6 +46,13 @@ const networkToChainId = {
 	kovan: 42,
 };
 
+const chainIdToNetwork = {
```
justinjmoses, line 49: [np] `Object.entries(networkToChainId).reduce...`

test/contracts/utils/connectContract.js
```diff
@@ -0,0 +1,30 @@
+const { artifacts } = require('@nomiclabs/buidler');
+const { getTarget } = require('../../../index.js');
```
justinjmoses, line 2: [np] `require('../../..')`

ajsantander: I actually like to be more explicit with the `index.js` in there. Let me know if you insist that this should be done and we could include it to our linter or something.

package.json
```diff
@@ -17,9 +17,11 @@
 		"pack": "webpack --mode production",
 		"prepublishOnly": "npm run generate-asts && npm run pack",
 		"postpublish": "node publish persist-tokens --yes",
+		"fork": "npde publish fork --reset --network mainnet",
```
i-stam, line 20: typo : npde

justinjmoses, line 20: typo

publish/src/commands/fork.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 const { ensureNetwork } = require('../util');
-const { getUsers } = require('../../../index.js');
+const { getUsers, networkToChainId } = require('../../../index.js');
```
justinjmoses, line 4: [np] can just require `../../..`

buidler.config.js
```diff
@@ -230,6 +243,13 @@ module.exports = {
 			},
 			baseNetworkConfig
 		),
+		localhost: Object.assign(
```
justinjmoses, line 246: we can dedupe this as it's used by `coverage` above

test/prod/TradingRewards.prod.js
```diff
@@ -0,0 +1,133 @@
+const { contract } = require('@nomiclabs/buidler');
+const { getUsers } = require('../../index.js');
```
justinjmoses, line 2: (ibid) (=> as above)

package.json
ajsantander: Ah good catch!

test/prod/utils/index.js
```diff
@@ -0,0 +1,16 @@
+const { detectNetworkName } = require('./detectNetwork');
+const { connectContract, connectContracts } = require('./connectContract');
+const { getEther, getSNX, getsUSD } = require('./getTokens');
+const { exchangeSynths } = require('./exchangeSynths');
+const { readSetting } = require('./systemSettings');
```
justinjmoses, line 5: it's a little confusing that these patterns don't hold... `connectContract` is destructed from `connectContract` and `exchangeSynths` from `exchangeSynths` whereas `readSetting` is from `systemSettings`. Not sure what's best here - it does make the files easier to read but by using different names for functions than the modules, it's not as easy to find the source if you need it... 🤔

test/prod/utils/connectContract.js
```diff
@@ -0,0 +1,30 @@
+const { artifacts } = require('@nomiclabs/buidler');
+const { getTarget } = require('../../../index.js');
+
+async function connectContract({ network, contractName, abiName = contractName }) {
```
justinjmoses, line 4: the abiName can be inferred by using `source` from the result of `getTarget`

test/prod/utils/exchangeSynths.js
```diff
@@ -0,0 +1,45 @@
+const { toBytes32 } = require('../../../index.js');
+const { connectContract, connectContracts } = require('./connectContract');
+const { getDecodedLogs } = require('../../contracts/helpers');
+
+async function getExchangeLogs({ network, exchangeTx }) {
+	const { TradingRewards, Synthetix } = await connectContracts({
+		network,
+		requests: [
+			{ contractName: 'TradingRewards' },
+			{ contractName: 'ProxyERC20', abiName: 'Synthetix' },
```
justinjmoses, line 10: can just use `Synthetix` - no need to worry about the proxy unless you want to test that here

test/prod/utils/index.js
ajsantander: Yeah, I see your point. However, I insist with the current status. The file name is the "topic" and the functions it exposes are all related to that topic.

test/prod/utils/getTokens.js
```diff
@@ -0,0 +1,61 @@
...
+
+	await web3.eth.sendTransaction({
+		from: fromAccount,
+		to: account,
+		value: amount,
+	});
+}
+
+async function getSNX({ network, amount, account, fromAccount }) {
+	const SNX = await connectContract({ network, contractName: 'ProxyERC20' });
```
justinjmoses, line 21: How is this working without the knowledge of the `Synthetix` source ABI?

test/contracts/utils/connectContract.js
justinjmoses: Shouldn’t you then change `require(‘./util’)` into require('./util/index.js') ? It's a slippery slope and I think `index.js` is obvious enough to any JS engineer here, no?

test/prod/utils/index.js
justinjmoses: Right but the topic isn't super consistent - "get tokens", "exchange synths", "connect contract" - shouldn't "system settings" be "get settings" or something similar?

test/prod/utils/connectContract.js
ajsantander: I tried that but in some cases I don't get the ABI that I want.

These are the mappings for `contractName` => `source`:
```
TradingRewards => TradingRewards
AddressResolver => AddressResolver
SystemSettings => SystemSettings
ProxyERC20 => ProxyERC20
SystemSettings => SystemSettings
SynthsUSD => Synth
ProxyERC20 => ProxyERC20
ProxyERC20 => ProxyERC20
```

Notice how for ProxyERC20 I sometimes need the Synthetix ABI.

test/prod/utils/getTokens.js
ajsantander: Because I only need to invoke the ERC20 ABI here, i.e. `transfer(...)`.

test/prod/utils/index.js
ajsantander: What I had in mind is for it to also be able to host a `setSetting` function. Perhaps what I could do to be more consistent is rename the files like this:
```
artifacts.js => connectContract, connectContracts
network.js => detectNetwork
exchange.js => getExchangeLogs, exchangeSynths
settings.js => readSetting, setSetting
tokens.js => getEther, getSNX, getsUSD
```

buidler.config.js
ajsantander: As discussed in chat: yeah, coverage and localhost effectively have the same props. However, I prefered to duplicate the code to avoid invoking prod tests like this `npx buidler test:prod --network coverage`.

test/contracts/utils/connectContract.js
ajsantander: Ok, you win!

publish/src/solidity.js
justinjmoses, line 68: I think this is because you didn't run `npm run install`

i-stam (approved):

publish/src/solidity.js
ajsantander: Ah cool.

PR #696, opened 2020-08-30
Fixing testnet test to show all synths, and adding standalone virtual prop to feeds

justinjmoses:
1. the `node test/testnet` script wasn't showing all synth rates due to filtering, this fixes that
2. also adds the virtual prop `standalone` to each entry on `feeds` to allow for quick filtering on those feeds that need to be supported even without having a synth using them


index.js
```diff
@@ -243,7 +263,7 @@ const getSynths = ({ network = 'mainnet', path, fs, deploymentPath } = {}) => {
 
 		if (feeds[synth.asset]) {
 			// mixing the feed
-			synth = Object.assign({}, feeds[synth.asset], synth);
+			synth = Object.assign({ feed: feeds[synth.asset].feed }, synth);
```
ajsantander, line 266: The code looks alright up to this point, but I fear its readability is decreasing with each update, which increases fragility as a consequence. The objective here is to post-process input data from json files into the returned objects. We might want to add a bit of architecture below this to make the code equally functional, but clearer, simpler.

justinjmoses: Yeah, I do agree with you, it's getting a lot less clear. I guess trying to add the `standalone` property to just the feed and not the synth is a bit much...

justinjmoses: It seemed cleaner to do this than export a function to filter all feeds with no associated synth and have it fetch both feeds and synths and then figure it out and mixin the `standalone` prop. But, the problem with doing it in `getFeeds` is that it needs to not add `standalone: xxx` into the `synths`, hence the need to only take that single prop.

ajsantander: Yeah, I totally agree with your choice. I'm just saying that I smell a bit of architecture here that could make things much simpler. Not for this PR tho, just to keep in mind.

PR #699, opened 2020-08-31
Ensure dest rate of 0 does not throw for settlement (part of SIP-86)

justinjmoses:


contracts/ExchangeRates.sol
```diff
@@ -272,6 +272,11 @@ contract ExchangeRates is Owned, SelfDestructible, MixinResolver, MixinSystemSet
 
         (uint srcRate, ) = _getRateAndTimestampAtRound(sourceCurrencyKey, roundIdForSrc);
         (uint destRate, ) = _getRateAndTimestampAtRound(destinationCurrencyKey, roundIdForDest);
+        if (destRate == 0) {
```
jacko125, line 275: good catch on this division edge case !

zyzek (approved):

contracts/ExchangeRates.sol
jacko125: And will revert if `destRate` is 0 for `effectiveValueAndRates`

https://github.com/Synthetixio/synthetix/blob/master/contracts/ExchangeRates.sol#L607

jacko125 (dismissed):

contracts/ExchangeRates.sol
justinjmoses: Hmmm. Should we prevent the revert in that case? Feels like it is the right move.

justinjmoses: Done.

jacko125: @justinjmoses  what was done ?

justinjmoses: Adds the fix to `effectiveValueAndRates`: https://github.com/Synthetixio/synthetix/pull/699/files#diff-3477eb31675538d20c3531997ee9d0d6R623-R625

clementbalestrat (approved):

jacko125 (approved):

PR #708, opened 2020-09-08
Ethercollat susd

hav-noms:
https://sips.synthetix.io/sips/sip-85


contracts/EtherCollateralsUSD.sol
```diff
@@ -36,7 +35,7 @@ contract EtherCollateral is Owned, Pausable, ReentrancyGuard, MixinResolver, IEt
     // ========== SETTER STATE VARIABLES ==========
 
     // The ratio of Collateral to synths issued
-    uint256 public collateralizationRatio = SafeDecimalMath.unit() * 125; // SCCP-27
+    uint256 public collateralizationRatio = SafeDecimalMath.unit() * 150;
```
jacko125, line 38: @hav-noms  Clinton, did we want borrowers to loan up to 200% , then they get liquidated at 150% or below ? 

Or allow them to go up to 150% if they want to risk it?

hav-noms: partial liquidations under 150%

contracts/EtherCollateralsUSD.sol
```diff
@@ -0,0 +1,683 @@
...
+    uint256 public accountLoanLimit = 50;
+
+    // If true then any wallet addres can close a loan not just the loan creator.
+    bool public loanLiquidationOpen = false;
+
+    // Time when remaining loans can be liquidated
+    uint256 public liquidationDeadline;
+
+    // Liquidation ratio when loans can be liquidated
+    uint256 public liquidationRatio = (150 * SafeDecimalMath.unit()) / 100; // 1.5 ratio
```
justinjmoses, line 71: Qu: do we want any of this in system settings?

contracts/EtherCollateralsUSD.sol
```diff
@@ -0,0 +1,683 @@
...
+import "./SafeDecimalMath.sol";
+
+// Internal references
+import "./interfaces/ISystemStatus.sol";
+import "./interfaces/IFeePool.sol";
+import "./interfaces/ISynth.sol";
+import "./interfaces/IERC20.sol";
+import "./interfaces/IExchangeRates.sol";
+
+import "@nomiclabs/buidler/console.sol";
```
justinjmoses, line 20: don't forget to remove this

justinjmoses: we should probably have a lint error for this...

justinjmoses: It exists! https://www.npmjs.com/package/solhint-plugin-buidler - being upgraded as we speak to include `import` checks as well as `console.log` usage.

contracts/interfaces/IEtherCollateralsUSD.sol
```diff
@@ -0,0 +1,18 @@
+pragma solidity >=0.4.24;
+
+
+interface IEtherCollateralsUSD {
```
justinjmoses, line 4: We can just use IEtherCollateral no? It's the same interface.

contracts/EtherCollateralsUSD.sol
hav-noms: Not for these trial contracts but yes for Multi-Collateral v1.0

contracts/EtherCollateralsUSD.sol
hav-noms: great! Yes still in Draft PR here. Expect it to be gone when PR submitted

contracts/interfaces/IEtherCollateralsUSD.sol
hav-noms: Different Interface now with liquidations involved. Will be thrown out for Multi-Collateral v1.0

justinjmoses:
Don't forget to add the new contracts to `releases.json` so our fork deploys and gas checks pick up these changes.

PR #711, opened 2020-09-10
Testing contracts in the OVM

K-Ho:
This PR adds a Buidler flag for compiling the contracts for the OVM and then running the tests against an OVM node. Tests with `@ovm-skip` are not run. The `native` flag is for easily testing changes to the OVM compiler and can be removed once the compiler is more stable. Enabling the optimizer for at least 1 run greatly decreases contract size, so it's recommended to enable the `optimizer` flag when running tests in the `OVM`

Note: the tests are incredibly slow right now and will require an update to the compiler to be sped up (we will get this fixed ASAP).


buidler.config.js
```diff
@@ -157,7 +160,20 @@ const optimizeIfRequired = ({ bre, taskArguments: { optimizer } }) => {
 task('compile')
 	.addFlag('showsize', 'Show size of compiled contracts')
 	.addFlag('optimizer', 'Compile with the optimizer')
+	.addFlag('ovm', 'Compile with the OVM Solidity compiler')
+	.addFlag('native', 'Compile with the native solc compiler')
```
justinjmoses, line 164: What's the reasoning behind having this `native` flag? Wouldn't not using `--ovm` yield the same result?

buidler.config.js
```diff
@@ -194,11 +210,31 @@ task('compile')
...
 	.setAction(async (taskArguments, bre, runSuper) => {
-		optimizeIfRequired({ bre, taskArguments });
+		const { gas, grep, ovm, native } = taskArguments;
+
+		if (ovm) {
+			console.log(gray('Compiling and running tests in the OVM...'));
+			bre.config.solc = {
+				path: path.resolve(__dirname, 'node_modules', '@eth-optimism', 'solc'),
+			};
+			await bre.config.startOvmNode();
```
justinjmoses, line 224: Where does this come from? It fails for me

K-Ho: Ahh I think I need to re-publish the latest `ovm-toolchain` package!

buidler.config.js
K-Ho: yes - native flag is only for us while we fix our compiler - compiling the native solidity compiler takes a few seconds, whereas compiling the Solidity compiler as `solc` takes ~25 minutes. We can definitely remove the native flag once the compiler is all fixed up :)

buidler.config.js
K-Ho: Should be fixed now, can you give it another try?

buidler.config.js
justinjmoses: OK - could you just indicate that it's temporary?

buidler.config.js
justinjmoses: Looking good thx

test/contracts/Depot.js
```diff
@@ -25,7 +25,7 @@ const { GAS_PRICE } = require('../../buidler.config');
 
 const { toBytes32 } = require('../..');
 
-contract('Depot', async accounts => {
+contract('Depot @ovm-skip', async accounts => {
```
justinjmoses, line 28: Ah I figured out why this breaks legacy tests. The legacy runner looks for a contract named by this string, so suffixing it ruins that assumption. I guess we started doing it with Binary Options without issues cause Binary Options has no legacy support. OK so the solution is to simply wrap the content in a describe block. I'll do that now. 

cc @zyzek

Issue #722, opened 2020-09-15
Gas-reporter fails to log gas consumption in the output file when prod tests are run


i-stam:
A mainnet fork is used and the prod tests are run instead.

Check #709 for instructions on spawining the environment and running the tests.

@cgewecke might be able to help with this one.

cgewecke:
@i-stam 

Hi. Out of curiosity, did the codechecks report in CI also stop working?

i-stam:
> @i-stam
> 
> Hi. Out of curiosity, did the codechecks report in CI also stop working?

Hi! The output report works fine for the unit tests:
`"test:gas": "buidler test --gas --optimizer || cat test-gas-used.log",`
which is what is being invoked in CI (`npm run test:gas && npx codechecks`)

The problem is when we try to run this:
```
"test:prod": "buidler test:prod --network localhost",
"test:prod:gas": "buidler test:prod --gas --network localhost --gas-output-file test-gas-used-prod.log",
```

justinjmoses:
The output is an empty file.

Basically to reproduce, run `npm run fork -- --network mainnet` in one shell and in another run `npm run test:prod:gas`. 

When you `cat test-gas-used-prod.log` you'll see:

```
·------------------------|----------------------------|-------------------|-------------------------------------·
|  [90mSolc version: 0.5.16[39m  ·  [90mOptimizer enabled: false[39m  ·  [90mRuns: undefined[39m  ·  [90mBlock limit: 9007199254740991 gas[39m  │
·························|····························|···················|······································
|  [32m[1mMethods[22m[39m                                                                                                      │
··············|··········|··············|·············|···················|·················|····················
|  [1mContract[22m   ·  [1mMethod[22m  ·  [32mMin[39m         ·  [32mMax[39m        ·  [32mAvg[39m              ·  [1m# calls[22m        ·  [1musd (avg)[22m        │
·-------------|----------|--------------|-------------|-------------------|-----------------|-------------------·
```

justinjmoses:
![image](https://user-images.githubusercontent.com/799038/93247729-bbeaca80-f75c-11ea-98f6-ecf2e5b7c73c.png)

PR #723, opened 2020-09-15
Fix buildler-gas-reporter production reports

cgewecke:
#722 

Synthetix is using a beta version of the gas-reporter that's buidlerevm-compatible when the `fast` option is enabled. It wraps the provider in a way which doesn't work for ganache. Disabling `fast` asks the reporter to fall back on it's default logic (e.g slow) ...

Screen Shot 2020-09-15 at 1 44 16 PM

Also saw the diff is missing from the codechecks report since the migration to CircleCI in #684. Codechecks has [trouble discovering the commit hash][1] of the merge target in CircleCI unless it knows the branch name - have added some config to the `.yml` to help with this. Hopefully it will start to record the run history again and make target comparison possible.

[1]: https://github.com/codechecks/docs/blob/master/configuration.md#settings

### DIFF
--- a/buidler.config.js
+++ b/buidler.config.js
@@ -257,7 +257,7 @@ task('test')
 				bre.config.mocha.invert = true;
 			}
 			// Tell buidler-gas-reporter not to wrap provider when using ganache
-			if (bre.network.name === 'localhost'){
+			if (bre.network.name === 'localhost') {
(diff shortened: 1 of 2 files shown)


cgewecke:
Also, not certain you're getting all the data you might expect from that test set. Lmk.

PR #725, opened 2020-09-16
Fix buildler-gas-reporter production reports (internal PR)

i-stam:
The same as #723. We are just using an internal branch to be able to bypass the errors returned by CI when using a branch from a forked repo.

### DIFF
--- a/codechecks.yml
+++ b/codechecks.yml
@@ -1,2 +1,6 @@
 checks:
   - name: eth-gas-reporter/codechecks
+settings:
+  branches:
+    - develop
+    - master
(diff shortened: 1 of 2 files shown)


codechecks.yml
```diff
@@ -1,2 +1,6 @@
 checks:
   - name: eth-gas-reporter/codechecks
+settings:
```
justinjmoses, line 3: oh niiiiice.

justinjmoses: thanks @cgewecke :pray:

PR #726, opened 2020-09-16
Adds a test for deploying two bridged instances in the same network

ajsantander:
We may end up deleting the actual test, but it will pave the way for deployments of an instance in L2, and testing it's interaction with L1 instances.

#### Overview
* New test in `test/multi/index.js`. Run it with `npm run test:multi`.
* `publish/deploy` now looks into a `params.json` file for default parameter overrides. Will prompt the operator for confirmation on any overiides.


publish/deployed/local-ovm/rewards.json
```diff
@@ -0,0 +1,27 @@
+[
```
justinjmoses, line 1: this can just be an empty `[]`

ajsantander: I thought we needed to tell the ovm instance to have no pre-minted tokens?

justinjmoses: no this is for LP staking rewards - not relevant here. It's used in regular local to test out the configuration of these but not relevant for L2-ovm

publish/src/commands/deploy.js
```diff
@@ -30,21 +30,7 @@ const {
...
-		TARGET_THRESHOLD,
-		LIQUIDATION_DELAY,
-		LIQUIDATION_RATIO,
-		LIQUIDATION_PENALTY,
-		RATE_STALE_PERIOD,
-		EXCHANGE_FEE_RATES,
-		MINIMUM_STAKE_TIME,
-		AGGREGATOR_WARNING_FLAGS,
-	},
+	defaults,
```
justinjmoses, line 33: this needs to be loaded in right?

ajsantander: Loaded in?

I'm just not destructuring it as a way to enforce going through `getDeployParameter()`

publish/deployed/local-ovm/synths.json
```diff
@@ -0,0 +1,284 @@
+[
+	{
+		"name": "sUSD",
+		"asset": "USD"
+	},
+	{
```
justinjmoses, line 6: only need sUSD

publish/deployed/local-ovm/config.json
```diff
@@ -0,0 +1,413 @@
...
+	"SynthUtil": {
+		"deploy": true
+	},
+	"DappMaintenance": {
+		"deploy": true
+	},
+	"SynthsUSD": {
+		"deploy": true
+	},
+	"SynthsEUR": {
```
justinjmoses, line 98: Can you remove all Synth*, TokenState* and Proxy* that isn't sUSD / Synthetix / FeePool

publish/deployed/local-ovm/feeds.json
```diff
@@ -0,0 +1,32 @@
+{
+	"SNX": {"asset": "SNX", "feed": "0x0"},
```
justinjmoses, line 2: only need SNX

publish/src/commands/deploy.js
justinjmoses: Sorry i misread this as index.js. Please ignore.

publish/deployed/local-ovm/rewards.json
ajsantander: Ah sorry, I miss-understood. Ok.

PR #728, opened 2020-09-16
Avoid timeouts in forks

ajsantander:
Our prod tests in ci (which use forking) were timing out. I noticed that deployments were not, which actually involve much more complex transactions... turns out, it was buidler tests timing out, not ganache.


buidler.config.js
```diff
@@ -36,6 +36,7 @@ const baseNetworkConfig = {
 	// default to allow unlimited sized so that if we run buidler EVM in isolation (via npx buidler node)
 	// it will use this setting and allow any type of compiled contracts
 	allowUnlimitedContractSize: true,
+	timeout: 0,
```
justinjmoses, line 39: can you add a comment here as to why it's necessary? 

also, should we set it by default or only during some conditions?

ajsantander: Yeah, thats a great point! Miss-use of this could hang some of the CI tests. I'll move this to the test:prod task with a comment.

justinjmoses (approved):

PR #732, opened 2020-09-20
[SIP-83] Debt snapshots

zyzek:


contracts/Issuer.sol
```diff
@@ -136,58 +149,76 @@ contract Issuer is Owned, MixinResolver, MixinSystemSettings, IIssuer {
...
-                    totalSynths = totalSynths.sub(etherCollateral().totalIssuedSynths());
-                }
-
-                // minus total issued synths from Ether Collateral from sUSD.totalSupply()
-                if (synth == "sUSD") {
-                    totalSynths = totalSynths.sub(etherCollateralsUSD().totalIssuedSynths());
-                }
+            bool isSUSD = key == sUSD;
+            if (isSUSD || key == sETH) {
+                IEtherCollateral etherCollateralContract = isSUSD
```
jacko125, line 169: EtherCollateralsUSD doesn't share the same interface with EtherCollateral (IEtherCollateral). 

But I believe if we want to just use the `totalIssuedSynths()` function then can cast to IEtherCollateral

contracts/Issuer.sol
```diff
@@ -136,58 +149,76 @@ contract Issuer is Owned, MixinResolver, MixinSystemSettings, IIssuer {
...
-            currencyRate = exchangeRates().rateForCurrency(currencyKey);
+    function _totalIssuedSynths(bytes32 currencyKey, bool excludeEtherCollateral)
+        internal
+        view
+        returns (uint totalIssued, bool anyRateIsInvalid)
+    {
+        IFlexibleStorage store = flexibleStorage();
+
+        bytes32[] memory keys = new bytes32[](2);
+        keys[0] = CACHED_SNX_ISSUED_DEBT;
```
jacko125, line 196: Could this be wrapped inside a struct like `cachedSNXIssuedDebtAndTimestamp` instead of being expressed as an array of key[] values?

contracts/Issuer.sol
```diff
@@ -434,256 +494,274 @@ contract Issuer is Owned, MixinResolver, MixinSystemSettings, IIssuer {
...
+        // Store their debtRatio against a fee period to determine their fee/rewards % for the period
+        _appendAccountIssuanceRecord(debtAccount);
+    }
 
-        // Add penalty
-        totalRedeemed = snxRedeemed.multiplyDecimal(SafeDecimalMath.unit().add(liquidationPenalty));
+    // If burning to target, `amount` is ignored, and the correct quantity of sUSD is burnt to reach the target
+    // c-ratio, allowing fees to be claimed. In this case, pending settlements will be skipped as the user
+    // will still have debt remaining after reaching their target.
+    function _voluntaryBurnSynths(
```
jacko125, line 735: @zyzek So  `_voluntaryBurnSynths` is replacing `_internalBurnSynths` ? could the func be named that signals more clearly that `burnToTarget` to allow / skip the minimum stake time ?

contracts/Issuer.sol
```diff
@@ -748,14 +823,12 @@ contract Issuer is Owned, MixinResolver, MixinSystemSettings, IIssuer {
...
     function _removeFromDebtRegister(
         address from,
-        uint amount,
+        uint debtToRemove,
         uint existingDebt,
         uint totalDebtIssued
     ) internal {
         ISynthetixState state = synthetixState();
 
-        uint debtToRemove = amount;
```
jacko125, line 757: Good to cleanup this one, the debtToRemove = amount assignment was redundant

zyzek: 👍

contracts/Issuer.sol
zyzek: Yeah, previously it was pretty hard to distinguish between `_internalBurnSynths` and `_burnSynths`.

I called it `voluntary` since it's used for intentionally-initiated burns, rather than synth burnings that happen through other processes such as liquidation. I am open to renaming it, but not sure what to call it instead. Suggestions?

contracts/Issuer.sol
zyzek: It's an array so that it can be passed into `FlexibleStorage.getUIntValues`. Do you have a particular alternative structure in mind that wouldn't require unpacking the values into an array later on anyway?

contracts/Issuer.sol
zyzek: Right, that's the intention here. It's using a ternary here just in order to save a bit of code space on duplication of the `totalIssuedSynths` call.

contracts/Issuer.sol
```diff
@@ -434,256 +519,289 @@ contract Issuer is Owned, MixinResolver, MixinSystemSettings, IIssuer {
...
+        uint susdAmount,
+        address liquidator
+    ) external onlySynthetix returns (uint totalRedeemed, uint amountToLiquidate) {
+        // Ensure waitingPeriod and sUSD balance is settled as burning impacts the size of debt pool
+        require(!exchanger().hasWaitingPeriodOrSettlementOwing(liquidator, sUSD), "sUSD needs to be settled");
 
-        // record issue timestamp
-        _setLastIssueEvent(from);
+        // Check account is liquidation open
+        require(liquidations().isOpenForLiquidation(account), "Account not open for liquidation");
```
jacko125, line 604: @zyzek Was it out of space when we set the liquidations() address locally as `ILiquidations _liquidations = liquidations();` ?

zyzek: The stack was getting too deep.

contracts/Issuer.sol
jacko125: I think what is there now is fine, using the same IEtherCollateral

contracts/Issuer.sol
jacko125: We have been discussing adding a way to compact and store Struct data into the FlexibleStorage's using `bytes32` as a function and storing struct data that way. 

I was thinking you can have a helper function `_getCachedSNXIssuedDebtAndTimestamp()` that wraps this logic of loading the 2x uint values, to rebuild and return a struct for use inside `issuer.sol`. So the low level logic of `keys[0] and keys[1]` can be abstracted away from the actual implementation of the function `_totalIssuedSynths()` for example.

zyzek: Good call. Done. The struct thing I guess is a discussion for later. Perhaps would have been a decent reason to have a `bytes` mapping in `FlexibleStorage`.

hav-noms (approved):

contracts/Issuer.sol
```diff
@@ -136,58 +156,83 @@ contract Issuer is Owned, MixinResolver, MixinSystemSettings, IIssuer {
...
+    function _issuedSynthValues(bytes32[] memory currencyKeys, uint[] memory rates) internal view returns (uint[] memory) {
+        uint numValues = currencyKeys.length;
+        uint[] memory values = new uint[](numValues);
+
+        for (uint i = 0; i < numValues; i++) {
+            bytes32 key = currencyKeys[i];
+
+            uint supply = IERC20(address(synths[key])).totalSupply();
+
+            bool isSUSD = key == sUSD;
```
hav-noms, line 174: Sorry about this nastyness. @mjspain12 lets make sure this can be refactored to a rolled up total like supply.sub(MultiCollateralManager.totalIssuedSynths(sUSD))

hav-noms (dismissed):

jacko125 (approved):
LGTM the changes

cgewecke:
@justinjmoses 

It looks like it might be a difference between BuidlerEVM and ganache (which solidity-coverage uses) wrt how time passes? 

For clarification, this test expects a later timestamp but the clock stopped ?
```
Contract: Issuer (via Synthetix)
       when integrating with modern contracts
         when minimum stake time is set to 0
           when the issuanceRatio is 0.2
             Debt snapshots
               cacheSNXIssuedDebt()
                 updates the debt snapshot timestamp:

      AssertionError: function(title, fn) {
      return common.suite.create({
        title: title,
        file: file,
        fn: fn
      });
    }: expected '1551831872' to not equal '1551831872'
      + expected - actual


      at Object.assertBNNotEqual [as bnNotEqual] (test/utils/index.js:348:10)
      at Context.<anonymous> (test/contracts/Issuer.js:2492:14)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

justinjmoses:
Sorry, I was pinging you about the prior failure before on compilation: https://app.circleci.com/pipelines/github/Synthetixio/synthetix/3901/workflows/1c4f9123-0ce2-4b0c-be33-1154dd1ea2d3/jobs/25869 

I got it passed that now by removing from instrumentation the one contract that has the experimental pragma as such: https://github.com/Synthetixio/synthetix/pull/732/files#diff-91b2992b964567d52aaf8a04e30f9fe4R11

The issue with timestamps is definitely a known issue with ganache vs buidler evm. I'm curious if you know anything about how the `pragma experimental` line might be affecting the instrumentation. 

Also, if you remove `EtherCollateralsUSD` from the ignore in `.solcover.js` you'll see compilation fails again, this time for another unknown reason, so something seems to be up with the instrumentation step.

cgewecke:
@justinjmoses Ah ok I see, I'll look into those rn.

PR #734, opened 2020-09-21
No forking on test:testnet

ajsantander:
When we introduced forking, we picked `test:testnet` as the easiest temporary entry point, but now that `prod` testing is in place, we no longer need it. This PR removes all forking related stuff from this test, and updates the CI to not use `test:testnet` in forking related workflows. Instead, the `fork-prod` tests are extended to other networks.


.circleci/config.yml
```diff
@@ -367,87 +327,61 @@ workflows:
...
-      #         only: develop
-      # - fork-deploy:
-      #     name: fork-deploy-ropsten
-      #     network: ropsten
-      #     requires:
-      #       - fork-prepare-deploy-ropsten
-      #     filters:
-      #       branches:
-      #         only: develop
-      # - fork-test-testnet:
```
justinjmoses, line 421: Let's remove this comment block for now - we can always pull it back out from history later. Though could you keep a comment in as to why there's no `ropsten` target?

justinjmoses (approved):
LGTM, though I'd like to hear from @jacko125 or @hav-noms re: EtherCollateral prod coverage

justinjmoses (approved):

PR #739, opened 2020-09-21
Use buidler's RC for compiling and testing

fvictorio:
Hi! This PR upgrades Buidler (and `buidler-truffle5`) to the RC of the next major version. This should let you remove the need of an extra setup for your legacy contracts. These are the things I had to do:

- I moved the legacy contracts in the `legacy` directory to the `contracts` directory.
- I changed the buidler config to use the new `solidity` configuration, that supports multiple compiler versions. I used the same versions that you were using before.
- I had to comment out the `buidler-ast-doc` and `buidler-gas-reporter` plugins because they don't work with the new pipeline yet.
- The hardest part was adapting the tests. One of the changes in buidler's new compilation pipeline is that artifacts don't live in a flat structure in the artifacts directory. Instead, they replicate the directory structure of the contracts. This means that now you can have contracts with the same name, without one artifact being written over the other.

    But this comes with a cost: now if you have two contracts named `Foo` you can't do `artifacts.require('Foo')`, because there's no way to know which one you want. In those scenarios, you have to use the Fully Qualified Name, for example `contracts/Foo.sol:Foo`. When there's only one contract with that name, this is not necessary.

    Since the tests have a lot of stuff that I don't understand, I added a helper `disambiguateContract` and I used it in several `require`s. Of course you are free to solve that issue however you want, I did it that way because it seemed like the less intrusive option.

Almost all tests pass with the new setup. The ones that don't are caused by a bug that we just found out, related to linking ambiguous artifacts with a truffle artifact. We'll fix that tomorrow, so I'll probably update this PR with that.

I only tried to compile the contracts and run the tests. There are probably other things you do in your normal workflow, and there's a good chance that some of them stop working. Please let me know when that happens and I'll try to help!

Hope you enjoy it :smiley:


justinjmoses:
Thanks @fvictorio 🙏  - moving this over to #914 to work on it piece by piece.

PR #747, opened 2020-09-24
Goerli OVM deploy preparation 

justinjmoses:


contracts/EmptyEtherCollateral.sol
```diff
@@ -0,0 +1,7 @@
+pragma solidity ^0.5.16;
+
+
+// Empty contract for ether collateral placeholder for OVM
+contract EmptyEtherCollateral {
+    function totalIssuedSynths() external pure returns (uint) {}
```
i-stam, line 6: could we just return 0 for clarity

ajsantander (dismissed):

PR #753, opened 2020-09-24
Merge EtherCollateralsUSD and rest of Fomalhaut into develop

jacko125:


.solcover.js
```diff
@@ -4,7 +4,7 @@ const {
 
 module.exports = {
 	port: 8545,
-	skipFiles: ['test-helpers', 'EscrowChecker.sol'],
+	skipFiles: ['test-helpers', 'EscrowChecker.sol', 'EtherCollateralsUSD.sol'],
```
ajsantander, line 7: Why are we skipping these Solidity files from coverage?

contracts/FeePool.sol
```diff
@@ -725,11 +733,12 @@ contract FeePool is Owned, Proxyable, SelfDestructible, LimitedSetup, MixinResol
...
 
     /* ========== Modifiers ========== */
-    modifier onlyExchangerOrSynth {
+    modifier onlyInternalContracts {
         bool isExchanger = msg.sender == address(exchanger());
         bool isSynth = issuer().synthsByAddress(msg.sender) != bytes32(0);
+        bool isEtherCollateralsUSD = msg.sender == address(etherCollateralsUSD());
 
-        require(isExchanger || isSynth, "Only Exchanger, Synths Authorised");
+        require(isExchanger || isSynth || isEtherCollateralsUSD, "Only Internal Contracts");
```
ajsantander, line 741: nit: Could save gas here by passing through to the function as soon as any condition is met.

ajsantander (changes requested):
Looks good. Would like to know more about the situation with test coverage tho.

.solcover.js
justinjmoses: Because there is some issue in `EtherCollateralsUSD` coverage, it was decided to skip it.

ajsantander (dismissed):

PR #754, opened 2020-09-24
Main index file understands ovm deployments

ajsantander:
This PR adds a `useOVM` flag to most of the interfaces in the main index.js file (i.e. the synthetix npm package).

For example, when retrieving addresses:
```
const target = syntheitx.getTarget({.network: 'goerli', contract: 'Syntheitx', useOVM: true });
```
Will actually resolve to `publish/deployed/goerli-ovm` instead of `publish/deployed/goerli`. Basically, all this does is add '-ovm' to the paths.

By default, all `useOVM` flags are set to false.


justinjmoses (approved):
Looks great. Would also love a `--ovm` flag for each request in `bin.js` if you wouldn't mind.

PR #759, opened 2020-09-27
Try catch aggregator getrounddata

justinjmoses:


contracts/ExchangeRates.sol
```diff
@@ -556,12 +557,23 @@ contract ExchangeRates is Owned, SelfDestructible, MixinResolver, MixinSystemSet
...
         if (aggregator != AggregatorV2V3Interface(0)) {
-            (, int256 answer, , uint256 updatedAt, ) = aggregator.latestRoundData();
-            return
-                RateAndUpdatedTime({
-                    rate: uint216(_rateOrInverted(currencyKey, _formatAggregatorAnswer(currencyKey, answer))),
-                    time: uint40(updatedAt)
-                });
+            // this view from the aggregator is the most gas efficient but it can throw when there's no data,
+            // so let's call it low-level to suppress any reverts
+            bytes memory payload = abi.encodeWithSignature("latestRoundData()");
```
clementbalestrat, line 562: @justinjmoses do we need to do that check for `latestRoundData()`?
I believe this issue was with `roundData(roundID)` only.

In `addAggregator` we've got a test for that already:

`require(aggregator.latestRound() >= 0, "Given Aggregator is invalid");`

clementbalestrat (approved):

contracts/ExchangeRates.sol
justinjmoses: From what we were told, both can throw. It's very unlikely that there will be no latest round but it's technically still possible. The tests also do this by looking up `getRoundData` with the latest `roundID`. We could remove this but unless there's a gas hit it seems more robust to keep it consistent.

contracts/ExchangeRates.sol
```diff
@@ -583,8 +595,19 @@ contract ExchangeRates is Owned, SelfDestructible, MixinResolver, MixinSystemSet
         AggregatorV2V3Interface aggregator = aggregators[currencyKey];
 
         if (aggregator != AggregatorV2V3Interface(0)) {
-            (, int256 answer, , uint256 updatedAt, ) = aggregator.getRoundData(uint80(roundId));
-            return (_rateOrInverted(currencyKey, _formatAggregatorAnswer(currencyKey, answer)), updatedAt);
+            // this view from the aggregator is the most gas efficient but it can throw when there's no data,
```
jacko125, line 598: For this internal view did we want to make it return a RateAndUpdatedTime struct like this function  ?

`    function _getRateAndUpdatedTime(bytes32 currencyKey) internal view returns (RateAndUpdatedTime memory) {
`

contracts/ExchangeRates.sol
jacko125: Actual implementation is here https://github.com/smartcontractkit/chainlink/blob/develop/evm-contracts/src/v0.6/FluxAggregator.sol.

It does seem to be safer to check that latestRoundData() could revert as it uses the getRoundData() function internally too but the `latestRoundID` is provided by the aggregator contract so it shouldn't revert / be missing data.

clementbalestrat: yeah my point was because we're checking if `latestRound` returns something when we add the aggregator, then it should never revert when trying to get the latest round ID. Also what @jacko125 said makes sense too @justinjmoses.

clementbalestrat: though I don't think adding this check will consume more gas so I'm happy to leave it

jacko125 (dismissed):

publish/deployed/mainnet/deployment.json
```diff
@@ -20,11 +20,11 @@
 		},
 		"ExchangeRates": {
 			"name": "ExchangeRates",
-			"address": "0xdB2Ae36C2e9C00070e5bF752Be1FA2d477E98BDa",
+			"address": "0xb01850a6c664FEfF9945c6f74ff86572794A2a1f",
```
justinjmoses, line 23: Ahhhh what's this?? Is this a mistaken merge? The correct one is 0xdB2

justinjmoses:
@clementbalestrat can you modify https://github.com/Synthetixio/synthetix/blob/develop/.circleci/config.yml#L326-L355 temporarily to make sure it runs successfully with your new prod test?

jacko125 (approved):

PR #761, opened 2020-09-28
Use freshDeploy in deploy script

ajsantander:
Removes redundant `network === 'local'` checks in the deploy script. The `freshDeploy` flag should be enough.

### DIFF
--- a/publish/deployed/local-ovm/config.json
+++ b/publish/deployed/local-ovm/config.json
@@ -29,6 +29,9 @@
 	"EtherCollateral": {
 		"deploy": true
 	},
+	"EtherCollateralsUSD": {
+		"deploy": true
+	},
 	"FeePool": {
 		"deploy": true
 	},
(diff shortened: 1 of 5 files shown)


publish/deployed/local-ovm/config.json
```diff
@@ -29,6 +29,9 @@
 	"EtherCollateral": {
 		"deploy": true
 	},
+	"EtherCollateralsUSD": {
+		"deploy": true
+	},
```
ajsantander, line 34: When we fixed this, we had forgotten to also update local-ovm.

publish/src/commands/deploy.js
```diff
@@ -92,7 +92,7 @@ const deploy = async ({
 	});
 
 	// Fresh deploy and deployment.json not empty?
-	if (freshDeploy && Object.keys(deployment.targets).length > 0) {
+	if (freshDeploy && Object.keys(deployment.targets).length > 0 && network !== 'local') {
```
ajsantander, line 95: This requirement can be ignored when using local.

i-stam (approved):

PR #762, opened 2020-09-28
Assimilate l2-comms into a scripts folder

ajsantander:
This PR introduces a `./scripts` folder with some scripts we used during the Goerli Optimism trial deployment. Te idea is to have a place to put these scripts, and have a very easy way to add new ones or extend the existing.


scripts/status.js
```diff
@@ -0,0 +1,173 @@
...
+	});
+
+	logItem(
+		'Synthetix.anySynthOrSNXRateIsInvalid:',
+		await Synthetix.anySynthOrSNXRateIsInvalid()
+	);
+	logItem('Synthetix.totalSupply', (await Synthetix.totalSupply()).toString() / 1e18);
+
+	/* ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ */
+	/* ~~~~ FixedSupplySchedule ~~~~ */
```
justinjmoses, line 53: Note: a lot of this is shared with SupplySchedule - perhaps we just use the SupplySchedule ABI and report on whatever both have in common?

scripts/status.js
```diff
@@ -0,0 +1,173 @@
...
+			wallet,
+		});
+
+		logItem(
+			'FixedSupplySchedule.inflationStartDate',
+			new Date((await FixedSupplySchedule.inflationStartDate()).toString() * 1000)
+		);
+
+		const supply = bre.ethers.utils.formatEther(await FixedSupplySchedule.mintableSupply());
+		logItem('FixedSupplySchedule.mintableSupply', supply);
```
justinjmoses, line 73: Also we should split these out into a file per contract I reckon. I'd also love one for ExchangeRates that shows the rates (like the testnet script)

scripts/status.js
ajsantander: Nice. Done.

scripts/status.js
ajsantander: Improvements for the future? My goal in this PR is to reproduce what we had in l2-comms in this repo, in a way that is extensible and easy to use.

buidler.config.js
```diff
@@ -11,6 +11,7 @@ require('@eth-optimism/ovm-toolchain/build/src/buidler-plugins/buidler-ovm-compi
 require('@eth-optimism/ovm-toolchain/build/src/buidler-plugins/buidler-ovm-node'); // add ability to start an OVM node
 
 usePlugin('@nomiclabs/buidler-truffle5'); // uses and exposes web3 via buidler-web3 plugin
+usePlugin('@nomiclabs/buidler-ethers');
```
justinjmoses, line 14: Is this still needed?

package.json
```diff
@@ -70,9 +70,10 @@
 	},
 	"devDependencies": {
 		"@codechecks/client": "0.1.10",
-		"@eth-optimism/solc": "0.5.16-alpha.2",
 		"@eth-optimism/ovm-toolchain": "0.0.1-alpha.2",
+		"@eth-optimism/solc": "0.5.16-alpha.2",
 		"@nomiclabs/buidler": "1.3.7",
+		"@nomiclabs/buidler-ethers": "^2.0.0",
```
justinjmoses, line 76: and this?

scripts/airdrop.js
```diff
@@ -0,0 +1,124 @@
...
+		gasPrice,
+		gasLimit,
+	};
+
+	for (const staker of data) {
+		const address = staker.address;
+
+		const collateral = parseEther(`${staker.collateral}`);
+
+		const l2Balance = await Synthetix.balanceOf(address);
```
justinjmoses, line 78: we should also check make sure there have been no transfer events in some recent window of time.

scripts/status.js
```diff
@@ -0,0 +1,200 @@
...
+		provider,
+	});
+
+	const logRate = async currency => {
+		const rate = await ExchangeRates.rateForCurrency(formatBytes32String('SNX'));
+		const updated = (await ExchangeRates.lastRateUpdateTimes(formatBytes32String('SNX'))) * 1000;
+		logItem(`${currency} rate:`, `${rate} (updated ${updated})`);
+	};
+
+	await logRate('SNX');
```
justinjmoses, line 174: could we also log all synths?

justinjmoses (changes requested):
A couple of minor changes

buidler.config.js
ajsantander: Good catch, thanks.

scripts/airdrop.js
ajsantander: Do you think this is still required, given that the script now demands that the system be suspended?

justinjmoses: Ah well you can't transfer SNX at all if the system is suspended. 🤔

scripts/status.js
```diff
@@ -0,0 +1,279 @@
...
+	const logRate = async currency => {
+		const rate = await ExchangeRates.rateForCurrency(formatBytes32String(currency));
+		const updated = (await ExchangeRates.lastRateUpdateTimes(formatBytes32String(currency))) * 1000;
+		logItem(`${currency} rate:`, `${formatEther(rate)} (${new Date(updated.toString() * 1000)})`);
+	};
+
+	await logRate('SNX');
+
+	const synthsPath = path.join(getPathToNetwork({ network, useOvm }), constants.SYNTHS_FILENAME);
+	const synths = JSON.parse(fs.readFileSync(synthsPath));
```
justinjmoses, line 249: Why not use `getSynths()` ?

ajsantander: Nize

justinjmoses (approved):
:shipit:

PR #763, opened 2020-09-29
L1::L2 Deposit interface

justinjmoses:
Run `npm run test:multi`


contracts/SecondaryDeposit.sol
```diff
@@ -0,0 +1,131 @@
...
+
+    function maximumDeposit() external view returns (uint) {
+        return getMaximumDeposit();
+    }
+
+    // ========== PUBLIC FUNCTIONS =========
+
+    // invoked by user on L1
+    function deposit(uint amount) external {
+        require(amount <= getMaximumDeposit(), "Cannot deposit more than the max");
```
K-Ho, line 77: Should this be limited ber account? Nothing's stopping someone from going to the UI and depositing 2500 SNX over and over again from the same account

justinjmoses: Could do. The limit is really supposed to be soft. Keep in mind the more that moves to L2 the more rewards left to L1 for rewards...

contracts/SecondaryDeposit.sol
```diff
@@ -0,0 +1,132 @@
...
+    /// ========= VIEWS =================
+
+    function maximumDeposit() external view returns (uint) {
+        return getMaximumDeposit();
+    }
+
+    // ========== PUBLIC FUNCTIONS =========
+
+    // invoked by user on L1
+    function deposit(uint amount) external {
```
i-stam, line 80: Should these methods (deposit/withdraw) be available only to L1/L2 users correspondingly? Currently, they will go all the way down to the other layer.

PR #764, opened 2020-09-30
Adds a past-events script

ajsantander:


scripts/past-events.js
```diff
@@ -0,0 +1,79 @@
...
+
+	/* ~~~~~~~~~~~~~~~~~~~ */
+	/* ~~~~~~ Setup ~~~~~~ */
+	/* ~~~~~~~~~~~~~~~~~~~ */
+
+	const { provider } = await setupProvider({ providerUrl });
+
+	const contract = await getContract({
+		contract: contractName,
+		abi: abiName || contractName,
```
justinjmoses, line 32: [np] I've referred to this as `source` in the code, not `abiName`

justinjmoses (approved):
Just one nitpick

i-stam (approved):

scripts/past-events.js
ajsantander: Ah, I see. Given that this change would affect multiple ongoing PRs, I'll wait until they are merged.

PR #765, opened 2020-10-01
Proposes a new way to perform L2 airdrops

ajsantander:
The main difference with the previous script is that this one doesn't check on-chain balances since these could change while the script is running. Instead, this produces a local file to keep track of what has already been transferred.

### DIFF
--- a/scripts/airdrop.js
+++ b/scripts/airdrop.js
@@ -106,6 +106,8 @@ async function airdrop({
 			receipt = await runTx(
 				await Synthetix.transfer(staker.address, parseEther(`${remaining}`), overrides)
 			);
+
+			if (!receipt) missedContenders++;
 		}
 
 		return {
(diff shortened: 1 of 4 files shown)


scripts/airdrop.js
```diff
@@ -70,36 +98,69 @@ async function airdrop({ filePath, network, useOvm, providerUrl, privateKey, gas
...
 	.option('-g, --gas-price <value>', 'Gas price to set when performing transfers', 1)
+	.option('-i, --in-file-path <value>', 'The path to the JSON file containing the target addresses')
 	.option(
 		'-k, --private-key <value>',
 		'The private key of the address that will be used to transfer tokens from'
 	)
 	.option('-l, --gas-limit <value>', 'Max gas to use when signing transactions', 8000000)
-	.option('-n, --network <value>', 'The network to run off.', x => x.toLowerCase(), 'mainnet')
+	.option('-n, --network <value>', 'The network to run off.', (x) => x.toLowerCase(), 'mainnet')
+	.option('-l, --reset', 'Clear all data in output file', false)
```
justinjmoses, line 159: `-l` is overloaded which can cause issues.

justinjmoses (approved):
Just one note to fix

scripts/airdrop.js
ajsantander: Ah, missed that one. Thanks!

i-stam (approved):

PR #766, opened 2020-10-01
Always perform prod tests for mainnet

ajsantander:
Previously, prod tests for all networks were run only when merging to develop. With this change, this remains true, except mainnet prod tests are done on *every* PR.


jacko125 (approved):
@ajsantander  LGTM ! Will be good to get these fork gas tests on each PR !

PR #772, opened 2020-10-05
Improves scripts

ajsantander:
Interactive script:
* can use forks
* can sign txs

Airdrop
* can use forks

### DIFF
--- a/scripts/airdrop.js
+++ b/scripts/airdrop.js
@@ -5,7 +5,7 @@ const path = require('path');
 const program = require('commander');
 const { gray, cyan, yellow, red } = require('chalk');
 const { parseEther, formatEther } = require('ethers').utils;
-	const { wrap } = require('..');
+const { wrap } = require('..');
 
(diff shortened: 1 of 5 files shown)


i-stam:
lgtm :)

scripts/airdrop.js
```diff
@@ -32,7 +34,17 @@ async function airdrop({
 	providerUrl = providerUrl.replace('network', network);
 	if (!providerUrl) throw new Error('Cannot set up a provider.');
 
-	if (!privateKey) throw new Error('No private key available.');
+	let publicKey;
+	if (useFork) {
```
i-stam, line 38: perhaps we could add an `else` statement here setting `providerUrl` that is done at line 34.

ajsantander: Thanks for the review @i-stam 

I'd rather merge this now so that this util is available today. We can improve on this on other PRs?

PR #776, opened 2020-10-07
SIP-84 - re-add paused synths to feeds.json and synths.json

jacko125:
SIP-84 - re-add paused synths to feeds.json and synths.json

- feeds.json
- synths.json


zyzek (approved):

justinjmoses:
`test-deployments` fails as expected as the mainnet synths in this PR do not match what's currently on mainnet

PR #780, opened 2020-10-08
Deploy.js - getExistingContracts to use versions.json for existing addresses

jacko125:
- [x] Pending fix to ensure `local` tests run with using versions.json


hav-noms (dismissed):

justinjmoses (approved):

publish/src/commands/deploy.js
```diff
@@ -1262,6 +1263,37 @@ const deploy = async ({
 
 	console.log(gray(`\n------ CONFIGURE INVERSE SYNTHS ------\n`));
 
+	// when the oldExrates exists - meaning there is a valid ExchangeRates in the existing deployment.json
+	// for this environment (true for all environments except the initial deploy in 'local' during those tests)
+
+	// Load previous ExchangeRates contract from versions.json
+	const exchangeRatesVersions = getVersions({ network, byContract: true }).ExchangeRates;
```
justinjmoses, line 1270: this won't work in local mode though where there are no historic versions

justinjmoses (changes requested):

publish/src/commands/deploy.js
justinjmoses: you should also include `path, fs` as well

jacko125: will update this to be more comprehensive and handle getExistingContracts()

justinjmoses: Instead of this approach, why not have `getExistingContract` use `getVersions` instead, it's what we should do now that we have access to historical versions.

justinjmoses:
Ping on this @jacko125

jacko125:
> Ping on this @jacko125

Ready for review. Merged latest changes in deploy.js in. Tested against `local:test` run

justinjmoses (approved):
LGTM Jacko - if you remove the lock file change the tests should all pass.

justinjmoses (approved):

PR #786, opened 2020-10-08
Upgrade solidity-coverage to 0.7.11

cgewecke:
Per [#732 comment][1]: compilation has been failing after instrumenting contracts for coverage. The error looks like:
```
Compiling...
Downloading compiler version 0.5.16
* Line 1, Column 1
  Syntax error: value, object or array expected.
* Line 1, Column 2
  Extra non-whitespace after JSON value.

> solidity-coverage cleaning up, shutting down ganache server
```

Believe this is solcjs crashing on the size of the JSON object passed to it for compilation. Instrumentation is adding a lot of Solidity code to the files and you have > 120 contracts, some of which are large. 

PR updates solidity-coverage to use some new options that reduce the instrumentation footprint by 
40 - 50%. It disables data collection for Istanbul's statement and function measures, (neither of which are tracked by codecov)

+ Statement coverage is like:
  ```solidity
  // one line, one statement
  uint x = 5;

  // one line, two statements
  uint x = 5; uint y = 7;
  ```

+ Function coverage measures whether or not every defined function is called. But you can tell whether or not a 
function was called by looking at the coverage of its lines, so it's not that useful...

PR removes `EtherCollateralsUSD.sol`, `BinaryOptionMarketData.sol` from the skipFiles list. 

(Lots of package-lock changes here, idk. Upgraded sc's web3 and ganache deps as part of new release and ran installation for this PR w/ Node 12.16.0)

[1]: https://github.com/Synthetixio/synthetix/pull/732#issuecomment-705092226

### DIFF
--- a/.solcover.js
+++ b/.solcover.js
@@ -7,8 +7,6 @@ module.exports = {
 	skipFiles: [
 		'test-helpers',
 		'EscrowChecker.sol',
-		'EtherCollateralsUSD.sol',
-		'BinaryOptionMarketData.sol',
 	],
 	providerOptions: {
 		default_balance_ether: 10000000000000, // extra zero just in case (coverage consumes more gas)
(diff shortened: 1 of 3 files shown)


justinjmoses (approved):
:pray: thanks so much @cgewecke !

PR #787, opened 2020-10-08
Working local L1 L2 multi test

K-Ho:
Working full multi test when spinning up local L1 chain at `localhost:9545` and L2 chain at `localhost:8545`. Tutorial to test here: https://hackmd.io/MLrKSCuvRLuzSCUuTcneVg?view

Note: To make the test run faster, I changed the `local` network `config.json` and `synths.json` - This is probably not the right way to do this, so feel free to revert that change and switch the multi-test to use the correct network


test/multi/index.js
```diff
@@ -54,29 +54,27 @@ describe('deploy multiple instances', () => {
...
+		wallets.push(l2Setup.wallet);
+		l2Provider = l2Setup.provider
 	});
 
 	before('compile if needed', async () => {
-		// if (isCompileRequired()) {
 		// OPTIMISM -- Always re-build to avoid accidentally deploying OVM contracts to L1
 		console.log('Found source file modified after build. Rebuilding...');
 
 		await commands.build({ showContractSize: true, testHelpers: true });
```
justinjmoses, line 77: shouldn't we compile the second time with `--ovm`?

PR #793, opened 2020-10-11
Upgrade solidity-coverage to 0.7.11 (#786)

justinjmoses:


.solcover.js
```diff
@@ -20,4 +18,9 @@ module.exports = {
 		grep: '@cov-skip', // Find everything with this tag
 		invert: true, // Run the grep's inverse set.
 	},
+	// Reduce instrumentation footprint - volume of solidity code
+	// passed to compiler causes it to crash (See discussion PR #732)
+	// Line and branch coverage will still be reported.
+	measureStatementCoverage: false,
+	measureFunctionCoverage: false,
```
cgewecke, line 25:
```suggestion
	measureFunctionCoverage: true,
```

Toggling this back on might be necessary. Looks like codecov is combining function and line coverage for the totals, and there are a lot of dropped "lines" in this report. 

The report for `Depot` from `develop` a few days ago looks like this:
Screen Shot 2020-10-11 at 12 49 46 PM

But the report in this PR looks like:
Screen Shot 2020-10-11 at 12 53 50 PM

cgewecke: Or I think you could also keep both new options set to false and your future reports will just be relative to the new baseline.

zyzek (commented):
Will this have a bearing on branch coverage? For example, I presume we will no longer be able to tell if a `require` assertion failing is tested.

cgewecke:
Hi @zyzek, 

Branch coverage is still being measured but AFAIK codecov doesn't really report what they call "partial lines" for Solidity. They [rely on their own JS parsing of the files][1] to generate the branch representation (and the files aren't JS)

Coveralls *does* report branch coverage for Solidity fwiw. There's a little more about this [in the solidity-coverage docs, here][2]

You can also look at the html report generated when you run the tool at the command line. 


[1]: https://docs.codecov.io/docs/node#partial-line-coverage
[2]: https://github.com/sc-forks/solidity-coverage/blob/master/docs/faq.md#continuous-integration

jacko125 (approved):

PR #796, opened 2020-10-12
Merge staging into develop update mainnet synths.json

jacko125:


publish/assets.json
```diff
@@ -107,8 +107,8 @@
 		"sign": "",
 		"description": "Kyber Network"
 	},
-	"LEND": {
-		"asset": "LEND",
+	"AAVE": {
```
i-stam, line 110: My guess from this is that LEND is now obsolete since both LEND and AAVE will coexist for an indefinite amount of time and there is no 1-1 correspondence (100 LEND => 1AAVE).

justinjmoses:
I already merged master into develop and missed this.

PR #800, opened 2020-10-13
Add unit tests for mintableSynthetix

i-stam:


test/contracts/MintableSynthetix.unit.js
```diff
@@ -0,0 +1,156 @@
...
+				await onlyGivenAddressCanInvoke({
+					fnc: mintableSynthetix.mintSecondary,
+					args: [account1, 100],
+					address: secondaryDeposit,
+					accounts,
+					reason: 'Can only be invoked by the SecondaryDeposit contract',
+				});
+			});
+
+			it('should NOT allow any account to call mintSecondary()', async () => {
```
ajsantander, line 96: Is this `it` identical to the `it` above?

test/contracts/MintableSynthetix.unit.js
```diff
@@ -0,0 +1,156 @@
...
+		});
+
+		describe('mintSecondary()', async () => {
+			let mintSecondaryTx;
+			before('when secondaryDeposit calls mintSecondary()', async () => {
+				mintSecondaryTx = await mintableSynthetix.mintSecondary(account1, 100, {
+					from: secondaryDeposit,
+				});
+			});
+			it('should tranfer the tokens to the right account', async () => {
```
ajsantander, line 114: nit: consistency with spaces between blocks

test/contracts/MintableSynthetix.unit.js
```diff
@@ -0,0 +1,156 @@
...
+		const SYNTHETIX_TOTAL_SUPPLY = toWei('100000000');
+		before('deploy a new instance', async () => {
+			resolverMock = await artifacts.require('GenericMock').new();
+			await mockGenericContractFnc({
+				instance: resolverMock,
+				mock: 'AddressResolver',
+				fncName: 'getAddress',
+				returns: [secondaryDeposit],
+			});
+			proxy = await setupContract({
```
ajsantander, line 38: I think setupContract is more intended for spec tests, because it gives you a contract in the context of the bigger system. Unit tests can mock everything else, so that you can just focus on the contract in question. This would simplify the setup of these tests a lot.

However, if this works, and it increases test coverage, I would leave it for now and maybe correct it with another PR?

ajsantander (approved):

test/contracts/MintableSynthetix.unit.js
i-stam: fixed

test/contracts/MintableSynthetix.unit.js
i-stam: Agreed, I think that we should agree in some common patterns that we should follow while developing unit tests etc.
In this case, I wanted to reuse as much code as possible.
We should invite @justinjmoses in the discussion.

PR #802, opened 2020-10-14
Deploy.js: Skip AddressResolver-dependent steps until addresses have been properly imported.

zyzek:


justinjmoses (approved):
LGTM

publish/src/commands/deploy.js
```diff
@@ -1300,444 +1301,472 @@ const deploy = async ({
...
+			console.log(gray('Snapshot complete.'));
+		} else {
+			console.log(gray('No snapshot required.'));
+		}
 	} else {
-		console.log(gray('No snapshot required.'));
+		console.log(
+			gray('Addresses are not set up, owner actions must be performed before continuing.')
+		);
 	}
```
ajsantander, line 1770: @zyzek maybe it would be a good idea to either (1) exit the script here, or (2) prompt the user for continuing after owner actions are set.

ajsantander (approved):
Just left one suggestion.

As a general comment, something to consider when refactoring this file, is that it is so big that a minor change diffed a lot of lines and made this diff hard to review, when it is not such a big change logically.

ajsantander (changes requested):
Sorry, changing my previous approval to "request changes" to make sure that my comment is addressed. Perhaps we should stop the script or hold it instead of just letting it continue with a console log.

publish/src/commands/deploy.js
zyzek: I think I prefer just exiting the script; prompting the user to continue also I think necessitates checking that everything has been imported properly before continuing successfully. I did start implementing this but I realised that we get the same result anyway by just re-running deploy.js.

It's already very close to the end of the script, so maybe it's a bit much to actually exit early, but at least I can change the printed message and exit with status code 1.

ajsantander: That's awesome. Thanks @zyzek

ajsantander (approved):

ajsantander (approved):

PR #803, opened 2020-10-14
Add unit tests for MintableSynthetix

i-stam:
- Rename the old .unit tests to .spec

- Use a new MinatbleSynthetixMock (to avoid conflicts with FakeMintableSynthetix for now)


test/contracts/MintableSynthetix.spec.js
```diff
@@ -0,0 +1,116 @@
...
+const { onlyGivenAddressCanInvoke, ensureOnlyExpectedMutativeFunctions } = require('./helpers');
+const { setupAllContracts } = require('./setup');
+const { toWei } = web3.utils;
+const { toBytes32 } = require('../..');
+const BN = require('bn.js');
+
+const MintableSynthetix = artifacts.require('MintableSynthetix');
+const SYNTHETIX_TOTAL_SUPPLY = toWei('100000000');
+
+contract('MintableSynthetix (unit tests)', accounts => {
```
justinjmoses, line 12: spec tests?

contracts/test-helpers/FakeMintableSynthetix.sol
```diff
@@ -1,11 +1,28 @@
...
+
+    constructor(
+        address payable _proxy,
+        TokenState _tokenState,
+        address _owner,
+        uint _totalSupply,
+        address _resolver
+    ) public MintableSynthetix(_proxy, _tokenState, _owner, _totalSupply, _resolver) {}
+
+    function emitTransfer(
```
justinjmoses, line 19: If you are going to override this and set params, why not check the params are set in the unit test? Or even just a flag to say it's called?

contracts/test-helpers/FakeMintableSynthetix.sol
```diff
@@ -1,11 +1,28 @@
...
-contract FakeMintableSynthetix {
-    address public mintSecondaryCall_account;
-    uint public mintSecondaryCall_amount;
+import "../MintableSynthetix.sol";
 
-    function mintSecondary(address account, uint amount) external {
-        mintSecondaryCall_account = account;
-        mintSecondaryCall_amount = amount;
+
+contract FakeMintableSynthetix is MintableSynthetix {
```
ajsantander, line 6: Hmmm... after looking at this code, I'm going back to the idea that FakeMintableSynthetix should be merged with MintableSynthetixMock. What does each one do?

FakeMintableSynthetix
* Mocks internal calls to emitTransfer

MintableSynthetixMock
* Mocks calls to mintSecondary

They're both mocking stuff, with the subtle difference that one copies the signature of the constructor, and the other doesn't. So, in that sense, none of these are fakes. A Fake, in the sense of what we discussed offline, is a the same contract with some functions stubbed out. A mock, is a contract that measures calls. In that sense, a Mock is a superset of a Fake.

So, again, my suggestion here is that these two are merged as MintableSinthetixMock. While we're at it, we should agree on nomenclature: put Fake/Mock at the end?

test/contracts/MintableSynthetix.spec.js
```diff
@@ -0,0 +1,116 @@
...
+	});
+
+	let mintableSynthetix;
+	let addressResolver;
+	describe('when system is setup', () => {
+		before('deploy a new instance', async () => {
+			({
+				MintableSynthetix: mintableSynthetix,
+				AddressResolver: addressResolver,
+			} = await setupAllContracts({
```
ajsantander, line 30: BOOM

yes 🙌

contracts/test-helpers/FakeMintableSynthetix.sol
i-stam: the only problem is that the mock overrides  `mintSecondary()`, so when the parent function needs to be called, the overriding one will be called instead. Maybe there is an easy way to distinguish between them, but I didn't think about this much, I just wanted to make the tests work without breaking the rest of the tests.

contracts/test-helpers/FakeMintableSynthetix.sol
i-stam: done

contracts/test-helpers/FakeMintableSynthetix.sol
ajsantander: Ah, I understand now. Ok, so then just nit about consistent naming, that's all.

@i-stam feel free to resolve this thread

ajsantander (approved):

test/contracts/MintableSynthetix.spec.js
```diff
@@ -0,0 +1,116 @@
...
+		describe('burnSecondary()', async () => {
+			let burnSecondaryTx;
+			const amount = 100;
+			before('when secondaryDeposit calls burnSecondary()', async () => {
+				burnSecondaryTx = await mintableSynthetix.burnSecondary(account1, amount, {
+					from: secondaryDeposit,
+				});
+			});
+			it('should tranfer the tokens to the right account', async () => {
+				assert.equal(0, await mintableSynthetix.balanceOf(account1));
```
justinjmoses, line 100: The actual should come first then the expected - you need to reverse the order of parameters here.

test/contracts/MintableSynthetix.unit.js
```diff
@@ -106,54 +93,41 @@ contract('MintableSynthetix (unit tests)', accounts => {
...
 			});
 
-			it('should emit a Transfer event', async () => {
-				assert.eventEqual(mintSecondaryTx, 'Transfer', {
-					from: mintableSynthetix.address,
-					to: account1,
-					value: amount,
-				});
+			it('should invoke emitTransfer', async () => {
+				assert.equal(mintableSynthetix.address, await mintableSynthetix.from());
```
justinjmoses, line 109: Here as well

justinjmoses (approved):

PR #804, opened 2020-10-15
Commented out unused code in SecondaryDeposit.sol

ajsantander:


i-stam (approved):

i-stam:
for completeness we should remove `import "./interfaces/IRewardEscrow.sol";` too ;)

PR #805, opened 2020-10-15
Implements two gas reports, one for unit tests, and another for prod tests

ajsantander:
I requested this feature in eth-gas-reporter: https://github.com/cgewecke/eth-gas-reporter/issues/217

This produces a new gas report in the "Checks" section and causes CI to fail if a method call consumes more than 25% than it did before.

Note that the "prod gas report" will only be produced on commits to the develop branch.

### DIFF
--- a/buidler.config.js
+++ b/buidler.config.js
@@ -306,6 +306,7 @@ module.exports = {
 		enabled: false,
 		showTimeSpent: true,
 		currency: 'USD',
+		maxMethodDiff: 25, // CI will fail if gas usage is > than this %
 		outputFile: 'test-gas-used.log',
 	},
 };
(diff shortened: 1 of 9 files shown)


buidler.config.js
```diff
@@ -306,6 +306,7 @@ module.exports = {
 		enabled: false,
 		showTimeSpent: true,
 		currency: 'USD',
+		maxMethodDiff: 25, // CI will fail if gas usage is > than this %
```
jacko125, line 309: How can we override this failure if the expected gas usage is going to be more than 25% (something that can't be optimised for example the recent exchange updates for debt snapshot might increase it by more than 25% ?)

jacko125 (approved):
lgtm, will be helpful !

buidler.config.js
ajsantander: I suggest that in those cases, the percentage is incremented in the PR, but it has to be done consciously, and that's the whole point. After merge, another PR could lower the percentage again.

PR #806, opened 2020-10-15
Enable withdrawal functionality in L2

i-stam:


contracts/test-helpers/CrossDomainMessengerMock.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    // mock xDomainMessageSender()
+    function xDomainMessageSender() external view returns (address) {
+        return xDomainMsgSender;
+    }
+
+    // Events
+    event MintedSecondary(address indexed account, uint amount);
+    event WithdrawalCompleted(address indexed account, uint amount);
+}
```
justinjmoses, line 54: Please make sure your IDE has add newlines at end of files enabled.

contracts/SecondaryDeposit.sol
```diff
@@ -23,14 +23,14 @@ contract SecondaryDeposit is Owned, MixinResolver, MixinSystemSettings, ISeconda
     bytes32 private constant CONTRACT_EXT_MESSENGER = "ext:Messenger";
     bytes32 private constant CONTRACT_SYNTHETIX = "Synthetix";
     bytes32 private constant CONTRACT_ISSUER = "Issuer";
-    bytes32 private constant CONTRACT_REWARDESCROW = "RewardEscrow";
+    // bytes32 private constant CONTRACT_REWARDESCROW = "RewardEscrow";
     bytes32 private constant CONTRACT_ALT_SECONDARYDEPOSIT = "alt:SecondaryDeposit";
```
justinjmoses, line 27: This needs to be `alt:SecondaryWithdrawal`

justinjmoses:
For naming convention, how about we call the L2 contract `SecondaryManager`? Because it does two things - one is relayed a message to mint on L2 and the other is to initiate a withdrawal onto L1?

So we'd have:
- L1 `SecondaryDeposit`, whose job is to manage deposits - holding onto the L1 `SNX`
- L2 `SecondaryManager` whose job it is to relay deposits to L2 Synthetix and to initiate withdraws back to L1.

In terms of labels, I'd suggest we keep both as unique, so for the AddressResolver we'd import `SecondaryDeposit` in L1 and `SecondaryManager` in L2. In the deploy script, we deploy based off config.json, which would configure which to deploy in which environment. In terms of linkage, L1 would have a link to `alt:SecondaryManager` and L2 would have a link to `alt:SecondaryDeposit`. 

Another alternative is to simply have L1 deposits relayed to Synthetix on L2 directly. But then L2 Synthetix (`MintableSynthetix`) needs to know about `ext:Messenger` in order to accept incoming messages, as well as `alt:SecondaryDeposit` as it's the sender via the messenger. 

Thoughts?

contracts/SecondaryDeposit.sol
i-stam: I left it like that till we decide on the naming convention, to me this could be just renamed to `companion`, thus, we use the same name on both layers.

i-stam:
> For naming convention, how about we call the L2 contract `SecondaryManager`? Because it does two things - one is relayed a message to mint on L2 and the other is to initiate a withdrawal onto L1?
> 
> So we'd have:
> 
> * L1 `SecondaryDeposit`, whose job is to manage deposits - holding onto the L1 `SNX`
> * L2 `SecondaryManager` whose job it is to relay deposits to L2 Synthetix and to initiate withdraws back to L1.
> 
> In terms of labels, I'd suggest we keep both as unique, so for the AddressResolver we'd import `SecondaryDeposit` in L1 and `SecondaryManager` in L2. In the deploy script, we deploy based off config.json, which would configure which to deploy in which environment. In terms of linkage, L1 would have a link to `alt:SecondaryManager` and L2 would have a link to `alt:SecondaryDeposit`.
> 
> Another alternative is to simply have L1 deposits relayed to Synthetix on L2 directly. But then L2 Synthetix (`MintableSynthetix`) needs to know about `ext:Messenger` in order to accept incoming messages, as well as `alt:SecondaryDeposit` as it's the sender via the messenger.
> 
> Thoughts?

I wanted to finish the tests in order to have a more concrete understanding. The naming you suggest sounds ok. Both contracts have a deposit/withdraw component, so in theory, they could be named L1/L2SecondaryManager which is more explicit in terms of where each contract "sits". As far as the messenger naming is concerned, I suggest removing the `alt:` prefix as it does not add any information (not used anywhere else) and rename them to just `companion` which is the way we refer to them everywhere else in the code. 
The 2nd approach I think is less optimal since we will have to edit the code and the flow we have already implemented and we may encounter new hurdles (unless there is a clear advantage).

contracts/SynthetixL1ToL2Bridge.sol
```diff
@@ -4,34 +4,34 @@ pragma solidity ^0.5.16;
...
     bool public activated;
 
     /* ========== ADDRESS RESOLVER CONFIGURATION ========== */
     bytes32 private constant CONTRACT_EXT_MESSENGER = "ext:Messenger";
     bytes32 private constant CONTRACT_SYNTHETIX = "Synthetix";
     bytes32 private constant CONTRACT_ISSUER = "Issuer";
-    bytes32 private constant CONTRACT_REWARDESCROW = "RewardEscrow";
-    bytes32 private constant CONTRACT_ALT_SECONDARYDEPOSIT = "alt:SecondaryDeposit";
+    // bytes32 private constant CONTRACT_REWARDESCROW = "RewardEscrow";
+    bytes32 private constant CONTRACT_ALT_SYNTHETIX_BRIDGE = "alt:SynthetixOptimisticBridge";
```
justinjmoses, line 27: This needs changing

contracts/SynthetixL2ToL1Bridge.sol
```diff
@@ -0,0 +1,76 @@
...
+
+// solhint-disable indent
+import "@eth-optimism/rollup-contracts/build/contracts/bridge/interfaces/CrossDomainMessenger.interface.sol";
+
+
+contract SynthetixL2ToL1Bridge is Owned, MixinResolver, ISynthetixL2ToL1Bridge {
+    /* ========== ADDRESS RESOLVER CONFIGURATION ========== */
+    bytes32 private constant CONTRACT_EXT_MESSENGER = "ext:Messenger";
+    bytes32 private constant CONTRACT_SYNTHETIX = "Synthetix";
+    bytes32 private constant CONTRACT_ALT_SYNTHETIX_BRIDGE = "alt:SynthetixOptimisticBridge";
```
justinjmoses, line 19: And here

contracts/SynthetixL2ToL1Bridge.sol
```diff
@@ -0,0 +1,76 @@
...
+    // invoked by user on L2
+    function initiateWithdrawal(uint amount) external {
+        // instruct L2 Synthetix to burn this supply
+        synthetix().burnSecondary(msg.sender, amount);
+
+        // create message payload for L1
+        bytes memory messageData = abi.encodeWithSignature("completeWithdrawal(address,uint256)", msg.sender, amount);
+
+        // relay the message to Bridge onL1  via L2 Messenger
+        messenger().sendMessage(synthetixBridge(), messageData, 3e6);
```
justinjmoses, line 54: We should probably make this a constant

contracts/SynthetixL1ToL2Bridge.sol
```diff
@@ -101,72 +101,41 @@ contract SecondaryDeposit is Owned, MixinResolver, MixinSystemSettings, ISeconda
         bytes memory messageData = abi.encodeWithSignature("mintSecondaryFromDeposit(address,uint256)", msg.sender, amount);
 
         // relay the message to this contract on L2 via Messenger1
-        messenger().sendMessage(companion(), messageData, 3e6);
+        messenger().sendMessage(synthetixBridge(), messageData, 3e6);
```
justinjmoses, line 104: We should use a constant for this gas limit

test/multi-same-chain/index.js
```diff
@@ -122,14 +145,28 @@ describe('deploy multiple instances', () => {
...
+				bridgeAlt = fetchContract({ contract: 'SynthetixL1ToL2Bridge', instance: 1 - i });
+				await resolver.importAddresses(
+					[toBytes32('alt:SynthetixOptimisticBridge')],
+					[bridgeAlt.address]
+				);
+			} else {
+				contract = fetchContract({ contract: 'SynthetixL1ToL2Bridge', instance: i });
+				bridgeAlt = fetchContract({ contract: 'SynthetixL2ToL1Bridge', instance: 1 - i });
+				await resolver.importAddresses(
+					[toBytes32('alt:SynthetixOptimisticBridge')],
```
justinjmoses, line 164: Oh you want to call it the same thing. Hmmm. 🤔

contracts/SynthetixL1ToL2Bridge.sol
i-stam: Please suggest another name if you have a specific preference. My 2 cents: this one is now skipped by the deploy script so it doesn't cause problems. Having something  that is the same on both sides is useful:
1.  no copy-paste errors while setting the resolver.
2. it reflects the dual nature of the bridges + the layer is implied by the alt prefix.
3. Using optimistic as you mentioned helps in case of future L2 additions, it contains all info required: a. OVM, b. bridge on the other (alt) layer c. bridge for depositing.withdrawing SNX (Synthetix prefix)

justinjmoses: `ovm:SynthetixBridgeToBase`

`base:SynthetixBridgeToOptimism`

test/multi-same-chain/index.js
```diff
@@ -122,14 +145,28 @@ describe('deploy multiple instances', () => {
...
-	before('tell each deposit contract about the other', async () => {
+	before('tell each contract about the other', async () => {
 		for (const i of [0, 1]) {
 			const resolver = fetchContract({ contract: 'AddressResolver', instance: i });
-			const deposit = fetchContract({ contract: 'SecondaryDeposit', instance: i });
-			const depositAlt = fetchContract({ contract: 'SecondaryDeposit', instance: 1 - i });
-			await resolver.importAddresses([toBytes32('alt:SecondaryDeposit')], [depositAlt.address]);
+			let contract;
+			let bridgeAlt;
+			if (i) {
```
justinjmoses, line 153: [np] if we are using the if branch then no need to use the for loop

justinjmoses (approved):
LGTM!

test/multi-same-chain/index.js
i-stam: loop unrolling huh?  we can do it, there are two extra instructions that need to be unrolled

justinjmoses (approved):

PR #809, opened 2020-10-16
Makes CI fail when prod tests on mainnet fail

ajsantander:
CI was not failing when test:prod failed in some cases like this one: https://app.circleci.com/pipelines/github/Synthetixio/synthetix/4075/workflows/cf1728df-edcd-44e2-842a-59184378973d/jobs/27939


package.json
```diff
@@ -24,7 +24,7 @@
 		"test:legacy": "npm run compile:legacy && buidler test:legacy",
 		"test:ovm": "buidler test --ovm --optimizer",
 		"test:prod": "buidler test:prod --network localhost",
-		"test:prod:gas": "buidler test:prod --gas --network localhost --gas-output-file test-gas-used-prod.log || cat test-gas-used-prod.log",
+		"test:prod:gas": "buidler test:prod --gas --network localhost --gas-output-file test-gas-used-prod.log",
```
justinjmoses, line 27: Can you add test-gas-used-prod.log to a `store_artifacts` step in Circle CI?

ajsantander: Yeah, this is interesting.

Actually, why are we not storing artifacts for `test-gas-used.log`? I.e. why would we need to store the prod version of this log file when we don't do so for the regular version?

As far as I can see, we're only storing artifacts for two jobs:
* test-publish
* pack

ajsantander: I'll store both

justinjmoses (approved):

PR #811, opened 2020-10-19
SIP-91 Debt snapshots L2 refactor

zyzek:
This PR pulls debt snapshot functionality out into its own contract, `DebtCache.sol`, in order to reduce the size of the `Issuer` so that it's deployable on L2, and so that the debt cache can be replaced on L2 with a version that does not require snapshots to be taken. Several new functions have been added to `DebtCache`, and a number of the functions that used to be in `Issuer` have been renamed to produce a cleaner interface. See [SIP-91](https://sips.synthetix.io/sips/sip-91) for additional details.

The L2 version of `DebtCache` is called `RealtimeDebtCache`. This version has the same interface as `DebtCache`, but reports real-time data instead of cached values from all functions that previously reported cached information. All cache updating operations have been turned into no-ops.

Additionally, this PR adds functions to allow multiple synths to be added and removed from the Issuer at once, which speeds up deployments, particularly when redeploying the Issuer itself.

- [x] Write [the SIP](https://github.com/Synthetixio/SIPs/pull/269)
- [x] Separate snapshot logic into new `DebtCache` contract and redo the interface
- [x] Update setup.js, fix existing tests, migrating logic into new test file (note no .spec/.unit distinction in DebtCache.js as the logic was just pulled out of Issuer.js which did not follow this pattern)
- [x] Realtime/L2 `DebtCache` version
- [x] Add tests for new functionality
- [x] Fix deploy.js (in particular replacing `IssuerWithoutUpdatableCache` with `RealtimeDebtCache` on `useOvm`)
- [x] Remove flexible storage from debt cache, add public getters, initialise invalidity to true
- [x] Add multi-synth add/remove functions to issuer and update deploy.js to use it
- [x] Ensure all tests are passing on CI


contracts/RealtimeDebtCache.sol
```diff
@@ -0,0 +1,48 @@
+pragma solidity ^0.5.16;
+
+// Inheritance
+import "./DebtCache.sol";
+
+
+// https://docs.synthetix.io/contracts/RealtimeDebtCache
```
justinjmoses, line 7: actually these urls are wrong - it should be /contracts/source/contracts/RealtimeDebtCache

publish/src/commands/deploy.js
```diff
@@ -689,9 +689,16 @@ const deploy = async ({
...
+		name: 'DebtCache',
+		source: useOvm ? 'RealtimeDebtCache' : 'DebtCache',
+		deps: ['AddressResolver'],
+		args: [account, resolverAddress],
+	});
+
 	const exchanger = await deployer.deployContract({
 		name: 'Exchanger',
-		deps: ['AddressResolver'],
+		deps: ['AddressResolver', 'DebtCache'],
```
justinjmoses, line 701: You don't need to do this - this is just for deps that would break contract instantiation. For those required by the address resolver, we don't include them in deps (otherwise we'd have to change the deploy script a whole lot)

publish/src/commands/deploy.js
```diff
@@ -753,8 +760,7 @@ const deploy = async ({
 
 	const issuer = await deployer.deployContract({
 		name: 'Issuer',
-		source: useOvm ? 'IssuerWithoutUpdatableCache' : 'Issuer',
-		deps: ['AddressResolver'],
+		deps: ['AddressResolver', 'DebtCache'],
```
justinjmoses, line 763: (ibid)

test/contracts/setup.js
```diff
@@ -780,8 +787,11 @@ const setupAllContracts = async ({
 		if (returnObj['Synth']) {
 			returnObj['Issuer'].addSynth(returnObj['Synth'].address, { from: owner });
 		}
-		for (const synthAddress of synthsToAdd) {
-			await returnObj['Issuer'].addSynth(synthAddress, { from: owner });
+
+		const chunkSize = 15;
```
justinjmoses, line 791: This is probably overkill - can just keep using the singular `addSynth` for the test setup script

justinjmoses (dismissed):
Just a few nps, otherwise LGTM

contracts/RealtimeDebtCache.sol
zyzek: Ah, right you are. I've fixed this for all the existing files.

publish/src/commands/deploy.js
zyzek: Noted, and fixed.

publish/src/commands/deploy.js
zyzek: Also fixed.

test/contracts/setup.js
zyzek: I've reverted this. The idea was to see if it sped up any test suites, but given there aren't any that add a lot of synths, it seems not to make a difference.

contracts/ContractStorage.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity ^0.5.16;
 import "./interfaces/IAddressResolver.sol";
 
 
-// https://docs.synthetix.io/contracts/source/contracts/ContractStorage
+// https://docs.synthetix.io/contracts/source/contracts/contractstorage
```
jacko125, line 7: Is this supposed to be Upper case for `ContractStorage` ?

contracts/RealtimeDebtCache.sol
```diff
@@ -0,0 +1,48 @@
...
+
+// Inheritance
+import "./DebtCache.sol";
+
+
+// https://docs.synthetix.io/contracts/source/contracts/realtimedebtcache
+contract RealtimeDebtCache is DebtCache {
+    constructor(address _owner, address _resolver) public DebtCache(_owner, _resolver) {}
+
+    // Report the current debt values from all all cached debt functions
```
ajsantander, line 11: nit: "all all" typo

ajsantander (changes requested):
Looking good @zyzek !

A couple minor requests (apart from those made via inline comments):
1) Add Issuer and DebtCache to releases.json
2) Remove Issuer and IssuerWithoutUpdatableCache from ovm-ignore.json

contracts/DebtCache.sol
```diff
@@ -0,0 +1,322 @@
...
+
+    /* ========== MUTATIVE FUNCTIONS ========== */
+
+    // This function exists in case a synth is ever somehow removed without its snapshot being updated.
+    function purgeCachedSynthDebt(bytes32 currencyKey) external onlyOwner {
+        require(issuer().synths(currencyKey) == ISynth(0), "Synth exists");
+        delete cachedSynthDebt[currencyKey];
+    }
+
+    function takeDebtSnapshot() external requireSystemActiveIfNotOwner {
```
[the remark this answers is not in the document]
zyzek: There is already a `_currentDebt`, but I am not using it, since `takeDebtSnapshot` on each loop iteration needs to cache the debt of the synth it's currently up to, so it seemed more graceful to recapitulate and extend the logic from `_currentDebt`.

contracts/RealtimeDebtCache.sol
zyzek: Nice catch; corrected.

zyzek:
Ok good call, I'm getting some mysterious reverts with the fork prod test when the tests try to issue or exchange synths through the `Synthetix` contract. No error messages and this doesn't affect issuance for some reason. I'll investigate...

zyzek:
Fixed the prod tests, ready for another look @ajsantander .

publish/releases.json
```diff
@@ -48,7 +48,14 @@
 			"major": 2,
 			"minor": 31
 		},
-		"sources": ["SystemSettings", "Synthetix", "SecondaryDeposit"],
-		"sips": []
+		"sources": [
+			"SystemSettings",
+			"Synthetix",
+			"SynthetixBridgeToOptimism",
```
ajsantander, line 54: Do we need to deploy SynthetixBridgeToOptimism in L1 mainnet right now?

ajsantander:
@zyzek I ran the prod tests after your latest changes and they all pass, awesome 🙌

ajsantander (approved):

publish/releases.json
zyzek: Removed this and SystemSettings.

PR #812, opened 2020-10-20
SIP 79: Deferred Transaction Gas Tank

0xclem:


contracts/test-helpers/StubFlexibleStorage.sol
```diff
@@ -0,0 +1,294 @@
+pragma solidity ^0.5.16;
+
+import "../interfaces/IFlexibleStorage.sol";
+
+
+contract StubFlexibleStorage is IFlexibleStorage {
```
justinjmoses, line 6: This seems like a huge stub. Why not just use the real `FlexibleStorage` instead of rewriting all this?

contracts/GasTank.sol
```diff
@@ -0,0 +1,186 @@
...
+import "./SafeDecimalMath.sol";
+
+// Internal references
+import "./interfaces/ISystemStatus.sol";
+import "./interfaces/ISystemSettings.sol";
+import "./interfaces/IDelegateApprovals.sol";
+import "./interfaces/IExchangeRates.sol";
+
+
+contract GasTank is Owned, MixinResolver, ReentrancyGuard, MixinSystemSettings {
```
justinjmoses, line 19: Please implement the `IGasTank` interface

contracts/GasTank.sol
```diff
@@ -0,0 +1,187 @@
...
+    function currentGasPrice() public view returns (uint currentGasPriceWei) {
+        return _exchangeRates().rateForCurrency("fastGasPrice");
+    }
+
+    function currentEtherPrice() public view returns (uint currentGasPriceWei) {
+        return _exchangeRates().rateForCurrency("ETH");
+    }
+
+    function executionCost(uint _gas) public view returns (uint etherCost) {
+        return (_gas + PAYGAS_COST) * currentGasPrice() + getKeeperFee().divideDecimal(currentEtherPrice());
```
justinjmoses, line 94: Can you use SafeMath operators

contracts/GasTank.sol
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    function _setMaxGasPrice(address _account, uint _gasPrice) internal {
+        require(_gasPrice > 0, "Gas Price must be greater than 0");
+        flexibleStorage().setUIntValue(CONTRACT_NAME, keccak256(abi.encodePacked(MAX_GAS_PRICE, _account)), _gasPrice);
+        emit MaxGasPriceSet(_account, _gasPrice);
+    }
+
+    function approveContract(bytes32 _contractName, bool _approve) external onlyOwner {
+        address contractAddress = resolver.requireAndGetAddress(_contractName, "Missing contract address");
```
justinjmoses, line 127: Note this won't handle upgrades of contracts, as you are persisting the address of the contract that has access, not the label. Perhaps the label is better to persist?

test/contracts/GasTank.spec.js
```diff
@@ -0,0 +1,333 @@
...
+			from: owner,
+		});
+		await systemSettings.setKeeperFee(keeperFeeDefault, { from: owner });
+		await gasTank.setResolverAndSyncCache(addressResolver.address, { from: owner });
+		await keeper.setResolverAndSyncCache(addressResolver.address, { from: owner });
+	});
+
+	addSnapshotBeforeRestoreAfterEach();
+
+	beforeEach(async () => {
```
justinjmoses, line 43: we still need the `ensureOnlyExpectedMutativeFunctions` checks

test/contracts/GasTank.spec.js
```diff
@@ -0,0 +1,333 @@
...
+			const keeperFee = toUnit(Number(keeperFeeDefault) / Number(ethPriceDefault));
+
+			assert.bnEqual(
+				await gasTank.executionCost(gasAmount),
+				toBN(transactionPrice).add(toBN(keeperFee))
+			);
+		});
+	});
+
+	describe('approveContract', () => {
```
justinjmoses, line 90: You need to use `onlyGivenAddressCanInvoke` to ensure it is protected

test/contracts/GasTank.spec.js
```diff
@@ -0,0 +1,333 @@
...
+					from: accountOne,
+				});
+				assert.bnEqual(await gasTank.maxGasPriceOf(accountTwo), toBN(maxGasPrice));
+			});
+		});
+	});
+
+	describe('payGas', () => {
+		it('should revert if contract has not been approved', async () => {
+			await assert.revert(
```
justinjmoses, line 270: please use `onlyGivenAddressCanInvoke` with no `address` so it checks the range of `accounts`

contracts/GasTank.sol
```diff
@@ -0,0 +1,187 @@
...
+        require(_amount > 0, "Deposit must be greater than 0");
+        _setDepositBalance(_account, _amount);
+        emit EtherDeposited(msg.sender, _account, _amount);
+    }
+
+    function _withdrawEther(
+        address _account,
+        address payable _recipient,
+        uint _amount
+    ) internal nonReentrant {
```
justinjmoses, line 113: We should check re-entrancy protection in a test.

test/contracts/setup.js
```diff
@@ -624,6 +625,17 @@ const setupAllContracts = async ({
...
 		},
+		{
+			contract: 'GasTank',
+			mocks: ['Synthetix'],
+			deps: [
+				'SystemStatus',
+				'SystemSettings',
+				'DelegateApprovals',
+				'ExchangeRates',
+				'SynthetixState',
```
justinjmoses, line 636: You can't mock `SynthetixState` as well?

justinjmoses (changes requested):
Comments inline

contracts/GasTank.sol
```diff
@@ -0,0 +1,187 @@
...
+
+    function payGas(
+        address _spender,
+        address payable _recipient,
+        uint _gas
+    ) external nonReentrant returns (uint) {
+        require(approved[msg.sender], "Contract is not approved");
+        uint etherSpent = executionCost(_gas);
+        require(tx.gasprice >= currentGasPrice(), "Gas price is too low");
+        if (maxGasPriceOf(_spender) > 0) {
```
zyzek, line 171: You can save `maxGasPriceOf(_spender)` to a local variable to save the overhead of a second function call inside the conditional.

While we're on it: does the extra constant in the `executionCost` call correspond to `maxGasPriceOf(_spender) == 0` or `maxGasPriceOf(_spender) > 0`? I guess it should be the latter.

contracts/GasTank.sol
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    function payGas(
+        address _spender,
+        address payable _recipient,
+        uint _gas
+    ) external nonReentrant returns (uint) {
+        require(approved[msg.sender], "Contract is not approved");
+        uint etherSpent = executionCost(_gas);
+        require(tx.gasprice >= currentGasPrice(), "Gas price is too low");
```
zyzek, line 170: We're doing two calls to `currentGasPrice`; one here and one inside `executionCost`. I'd probably compute this once an then pass it into an internal version of `executionCost` which takes in the gas price as an argument, and is also called by the public function.

I'd also check `tx.gasprice >= currentGasPrice()` before computing `etherSpent`, which will save a bit of gas in case the tx fails... Though given that this `payGas` operation is happening at the very end of the execution maybe that's pretty negligible.

I forget if we made any decisions about contracts being able to "lock" a value of the user's ether; e.g. you submit a limit order and you have to keep at least enough ether in there to cover the execution cost of that order.

contracts/SystemSettings.sol
```diff
@@ -38,6 +38,9 @@ contract SystemSettings is Owned, MixinResolver, MixinSystemSettings, ISystemSet
     // Minimum Stake time may not exceed 1 weeks.
     uint public constant MAX_MINIMUM_STAKE_TIME = 1 weeks;
 
+    // Max USD value of ether to be awarded to keepers
+    uint public constant MAX_KEEPER_FEE = 1e18 * 10;
```
zyzek, line 42: Could just be `10e18` I think?

contracts/GasTank.sol
```diff
@@ -0,0 +1,187 @@
...
+
+// Internal references
+import "./interfaces/ISystemStatus.sol";
+import "./interfaces/ISystemSettings.sol";
+import "./interfaces/IDelegateApprovals.sol";
+import "./interfaces/IExchangeRates.sol";
+import "./interfaces/IGasTank.sol";
+
+
+contract GasTank is Owned, MixinResolver, ReentrancyGuard, MixinSystemSettings, IGasTank {
```
zyzek, line 20: Add a link to the documentation page that will be generated for this contract: https://docs.synthetix.io/contracts/source/contracts/gastank. You should also do the same for the `IGasTank` interface.

contracts/SystemSettings.sol
zyzek, line 42: We probably should add this parameter as a value in the `defaults` object defined in `index.js`. Not sure what we agreed the initial default value should be. Was it zero?

With the default, we'll also need to add a test that it's been set up properly in `test/publish/index.js`, along with all the other default value tests. (line 260 or so)

test/contracts/GasTank.spec.js
```diff
@@ -0,0 +1,333 @@
...
+
+			const tx = await keeper.spendGas(accountTwo, {
+				from: keeperAccount,
+				gasPrice: fastGasPriceDefault,
+			});
+
+			const keeperAccountBalance = await web3.eth.getBalance(keeperAccount);
+
+			// the final balance should be greater than initial balance (tx refund + keeper fee)
+			assert.bnGt(toBN(keeperAccountBalance), toBN(keeperAccountInitialEthBalance));
```
zyzek, line 324: We should be able to check that this is sending back the right amount by checking something like:

```
const gasSpent = tx.receipt.gasUsed;
const gasRefund = gasSpent.mul(fastGasPriceDefault);
const keeperFee = divideDecimalRound(keeperFeeDefault, ethPriceDefault); // imported from test/utils/index.js

assert.bnClose(keeperAccountBalance, keeperAccountInitialEthBalance.add(gasRefund).add(keeperFee));
```

Some of the particular names and logic might be off but that's the general idea.

zyzek: Might have to pass bnClose a slightly larger threshold value as the third argument, but I think it's important to make sure the gas being refunded is not too far off its intended value.

contracts/test-helpers/Keeper.sol
```diff
@@ -0,0 +1,33 @@
...
+    function _gasTank() internal view returns (IGasTank) {
+        return IGasTank(requireAndGetAddress(CONTRACT_GASTANK, "Missing GasTank address"));
+    }
+
+    function _synthetix() internal view returns (ISynthetix) {
+        return ISynthetix(requireAndGetAddress(CONTRACT_SYNTHETIX, "Missing Synthetix address"));
+    }
+
+    function spendGas(address payable _account) public {
+        uint gasConsummed = gasleft();
```
zyzek, line 27: nitpick: Consummed -> Consumed

contracts/GasTank.sol
```diff
@@ -0,0 +1,187 @@
...
+        uint _gas
+    ) external nonReentrant returns (uint) {
+        require(approved[msg.sender], "Contract is not approved");
+        uint etherSpent = executionCost(_gas);
+        require(tx.gasprice >= currentGasPrice(), "Gas price is too low");
+        if (maxGasPriceOf(_spender) > 0) {
+            require(tx.gasprice <= maxGasPriceOf(_spender), "Spender gas price limit is reached");
+        }
+        _setDepositBalance(_spender, balanceOf(_spender).sub(etherSpent));
+        _recipient.transfer(etherSpent);
```
zyzek, line 175: So consensys recommends that we don't use `.transfer` anymore. https://diligence.consensys.net/blog/2019/09/stop-using-soliditys-transfer-now/

Should we remove it here, and in `_withdrawEther`?

Thoughts on this @justinjmoses @ajsantander ?

zyzek (changes requested):

test/contracts/setup.js
clementbalestrat: Actually @justinjmoses I realised I can't mock `Synthetix`. Some of the tests fail with `Transaction reverted: function returned an unexpected amount of data`

contracts/GasTank.sol
clementbalestrat: hmmm I'm not sure. 

How do you handle `require(approved[msg.sender], "Contract is not approved");` in `PayGas` then?

contracts/GasTank.sol
clementbalestrat: Then it probably makes sense to test every contract using this protection?

contracts/GasTank.sol
clementbalestrat: regarding your last point, I can't remember either. Let's decide when we actually implement a keeper.

contracts/SystemSettings.sol
clementbalestrat: Probably not `MAX_KEEPER_FEE` as it is just a limit. I guess you meant the default `KEEPER_FEE`.

test/contracts/GasTank.spec.js
clementbalestrat: yeah good call

PR #813, opened 2020-10-20
Improve OVM contract compilation

ajsantander:
Experimental PR for OVM contract compilation using the stricter Optimism solc compiler that will throw when invalid opcodes like selfdestruct are used.

Contracts that are not intended for OVM compilation are ignored using an @ovm-ignore flag in the solidity code. A Buidler internal task is overriden to filter the source paths.


ajsantander:
The filtering is currently working, but there's a bit of a technical problem: SelfDestructible contains an opcode which is forbidden in the OVM, `selfdestruct` and we can easily mark it so that it's ignored in OVM compilations. However, it is part of Synthetix's inheritance tree:

Screen Shot 2020-10-20 at 16 45 16

There are other cases in which this can happen.

ajsantander:
This is a log of compilation errors/warnings when using the new optimism-solc compiler: https://hackmd.io/uH8oskKsQ0uqEY5Ry8hu9Q

ajsantander:
This PR was good to understand the problem and possible ways to address it. So, closing it in favor or smaller PRs.

PR #814, opened 2020-10-22
Use strict OVM compiler (removes SelfDestructible)

ajsantander:
Second attempt for #813 

#### Key changes
* Uses the stricter Optimism solc compiler that will throw when invalid opcodes like selfdestruct are used.
* Contracts that are not intended for OVM compilation are ignored by adding their names to `publish/ovm-ignore.json`.

#### Completed tasks / checks
- [x] Update eth-optimism/solc to v0.5.16-alpha.3
- [x] Filter sources in `npx buidler compile`
- [x] Remove SelfDestructible
  - [x] Remove inheritance from all contracts, one by one, and pass tests
  - [x] Delete SelfDestruct.sol, SelfDestruct.js, SelfDestruct from all tests and pass tests
  - [x] SelfDestruct cannot be found in sources (except legacy)
  - [x] Compiles in normal mode and passes all tests
- [x] Successful buidler compile with ovm flag
- [x] Filter sources in `node publish build`
- [x] Successful build with ovm flag
- [x] Can we do anything about duplicate log messages in buidler compilation filter?


buidler.config.js
```diff
@@ -165,6 +166,30 @@ const optimizeIfRequired = ({ bre, taskArguments: { optimizer } }) => {
 	bre.optimizer = !!optimizer;
 };
 
+// This overrides a buidler internal task, which is part of its compile task's lifecycle.
+// This allows us to filter out non OVM compatible contracts from the compilation list,
+// which are entries in publish/deployed/ovm-ignore.json.
+internalTask('compile:get-source-paths', async (_, { config }, runSuper) => {
```
ajsantander, line 172: This is how we filter out non-OVM compatible contracts with Buidler compilation

ajsantander:
Screen Shot 2020-10-22 at 12 03 28

Note on this ☝️ 

Asked the Buidler team if it is normal that the `compile:get-source-paths` is called twice in the `compile` task's lifecycle and they said that unfortunately yes. Gonna be fixed in the upcoming hardhat release tho.

ajsantander:
@justinjmoses have you ever encountered this problem if coverage dropping because a file was removed?

publish/ovm-ignore.json
```diff
@@ -0,0 +1,14 @@
+[
```
justinjmoses, line 1: [np] Maybe just an array of strings, unless you think there's value in a full object?

publish/src/commands/build.js
```diff
@@ -48,6 +48,24 @@ const build = async ({
 		ignore: [].concat(!testHelpers ? /^test-helpers\// : []),
 	});
 
+	if (useOVM) {
+		const ovmIgnored = JSON.parse(fs.readFileSync('publish/ovm-ignore.json'));
```
justinjmoses, line 52: should we put this in a require in the top level index file and then load it from there? (as above with the buidler.config.js)

justinjmoses:
qu: why `--force` in the example above? Because nothing had changed but you wanted to compile regardless?

justinjmoses:
> have you ever encountered this problem of coverage dropping because a file was removed?

I imagine it's smart enough to know that overall code has decreased - did see if you could find anything missing in the line diff in codecov? Worst case we can just ignore the diff

justinjmoses (approved):
LGTM, with a few small suggestions. Would like other eyes on it though.

buidler.config.js
```diff
@@ -165,6 +166,66 @@ const optimizeIfRequired = ({ bre, taskArguments: { optimizer } }) => {
...
+	let filePaths = await runSuper();
+
+	if (config.ignoreNonOvmContracts) {
+		const ovmIgnored = JSON.parse(fs.readFileSync('publish/ovm-ignore.json'));
```
justinjmoses, line 176: see comment below on `index.js` suggestion

contracts/BinaryOptionMarketFactory.sol
```diff
@@ -2,14 +2,13 @@ pragma solidity ^0.5.16;
...
 import "./Owned.sol";
-import "./SelfDestructible.sol";
 import "./MixinResolver.sol";
 
 // Internal references
 import "./BinaryOptionMarket.sol";
 
 
-contract BinaryOptionMarketFactory is Owned, SelfDestructible, MixinResolver {
+contract BinaryOptionMarketFactory is Owned, MixinResolver {
```
zyzek, line 11: This is only informational; we can worry about it if/when we ever come to deploying binary options to L2, but the underlying `BinaryOptionMarket` and `BinaryOption` contracts still use the selfdestruct opcode, though not through inheriting `SelfDestructible`.

buidler.config.js
```diff
@@ -165,6 +166,66 @@ const optimizeIfRequired = ({ bre, taskArguments: { optimizer } }) => {
 	bre.optimizer = !!optimizer;
 };
 
+// This overrides a buidler internal task, which is part of its compile task's lifecycle.
+// This allows us to filter out non OVM compatible contracts from the compilation list,
+// which are entries in publish/deployed/ovm-ignore.json.
```
zyzek, line 171: File path should be `publish/ovm-ignore.json`, no?

zyzek (dismissed):
Looks good. Nothing much to add to Justin's remarks except a miniscule comment.

jacko125:
@ajsantander Have you ran through what's the implications of removing SelfDestructible from 

- Depot.sol
- FeePool.sol
- FeePoolState.sol
- ExchangeRates.sol
- ExternStateToken.sol

I don't see any impediment to removing them other than we would lose the ability selfdestruct and clean those contracts up on mainnet when we release a new version of those contracts in the future. 

Just a note, not being able to selfDestruct these contracts would mean if theres any ether stuck in the contracts we won't be able to retrieve the ether by self destructing them. 

Otherwise the changes look good!

jacko125 (approved):

publish/ovm-ignore.json
ajsantander: Yeah, not sure why I went the object route. Changed to an array of strings 👍

publish/src/commands/build.js
ajsantander: Done, top level index now exposes an `ovmIgnored` object which both build and buidler use.

ajsantander:
> qu: why `--force` in the example above? Because nothing had changed but you wanted to compile regardless?

@justinjmoses 
If you are referring to the screenshot I posted, then no particular reason. I must have been using it to be 100% sure that the filtering worked.

ajsantander:
> > have you ever encountered this problem of coverage dropping because a file was removed?
> 
> I imagine it's smart enough to know that overall code has decreased - did see if you could find anything missing in the line diff in codecov? Worst case we can just ignore the diff

@justinjmoses 
Yes, I did go in and look at that, and I saw that no new dark red blocks can be found in the diff. So, I guess its not smart enough. I can ping gcewecke, or we could ignore.

buidler.config.js
ajsantander: Applied!

contracts/BinaryOptionMarketFactory.sol
ajsantander: Yep! I saw it. However, since we are ignoring all binary option stuff with the new filtering mechanism introduced in this PR, we can compile just fine with OVM.

buidler.config.js
ajsantander: Ah. Fixed.

ajsantander:
> @ajsantander Have you ran through what's the implications of removing SelfDestructible from
> 
>     * Depot.sol
> 
>     * FeePool.sol
> 
>     * FeePoolState.sol
> 
>     * ExchangeRates.sol
> 
>     * ExternStateToken.sol
> 
> 
> I don't see any impediment to removing them other than we would lose the ability selfdestruct and clean those contracts up on mainnet when we release a new version of those contracts in the future.
> 
> Just a note, not being able to selfDestruct these contracts would mean if theres any ether stuck in the contracts we won't be able to retrieve the ether by self destructing them.
> 
> Otherwise the changes look good!

@jacko125 let's discuss offline

jacko125 (dismissed):
LGTM!

publish/src/commands/build.js
ajsantander: Hmmm... a kinda weird implication of it being at the top level is that I was required to publish the file (see package.json) to pass linter checks.

.circleci/config.yml
```diff
@@ -276,8 +286,8 @@ jobs:
                 at: .
             - run:
                 command: |
-                    npx buidler compile --ovm --optimizer
-                name: Compile in OVM, enforcing size limit
+                    npx buidler compile --use-ovm --optimizer
```
justinjmoses, line 289: Should this not do the `fail-oversize` instead of the compile task?

publish/src/commands/build.js
```diff
@@ -212,5 +228,6 @@ module.exports = {
 			.option('-s, --show-contract-size', 'Show contract sizes')
 			.option('-t, --test-helpers', 'Also compile the test-helpers')
 			.option('-w, --show-warnings', 'Show warnings')
+			.option('-z, --use-ovm', 'Use Optimism OVM-compatible compiler')
```
justinjmoses, line 231: :pray:

justinjmoses (dismissed):
Just one small comment about the oversize failure on compile - just a np. LGTM.

i-stam (approved):

jacko125 (dismissed):

.circleci/config.yml
ajsantander: My idea was that only `job-compile` checks for oversizes, and this `job-coverage-report` does not.

justinjmoses (approved):

PR #815, opened 2020-10-22
Ci compile job

ajsantander:
#### Changes overview
* Adds a new `--fail-oversize` flag to the buidler compile task, that causes the task to throw if any contract is oversized.
* Creates a new `job-compile` circleci job that compiles contracts in both OVM and non-OVM modes using the new flag and adds it into `workflow-dev`.
* Note: This PR branches off #814, so that one needs to be merged first, and then this one needs to be repointed to `develop`.
* Note: The new job actually fails, because we haven't merged in zizek's refactor of Issuer.

#### Details
- [x] Disable `job-test-ovm` for now, since it's not being used
- [x] Create the new `job-compile` and add it to `workflow-dev`
- [x] Add `--fail-oversize` flag to buidler compilation


justinjmoses (approved):

zyzek (approved):

.circleci/config.yml
```diff
@@ -277,7 +287,7 @@ jobs:
             - run:
                 command: |
                     npx buidler compile --ovm --optimizer
-                name: Compile in OVM, enforcing size limit
+                name: Run unit tests on OVM compiled output (TODO enable tests)
```
zyzek, line 290: Approved; the nitpickest: not sure if we want to leave these TODOs around or not.

ajsantander: Thanks @zyzek 

Don't worry, I'm tracking this todo within the same epic that tracks the issue for this PR. As soon as we merge this in, I will be enabling OVM tests and removing this TODO.

PR #818, opened 2020-10-26
Send rewards into l2

justinjmoses:
Also updates `SynthetixBridgeToOptimism` to use new smockit testing library


contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -120,9 +141,28 @@ contract SynthetixBridgeToOptimism is Owned, MixinResolver, ISynthetixBridgeToOp
         emit BridgeMigrated(address(this), newBridge, contractBalance);
     }
 
+    // invoked by RewardsDistribution on L1 (takes SNX)
+    function notifyRewardAmount(uint256 reward) external {
```
i-stam, line 145: It seems that `notifyRewardAmount() `and `rewardDeposit() `can be combined and merged

i-stam (approved):

PR #820, opened 2020-10-28
Use smock to test L2 contracts

i-stam:


contracts/test-helpers/FakeSynthetixBridgeToBase.sol
```diff
@@ -17,7 +17,7 @@ contract FakeSynthetixBridgeToBase is SynthetixBridgeToBase {
     ) public SynthetixBridgeToBase(_owner, _resolver) {
         mintableSynthetix = ISynthetix(_mockMintableSynthetix);
         xChainBridge = _bridge;
-        crossDomainMessengerMock = address(new MockCrossDomainMessenger(_bridge));
+        crossDomainMessengerMock = address(new MockCrossDomainMessenger());
```
justinjmoses, line 20: We can delete this file now no?

contracts/test-helpers/MockCrossDomainMessenger.sol
```diff
@@ -5,16 +5,10 @@ import "../interfaces/ISynthetixBridgeToBase.sol";
 
 
 contract MockCrossDomainMessenger {
-    address public xDomainMsgSender;
```
justinjmoses, line 8: We should be able to delete this as well no?

test/contracts/SynthetixBridgeToBase.unit.js
```diff
@@ -25,152 +22,161 @@ contract('SynthetixBridgeToBase (unit tests)', accounts => {
...
+			artifacts.require('SynthetixBridgeToOptimism').abi.find(({ name }) => name === fnc),
+			args
+		);
+
+	describe('when all the deps are (s)mocked', () => {
+		let messenger;
+		let mintableSynthetix;
+		let resolver;
+		beforeEach(async () => {
+			messenger = await smockit(artifacts.require('MockCrossDomainMessenger').abi, {
```
justinjmoses, line 36: Instead of mocking this target, you can mock the interface and it should work as expected

justinjmoses (approved):
Looking good - just please remove those unused testing files. Otherwise LGTM.

contracts/test-helpers/FakeSynthetixBridgeToBase.sol
i-stam: yes, we don't need all those "fake" contracts anymore

justinjmoses (approved):

PR #821, opened 2020-10-28
SIP 89 Virtual Synths

justinjmoses:


justinjmoses:
Example of running `Synthetix.prod.js`

image

at the same time, here was the rate on 1inch

image

contracts/Exchanger.sol
```diff
@@ -527,6 +589,37 @@ contract Exchanger is Owned, MixinResolver, MixinSystemSettings, IExchanger {
...
+
+        if (virtualSynth) {
+            vSynth = _createVirtualSynth(dest, recipient, amountReceived);
+            dest.issue(address(vSynth), amountReceived);
+        } else {
+            dest.issue(recipient, amountReceived);
+        }
+    }
+
+    function _createVirtualSynth(
```
jacko125, line 615: Is the concept with `ExchangerWithVirtualSynth` that overrides the internal function `_createVirtualSynth` in the Exchanger class supposed to be that the former will be whats deployed on L1 and the normal "Exchanger" contract will be only be on OVM ?

contracts/Exchanger.sol
```diff
@@ -377,19 +396,40 @@ contract Exchanger is Owned, MixinResolver, MixinSystemSettings, IExchanger {
...
+            exchangeForAddress,
+            false
         );
 
         _processTradingRewards(fee, originator);
 
         _emitTrackingEvent(trackingCode, destinationCurrencyKey, amountReceived);
     }
 
+    function exchangeWithVirtual(
```
jacko125, line 413: I'd split the VirtualSynth creation outside of the internal `_exchange` function for separation of concerns so that `exchangeWithVirtual()` would create a VirtualSynth instead of relying on a bool flag to determine if an VirtualSynth is created within `_exchange`.

This could be.a limitation with the stack too deep issue instead of returning the required variables to setup a virtual synth not inside `_exchange` ?

contracts/Exchanger.sol
```diff
@@ -527,6 +589,37 @@ contract Exchanger is Owned, MixinResolver, MixinSystemSettings, IExchanger {
...
+    ) internal returns (IVirtualSynth vSynth) {
+        // Burn the source amount
+        issuer().synths(sourceCurrencyKey).burn(from, sourceAmountAfterSettlement);
+
+        // Issue their new synths
+        ISynth dest = issuer().synths(destinationCurrencyKey);
+
+        if (virtualSynth) {
+            vSynth = _createVirtualSynth(dest, recipient, amountReceived);
+            dest.issue(address(vSynth), amountReceived);
```
zyzek, line 609: It's a bit of a nitpick, but a minor amount of space can be saved by reassigning the recipient address and using only one `issue` call, as follows:

```solidity
if (virtualSynth) {
    recipient = address(_createVirtualSynth(dest, recipient, amountReceived);
}

dest.issue(recipient, amountReceived);
```

contracts/Exchanger.sol
zyzek: I was thinking this too; however the destination account needs to be known before the exchange (which is either the vSynth or the original recipient), and the vSynth needs to know the amount received after the exchange.
It's possible to do this by breaking the `_exchange` function up into the bits before the call to `_convert`, and the bits after, but I'm not sure this is any cleaner than what's here. Perhaps we can just do the exchange quantity calculation by itself and construction of the vSynth before rather than after the exchange itself, and pass these in as parameters?

contracts/Exchanger.sol
zyzek: Alternately, `_createVirtualSynth` on this layer could just return the recipient or zero address, and then the if statement can be removed, and the `virtualSynth` bool argument removed entirely in the enclosing call to _exchange by relying on the function overriding. E.g.:

```solidity
vSynth = _createVirtualSynth(dest, recipient, amountReceived); // on layer 2 returns address(0);
address recipientOrSynth = vSynth == address(0) ? recipient : vSynth;
dest.issue(recipient, amountReceived);
```

justinjmoses:
This needs to be moved to a PR with a new branch name to trigger the `workflow-develop` CI flow

contracts/ExchangerWithVirtualSynth.sol
```diff
@@ -0,0 +1,27 @@
...
+        uint amount
+    ) internal returns (IVirtualSynth vSynth) {
+        // Note: should actually pass ReadProxyAddressResolver here to future-proof this vSynth,
+        // but we don't have it currently in the AddressResolver...
+        // Solution
+        vSynth = new VirtualSynth(synth, resolver, recipient, amount);
+        emit VirtualSynthCreated(address(vSynth), address(synth), synth.currencyKey(), amount);
+    }
+
+    event VirtualSynthCreated(address vSynth, address synth, bytes32 currencyKey, uint amount);
```
zyzek, line 26: Does this strictly need the currency key? One assumes it could always be recovered from the synth itself at the block this was emitted.

zyzek (changes requested):
Leaving several comments now, but will continue reviewing.

contracts/VirtualSynth.sol
```diff
@@ -0,0 +1,142 @@
...
+
+    function name() external view returns (string memory) {
+        return string(abi.encodePacked("Virtual Synth ", synth.currencyKey()));
+    }
+
+    function symbol() external view returns (string memory) {
+        return string(abi.encodePacked("v", synth.currencyKey()));
+    }
+
+    function decimals() external pure returns (uint8) {
```
zyzek, line 89: Do we need both a public `DECIMALS` constant and this external `decimals` function? Isn't renaming `DECIMALS` to `decimals` sufficient to satisfy the ERC20 extended interface? Or am I missing some integration that expects the `DECIMALS` form?

contracts/interfaces/IVirtualSynth.sol
```diff
@@ -0,0 +1,20 @@
+pragma solidity >=0.4.24;
+
+import "./ISynth.sol";
+
+
+interface IVirtualSynth {
```
zyzek, line 6: `secsLeftInWaitingPeriod` seems to be missing from the interface; `resolver`, as well, if we think that's needed at all.

zyzek (changes requested):

contracts/Exchanger.sol
justinjmoses: Yeah Anton's point answers your Jacko. But doing the creation first could definitely work and might be a lot nicer. Let me look into this.

contracts/ExchangerWithVirtualSynth.sol
justinjmoses: No not strictly - it just seemed like it might be better for reporting - but def not critical.

contracts/VirtualSynth.sol
```diff
@@ -0,0 +1,142 @@
...
+    constructor(
+        ISynth _synth,
+        IAddressResolver _resolver,
+        address _recipient,
+        uint _amount
+    ) public ERC20() {
+        synth = _synth;
+        resolver = _resolver;
+
+        // Note: we can do this as Exchanger will issue this amount to us
```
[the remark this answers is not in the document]
justinjmoses: yep

contracts/VirtualSynth.sol
justinjmoses: I was just following the convention in `Synth.sol` but yeah, different scenario. I'll fix it up.

contracts/interfaces/IVirtualSynth.sol
justinjmoses: Agreed on `secs` but not necessary on the `resolver`.

justinjmoses:
Closed for #857

PR #822, opened 2020-10-28
Deploy script checks compiled contracts expected compiler version

ajsantander:
Since we're using solc, or a modified version of solc when we're targeting an Optimism deploy, our contract artifacts can be significantly different depending on which compiler was used.

Atm, the consequences of deploying incorrect artifacts to a network (i.e. OVM artifacts to Mainnet L1, or regular artifacts to Mainnet L2) are no entirely understood, but we are quite sure that they will be undesirable.

This PR adds a check to the deploy script, targeting an expected solc version, depending on the `--use-ovm` flag.

** UPDATE**
* Optimism has implemented our request of including "ovm" in their compiler versions, so instead of checking for specific compiler versions, the code now checks for "ovm" in the version. 
* Additionally, I've added a check on deploymentPath, to make sure that we don't accidentally target the 'goerli' deployment path when using ovm.
* Finally, I've fixed a problem with update-versions.js, so that ovm is also considered.


justinjmoses (dismissed):

i-stam:
Ah, now the same chain multi tests fail, because they are using the same compiler version. but the PR looks like a great idea.

publish/deployed/goerli-ovm/versions.json
```diff
@@ -1 +1,207 @@
-{}
+{
+	"v2.30.0": {
+		"tag": "v2.30.0",
+		"fulltag": "v2.30.0",
+		"release": "Deneb",
+		"network": "goerli",
```
ajsantander, line 6: Needs to set for goerli-ovm?

ajsantander: No, the network is ok, and the versions.json file is in the correct path.

publish/src/commands/deploy.js
```diff
@@ -91,11 +88,33 @@ const deploy = async ({
 		network,
 	});
 
-	// Fresh deploy and deployment.json not empty?
-	if (freshDeploy && Object.keys(deployment.targets).length > 0 && network !== 'local') {
-		throw new Error(
-			`Cannot make a fresh deploy on ${deploymentPath} because a deployment has already been made on this path. If you intend to deploy a new instance, use a different path or delete the deployment files for this one.`
-		);
+	if (!ignoreSafetyChecks) {
```
jacko125, line 91: @ajsantander  What was broken with the deploy.js script that was introduced with #780  ?

ajsantander: Replied in PM with the exact location and reason.

publish/src/commands/versions-update.js
```diff
@@ -10,16 +10,23 @@ const { stringify, loadAndCheckRequiredSources } = require('../util');
...
+const versionsUpdate = async ({ versionTag, release, useOvm }) => {
 	console.log(gray('Checking deployments for version:', versionTag));
 
 	// prefix a "v" to the tag
 	versionTag = /^v/.test(versionTag) ? versionTag : 'v' + versionTag;
 
-	for (const network of networks.filter(n => n !== 'local')) {
+	let applicableNetworks = networks;
+	if (useOvm) {
+		applicableNetworks = ['goerli'];
```
justinjmoses, line 21: it would be much nicer to not have to hard-code this. Perhaps just check if `getPathToNetwork` exists and `continue` if not?

justinjmoses (dismissed):
Just one small fix and :shipit:

publish/src/commands/versions-update.js
ajsantander: Yes, that's great. Implemented.

justinjmoses (approved):

justinjmoses (approved):

justinjmoses (approved):

PR #824, opened 2020-10-28
Fixing broken tests from latest clean install

justinjmoses:


cgewecke:
Peeking over your shoulder here sorry - we used Synthetix as a sanity check at Web3 for a bit and saw these errors with the Web3@1.2.8 release which upgraded its @ethers/abi package. 

There were a couple behavioral changes that affected you:

**Parameter bounds checking**
```
4) Contract: SystemSettings
       when integrating with modern contracts
         setTargetThreshold()
           reverts when owner set the Target threshold to negative:
     AssertionError: expected 'value out-of-bounds (argument="_percent", value=-1, code=INVALID_ARGUMENT, version=abi/5.0.0-beta.153)' to include 'revert'
      at Object.assertRevert [as revert] (test/utils/index.js:477:11)
```
Here `-1` is passed as a `uint`. Previously Web3 accepted this but let the value underflow to MAX_INT - 1. The new Ethers encoder catches it and complains.
 
**Immutable arrays**
```
1) Contract: BinaryOptionMarketManager @gas-skip @ovm-skip
       when integrating with modern contracts
         Market tracking
           Adding, resolving, and expiring markets properly updates market lists:
     TypeError: Cannot assign to read only property '0' of object '[object Array]'
      at Array.sort (<anonymous>)
```
The new Ethers returns arrays as immutable objects. In this test [a native JS sort is run directly on the return result][1]. It just needs to be copied before sorting, e.g
```js
const sorted = [...immutableArray].sort()
```

[1]:https://github.com/Synthetixio/synthetix/blob/b4771e9668b64fafc2526aaacde6ce2d1bd9d006/test/contracts/BinaryOptionMarketManager.js#L1000

justinjmoses:
Wow Chris. Just wow. Thank you so damn much.

test/contracts/BinaryOptionMarketManager.js
```diff
@@ -997,11 +997,13 @@ contract('BinaryOptionMarketManager @gas-skip @ovm-skip', accounts => {
 				.sort();
 
 			const createdMarkets = markets.map(m => m.address).sort();
-			let recordedMarkets = (await manager.activeMarkets(0, 100)).sort();
 
+			let recordedMarkets = await manager.activeMarkets(0, 100);
+			let recordedMarketsSorted = [...recordedMarkets].sort();
+			console.log(recordedMarketsSorted);
```
justinjmoses, line 1003: we can remove these logs no?

i-stam: done!

justinjmoses:
@cgewecke we're seeing coverage now fail repeatedly using the latest `ganache-core` (`2.13.1`) it seems:

https://app.circleci.com/pipelines/github/Synthetixio/synthetix/4216/workflows/3875c150-7730-4168-accd-046eba146a87/jobs/29805

Have any ideas why that might be?

justinjmoses:
Actually scratch that - it's due to the `smock` library somehow.

PR #825, opened 2020-10-29
New messenger interface

i-stam:
- Use the latest npm optimism contracts package (0.0.2-alpha.6)

- Update the messenger interface in the contract.


justinjmoses (approved):
Yeah I like this 👍

PR #826, opened 2020-10-30
Use preferred mainnet provider if available

ajsantander:
Adds a new PROVIDER_URL_MAINNET to .env. Scripts like deploy, fork, etc will use this provider instead of PROVIDER_URL if available.

Should be handy for using a specialized provider for mainnet in CI production tests.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,6 @@
 ETHERSCAN_KEY=SomeEtherscanKey
-PROVIDER_URL=https://network.infura.io/v3/SomeInfuraKey
-DEPLOY_PRIVATE_KEY= new one SomeEthereumPrivateKey (including 0x prefix)
-TESTNET_DEPLOY_PRIVATE_KEY=new one SomeEthereumPrivateKey (including 0x prefix)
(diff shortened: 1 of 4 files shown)


DONALDSULLENDER (approved):

DONALDSULLENDER (approved):

jacko125:
@ajsantander  PROVIDER_URL_MAINNET could be better named `PREFERRED_PROVIDER_URL` as that is more clear the intention ?

ajsantander:
> @ajsantander PROVIDER_URL_MAINNET could be better named `PREFERRED_PROVIDER_URL` as that is more clear the intention ?

@jacko125 I did give this plenty of thought, and I still prefer my choice. `PREFERRED_PROVIDER_URL` does not specify that its mainnet only, and something like `PREFERRED_MAINNET_PROVIDER_URL` would just be too long. ` PROVIDER_URL_MAINNET` might not be super clear right away, but makes sense once you think about it.

justinjmoses (approved):
LGTM but fine to wait for @jacko125's feedback on his suggestion

jacko125:
Fine with the parameter name for setting preferred mainnnet provider

PR #830, opened 2020-11-02
ci: add dockerhub auth

drptbl:
Added docker authentication while pulling images.

https://support.circleci.com/hc/en-us/articles/360050623311-Docker-Hub-rate-limiting-FAQ

Signed-off-by: Jakub Mucha <jakub.mucha@icloud.com>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,6 +79,9 @@ jobs:
     job-compile:
         docker:
             - image: circleci/node:12.18
+              auth:
+                username: $DOCKERHUB_USERNAME
+                password: $DOCKERHUB_TOKEN
         steps:
             - checkout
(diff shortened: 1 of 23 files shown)


.circleci/config.yml
```diff
@@ -79,6 +79,9 @@ jobs:
     job-compile:
         docker:
             - image: circleci/node:12.18
+              auth:
```
justinjmoses, line 82: one way to make this easier is to use yaml anchors. @ajsantander do you know if your concatenation script can easily support anchors or something similar to make this less painful?

ajsantander:
Closing this in favor of #924 

Thx @drptbl 
We we're in the process of implementing a templating system to ci. 924 simply re-applies this PR in those terms.

PR #851, opened 2020-11-06
SIP-91 audit remediation + realtime debt cache external overrides

zyzek:


i-stam (approved):
lgtm

contracts/DebtCache.sol
```diff
@@ -90,6 +90,22 @@ contract DebtCache is Owned, MixinResolver, MixinSystemSettings, IDebtCache {
         return getDebtSnapshotStaleTime();
     }
 
+    function cachedDebt() external view returns (uint) {
```
justinjmoses, line 93: [qu] Any reason to add these given they are auto-added by the compiler?

justinjmoses (approved):
:shipit:

contracts/DebtCache.sol
zyzek: Solidity didn't like me overriding the synthesised getters for these public members the public members in `RealtimeDebtCache` unless I explicitly exposed them this way.

PR #854, opened 2020-11-06
Adding fork to settlement and synth filtering

justinjmoses:


publish/src/commands/settle.js
```diff
@@ -71,6 +74,9 @@ const settle = async ({
 	const user = web3.eth.accounts.wallet.add(privateKey);
```
ajsantander, line 74: You might want to pwn the owner address here:
```	let account;
	if (useFork) {
		account = getUsers({ network, user: 'owner' }).address; // protocolDAO
	} else {
		web3.eth.accounts.wallet.add(privateKey);
		account = web3.eth.accounts.wallet[0].address;
	}
	console.log(gray(`Using account with public key ${account}`));
```

ajsantander (approved):

publish/src/commands/settle.js
justinjmoses: Hmm - useful but does ignore the privateKey passed to the command which may not always be intended...

PR #866, opened 2020-11-10
Develop sccp 56 set s eth s btc fee 30bps

jacko125:
- Updated deploy script to check the current Ex-rates matches what the defaults are we want. 

- Stages the new rates if the current rates deviated from the defaults.


publish/src/commands/deploy.js
```diff
@@ -1536,7 +1539,7 @@ const deploy = async ({
 						synth
 					)
 				)
-				.filter(({ currentRate }) => currentRate === '0');
+				.filter(({ currentRateWei, targetRate }) => currentRateWei !== targetRate);
```
jacko125, line 1542: @justinjmoses  Currently we only set defaults exhange Rates for a fee rate that isn't there `currentRate === '0'` 

We may want to keep that logic instead of having it compare the `currentRate with the TargetRate` to determine if we want to  update them. 

I can remove all the changes if so

justinjmoses (approved):

PR #871, opened 2020-11-11
sOIL & iOIL

hav-noms:
https://sips.synthetix.io/sips/sip-62


publish/deployed/mainnet/config.json
```diff
@@ -485,6 +485,24 @@
 	"SynthiADA": {
 		"deploy": false
 	},
+	"TokenStatesOIL": {
```
justinjmoses, line 488: so with new synths, you don't actually add these. You add to `synths.json` and add a `--add-new-synths` flag to the deploy script. I will remove these

justinjmoses (approved):

PR #876, opened 2020-11-11
Releasing Mimosa to rinkeby

justinjmoses:


jacko125 (approved):
LGTM , will keep eye out on edge cases when we are resetting isynths

jacko125 (approved):

PR #881, opened 2020-11-12
SIP-94 Fixing underlying issue with settler on vSynth

justinjmoses:
1. Major: synth settlement wasn't being applied to the vSynth but rather the account who began the exchange, rendering the vSynth fairly useless.
2. Minor: vSynth to use proxy of synth to allow for future upgrades.
3. Incidental: optional tracking code for `exchangeWithVirtual`


jacko125 (dismissed):
Approved.

contracts/Exchanger.sol
```diff
@@ -580,7 +581,7 @@ contract Exchanger is Owned, MixinResolver, MixinSystemSettings, IExchanger {
 
         // persist the exchange information for the dest key
         appendExchange(
-            destinationAddress,
```
zyzek, line 583: Do we not also need to do this when emitting the exchange event just above as well?

zyzek (changes requested):
One small query. Also just to point out that we're going to need to be careful about settling all vSynths if we ever purge any Synths with outstanding settlements in future.

jacko125:
> One small query. Also just to point out that we're going to need to be careful about settling all vSynths if we ever purge any Synths with outstanding settlements in future.

I think we should have a keeper settle and execute the vSynths to their final destinations as soon as possible. That was something discussed before so for the trader / user they wouldn't have to do anything manually to get the wBTC for example.

zyzek (approved):

jacko125 (approved):
Lgtm the changes

PR #883, opened 2020-11-12
Build circleci config with Mustache

ajsantander:
The CircleCI CLI has a command for packing a config file called "pack", but we found it to be very limited in terms of what it can do. Using Mustache would seem to be a much more powerful approach.

This PR also removes the complex circleci hook that often caused weird situations. Devs tweaking ci should know what they're doing (it's documented in the README within the .circleci folder).

PLEASE IGNORE FAILING NON-REQUIRED CI WORKFLOWS


package.json
```diff
@@ -32,7 +32,7 @@
 	},
 	"husky": {
 		"hooks": {
-			"pre-commit": "./hooks/circleci && npm run lint && pretty-quick --staged"
+			"pre-commit": "npm run lint && pretty-quick --staged"
```
justinjmoses, line 35: Qu: don't you want to validate at least on pre-commit?

package.json
```diff
@@ -129,6 +129,7 @@
 		"@eth-optimism/contracts": "0.0.2-alpha.7",
 		"abi-decoder": "2.3.0",
 		"commander": "5.1.0",
+		"mustache": "^4.0.1",
```
justinjmoses, line 132: Isn't this more of a dev dependency? It's really just to build CI config when we change it.

justinjmoses (changes requested):
Overall a good direction for sure, but a few things to consider

package.json
ajsantander: Moved to devDependencies

package.json
ajsantander: Actually, went a little bit further. This is now handled in the pack.js script, and will only run circleci if available in the OS.

justinjmoses: How is that happening?

ajsantander: It uses execa to run the circleci command, and catches failures: https://github.com/Synthetixio/synthetix/pull/883/files#diff-7906ff03d65275e67cb294aedd9c5f682a4301f8ba0d0ebeeb44aac8cc9a07c1R49-R54

Of course, it can also fail because the circleci command was run and exited with 1. In that case, the developer will have to pay attention to the output.

justinjmoses (approved):

PR #884, opened 2020-11-12
Smarter lint hook

ajsantander:
Uses `git diff` to only run linters if associated files have changed.

Additionally, we should consider using `npm run lint:fix` instead of `npx prettier-quick`.


jacko125 (approved):

hooks/lint
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+if ! git diff --name-only --exit-code -- '***.js'; then
+	npm run lint:js
+	npx pretty-quick --staged
```
justinjmoses, line 5: `pretty-quick` works on both `.js` and `.sol` so should run if either of the patterns were changed

justinjmoses (changes requested):

hooks/lint
ajsantander: @justinjmoses actually, you made me realize a better way to do this. Will open a new PR.

PR #890, opened 2020-11-13
Improved linting in pre-commit hook

ajsantander:
After @justinjmoses' comment in a previous attempt to speed up linting in the pre-commit hook, I understood how `pretty-quick` works.

We don't need to run `npm run lint` in the hook. Running `pretty-quick --staged --verbose` (notice the extra `--verbose` is more than enough) and is incredibly fast.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
 	},
 	"husky": {
 		"hooks": {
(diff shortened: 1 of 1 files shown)


justinjmoses (approved):
If it lints as well then i'm all for this

PR #896, opened 2020-11-16
Adds a --clean-build flag to the build command

ajsantander:
This is particularly important when compiling contracts for an ovm deploy. Some contracts may be ignored in `ovm-ignore.json` but included in a target ovm deployment like `publish/deployed/goerli-ovm`, which may result in evm code being deployed in the ovm.


i-stam (approved):
thanks for this!

PR #899, opened 2020-11-16
RFC: Resolver cache overhaul

justinjmoses:
**Goal**
To reduce the load on deploys - processing many `setResolverAndSyncCache` calls is simply not scalable. 

**Solidity Improvements**
* `AddressResolver.importAddresses()` now invokes the protected `invalidateCache()` function (protected to prevent griefing) for each `MixinResolver` target, where implemented
* `AddressResolver.importAddresses()` emits an event `AddressImported` for each address added
* `MixinResolver` has flag `isCacheValid`. When not valid, the view `requireAndGetAddress()` will revert (ensuring cache's are always up to date). This ensures caches are always rebuilt after imports otherwise the system will fail
* `MixinResolver` has a `rebuildCache()` function (publicly available) for the contract to rebuild it's cache and flip the cache flag
* `AddressResolver.rebuildCaches` is exposed (and non-protected), for use on deployment to batch up
* `MixinResolver` no longer needs the `Owner` pattern
* `MixinResolver.requireAndGetAddress()` uses `abi.encodePacked` for string concatenation and a simpler API.

**Deployment Improvements**
* All targets added to the `AddressResolver`
* All targets that have `MixinResolver.rebuildCache` exposed, will be added to a batched `AddressResolver.rebuildCaches` invocation

**Stats**
* `AddressResolver.importAddresses(all 140 local targets)` takes approximately **4.8m gas**
* `AddressResolver.rebuildCaches(all 47 local contracts with MixinResolver)` takes approximately **7.8m gas**

**Known Issues**
* `SynthetixBridge*` contracts will fail on `rebuildCache()` for those contracts with `:` in them. One potential solution is add checks to `rebuildCache` that will pass on anything with a `:` in it (a bit more gas intensive as it would have to loop over each character in the `bytes` array). c @i-stam @ajsantander

### DIFF
--- a/contracts/AddressResolver.sol
+++ b/contracts/AddressResolver.sol
@@ -2,9 +2,11 @@ pragma solidity ^0.5.16;
 
 // Inheritance
 import "./Owned.sol";
-
 import "./interfaces/IAddressResolver.sol";
+
+// Internal references
 import "./interfaces/IIssuer.sol";
+import "./MixinResolver.sol";
 
 
 // https://docs.synthetix.io/contracts/source/contracts/addressresolver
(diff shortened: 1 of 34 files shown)


justinjmoses:
Closed for #904

PR #901, opened 2020-11-16
Preparations for L2 phase c deployment

ajsantander:
The following changes were needed to be able to deploy in L2 uat:

* New optimism/solc compiler required by the new ovm
* TradingRewards is required as an AddressResolver dependency of Exchanger. We thought it best to remove functionality from TradingRewards for an extremely edge case scenario, instead of having a 2nd TradingRewards contract, or a third Exchanger contract.
* Synthetix is split into SynthetixBase + Synthetix so that the contracts fit when compiled with the new optimism/solc compiler.


ajsantander:
`job-compile` is failing because new OVM compiler produces larger contract sizes. Attempting to reduce resulting size in `MintableSynthetix` here https://github.com/Synthetixio/synthetix/pull/903

contracts/TradingRewards.sol
```diff
@@ -209,16 +209,6 @@ contract TradingRewards is ITradingRewards, ReentrancyGuard, Owned, Pausable, Mi
         emit NewPeriodStarted(_currentPeriodID);
     }
 
-    // Note: Contract does not accept ETH, but still could receive via selfdestruct.
```
ajsantander, line 212: Eth transfers are not compatible with L2, and Exchanger requires TradingRewards as an AddressResolver dependency. Instead of creating a new version of Exchanger, specific for L2, we decided to remove this extreme edge case feature from TradingRewards.

publish/deployed/goerli-ovm/deployment.json
```diff
@@ -2,18 +2,18 @@
 	"targets": {
 		"SafeDecimalMath": {
 			"name": "SafeDecimalMath",
-			"address": "0x212dD0505eDECA3947b07b86e0b0d44ba8eE81c4",
```
ajsantander, line 5: When Optimism upgraded their compiler, they had to recompile and perform surgery on the already deployed contracts, changing their code (remember, this is a super alpha testnet!). While doing so, they had to change the addresses of our libraries. This change makes sure that newly deployed contracts use the new libraries.

publish/deployed/goerli-ovm/params.json
```diff
@@ -1 +1 @@
-[{"name": "DEBT_SNAPSHOT_STALE_TIME", "value": "999999999999999999"}]
```
ajsantander, line 1: This parameter is no longer needed with the new ovm specific DebtCache.

publish/ovm-ignore.json
```diff
@@ -8,10 +8,8 @@
 	"EtherCollateralsUSD",
 	"ExchangerWithVirtualSynth",
 	"FakeTradingRewards",
-	"Issuer",
-	"IssuerWithoutUpdatableCache",
 	"MockBinaryOptionMarket",
 	"MockBinaryOptionMarketManager",
 	"TestableBinaryOptionMarket",
-	"TradingRewards"
+	"Synthetix"
```
ajsantander, line 14: These are contracts that are ignored for ovm compilation. We have a CI process that compiles stuff and errors if any is oversized. Compiling these with the ovm compiler would make that CI process fail.

publish/src/commands/connect-bridge.js
```diff
@@ -187,7 +187,7 @@ const bootstrapConnection = ({
 	});
 
 	// allow local deployments to use the private key passed as a CLI option
-	if (network !== 'local' || !privateKey) {
```
ajsantander, line 190: Minor fix to a new script we have for connecting up the bridge between L1 and L2.

contracts/TradingRewards.sol
jacko125: Sounds good !

jacko125 (approved):
LGTM these changes !

contracts/BaseSynthetix.sol
```diff
@@ -0,0 +1,328 @@
...
+    function mintSecondary(address, uint) external {
+        revert("Cannot be run on this layer");
+    }
+
+    function mintSecondaryRewards(uint) external {
+        revert("Cannot be run on this layer");
+    }
+
+    function burnSecondary(address, uint) external {
+        revert("Cannot be run on this layer");
```
zyzek, line 303: Very minor, but we could probably save a bit of code size by putting this duplicated revert statement inside a private function which gets called repeatedly.

zyzek (approved):
The PR looks fine to my eyes.

In general I feel as if we should just start moving away from this god Synthetix contract pattern and just move the various functions out to the issuer, exchanger, and so on. The less we need distinct contracts for the different layers the easier it will be to turn on more functionality on L2 without much effort.

ajsantander:
> The PR looks fine to my eyes.
> 
> In general I feel as if we should just start moving away from this god Synthetix contract pattern and just move the various functions out to the issuer, exchanger, and so on. The less we need distinct contracts for the different layers the easier it will be to turn on more functionality on L2 without much effort.

100%
Something to consider for v3 is not wasting modularity by having all the functions pass through a central contract.

contracts/BaseSynthetix.sol
ajsantander: Yeah. Good one. Applied.

PR #903, opened 2020-11-17
Fixes compilation issues with MintableSynthetix on L2

ajsantander:
Optimism's latest compiler, `@eth-optimism/solc 0.5.16-alpha.6`, produces slightly larger contract sizes. As a result, MintableSynthetix is ~106%. This PR attempts to remedy this by:
* Allowing oversized contracts in the inheritance chain if they're not intended for direct deployment. E.g. if MintableSynthetix inherits Synthetix, and sizeOf(MintableSynthetix) < sizeOf(Synthetix), and MintableSynthetix is the deployment target, then `ovm-ignore.json` can be used to indicate that we want to compile Synthetix, but exclude it from the oversize checks.
* Stubbing some methods in MintableSynthetix, that _probably_ not needed in L2.


ajsantander:
Splitting this PR into smaller PRs

PR #912, opened 2020-11-18
BaseSynthetix tests

i-stam:


test/contracts/setup.js
```diff
@@ -612,6 +658,27 @@ const setupAllContracts = async ({
 				'ExchangeRates',
 			],
 		},
+		{
+			contract: 'BaseSynthetix',
+			mocks: [
```
justinjmoses, line 663: [np] I'd probably create a base object for this and clone it, but we can do that another time

justinjmoses (approved):

PR #918, opened 2020-11-20
Improved lint pre-commit hook

ajsantander:
Third attempt! I think I got it right this time.

I'm quite happy with the result. It:
* Always runs pretty-quick
* Detects changes in js/sol and only runs linter on the changed files

Even if there is content to lint, it is still super fast because it only runs linters on the changed files.


package.json
```diff
@@ -31,7 +31,7 @@
 	},
 	"husky": {
 		"hooks": {
-			"pre-commit": "./hooks/circleci && npm run lint && pretty-quick --staged"
+			"pre-commit": "./hooks/circleci && ./hooks/circleci"
```
justinjmoses, line 34: shouldn't `hook/lint` be in here?

ajsantander: oops!

ajsantander: Fixed. Good catch. I should stop doing drugs.

PR #920, opened 2020-11-20
Deploy bug fix: adding synths

justinjmoses:
This bug prevented `sUSD` being added to `Issuer` in the recent OVM deploy of Phase C.

The bug occurs because the reduce needs a default state. Without a default and with a single entry, it will always return the single entry, which is a string and which coerces to `true`.

E.g.
```javascript
['0x0'].reduce((acc, cur) => acc && cur === '0x1';}) 
// 0x0 (which coerces to true)!
['0x0'].reduce((acc, cur) => acc && cur === '0x1', true) 
// false
```

Interestingly another bug in this code appears if the last entry matches but ignores the previous ones:

```javascript
['0x0', '0x1'].reduce((acc, cur) => acc && cur === '0x0'); 
// false
['0x0', '0x1'].reduce((acc, cur) => acc && cur === '0x1');
// true! 
```

Might be worth adding a line to the publish deploy script to test this condition?


zyzek (approved):
I believe the second condition is also addressed by the adding the default state, so your proposed fix looks good. 

Approving, but an alternative which is perhaps slightly clearer is to use `Array.every` instead of the `reduce`. I might also add a length check just in case there's some problem with the `getSynths` call:

```javascript
expected: input =>
    input.length === chunk.length &&
    input.every((cur, idx) => cur === addressOf(chunk[idx].synth)),
```

PR #924, opened 2020-11-23
Added dockerhub authetication to circleci

ajsantander:
Re-applies #830 with mustache templating.

Notice that the only src file changed was `.circleci/src/snippets/job-header.yml`, the rest are build artifacts.


drptbl (approved):

justinjmoses (approved):
So much better

PR #925, opened 2020-11-23
New command: migrate-bridge

ajsantander:
Calling `SynthetixBridgeToOptimism.migrateBridge(newBridgeAddress)` is a delicate operation. A mistake could potentially brick a lot of SNX.

This PR proposes a command to use instead of manually calling the function, which performs a couple of safety checks before actually executing the tx.


publish/src/commands/migrate-bridge.js
```diff
@@ -0,0 +1,203 @@
...
+} = require('../util');
+
+const migrateBridge = async ({
+	network,
+	deploymentPath,
+	privateKey,
+	useFork,
+	gasPrice,
+	gasLimit,
+	newBridgeAddress,
```
justinjmoses, line 21: Instead of passing the new bridge address, we can use `getVersions` like we do in the `import-fee-periods` script to figure out what was the last version of the bridge, and then use the `getTargets` for the new version.

publish/src/commands/migrate-bridge.js
```diff
@@ -0,0 +1,203 @@
...
+		await confirmAction(
+			yellow("⚠⚠⚠ WARNING: You're about to perform the SNX migration. Are you sure?")
+		);
+	} catch (err) {}
+
+	// -----------------------------------
+	// Ok, let's migrate...
+	// -----------------------------------
+
+	const tx = await oldBridge.methods.migrateBridge(newBridge.options.address).send(params);
```
jacko125, line 125: @ajsantander  If we're super cautious / concerned on the `migrateBridge` call that passes in a new Bridge address to transfer the SNX balance,  my advice would be to use an Approve and Withdraw 2 step process so that the new bridge would be able to withdraw / transfer the SNX by signing another 2nd tx. 

```
function migrateBridge(address newBridge) external onlyOwner requireActive {
        require(newBridge != address(0), "Cannot migrate to address 0");
        activated = false;

        IERC20 ERC20Synthetix = synthetixERC20();
        // get the current contract balance and transfer it to the new SynthetixL1ToL2Bridge contract
        uint contractBalance = ERC20Synthetix.balanceOf(address(this));
        ERC20Synthetix.transfer(newBridge, contractBalance);

        emit BridgeMigrated(address(this), newBridge, contractBalance);
    }  
```

ajsantander: Thanks @jacko125 I see your point.

@justinjmoses should we take it a step further and make the process safer by having the new bridge pull the SNX after the old bridge has staged the operation?

publish/src/commands/migrate-bridge.js
ajsantander: Implemented

publish/src/commands/migrate-bridge.js
```diff
@@ -0,0 +1,253 @@
...
+	const bridgeVersions = getVersions({ network, deploymentPath, byContract: true })
+		.SynthetixBridgeToOptimism;
+	if (bridgeVersions.length < 2) {
+		throw new Error(
+			red(
+				'❌ At least two bridge versions are needed for a migration. Please make sure to deploy new versions before migrating, and that they are registered versions'
+			)
+		);
+	}
+	const [secondLastEntry, lastEntry] = bridgeVersions.slice(-2);
```
justinjmoses, line 50: To be in `versions.json`, this means a version must have been released already, whereas it's likely this will be called *before* the release but after the deployment. 

See https://github.com/Synthetixio/synthetix/blob/develop/publish/src/commands/import-fee-periods.js#L82-L104

publish/src/commands/migrate-bridge.js
```diff
@@ -0,0 +1,253 @@
...
+	});
+	const newBridge = getContract({
+		contract: 'SynthetixBridgeToOptimism',
+		deploymentPath,
+		getTarget,
+		getSource,
+		address: newBridgeAddress,
+		web3,
+	});
+	if (newBridge.options.address === oldBridge.options.address) {
```
justinjmoses, line 91: why not just lowercase and compare the addresses? creating the contract instances isn't necessary for this...

publish/src/commands/migrate-bridge.js
justinjmoses: Hmmm. It's an interesting idea. The old contract would need a function to approve spending the amount for the new address (which would also mark the old as inactive) and then the new would need a function to `transferFrom` on behalf of the old. It's probably something we can/should add to go in with the escrow upgrade. 

I can get behind this two step process, though it's an additional transaction for the pDAO.

publish/src/commands/migrate-bridge.js
ajsantander: True, but I'm not creating the instances just for this. It would be the same to compare the minimized addresses, I just went with this way.

publish/src/commands/migrate-bridge.js
ajsantander: @jacko125 discussed and will implement in another PR in the future

publish/src/commands/migrate-bridge.js
```diff
@@ -0,0 +1,257 @@
...
+
+	// -----------------------------------
+	// Ok, let's migrate...
+	// -----------------------------------
+
+	const tx = await oldBridge.methods.migrateBridge(newBridge.options.address).send(params);
+	console.log(tx);
+};
+
+const bootstrapConnection = ({ network, deploymentPath, privateKey, useFork }) => {
```
justinjmoses, line 180: are these not external utils somewhere?

justinjmoses (approved):

publish/src/commands/migrate-bridge.js
ajsantander: Yes, we certainly are suffering of a sort of utility bloating. I think it started because of having ethers _and_ web3. There is a tendency to just re-write the needed functionality custom fitted to the situation at hand instead of making common tools be flexible to accommodate it. Flexibility also makes shared utils harder or weirder to setup, so its arguably better to just re-implement in place without having other concerns. But I agree, this is something that needs to be agreed upon, just maybe not in this PR.

PR #928, opened 2020-11-25
Improved connect-bridge script

ajsantander:
Improves the connect-bridge script by having it read on-chain values before writing, and only writing diffs.

### DIFF
--- a/publish/src/commands/connect-bridge.js
+++ b/publish/src/commands/connect-bridge.js
@@ -1,7 +1,7 @@
 const fs = require('fs');
 const path = require('path');
 const Web3 = require('web3');
-const { gray, red } = require('chalk');
+const { gray, red, yellow } = require('chalk');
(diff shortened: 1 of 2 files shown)


i-stam (approved):

justinjmoses (approved):

publish/src/commands/connect-bridge.js
```diff
@@ -56,62 +64,108 @@ const connectBridge = async ({
...
+	SynthetixBridge,
+	dryRun,
+}) => {
+	// ---------------------------------
+	// Compare with on-chain values
+	// ---------------------------------
+
+	const filteredNames = [];
+	const filteredAddresses = [];
+	for (let i = 0; i < names.length; i++) {
```
justinjmoses, line 116: [np] you can use `for (const [i, name] of Object.entries(names))` instead of a regular for loop if you want

justinjmoses (approved):

PR #936, opened 2020-11-30
sip99

hav-noms:
- Update the OIL feeds.json for mainnet to use the proxy above in develop
- Update the iOIL inverse rates in synths.json for mainnet to suit the current OIL price
- Update OIL in assets.json to not say Brent Crude Oil  but rather Perpetual Oil Futures


jacko125 (approved):
LGTM.

PR #937, opened 2020-11-30
Split RewardEscrowV2 contract in L1 and L2 counterpart

i-stam:
Former RewardEscrowV2 contract is split in two:
- RewardEscrowV2 residing on L1.
- OptimisticRewardEscrowV2 residing on L2.
- Both inherit from BaseRewardEscrowV2 the basic functionality. 

The L1/L2 specific functions revert in the base contract.

### DIFF
--- a/contracts/OptimisticRewardEscrowV2.sol
+++ b/contracts/OptimisticRewardEscrowV2.sol
@@ -6,7 +6,7 @@ import "./BaseRewardEscrowV2.sol";
 
 
 // https://docs.synthetix.io/contracts/RewardEscrow
-contract RewardEscrowV2Optimism is BaseRewardEscrowV2 {
+contract OptimisticRewardEscrowV2 is BaseRewardEscrowV2 {
     /* ========== ADDRESS RESOLVER CONFIGURATION ========== */
     bytes32 private constant CONTRACT_SYNTHETIX_BRIDGE_BASE = "SynthetixBridgeToBase";
 
--- a/contracts/RewardEscrow.sol
+++ b/contracts/RewardEscrow.sol
@@ -1,4 +1,5 @@
 pragma solidity ^0.5.16;
+pragma experimental ABIEncoderV2;
 
 // Inheritance
 import "./Owned.sol";
--- a/contracts/RewardEscrow.sol
+++ b/contracts/RewardEscrow.sol
@@ -1,5 +1,4 @@
 pragma solidity ^0.5.16;
-pragma experimental ABIEncoderV2;
 
 // Inheritance
 import "./Owned.sol";
--- a/contracts/interfaces/IRewardEscrowV2.sol
+++ b/contracts/interfaces/IRewardEscrowV2.sol
@@ -2,7 +2,7 @@ pragma solidity >=0.4.24;
 pragma experimental ABIEncoderV2;
 
 
-interface IRewardEscrowV2 {
+library VestingEntries {
     struct VestingEntry {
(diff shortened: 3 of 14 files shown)


contracts/OptimisticRewardEscrowV2.sol
```diff
@@ -0,0 +1,51 @@
...
+
+// https://docs.synthetix.io/contracts/RewardEscrow
+contract OptimisticRewardEscrowV2 is BaseRewardEscrowV2 {
+    /* ========== ADDRESS RESOLVER CONFIGURATION ========== */
+    bytes32 private constant CONTRACT_SYNTHETIX_BRIDGE_BASE = "SynthetixBridgeToBase";
+
+    /* ========== CONSTRUCTOR ========== */
+
+    constructor(address _owner, address _resolver) public BaseRewardEscrowV2(_owner, _resolver) {
+        appendToAddressCache(CONTRACT_SYNTHETIX_BRIDGE_BASE);
```
jacko125, line 16: This is good to split the address to cache only specific for L2.

contracts/interfaces/IRewardEscrowV2.sol
```diff
@@ -2,15 +2,18 @@ pragma solidity >=0.4.24;
 pragma experimental ABIEncoderV2;
 
 
-interface IRewardEscrowV2 {
+library VestingEntries {
```
jacko125, line 5: Good idea to move this here into the interface. 

We've also previously just kept it inside the `interface IRewardEscrowV2` as a struct

jacko125 (approved):
lgtm! great effort splitting these out, will be really helpful.

contracts/OptimisticRewardEscrowV2.sol
```diff
@@ -0,0 +1,51 @@
+pragma solidity ^0.5.16;
+pragma experimental ABIEncoderV2;
+
+// Inheritance
+import "./BaseRewardEscrowV2.sol";
+
+
+// https://docs.synthetix.io/contracts/RewardEscrow
+contract OptimisticRewardEscrowV2 is BaseRewardEscrowV2 {
```
justinjmoses, line 9: I think for naming conventions we call this `ImportableRewardEscrowV2`

contracts/interfaces/IRewardEscrowV2.sol
i-stam: I was actually getting an error. Unfortunately, interface inheritance is not supported in the current version. We had to declare it inside the interface to be able to use it there as a return a parameter but the compiler was complaining about the inheriting contract not declaring. Using a library was the easiest solution I came up with.

contracts/OptimisticRewardEscrowV2.sol
i-stam: Yeap, I wanted to ask about opinions on naming. I like that `ImportableRewardEscrowV2` follows the same pattern as BaseSynthetix-MintableSynthetix.

contracts/RewardEscrow.sol
```diff
@@ -1,4 +1,5 @@
 pragma solidity ^0.5.16;
+pragma experimental ABIEncoderV2;
```
jacko125, line 2: @i-stam  We don't need to update the old RewardEscrow contract with the `pragma experimental ABIEncoderV2`

i-stam: You are right, I was testing something and forgot tor remove it. Fixed.

PR #938, opened 2020-11-30
POC legacy mechanism alternative

ajsantander:
Considers the removal of legacy contracts, their compilation, and their inclusion in tests.

In short, a legacy contract is an old contract, that is not supposed to ever be upgraded. This is because of the proxy mechanism used in Synthetix v2, which cannot persist state between upgrades, and thus needs central state to be isoplated in separate contracts. Legacy contracts are compiled when running `npm run test:legacy`. This runs the unit tests, but hacks the mocha require statement to go fetch legacy contracts, instead of official ones, which ensures that new contract code will work with these non-upgradeable contracts.

The mechanism described above is rather complex, and might be overkill. Instead, we propose an alternative lightweight mechanism which "complains" if a developer attempts to change a contract marked as non-upgradeable.


publish/non-upgradeable.json
```diff
@@ -0,0 +1,16 @@
+[
+	"Depot",
+	"EternalStorage",
+	"ExchangeState",
+	"FeePoolEternalStorage",
+	"FeePoolState",
+	"Proxy",
```
justinjmoses, line 7: So this should be the name of the actual contract label - not just the source - so `ProxySynthetix` for instance. Though it would be cumbersome to list all possible proxies. Perhaps instead these entries can be a regex?

ajsantander: Correct, and this is the current functionality. For example, `TokenState` in this list will cause `TokenStateSynthetix`, `TokenStatesUSD`, `TokenStateJPY` etc to be considered non-upgradeable.

justinjmoses: Ah right sorry I missed the `includes`. Perhaps it's better to have a pattern match of `/^.../` to ensure it starts with not just includes?

ajsantander: Hmmm... would `ReadProxyAddressResolver` be upgradeable?

justinjmoses: No it wouldn't

justinjmoses: But a new `ReadProxy` would be ok if we needed one. So perhaps the whole `ReadProxyAddressResolver` is best?

PR #940, opened 2020-12-01
merge 2.34.0 release into develop

jacko125:
- Potentially need to revise the `versions.json` and `release.json` to match `v2.34.0` instead of `v2.33.0` originally released with.


justinjmoses:
OK, I'm going to fix these files in `master` then merge into this branch

PR #944, opened 2020-12-02
Check if resolver needs imports and if bridge needs sync separately

ajsantander:
The connect-bridge was not syncing cache on the bridge if its resolver didnt need to import any addresses. However, if for example the bridge has a new dependency, its cache might be out of sync. So the script now separately checks if the bridge's cache is out of sync.


i-stam (approved):
lgtm, perhaps we could somehow test it to be sure that it works as expected

publish/src/commands/connect-bridge.js
```diff
@@ -127,41 +127,68 @@ const connectLayer = async ({
...
+	}
+
+	// ---------------------------------
+	// Sync cache on bridge if needed
+	// ---------------------------------
+
+	let needToSyncCacheOnBridge = needToImportAddresses;
+	if (!needToSyncCacheOnBridge) {
+		const isResolverCached = await SynthetixBridge.methods
+			.isResolverCached(AddressResolver.options.address)
```
justinjmoses, line 171: This has changed now due to SIP-100

publish/src/commands/connect-bridge.js
```diff
@@ -127,41 +127,68 @@ const connectLayer = async ({
...
+		}
+	}
+
+	if (needToSyncCacheOnBridge) {
+		console.log(yellow('  > Syncing cache on bridge...'));
+
+		if (!dryRun) {
+			console.log(yellow('  > SynthetixBridge.setResolverAndSyncCache()...'));
+			tx = await SynthetixBridge.methods
+				.setResolverAndSyncCache(AddressResolver.options.address)
```
justinjmoses, line 184: (ibid - from SIP-100)

justinjmoses (changes requested):

publish/src/commands/connect-bridge.js
ajsantander: Updated

publish/src/commands/connect-bridge.js
ajsantander: Updated

justinjmoses (approved):

PR #949, opened 2020-12-03
Prevent withdrawals on a suspended L2 system

ajsantander:


justinjmoses (approved):
Looks good, though can we also add `systemActive` for the minting functions as well?

ajsantander:
> Looks good, though can we also add `systemActive` for the minting functions as well?



> Looks good, though can we also add `systemActive` for the minting functions as well?

@justinjmoses I considered this, and intentionally did not add the modifier to minting functions. My reasoning was that we want to keep users from initiating actions when a system is suspended, but if that action was previously initiated in another chain, it should be allowed to finalize. What do you think?

justinjmoses:
> My reasoning was that we want to keep users from initiating actions when a system is suspended, but if that action was previously initiated in another chain, it should be allowed to finalize. What do you think?

Yeah it's tricky. I think because the message to L2 will keep trying to be replayed until such time that the system is resumed on L2 that this works. (You were the one who mentioned that initially IIRC). @K-Ho thoughts?

PR #951, opened 2020-12-04
SIP 100 Resolver & cache improvements

justinjmoses:
Replaces #904 

**Goals**
1. To reduce the load on deploys - processing many `setResolverAndSyncCache` calls is simply not scalable. 
1. To minimize the use of superuser owner functionality
1. To reduce deployment sizes of contracts by reducing constructor logic (in particular for the OVM deployment of contracts)

**Solidity Improvements**
* `AddressResolver.importAddresses()` emits an event `AddressImported` for each address added
* `AddressResolver.rebuildCaches` is exposed (and non-protected), for ease-of use
* `MixinResolver` no longer requires `Owned`. `resolver` is set once only on construction (with the expectation that `ReadProxyAddressResolver` will be used. 
* `MixinResolver` has a predicate view `isResolverCached()`  
* `MixinResolver` has a `rebuildCache()` function (publicly available) for the contract to rebuild its cache. This function also emits a `CacheUpdated` event for each item added to it's private cache. This replaces the onlyOwner function:  `setResolverAndSyncCache(address _resolver)`
* `MixinResolver.requireAndGetAddress()` uses `abi.encodePacked` for string concatenation and a simpler API.
* `MixinResolver` no longer performs tasks in the constructor. Contracts using the mixin must implement the abstract `resolverAddressesRequired` function.
* `MixinSystemSettings` now wraps `MixinResolver` and returns the default `resolverAddressesRequired()` containing `FlexibleStorage`

**Deployment Improvements**
* `AddressResolver.importAddresses` is invoked with every contract within the Synthetix protocol

**Stats**
* `AddressResolver.importAddresses(all 140 local targets)` takes approximately **3.5m gas**
* `AddressResolver.rebuildCaches(47 targets - most of the synths)` takes approximately **6.5m gas**

**Known Limitations**
1. As before, the deploy script will skip any contract with a dependency that has a `:` inside it. This is for the OVM bridge contracts, and `connect-bridge` must be run


ajsantander (approved):

contracts/MixinResolver.sol
```diff
@@ -5,64 +5,63 @@ import "./Owned.sol";
...
-    // (used for deployment)
-    function isResolverCached(AddressResolver _resolver) external view returns (bool) {
-        if (resolver != _resolver) {
-            return false;
-        }
-
-        // otherwise, check everything
-        for (uint i = 0; i < resolverAddressesRequired.length; i++) {
-            bytes32 name = resolverAddressesRequired[i];
+    function isResolverCached() external view returns (bool) {
```
jacko125, line 61: @justinjmoses  How will the system / deploy script be able to use `isResolverCached` after a release / upgrade to make sure everything is correctly cached and updated?

jacko125 (approved):
changes LGTM.

contracts/MixinResolver.sol
justinjmoses: yep.

PR #954, opened 2020-12-08
Adds a command that deploys an OVM pair of instances on local chains

ajsantander:
Assuming that docker is used to start a local L1 chain and a local L2 chain using https://github.com/ethereum-optimism/optimism-integration, this PR adds a `deploy-ovm-pair` command that deploys and connects a pair of instances to be used for local production tests.

The PR also introduces OVM safety checking in the deployment process, to ensure that no dangerous `5b` or `JUMPDEST` opcodes exist in appended deployment constructor parameters.

Another interesting addition of the PR, is the ability for the deployment script to split addresses being imported in the AddressResolver into chunks. This was needed for OVM to avoid that call to exceed the block limit.

To try the script out:
1. `git clone git@github.com:ethereum-optimism/optimism-integration.git`
2. `cd optimism-integration`
3. `docker-compose pull`
4. `./up.sh`
5. Wait 30 seconds or so for the chains to kick off
6. `cd ~/synthetix` (or wherever)
7. `node publish deploy-ovm-pair`

### DIFF
--- a/publish/src/Deployer.js
+++ b/publish/src/Deployer.js
@@ -52,7 +52,7 @@ class Deployer {
 
 		if (useFork) {
 			this.web3.eth.defaultAccount = getUsers({ network, user: 'owner' }).address; // protocolDAO
-		} else if (network === 'local') {
+		} else if (!privateKey && network === 'local') {
 			// Deterministic account #0 when using `npx buidler node`
 			this.web3.eth.defaultAccount = '0xc783df8a850f42e7F7e57013759C285caa701eB6';
 		} else {
@@ -147,6 +147,32 @@ class Deployer {
 
 		let deployedContract;
 
+		const areBytesSafeForOVM = bytes => {
+			const idx = bytes.toLowerCase().indexOf('5b');
+
+			const containsJumpdestOpcode = idx >= 0;
+			const isOpcodeAtAnEvenLoc = idx % 2 === 0;
+
+			return !(containsJumpdestOpcode && isOpcodeAtAnEvenLoc);
+		};
+
+		const getEncodedDeploymentParameters = ({ abi, params }) => {
+			let encodedParams = '0x';
+
+			const constructorABI = abi.find(item => item.type === 'constructor');
+			if (!constructorABI) {
+				return encodedParams;
+			}
+
+			const inputs = constructorABI.inputs;
+			if (!inputs || inputs.length === 0) {
+				return encodedParams;
+			}
+
+			const types = inputs.map(input => input.type);
+			return this.web3.eth.abi.encodeParameters(types, params);
+		};
+
 		if (deploy) {
 			console.log(gray(` - Attempting to deploy ${name}`));
 			let gasUsed;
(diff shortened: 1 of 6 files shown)


publish/src/Deployer.js
```diff
@@ -165,11 +199,26 @@ class Deployer {
...
+				});
+
+				// Check if the deployment parameters are safe in OVM
+				// (No need to check the metadata hash since its stripped with the OVM compiler)
+				if (this.useOvm) {
+					const encodedParameters = getEncodedDeploymentParameters({
+						abi: compiled.abi,
+						params: args,
+					});
+					if (!areBytesSafeForOVM(encodedParameters)) {
```
justinjmoses, line 214: This could be thrown before the contract is even created no?

publish/src/Deployer.js
```diff
@@ -178,6 +227,29 @@ class Deployer {
...
+				}
+			}
+
+			// If the resulting contract address is unsafe for OVM,
+			// it is possible that the address may be used as a constructor parameter of
+			// another contract. If it is unsafe, warn but don't fail.
+			if (this.useOvm && !areBytesSafeForOVM(deployedContract.options.address)) {
+				console.log(
+					yellow(
+						`⚠ WARNING: Deployed contract address ${deployedContract.options.address} contains unsafe opcodes for OVM. If it is used as a constructor argument in the deployment of another contract, it will make that deployment fail.`
```
justinjmoses, line 248: Should we add a retry function here? Or perhaps just throw and force a re-run of the deployment script?

publish/src/commands/connect-bridge.js
```diff
@@ -189,8 +191,9 @@ const setupInstance = async ({
 
 	let account;
 	if (privateKey) {
+		const idx = web3.eth.accounts.wallet.length;
 		web3.eth.accounts.wallet.add(privateKey);
-		web3.eth.defaultAccount = web3.eth.accounts.wallet[0].address;
+		web3.eth.defaultAccount = web3.eth.accounts.wallet[idx].address;
```
justinjmoses, line 196: Is this intentional? The defaultAccount index is the length of the number of wallet?

publish/src/Deployer.js
ajsantander: Ah yeah, I split the deployment call and put this in between because I initially intended to retrieve the encoding from the bytecode itself.

Will change.

publish/src/Deployer.js
ajsantander: We thought about this... the thing is, we might retry for a contract whose address is not used as a deployment parameter.

What could work is try to predict the address of the contract-to-be using the EOA and its nonce. If the test fails, have the EOA make a "blank" transaction to increase its nonce and try again.

However, this might prove to be overkill since its not only addresses in constructor parameters that can trigger the second order effect failure, for example the number `91`, which is `0x5b`.

ajsantander (commented):
Rename to `deploy-ovm-test-pair`

publish/src/Deployer.js
ajsantander: Fixed!

publish/src/Deployer.js
ajsantander: Implemented!

The deployer will now use dummy txs to increment nonces and produce safe contract addresses when deploying to OVM.

justinjmoses (approved):

PR #955, opened 2020-12-09
Safety check for ensuring gas deployment limit

i-stam:
- This adds a new parameter, namely `ensureOvmDeploymentGasLimit`,  to the `deploy` and `deployer` scripts to ensure that each contract deployment does not exceed the 9mil gas cap when deploying to the OVM.


publish/src/commands/deploy.js
```diff
@@ -1822,6 +1824,10 @@ module.exports = {
 				'-d, --deployment-path <value>',
 				`Path to a folder that has your input configuration file ${CONFIG_FILENAME}, the synth list ${SYNTHS_FILENAME} and where your ${DEPLOYMENT_FILENAME} files will go`
 			)
+			.option(
+				'-e, --ensure-ovm-deployment-gas-limit <value>',
```
ajsantander, line 1828: I'd rename this to something sligthly more convenient, like `--max-ovm-deployment-gas` and give it a default value of `9000000`, which is something you can do with commander. See the other params.

publish/src/Deployer.js
```diff
@@ -172,7 +174,11 @@ class Deployer {
 					})
 					.send(await this.sendParameters('contract-deployment'))
 					.on('receipt', receipt => (gasUsed = receipt.gasUsed));
-
+				if (this.useOvm && this.ensureOvmDeploymentGasLimit) {
+					if (gasUsed > 9000000) {
```
ajsantander, line 178: The variable incoming from deploy.js is not a boolean. I believe its a number, so you should use `gas > this.ensureOvmDeploymentGasLImit`.

ajsantander (changes requested):
This is the right way to do it @i-stam , just some minor necessary tweaks.

Also, this makes me think... should we make this restriction applicable to any tx, that is not just deployment transactions? I can imagine how an `AddressResolver.rebuildCaches()` call could easily exceed the limit.

publish/src/commands/deploy.js
justinjmoses: should we not make the param just a flag and hardcode in the `9m` amount?

ajsantander:
Also @i-stam we already have a `contractDeploymentGasLimit` parameter on the deploy script. Couldn't this somehow be used for what we need, without having to add a new parameter?

ajsantander:
We should be able to control this using the deploy command's `contractDeploymentGasLimit` parameter. There's no need to add a new parameter.

PR #975, opened 2020-12-14
Fix resolverAddressesRequired() bug and Update deploy script

i-stam:


contracts/Synthetix.sol
```diff
@@ -26,7 +26,7 @@ contract Synthetix is BaseSynthetix {
 
     function resolverAddressesRequired() public view returns (bytes32[] memory addresses) {
         bytes32[] memory existingAddresses = BaseSynthetix.resolverAddressesRequired();
-        bytes32[] memory newAddresses = new bytes32[](1);
+        bytes32[] memory newAddresses = new bytes32[](2);
```
jacko125, line 29: @i-stam  Good catch !

jacko125 (approved):

PR #978, opened 2020-12-14
Ensures OVM deployment txs are < 9m gas

ajsantander:
Several delicate changes in this PR:

* ExchangeRates stubbed for L2
* Issuer stubbed for L2
* No optimizations for L2 (1 run only)
* The deploy script throws if you attempt to set gas limits > 9m when deploying for Optimism

To verify that this deploys:
 1. git clone git@github.com:ethereum-optimism/optimism-integration.git
 2. cd optimism-integration
 3. docker-compose pull
 4. ./up.sh
 5. cd ../synthetix
 6. node publish deploy-ovm-pair
 7. npm run test:prod:ovm

### DIFF
--- a/publish/src/commands/deploy-ovm-pair.js
+++ b/publish/src/commands/deploy-ovm-pair.js
@@ -46,7 +46,7 @@ const deployOvmPair = async () => {
 };
 
 const deployInstance = async ({ useOvm, privateKey }) => {
-	await commands.build({ useOvm });
+	await commands.build({ useOvm, optimizerRuns: useOvm ? 1 : 200 });
 
 	await commands.deploy({
 		network: 'local',
--- a/publish/src/commands/deploy-ovm-pair.js
+++ b/publish/src/commands/deploy-ovm-pair.js
@@ -56,7 +56,7 @@ const deployInstance = async ({ useOvm, privateKey }) => {
 		gasPrice: '0',
 		useOvm,
 		methodCallGasLimit: '3500000',
-		contractDeploymentGasLimit: useOvm ? '11000000' : '9500000',
+		contractDeploymentGasLimit: useOvm ? '8900000' : '9500000',
 		privateKey,
 		ignoreCustomParameters: true,
 	});
--- a/publish/src/commands/deploy-ovm-pair.js
+++ b/publish/src/commands/deploy-ovm-pair.js
@@ -6,6 +6,10 @@ const commands = {
 	connectBridge: require('./connect-bridge').connectBridge,
 };
 
+const {
+	constants: { OVM_MAX_GAS_LIMIT },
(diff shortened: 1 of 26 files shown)


justinjmoses (approved):

i-stam (approved):

publish/src/commands/deploy.js
```diff
@@ -106,6 +106,20 @@ const deploy = async ({
 			throw new Error(`Deploying on Goerli needs to be performed with --manage-nonces.`);
 		}
 
+		// Every transaction in Optimism needs to be below 9m gas, to ensure
+		// there are no deployment out of gas errors during fraud proofs.
+		if (useOvm) {
+			const maxOptimismGasLimit = 8900000;
```
justinjmoses, line 112: can we not make this a constant in `index.js`?

publish/src/commands/deploy.js
```diff
@@ -509,13 +522,8 @@ const deploy = async ({
 
 	const exchangeRates = await deployer.deployContract({
 		name: 'ExchangeRates',
-		args: [
```
justinjmoses, line 512: We should call out this change. FYI @jacko125 @hav-noms - no longer deploying ExchangeRates with a `0.2` rate of `SNX`.

justinjmoses, line 512: We should be clear about this change - it means there will be no `SNX` rate initially, preventing any issuance. I know the purpose is to reduce gas on contract instantiation, but we still should be clear this change is happening. cc @jacko125 @hav-noms

publish/src/commands/deploy.js
ajsantander: Done

publish/src/commands/deploy.js
ajsantander: Correct! My initial intention was to remove it to reduce gas on contract deployment, yes, but afterwards we discussed if it was better to set the price of SNX to 20c on a fresh deploy, or to read it from the ExchangeRates being replaced by the new contract.

I can return the old functionality, but it might be a good time to review it.

publish/src/commands/deploy-ovm-pair.js
```diff
@@ -56,7 +56,7 @@ const deployInstance = async ({ useOvm, privateKey }) => {
 		gasPrice: '0',
 		useOvm,
 		methodCallGasLimit: '3500000',
-		contractDeploymentGasLimit: useOvm ? '11000000' : '9500000',
+		contractDeploymentGasLimit: useOvm ? '8900000' : '9500000',
```
justinjmoses, line 59: pls use constant

ajsantander: Done!

PR #981, opened 2020-12-15
Always return named parameter for resolverAddressesRequired()

ajsantander:
As far as I know, naming a return variable and using the return statement does not have unwanted effects. I've also verified this in Remix.

However, I believe that the code is semantically incorrect like this, and deviations between what's written and what's intended should always be avoided.

### DIFF
--- a/contracts/SynthetixBridgeToBase.sol
+++ b/contracts/SynthetixBridgeToBase.sol
@@ -60,12 +60,12 @@ contract SynthetixBridgeToBase is Owned, MixinSystemSettings, ISynthetixBridgeTo
 
     function resolverAddressesRequired() public view returns (bytes32[] memory addresses) {
         bytes32[] memory existingAddresses = MixinSystemSettings.resolverAddressesRequired();
-        addresses = new bytes32[](4);
-        addresses[0] = CONTRACT_EXT_MESSENGER;
-        addresses[1] = CONTRACT_SYNTHETIX;
-        addresses[2] = CONTRACT_BASE_SYNTHETIXBRIDGETOOPTIMISM;
-        addresses[3] = CONTRACT_ISSUER;
-        addresses = combineArrays(existingAddresses, addresses);
+        bytes32[] memory newAddresses = new bytes32[](4);
+        newAddresses[0] = CONTRACT_EXT_MESSENGER;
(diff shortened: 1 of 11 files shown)


contracts/SynthetixBridgeToBase.sol
```diff
@@ -65,7 +65,7 @@ contract SynthetixBridgeToBase is Owned, MixinSystemSettings, ISynthetixBridgeTo
         addresses[1] = CONTRACT_SYNTHETIX;
         addresses[2] = CONTRACT_BASE_SYNTHETIXBRIDGETOOPTIMISM;
         addresses[3] = CONTRACT_ISSUER;
-        return combineArrays(existingAddresses, addresses);
+        addresses = combineArrays(existingAddresses, addresses);
```
justinjmoses, line 68: it's not named above :(

contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -82,7 +82,7 @@ contract SynthetixBridgeToOptimism is Owned, MixinSystemSettings, ISynthetixBrid
         addresses[2] = CONTRACT_ISSUER;
         addresses[3] = CONTRACT_REWARDSDISTRIBUTION;
         addresses[4] = CONTRACT_OVM_SYNTHETIXBRIDGETOBASE;
-        return combineArrays(existingAddresses, addresses);
+        addresses = combineArrays(existingAddresses, addresses);
```
justinjmoses, line 85: as above.

justinjmoses (changes requested):
The signature isn't accurate in a couple of contracts - I suspect it compiles because it's using the signature from `MixinResolver` but still it's good to be accurate. Can you please double check and fix those?

jacko125 (approved):

jacko125 (approved):

contracts/SynthetixBridgeToBase.sol
ajsantander: Oops! Fixed.

contracts/SynthetixBridgeToOptimism.sol
ajsantander: Also fixed.

justinjmoses (approved):

PR #985, opened 2020-12-16
Add historical block numbers to versions and use in settler for historical debt

justinjmoses:


jacko125 (approved):
approved

publish/src/commands/settle.js
```diff
@@ -104,20 +104,29 @@ const settle = async ({
...
-		);
+	const versions = getVersions({ byContract: true });
+
+	const getContract = ({ label, source = label, blockNumber }) => {
+		let { address } = getTarget({ contract: label });
+
+		if (blockNumber) {
+			// look for the right contract based off the block
+			for (const entry of versions[source]) {
+				// assumes sorted by block number ascending
```
ajsantander, line 115: nit: Why make the assumption? Should be easy to also keep a minBlock variable around and only change the address of entry.block < minBlock

ajsantander, line 115: nit: Why make this assumption? If getVersions ever changes and there is not test for this, this will fail silently.

E.g. In the loop, record a distance to the required block, and pick the entry with the minimum (signed) distance.

publish/src/commands/settle.js
```diff
@@ -104,20 +104,29 @@ const settle = async ({
...
+				if (entry.block < blockNumber) {
+					address = entry.address;
+				}
+			}
+		}
+		// console.log(`For ${label} using ${address}`);
+		return new web3.eth.Contract(getSource({ contract: source }).abi, address);
```
ajsantander, line 122: nit: I guess it's alright to assume that the interface hasn't changed in this case. For a future deployment mechanism, it would be good to store historical abis as well.

ajsantander (approved):
Looks good!

Just a nit.

publish/src/commands/settle.js
justinjmoses: yeah, it would for sure.

publish/src/commands/settle.js
justinjmoses: it's a fair nit. Updated.

ajsantander (approved):

PR #989, opened 2020-12-17
Multiple xDomain gas limits

i-stam:


contracts/MixinSystemSettings.sol
```diff
@@ -24,10 +24,14 @@ contract MixinSystemSettings is MixinResolver {
     bytes32 internal constant SETTING_AGGREGATOR_WARNING_FLAGS = "aggregatorWarningFlags";
     bytes32 internal constant SETTING_TRADING_REWARDS_ENABLED = "tradingRewardsEnabled";
     bytes32 internal constant SETTING_DEBT_SNAPSHOT_STALE_TIME = "debtSnapshotStaleTime";
-    bytes32 internal constant SETTING_CROSS_DOMAIN_MESSAGE_GAS_LIMIT = "crossDomainMessageGasLimit";
+    bytes32 internal constant SETTING_CROSS_DOMAIN_DEPOSIT_GAS_LIMIT = "crossDomainDepositGasLimit";
+    bytes32 internal constant SETTING_CROSS_DOMAIN_ESCROW_GAS_LIMIT = "crossDomainEscrowGasLimit";
+    bytes32 internal constant SETTING_CROSS_DOMAIN_REWARD_GAS_LIMIT = "crossDomainRewardGasLimit";
```
ajsantander, line 29: Shouldn't there be one for SETTING_CROSS_DOMAIN_WITHDRAW_GAS_LIMIT too?

ajsantander: Oh, I see that you re-use the deposit limit for withdrawals. Perhaps rename the variable alternatively? IMO, they should be separate things.

contracts/MixinSystemSettings.sol
```diff
@@ -39,8 +43,19 @@ contract MixinSystemSettings is MixinResolver {
...
     }
 
-    function getCrossDomainMessageGasLimit() internal view returns (uint) {
-        return flexibleStorage().getUIntValue(SETTING_CONTRACT_NAME, SETTING_CROSS_DOMAIN_MESSAGE_GAS_LIMIT);
+    function _getGasLimit(CrossDomainMessageGasLimits gasLimitType) internal pure returns (bytes32) {
+        if (gasLimitType == CrossDomainMessageGasLimits.Deposit || gasLimitType == CrossDomainMessageGasLimits.Withdrawal) {
+            return SETTING_CROSS_DOMAIN_DEPOSIT_GAS_LIMIT;
+        } else if (gasLimitType == CrossDomainMessageGasLimits.Reward) {
+            return SETTING_CROSS_DOMAIN_REWARD_GAS_LIMIT;
+        } else {
```
ajsantander, line 51: nit: I would use `else if (gasLimitType == CrossDomainMessageGasLimits.Escrow)` here and in an additional else, revert.

contracts/MixinSystemSettings.sol
i-stam: Yes, I did it for efficiency but happy to add it is going to improve clarity.

contracts/MixinSystemSettings.sol
```diff
@@ -39,8 +43,19 @@ contract MixinSystemSettings is MixinResolver {
         return IFlexibleStorage(requireAndGetAddress(CONTRACT_FLEXIBLESTORAGE));
     }
 
-    function getCrossDomainMessageGasLimit() internal view returns (uint) {
-        return flexibleStorage().getUIntValue(SETTING_CONTRACT_NAME, SETTING_CROSS_DOMAIN_MESSAGE_GAS_LIMIT);
+    function _getGasLimit(CrossDomainMessageGasLimits gasLimitType) internal pure returns (bytes32) {
```
ajsantander, line 46: Perhaps rename to `_getGasLimitVar`?

index.js
```diff
@@ -103,6 +103,9 @@ const defaults = {
 		rinkeby: '0xEDC0C23864B041607D624E2d9a67916B6cf40F7a',
 	},
 	INITIAL_ISSUANCE: w3utils.toWei(`${100e6}`),
+	CROSS_DOMAIN_DEPOSIT_GAS_LIMIT: `${3e6}`,
```
ajsantander, line 106: Do we, at any point, enforce a way for developers in the future not to set invalid values here? Might be overkill, but perhaps no. Now, we know that a bad value could produce always failing txs on the other side, but future devs may not. Not thinking only about default values, but perhaps a warning in the deploy script, or in the contracts?

test/contracts/SynthetixBridgeToBase.spec.js
```diff
@@ -30,7 +30,9 @@ contract('SynthetixBridgeToBase (spec tests)', accounts => {
 		});
 
 		it('shows the expected cross domain message gas limit', async () => {
-			assert.bnEqual(await systemSettings.crossDomainMessageGasLimit(), 3e6);
+			assert.bnEqual(await systemSettings.crossDomainMessageGasLimit(0), 3e6);
```
ajsantander, line 33: These numbers should be imported from index.js, otherwise changing the setting will cause tests to fail

index.js
i-stam: this is a good idea!

justinjmoses (approved):

PR #992, opened 2020-12-19
Sip-60  Update

i-stam:
Just a few changes and bugs I found while working on the kovan deployment branch, on order to keep the sip-60 updated.


jacko125 (approved):
thanks !

PR #994, opened 2020-12-21
Use abi.selector() for encoding L2 messages + event renaming

i-stam:
These are the changes required to use abi`.encodeWithSelector`, encodeWithSignature had the advantage that you could see the arguments directly in the encoding e.g. `abi.encodeWithSignature("completeDeposit(address,uint256)", msg.sender, amount);` , whereas now you have to check the interface.

I renamed a couple of events too, just to be more consistent with the rest of them.


justinjmoses (approved):
I'm OK with this in principle, @ajsantander ?

ajsantander (approved):
Looks good!

PR #996, opened 2020-12-22
SIP-97 and SIP 103 - Multi Collateral Loans and sUSD Shorts.

mjs-12:
This PR adds the functionality described in SIP-97 and SIP-103.

The vast majority of the logic is defined in Collateral.sol. This contract manages all the loans associated with its collateral asset. The loans are stored in a corresponding CollateralState.sol contract, to allow us to upgrade functionality without losing the loans.

There are three child contracts that inherit from Collateral.sol, they are:

CollateralEth.sol --> For ETH collateral, handles payable aspects and withdraw patterns.
CollateralErc20.sol --> For ERC20 collaterals, checks allowances and performs transferFrom.
CollateralShort.sol --> For sUSD collateral, allows users to create a loan but receive back sUSD, otherwise known as "short selling".
There is a contract CollateralManager.sol which manages the various collateral contracts and reports information to the Issuer and DebtCache. This contract is also responsible for working out the interest rates that accrue on loans. These rates are stored in CollateralManagerState.sol.


publish/src/commands/deploy.js
```diff
@@ -1224,6 +1223,141 @@ const deploy = async ({
...
+			source: 'CollateralShort',
+			args: [
+				addressOf(collateralShortState),
+				account,
+				addressOf(collateralManager),
+				addressOf(readProxyForResolver),
+				toBytes32('sUSD'),
+				w3utils.toWei('1.5'),
+				w3utils.toWei('500'),
+				addressOf(deployer.getExistingContract({ contract: 'ProxyERC20sUSD' })),
```
jacko125, line 1344: @justinjmoses  We had issues with the prod tests failing on `network = local` when using `ProxysUSD` here for setting up the collateralShort contract and also `ProxyERC20sUSD` was added to kovan / mainnet `versions.json` to allow `deployer.getExistingContract` to pass. 

It seems that on the mainnet and kovan versions.json the `ProxyERC20sUSD` contract is recorded as the `ProxysUSD` because of the legacy migration of the old proxy contracts.

jacko125 (approved):
LGTM the latest changes.

publish/src/commands/deploy.js
justinjmoses: this was a great catch - thanks for adding @jacko125

PR #999, opened 2020-12-22
Increases multicollateral production test coverage

ajsantander:


mjs-12:
Looks good now, we just needed a bit more renBTC for our user to deposit after opening, and the shorted needed some more sETH to cover the small amount of interest accrued on his short.

jacko125 (approved):
LGTM

PR #1003, opened 2020-12-22
Improving deploy of collateral with predicates

justinjmoses:
Passing in the
1. Switched `MixinResolver.rebuildCache` to a `public` function so it can be called within the same contract and not be prohibitive with gas
2. Added synth keys as arguments to `Collateral.addSynths` to reduce contract gas usage so no need to fetch synth keys
3. `Collateral.addSynths` calls to `rebuildCache` once done.
4. Added predicate function `Collateral.areSynthsAndCurrenciesSet`  to prevent extra no-op txns during future deployments

TODO
- [x] `CollateralManager` needs be given the same treatment


contracts/Collateral.sol
```diff
@@ -184,6 +184,28 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
...
+        view
+        returns (bool)
+    {
+        if (synths.length != _synthNamesInResolver.length) {
+            return false;
+        }
+
+        for (uint i = 0; i < _synthNamesInResolver.length; i++) {
+            bytes32 synthName = _synthNamesInResolver[i];
+            if (synths[i] != synthName) {
```
jacko125, line 198: @justinjmoses  There is an assumption here that synths[] / synths[i] is appended in the same order as always / or in the same order as the input `_synthNamesInResolver` here

justinjmoses: yes, I was cognizant there's not enough contract space to be more specific

mjs-12 (approved):

test/contracts/TradingRewards.spec.js
```diff
@@ -159,7 +159,7 @@ contract('TradingRewards', accounts => {
 			});
 		});
 
-		describe('when SystemSettings tradingRewardsEnabled is set to true', () => {
+		describe.only('when SystemSettings tradingRewardsEnabled is set to true', () => {
```
jacko125, line 162: .only needs to be removed

PR #1004, opened 2020-12-23
update CollateralManager to addSynths + currencyKeys

jacko125:
-  debtCache to check if synthsByKey exists on collateralManager


mjs-12 (approved):
Looks good.

PR #1005, opened 2020-12-23
Deploys a mock renBTC for local multicollateral production tests

ajsantander:


contracts/test-helpers/SwapWithVirtualSynth.sol
```diff
@@ -11,7 +11,39 @@ import "../interfaces/ISynthetix.sol";
 import "../interfaces/IAddressResolver.sol";
 import "../interfaces/IVirtualSynth.sol";
 import "../interfaces/IExchanger.sol";
-import {IERC20 as IERC20Detailed} from "../interfaces/IERC20.sol";
+
+
+interface IERC20Detailed {
```
justinjmoses, line 16: yeah fair enough

justinjmoses (approved):
It's certainly making the deploy script gnarly, but it's probably the best alternative right now

PR #1009, opened 2020-12-23
Avoids multiple minor errors in prod tests

ajsantander:


ajsantander:
This PR might actually be useless, since SystemSettings.rateStalePeriod is alread ~90000 (25hs), so if rates are showing as stale in prod tests, its probably to do with something else.

ajsantander:
Thinking about it a bit more, this may be useful since prod tests may fast forward

test/prod/utils/ensureAccountHasBalance.js
```diff
@@ -15,6 +15,11 @@ function getUser({ network, deploymentPath, user }) {
 }
 
 async function ensureAccountHasEther({ network, deploymentPath, amount, account }) {
+	const currentBalance = web3.utils.toBN(await web3.eth.getBalance(account));
```
ajsantander, line 18: Changes to this file are actually not to avoid errors, I'm just avoiding transfers if the user already has the target balance

justinjmoses (approved):

PR #1010, opened 2020-12-23
Use chunking in the deployer's rebuildCaches()

ajsantander:


publish/src/commands/deploy.js
```diff
@@ -1453,6 +1453,25 @@ const deploy = async ({
...
+	for (let i = 0; i < contractsToRebuildCache.length; i += addressChunkSize) {
+		const chunk = contractsToRebuildCache.slice(i, i + addressChunkSize);
+		await runStep({
+			gasLimit: 7e6, // higher gas required
+			contract: `AddressResolver`,
+			target: addressResolver,
+			write: 'rebuildCaches',
+			writeArg: [chunk],
+		});
+	}
```
justinjmoses, line 1473: we also need to add a loop down below as well like `useOvm`

ajsantander: Solved

ajsantander:
Not needed for now

PR #1014, opened 2020-12-24
remove system status from collateral manager

mjs-12:
Removes system status from collateral manager.


jacko125 (approved):
lgtm removing systemStatus checks

PR #1022, opened 2021-01-04
Shorting rewards develop

mjs-12:


contracts/ShortingRewards.sol
```diff
@@ -50,11 +49,9 @@ contract ShortingRewards is IShortingRewards, RewardsDistributionRecipient, Reen
         address _resolver,
         address _rewardsDistribution,
         address _rewardsToken,
-        address _short,
```
jacko125, line 53: @mjspain12  What happens when there are more than one `CollateralShort` contract in the system? Is there going to be only the `sUSD` collateralShort which enables shorting in any allowed currency, which means resolver will handle only one `CollateralShort` ?

contracts/ShortingRewards.sol
```diff
@@ -102,22 +99,24 @@ contract ShortingRewards is IShortingRewards, RewardsDistributionRecipient, Reen
...
+    function withdraw(address account, uint256 amount) external onlyShortContract nonReentrant updateReward(account) {
         require(amount > 0, "Cannot withdraw 0");
         _totalSupply = _totalSupply.sub(amount);
         _balances[account] = _balances[account].sub(amount);
         emit Withdrawn(account, amount);
     }
 
-    function getReward(address account) public nonReentrant updateReward(account) {
+    // this can be called by anyone on the short contract to claim the rewards.
+    function getReward(address account) external onlyShortContract nonReentrant updateReward(account) {
```
jacko125, line 119: Doesn't seem necessary to protect `getReward` to only the shortContract. It could be a public function that allows someone to collect their reward either directly calling the ShortingRewards.sol or going through the `CollateralShort` contract.

contracts/ShortingRewards.sol
mjs-12: @jacko125 Yeah the assumption I've made here is that the only short collateral allowed in the system will be `sUSD` and hence only one contract. I suppose it is possible that in the future we could have another short contract that accepted a different collateral type. However, I don't see this happening for a while which is why I've opted to use the resolver.

contracts/ShortingRewards.sol
mjs-12: Yeah I was just trying to keep a consistent interaction pattern through the contract.

jacko125 (approved):
lgtm!

contracts/ShortingRewards.sol
```diff
@@ -48,12 +47,10 @@ contract ShortingRewards is IShortingRewards, RewardsDistributionRecipient, Reen
         address _owner,
         address _resolver,
         address _rewardsDistribution,
-        address _rewardsToken,
-        bytes32 _synth
+        address _rewardsToken
```
jacko125, line 50: Good cleanup, synth wasnt used at all.

PR #1032, opened 2021-01-10
Multi collateral liquidation fix develop

### LINKED ISSUE #1025: Multi-collateral liquidation bug
**Overview**

At a high level, the bug arises from comparing an sUSD value with a value may be denominated in another synth. The resulting comparison flags some loans for total liquidation when only partial liquidation is required.

**Technicals**

The function `liquidationAmount()` determines the amount of sUSD that needs to be liquidated to return a loan to its minimum collateralisation ratio.

This function is used by `liquidateInternal()` to determine how much of the loan can be liquidated. However, the parameter `payment` passed to the function by the liquidator is denominated in the synth that was borrowed. The following lines of code in `liquidateInternal()` are where the issue arises:

```solidity
// 8. Determine how much needs to be liquidated to fix their c ratio.
uint liqAmount = liquidationAmount(loan);

// 9. Only allow them to liquidate enough to fix the c ratio.
uint amountToLiquidate = liqAmount < payment ? liqAmount : payment;

// 10. Work out the total amount owing on the loan.
uint amountOwing = loan.amount.add(loan.accruedInterest);

// 11. If its greater than the amount owing, we need to close the loan.
if (amountToLiquidate >= amountOwing) {
    return closeByLiquidationInternal(borrower, msg.sender, loan);
}
```

In the case of an sUSD backed sETH short, the value passed as `payment` is denominated in sETH. However, the value assigned to `liqAmount` is denominated in sUSD. This means that `amountToLiquidate` is assigned to `payment` in most cases because the sUSD value `liqAmount` will in most cases exceed the value of `payment`.

This means that by passing in a large enough value for `payment`, `amountToLiquidate` will exceed `amountOwing` and therefore trigger `closeByLiquidationInternal`, closing a loan that may only require partial liquidation.

**Fix**

The function `liquidationAmount()` should convert the sUSD amount to be liquidated back into the borrowed currency. It is not useful for non sUSD positions to receive an sUSD value from this function. 

```solidity
function liquidationAmount(Loan memory loan) public view returns (uint amount) {
    uint liquidationPenalty = getLiquidationPenalty();
    uint debtValue = loan.amount.add(loan.accruedInterest).multiplyDecimal(
        _exchangeRates().rateForCurrency(loan.currency)
    );
    uint collateralValue = loan.collateral.multiplyDecimal(_exchangeRates().rateForCurrency(collateralKey));
    uint unit = SafeDecimalMath.unit();

    uint dividend = debtValue.sub(collateralValue.divideDecimal(minCratio));
    uint divisor = unit.sub(unit.add(liquidationPenalty).divideDecimal(minCratio));

    uint sUSDamount = dividend.divideDecimal(divisor);

    return _exchangeRates().effectiveValue(sUSD, sUSDamount, loan.currency);
}
```

 **Scope of bug**

Normal sUSD loans against ETH or renBTC are not vulnerable, since both values are denominated in sUSD. Loans such as sETH against ETH or sBTC against renBTC are vulnerable, however they are protected from movements in their collateralisation ratios and will only be at risk if enough interest accrues to put the position under the minimum collateralisation requirement.

The most vulnerable positions are sUSD shorts, since the shorts are denominated in sBTC/sETH and the collateralisation ratios of these positions varies with the BTC and ETH prices.

mjs-12:
This PR closes #1025.

It also removes the underlying contract from `CollateralShort`. Since the logic assumes the underlying is the `SynthsUSD`, there is no need to handle decimals and scaling like we do for `CollateralErc20`. This was done because the contract was over the size limit after the change made to fix the liquidation bug.


justinjmoses (approved):

justinjmoses:
Closing in favor of #1043

PR #1033, opened 2021-01-11
Sip 60 reward escrow v2 migrate escrow in batches

i-stam:
- Introduces a new way to migrate escrow entries: `initiateEscrowMigration` is expecting a batch of entries to be passed  which is relayed in separate messages e.g.
  `entryIDs =[[1,2,3],[4,5,6]]`  will relay two `completeEscrowMigration` L2 messages that will import the entries on L2.
  Max number of entries per message: 26 (fits safely into the 8mil L2 gas limit)
- There is also `depositAndMigrateEscrow` which can be used in cases efficiency and gas savings is the goal.
- `initiateWithdrawal` now checks the user's transferrable SNX and not the outstanding debt in order to avoid rounding errors.
- `requireZeroDebt` modifier introduced.
- Prod tests that work with forked envrionments.

Notes for the reviewers:

- importVestingEntries was renamed to completeEscrowMigration but the related events were not renamed. We are still using `ExportedVestingEntries` (L1) and `ImportedVestingEntries` (L2).
- `MintedSecondary` is emitted when a deposit is completed but SNX is also minted on escrow migration too. All minting can be tracked via `emitTransfer` event emission from MintableSynthetix.
- `initiateEscrowMigration` and `depositAndMigrateEscrow` are declared public in order to bypass a compiler bug when casting multidimensional arrays from calldata to memory (external to private function).
- `depositAndMigrateEscrow` has an` if => omit calling` pattern (e.g. if `depositAmount == 0` skip calling the deposit function which is not used in the rest of the code. Feedback?


contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -176,6 +155,20 @@ contract SynthetixBridgeToOptimism is Owned, MixinSystemSettings, ISynthetixBrid
         _initiateRewardDeposit(amount);
     }
 
+    function depositAndMigrateEscrow(uint256 depositAmount, uint256[][] memory entryIDs)
```
jacko125, line 158: @i-stam  So here the dapp / interface will pass in the `entryIDs` as batches of 26 entries in an array ? Like, 

`[[entryID1, entryID2, entryID3, entryID4 ...], [entryID26, entryID27, entryID28, entryID29, entryID30 ....]]`

i-stam: Yes, you can use also `initiateEscrowMigration` (ln 110). The batches don't necessarily need to consist of 26 entries. For example,  if you want to migrate 50 entries, it would be better to send 25+25.

jacko125: @clementbalestrat  This could be done via the staking dapp to split the list of entryIDs to migrate to L2 into batches

contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -207,6 +200,32 @@ contract SynthetixBridgeToOptimism is Owned, MixinSystemSettings, ISynthetixBrid
         emit Deposit(msg.sender, _depositAmount);
     }
 
+    function _initiateEscrowMigration(uint256[][] memory _entryIDs) private {
```
justinjmoses, line 203: [np] typically just use `internal` unless there's a reason for `private`?

contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -207,6 +200,32 @@ contract SynthetixBridgeToOptimism is Owned, MixinSystemSettings, ISynthetixBrid
...
+            require(_entryIDs[i].length <= MAX_ENTRIES_MIGRATED_PER_MESSAGE, "Exceeds max entries per migration");
+            // Burn their reward escrow first
+            // Note: escrowSummary would lose the fidelity of the weekly escrows, so this may not be sufficient
+            uint256 escrowedAccountBalance;
+            VestingEntries.VestingEntry[] memory vestingEntries;
+            (escrowedAccountBalance, vestingEntries) = rewardEscrowV2().burnForMigration(msg.sender, _entryIDs[i]);
+            // if there is an escrow amount to be migrated
+            if (escrowedAccountBalance > 0) {
+                // create message payload for L2
+                bytes memory messageData = abi.encodeWithSignature(
```
justinjmoses, line 216: was there a reason to not use `encodeWithSelector` here?

publish/releases.json
```diff
@@ -108,5 +108,14 @@
...
 		"sips": [97, 100, 103]
+	},
+	{
+		"name": "Castor",
+		"version": {
+			"major": 2,
+			"minor": 36
+		},
+		"sources": ["SynthetixBridgeToOptimism"],
+		"sips": [60, 102]
```
justinjmoses, line 119: 👍🏼

justinjmoses (approved):
A few comments but overall LGTM

contracts/SynthetixBridgeToOptimism.sol
i-stam: Just because the PR introducing `encodeWithSelector` was created against `develop` and it hasn't been merged. I though that we can add it after we merge everything since it has minimal impact.

contracts/SynthetixBridgeToOptimism.sol
i-stam: Since it is not going to be inherited anytime soon I used `private` instead.

PR #1036, opened 2021-01-12
Sip 60 escrowed snx liquidations

jacko125:
We currently have `liquidateDelinquentAccount` which allows you to liquidate all transferable SNX (Non escrowed SNX). If you tried to liquidate more than the account’s transferable SNX it would fail.
    
```
    function liquidateDelinquentAccount(
        address account,
        uint susdAmount,
        address liquidator
    )
```
    
Adding `liquidateEscrowedSNX` allows liquidator to specify the max amount of sUSD they want to liquidate, and also the vesting entryIDs for an account that they want to liquidate.
    
```
    function liquidateEscrowedSNX(
        address account,
        uint[] calldata entryIDs,
        uint susdAmount,
        address liquidator
    )
```
    
When liquidating escrowed SNX, we still apply the 10% liquidation bonus / penalty.

Escrow liquidation iterates over the list of entryID’s the liquidator passes in to fulfill the liquidation sUSD equivalent amount they want to liquidate / redeem in SNX.

The remainingAmounts of each liquidated escrowed SNX entry will be transferred to the liquidator as an escrowed VestingEntry (similar to account merging). The vesting terms (duration and ratePerSecond) will be the same.
The liquidator can then claim any already vested amounts once it’s in their ownership, but will have to wait for the remaining amounts to be streamed before claiming more.

If there is a partial liquidation of an escrow entry, in this edge case we reduce the remaining amount and create a new VestingEntry for the liquidator with the same vesting terms + redeemed amount.Tthe account being liquidated will have an updated `remainingAmount = remainingAmount - liquidated amounts` in their vesting entry to claim afterwards.  

- OVM compile oversize on issuer.sol

image


justinjmoses:
So the problem with oversize is actually not an issue with OVM deployment but rather how we check. The checking code can't ignore something that's a dependency, but since we deploy `IssuerWithoutLiquidations` on L2 as the `Issuer` it isn't an issue. @ajsantander any thoughts on having this not complain here in this instance?

justinjmoses (commented):
Can you change the branch to trigger the tests on these with `-develop`? 🙏🏼

PR #1060, opened 2021-01-17
Owner and nominate fixes for OVM support

justinjmoses:
This upgrade ensures the nominate and accept ownership steps function on L2.


publish/src/commands/owner.js
```diff
@@ -86,7 +86,7 @@ const owner = async ({
 	const web3 = new Web3(new Web3.providers.HttpProvider(providerUrl));
 
 	const code = await web3.eth.getCode(newOwner);
-	const isContract = code !== '0x';
+	const isContract = !useOvm && code !== '0x';
```
ajsantander, line 89: So, if it has code, it's not considered a contract if its on Optimism?
Why?

ajsantander (changes requested):

publish/src/commands/owner.js
justinjmoses: It basically a hacky way of saying don't use contract owners on L2 - which honestly won't stick around anyhow. I can't recall why this was necessary at the time though frankly - why was `getCode` not return `0x` when this was run?

ajsantander: Sorry, totally confused with what you're saying here 😅 

What you're doing is basically saying that "contract accounts in Optimism are considered EOAs" which I think is wrong.

ajsantander (approved):

publish/src/commands/owner.js
i-stam: I talked to Kevin about this. All accounts are initially EOAs (i.e. no code), but they are turned into a contract (proxy to the smart wallet) after they send the first tx. I 'm working to address this on the new owner-multisig script.

PR #1065, opened 2021-02-02
SIP-109 Synth Exchange Suspension for Market Closures

justinjmoses:


contracts/interfaces/ISystemStatus.sol
```diff
@@ -24,10 +24,14 @@ interface ISystemStatus {
...
     function requireExchangeActive() external view;
 
+    function requireExchangeBetweenSynthsAllowed(bytes32 sourceCurrencyKey, bytes32 destinationCurrencyKey) external view;
+
     function requireSynthActive(bytes32 currencyKey) external view;
 
     function requireSynthsActive(bytes32 sourceCurrencyKey, bytes32 destinationCurrencyKey) external view;
 
+    function synthExchangeSuspension(bytes32 currencyKey) external view returns (bool suspended, uint248 reason);
+
```
jacko125, line 34: Do we want to expose the new functions in the Interface ? 

- resumeSynths(bytes32[] calldata currencyKeys)
- resumeSynthExchange(bytes32 currencyKey)
- resumeSynthsExchange(bytes32[] calldata currencyKeys)
- suspendSynthExchange(bytes32 currencyKeys, uint256 reason)
- suspendSynthsExchange(bytes32[] calldata currencyKeys, uint256 reason)

jacko125 (commented):
SystemStatus changes LGTM. Only comment is whether the new functions need to be on the ISystemStatus ?

contracts/SystemStatus.sol
```diff
@@ -39,35 +43,72 @@ contract SystemStatus is Owned, ISystemStatus {
...
+        // and finally, the synths cannot be suspended
+        _internalRequireSynthActive(sourceCurrencyKey);
+        _internalRequireSynthActive(destinationCurrencyKey);
     }
 
     function isSystemUpgrading() external view returns (bool) {
         return systemSuspension.suspended && systemSuspension.reason == SUSPENSION_REASON_UPGRADE;
     }
 
+    function getSynthExchangeSuspensions(bytes32[] calldata synths)
```
jacko125, line 98: Thanks this will be useful for the Market closure to read all the Synth exchange suspension reasons

contracts/interfaces/ISystemStatus.sol
justinjmoses: Typically there's no need to expose the protected functions in the interface (as it's meant for third parties) though `suspendSynth` is an exception as it's used internally for [SIP-65](https://sips.synthetix.io/sips/sip-65)

contracts/SystemStatus.sol
```diff
@@ -22,12 +23,15 @@ contract SystemStatus is Owned, ISystemStatus {
...
 
+    mapping(bytes32 => Suspension) public synthExchangeSuspension;
+
     mapping(bytes32 => Suspension) public synthSuspension;
 
     constructor(address _owner) public Owned(_owner) {
         _internalUpdateAccessControl(SECTION_SYSTEM, _owner, true, true);
         _internalUpdateAccessControl(SECTION_ISSUANCE, _owner, true, true);
         _internalUpdateAccessControl(SECTION_EXCHANGE, _owner, true, true);
+        _internalUpdateAccessControl(SECTION_SYNTH_EXCHANGE, _owner, true, true);
```
ajsantander, line 34: Since SystemStatus is not really upgradeable, care will have to be taken for reconstructing the old contract's state. I imagine that non-owner access control for things like hot wallet suspension would be the most sensitive.

contracts/interfaces/ISystemStatus.sol
ajsantander: Agree with Jacko here. Justin, why do you say that these functions are protected? They are external, and thus part of the interface IMO.

ajsantander (commented):
Looks pretty good!

Question: what about Synths that call `Exchanger.exchange()`? This could be an edge case where Synth transfers may fail if exchange is suspended.

test/contracts/Exchanger.spec.js
```diff
@@ -636,6 +636,20 @@ contract('Exchanger (spec tests)', async accounts => {
...
+				beforeEach(async () => {
+					await setStatus({
+						owner,
+						systemStatus,
+						section: 'SynthExchange',
+						suspend: true,
+						synth: sBTC,
+					});
+				});
+				it('then settling it still works', async () => {
```
ajsantander, line 649: nit: typo

test/contracts/helpers.js
```diff
@@ -232,6 +232,12 @@ module.exports = {
 			} else {
 				await systemStatus.resumeExchange({ from: owner });
 			}
+		} else if (section === 'SynthExchange') {
+			if (suspend) {
+				await systemStatus.suspendSynthExchange(synth, reason, { from: owner });
+			} else {
+				await systemStatus.resumeSynthExchange(synth, { from: owner });
+			}
```
ajsantander, line 240: Could add a few more tests for:
* Trying to exchange when ore or more synths are suspended
* Recovering from an exchange suspension
* Trying to exchange with a regular suspension
* etc

test/contracts/RewardsIntegrationTests.js
```diff
@@ -147,7 +147,10 @@ contract('Rewards Integration Tests', async accounts => {
 
 	// run this once before all tests to prepare our environment, snapshots on beforeEach will take
 	// care of resetting to this state
-	before(async () => {
+	before(async function() {
+		// set a very long timeout for these (requires a non-fat-arrow above)
+		this.timeout(180e3);
```
ajsantander, line 152: How was this needed in this PR?

test/contracts/helpers.js
justinjmoses: I have that all in `Exchanger.spec.js`

test/contracts/RewardsIntegrationTests.js
justinjmoses: Unrelated but it keeps breaking coverage

contracts/SystemStatus.sol
justinjmoses: yep, already noted

justinjmoses:
> Question: what about Synths that call Exchanger.exchange()? This could be an edge case where Synth transfers may fail if exchange is suspended.
That's explicitly for purge, and it's allowed to bypass this - in fact it's particularly useful during iSynth repricing situations.

justinjmoses:
> Agree with Jacko here. Justin, why do you say that these functions are protected? They are external, and thus part of the interface IMO.
Typically we haven't included restricted external functions in the interfaces unless needed within the protocol. We've avoided it to make it easier for third parties to quickly tell which functions are available to them. In fact we'd probably want two different interfaces - one internal and one external - but we need Solc 0.6+ to get interface inheritance. So v3?

jacko125 (approved):

PR #1070, opened 2021-02-03
Hardhat asts develop

ajsantander:


package.json
```diff
@@ -18,6 +18,7 @@
 		"prepublishOnly": "npm run generate-asts && npm run pack",
 		"fork": "node --max-old-space-size=4096 ./node_modules/.bin/hardhat node --target-network mainnet",
 		"test": "hardhat test",
+		"describe": "hardhat describe",
```
justinjmoses, line 21: can you remove `"generate-asts": "hardhat compile",` and we'll update the releases to call describe instead

ajsantander: Done

justinjmoses (approved):

PR #1086, opened 2021-02-12
SIP-117 Add OVM exchanging functionality

i-stam:


contracts/BaseSynthetix.sol
```diff
@@ -307,7 +315,7 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         _;
     }
 
-    function _systemActive() private {
+    function _systemActive() private view {
```
ajsantander, line 318: Why are you adding this visibility identifier?

SystemStatus.requireSystemActive() can revert, thus its not just a view.

contracts/BaseSynthetix.sol
```diff
@@ -316,7 +324,57 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         _;
     }
 
-    function _issuanceActive() private {
+    function _issuanceActive() private view {
```
ajsantander, line 327: Same here.

contracts/BaseSynthetix.sol
```diff
@@ -316,7 +324,57 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
         systemStatus().requireIssuanceActive();
     }
+
+    modifier exchangeActive(bytes32 src, bytes32 dest) {
+        _exchangeActive(src, dest);
+        _;
+    }
+
+    function _exchangeActive(bytes32 src, bytes32 dest) private view {
```
ajsantander, line 336: The original `Synthetix._exchangeActive(...)` required these conditions:
* requireExchangeBetweenSynthsAllowed(src, dest)

And you are suggesting to change the conditions to:
* requireExchangeActive()
* requireSynthsActive(src, dest)

Which I think is a regression because the number of calls increases to two, and the actual internal checks become redundant and poorer:

Original (6 checks):
* requireExchangeBetweenSynthsAllowed(src, dest)
  * _internalRequireSystemActive();
  * _internalRequireExchangeActive();
  * _internalRequireSynthExchangeActive(sourceCurrencyKey);
  * _internalRequireSynthExchangeActive(destinationCurrencyKey);
  * _internalRequireSynthActive(sourceCurrencyKey);
  * _internalRequireSynthActive(destinationCurrencyKey);

New (4 checks, 1 redundant):
* requireExchangeActive()
  * _internalRequireSystemActive();
  * _internalRequireExchangeActive();
* requireSynthsActive(src, dest)
  * _internalRequireSystemActive(); <------- AGAIN
  * _internalRequireSynthActive(sourceCurrencyKey);
  * _internalRequireSynthActive(destinationCurrencyKey);
  
  In fact, the new suggestion would allow an exchange for two synths whose exchanging is disabled, which is a critical bug.

contracts/Collateral.sol
```diff
@@ -762,7 +762,7 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
         _;
     }
 
-    function _requireRateIsValid() private {
+    function _requireRateIsValid() private view {
```
ajsantander, line 765: As commented above, I think this is semantically incorrect.

contracts/Exchanger.sol
```diff
@@ -334,7 +334,7 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
             false
         );
 
-        _processTradingRewards(fee, destinationAddress);
+        // _processTradingRewards(fee, destinationAddress);
```
ajsantander, line 337: We should definitely not do this.

If we were to just comment out calls to this internal function, we should consistently comment out all calls. I.e. `exchangeOnBehalf` also calls it, as do others.

The proper way to do this would be to simply set `tradingRewardsEnabled` to false in all deployed instances.
2.

contracts/RealtimeDebtCache.sol
```diff
@@ -65,11 +65,7 @@ contract RealtimeDebtCache is DebtCache {
 
     function takeDebtSnapshot() external {}
 
-    function updateCachedSynthDebts(bytes32[] calldata currencyKeys) external {}
```
ajsantander, line 68: We should probably get @zyzek to validate this too.

test/optimism/index.js
```diff
@@ -64,10 +65,10 @@ describe('Layer 2 production tests', () => {
 		});
 	});
 
-	after('exit', async () => {
-		// TODO: Optimism watchers leave the process open, so we explicitely kill it
-		process.exit(0);
-	});
+	// after('exit', async () => {
```
ajsantander, line 68: Please try to avoid commenting out code. If its not needed, remove it.

test/optimism/synthExchange.test.js
```diff
@@ -0,0 +1,235 @@
+const ethers = require('ethers');
+const { assert } = require('../contracts/common');
+const { connectContract } = require('./utils/connectContract');
+const { toBytes32 } = require('../..');
+const { assertRevertOptimism } = require('./utils/revertOptimism');
+
+const itCanPerformSynthExchange = ({ ctx }) => {
+	describe.only('[SYNTEXCHANGE] when exchanging synths on L2', () => {
```
ajsantander, line 8: Remove the only

test/optimism/synthExchange.test.js
```diff
@@ -0,0 +1,235 @@
...
+										const tx = await SynthetixL2.settle(toBytes32('sETH'));
+
+										await assertRevertOptimism({
+											tx,
+											reason: 'Cannot settle during waiting period',
+											provider: ctx.providerL2,
+										});
+									});
+
+									// before('settle', async () => {
```
ajsantander, line 215: As before. Commented out code suggests that this is incomplete? If it is complete, it should be removed in a PR that is ready for review.

ajsantander: Btw, it might actually be a good idea to test settlement in prod right?

ajsantander (changes requested):
Great job @i-stam 

Reported a couple of medium things on the contracts, and general aesthetic things elsewhere.

contracts/BaseSynthetix.sol
i-stam: The compiler gives warnings, reverting does not technically invalidate it as a view. Can remove if it is confusing.

contracts/BaseSynthetix.sol
i-stam: This change wasn't made by me, it was introduced by another PR, I just updated the branch because it was lost during a merge.

jacko125 (approved):

contracts/Exchanger.sol
i-stam: It has been added back.

contracts/RealtimeDebtCache.sol
i-stam: It has been added back, no change has been made.

test/optimism/synthExchange.test.js
i-stam: Settle revert is tested, I couldn't  fast forward in ovm.

justinjmoses (approved):
Nice work

PR #1093, opened 2021-02-19
Configurable waiting period

i-stam:


publish/src/commands/deploy.js
```diff
@@ -1814,7 +1814,7 @@ const deploy = async ({
 			read: 'waitingPeriodSecs',
 			expected: input => input !== '0',
 			write: 'setWaitingPeriodSecs',
-			writeArg: await getDeployParameter('WAITING_PERIOD_SECS'),
+			writeArg: useOvm ? '0' : await getDeployParameter('WAITING_PERIOD_SECS'),
```
justinjmoses, line 1817: this can be overridden in the params per environment instead of coding this here, no?

contracts/BaseSynthetix.sol
```diff
@@ -171,6 +171,22 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, IBaseSyntheti
         return exchanger().exchange(messageSender, sourceCurrencyKey, sourceAmount, destinationCurrencyKey, messageSender);
     }
 
+    function exchangeOnBehalf(
```
justinjmoses, line 174: any reason to move this?

publish/src/commands/deploy.js
i-stam: Done! I modified only the `params.json` file in local-ovm deployment folder because it is used by the `deploy-ovm-pair` script.

contracts/Exchanger.sol
```diff
@@ -587,15 +587,23 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
-            destinationCurrencyKey,
-            amountReceived,
-            exchangeFeeRate
-        );
+        // iff the waiting period is gt 0
+        if (getWaitingPeriodSecs() > 0) {
+            // Nothing changes as far as issuance data goes because the total value in the system hasn't changed.
+            // But we will update the debt snapshot in case exchange rates have fluctuated since the last exchange
+            // in these currencies
+            _updateSNXIssuedDebtOnExchange([sourceCurrencyKey, destinationCurrencyKey], [sourceRate, destinationRate]);
```
justinjmoses, line 595: ⚠️ This should still exist above - we shouldn't mix no waiting period with an assumption that the real time debt caching is being used.

publish/deployed/local-ovm/params.json
```diff
@@ -2,5 +2,9 @@
 	{
 		"name": "INITIAL_ISSUANCE",
 		"value": "0"
+	},
+	{
```
justinjmoses, line 6: 🙏🏼 we should probably add these for kovan-ovm and mainnet-ovm as well in this PR

justinjmoses (changes requested):
a few small comments

i-stam:
Opened a new one that triggers CI unit testing

PR #1099, opened 2021-02-20
Refactor debtCache

i-stam:
This PR follows the pattern introduced for splitting code in L1 and L2 counterparts by introducing a BaseDebtCache contract.
In this way: 
1.  Necessary functionality is explicitly stated instead of stubbing every function available.
2. It allows for further optimizations (code elimination and further code separation e.g. in Exchanger) 
3. It also allows for future decoupling/renaming. (there is no such thing as cache on L2)


contracts/Exchanger.sol
```diff
@@ -587,15 +586,17 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
-        appendExchange(
-            destinationAddress,
-            sourceCurrencyKey,
-            sourceAmountAfterSettlement,
-            destinationCurrencyKey,
-            amountReceived,
-            exchangeFeeRate
-        );
+        // persist the exchange information for the dest key iff the waiting period is gt 0
+        if (getWaitingPeriodSecs() > 0) {
```
justinjmoses, line 590: want to remove this from here as it's in the other PR?

contracts/interfaces/IBaseDebtCache.sol
```diff
@@ -0,0 +1,47 @@
+pragma solidity >=0.4.24;
```
justinjmoses, line 1: I think we skip the creation of this file - it's not needed

i-stam:
Opened a new one with a `-develop` suffix to trigger CI unit testing

PR #1101, opened 2021-02-22
Change max debt default to 30m

mjs-12:
Fixes failing fork test.


jacko125 (approved):
Thanks @mjspain12 for updating this. We should keep track of this next SCCP if the max debt increases.

PR #1102, opened 2021-02-23
One CI workflow to rule them all

ajsantander:
Also removes the branch naming filters, so this one workflow to rule them all is always triggered.


justinjmoses (approved):

justinjmoses:
That's a weird failure - looks like it was killed? `Error: Invalid JSON RPC response: ""` used to happen from time to time, that why the test publish script has retries on it...

justinjmoses (approved):

hardhat/tasks/task-compile.js
```diff
@@ -25,7 +25,7 @@ task('compile')
 		if (taskArguments.useOvm) {
 			console.log(gray('Compiling with OVM Solidity compiler...'));
 
-			require('@eth-optimism/plugins/hardhat/compiler/0.5.16');
```
K-Ho, line 28: does this work? I think you may have to keep it at `@eth-optimism/plugins/hardhat/compiler/0.5.16`

K-Ho: nevermind!

ajsantander: Yes, according to the docs!
https://hardhat.org/plugins/eth-optimism-plugins-hardhat-compiler.html

justinjmoses (approved):
:boom:

ajsantander:
> That's a weird failure - looks like it was killed? `Error: Invalid JSON RPC response: ""` used to happen from time to time, that why the test publish script has retries on it...

This turned out to be the CI running out of memory, but not reporting the error properly because of stdout and stderr re-routing done in the tests.

PR #1103, opened 2021-02-23
Configurable waiting period

i-stam:
Allows for skipping certain functionality when `WAITING_PERIOD_SECS` is set to 0 i.e. no settlement needed.

Notes: 
- `sourceAmountAfterSettlement` is eliminated and the function argument `sourceAmount` is used instead in order to avoid stack too deep errors. (https://github.com/Synthetixio/synthetix/blob/dbe0d96d8b5b13bf2c101d64c589d1467061579a/contracts/Exchanger.sol#L519). An alternative solution would be to use structs as parameters but it would be much more intrusive.
- not assuming anymore that a 0 waiting period implies that RealtimeDebtCache is used.


contracts/Exchanger.sol
```diff
@@ -512,11 +512,16 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
     {
         _ensureCanExchange(sourceCurrencyKey, sourceAmount, destinationCurrencyKey);
 
-        uint sourceAmountAfterSettlement = _settleAndCalcSourceAmountRemaining(sourceAmount, from, sourceCurrencyKey);
+        bool needsSettlement = getWaitingPeriodSecs() > 0 ? true : false;
+
+        // If settlement is needed then update the source amount
+        if (needsSettlement) {
+            sourceAmount = _settleAndCalcSourceAmountRemaining(sourceAmount, from, sourceCurrencyKey);
```
justinjmoses, line 519: The only issue here is if we turn off the waiting period but there are still historical exchanges to be settled. It would be worth seeing what the gas cost is of just keeping this check regardless.

i-stam: This is a good point, will do.

contracts/BaseSynthetix.sol
```diff
@@ -163,6 +163,42 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
 
     // ========== MUTATIVE FUNCTIONS ==========
 
+    function exchange(
```
justinjmoses, line 166: [np] was there a reason to reorder the functions in this contract?

justinjmoses, line 166: so moving this in here is actually a SIP-114 concern not this - can you reduce this to just the waiting period conditional logic?

i-stam: This is due to the fact that this branch was created out of the synth-exchange one.

i-stam:
opened #1110 instead

PR #1105, opened 2021-02-23
SIP-116 Implementation

ajsantander:
Implementation of SIP 116:
https://sips.synthetix.io/sips/sip-116


contracts/interfaces/ISynthetixBridgeToBase.sol
```diff
@@ -6,10 +6,12 @@ import "./IRewardEscrowV2.sol";
 
 interface ISynthetixBridgeToBase {
     // invoked by users on L2
-    function initiateWithdrawal(uint amount) external;
+    function withdraw(uint amount) external;
```
justinjmoses, line 9: Hmmm - I think we should be using the official interfaces instead of this. 

So I propose reduding this interface down to just the non-standard functions and then have the deposit and withdrawal extend both these and their corresponding Optimism interface.

test/contracts/SynthetixBridgeToOptimism.spec.js
```diff
@@ -115,9 +115,59 @@ contract('SynthetixBridgeToOptimism (spec tests)', accounts => {
...
+
+						assert.bnEqual(userBalanceBefore.sub(toBN(amountToDeposit)), userBalanceAfter);
+					});
+
+					it("increases the contract's balance", async () => {
+						assert.bnEqual(
+							await synthetix.balanceOf(synthetixBridgeToOptimism.address),
+							contractBalanceBefore.add(amountToDeposit)
+						);
+					});
```
ajsantander, line 163: Further testing needed

contracts/interfaces/ISynthetixBridgeToBase.sol
ajsantander: Done

i-stam:
lgtm, since we are importing the optimism interfaces should we change argument naming e.g. `DepositInitiated(account, to, uint256 amount)` => `DepositInitiated(from,to,amount)`.  The functionality does not change but it is a bit more intuitive to the reader.

ajsantander:
> lgtm, since we are importing the optimism interfaces should we change argument naming e.g. `DepositInitiated(account, to, uint256 amount)` => `DepositInitiated(from,to,amount)`. The functionality does not change but it is a bit more intuitive to the reader.

Will do!

i-stam (approved):

contracts/SynthetixBridgeToBase.sol
```diff
@@ -69,18 +71,30 @@ contract SynthetixBridgeToBase is Owned, MixinSystemSettings, ISynthetixBridgeTo
         addresses = combineArrays(existingAddresses, newAddresses);
     }
 
+    function getFinalizeWithdrawalL1Gas() external view returns (uint32) {
```
ajsantander, line 74: I believe this was removed from the standard interface

contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -89,6 +91,10 @@ contract SynthetixBridgeToOptimism is Owned, MixinSystemSettings, ISynthetixBrid
         addresses = combineArrays(existingAddresses, newAddresses);
     }
 
+    function getFinalizeDepositL2Gas() external view returns (uint32) {
```
ajsantander, line 94: This too may have been removed from the standard interface

contracts/SynthetixBridgeEscrow.sol
```diff
@@ -0,0 +1,34 @@
...
+        return IERC20(requireAndGetAddress(CONTRACT_SYNTHETIX));
+    }
+
+    function resolverAddressesRequired() public view returns (bytes32[] memory addresses) {
+        addresses = new bytes32[](1);
+        addresses[0] = CONTRACT_SYNTHETIX;
+    }
+
+    function approveBridge(address _bridge, uint256 _amount) external onlyOwner {
+        IERC20(address(synthetixERC20())).approve(_bridge, _amount);
```
justinjmoses, line 28: qu: is it worth removing the MixinResolver altogether and simply passing through the token to approve? This way if there's ever other tokens mistakenly sent, we could recover them. (It also makes this contract super slim).

contracts/SynthetixBridgeEscrow.sol
```diff
@@ -0,0 +1,34 @@
...
+        emit Approval(msg.sender, _bridge, _amount);
```
justinjmoses, line 29: Let's call this `BridgeApproval` or some such

contracts/SynthetixBridgeToBase.sol
```diff
@@ -126,7 +112,10 @@ contract SynthetixBridgeToBase is Owned, MixinSystemSettings, ISynthetixBridgeTo
...
         VestingEntries.VestingEntry[] vestingEntries
     );
-    event MintedSecondary(address indexed account, uint256 amount);
+
+    event DepositFinalized(address indexed to, uint256 amount);
+
     event MintedSecondaryRewards(uint256 amount);
-    event WithdrawalInitiated(address indexed account, uint256 amount);
+
+    event WithdrawalInitiated(address indexed from, address to, uint256 amount);
```
justinjmoses, line 120: Should we redeclare these here? Would it not be better to use the version already defined in the interface? https://github.com/ethereum-optimism/optimism/blob/master/packages/contracts/contracts/optimistic-ethereum/iOVM/bridge/tokens/iOVM_L2DepositedToken.sol#L14-L23

package.json
```diff
@@ -129,7 +130,7 @@
 	},
 	"dependencies": {
 		"@chainlink/contracts-0.0.10": "npm:@chainlink/contracts@0.0.10",
-		"@eth-optimism/contracts": "0.0.2-alpha.7",
+		"@eth-optimism/contracts": "0.2.7",
```
justinjmoses, line 133: I notice 0.2.8 was published - shall we upgrade: https://github.com/ethereum-optimism/optimism/blob/master/packages/contracts/CHANGELOG.md#028

contracts/SynthetixBridgeEscrow.sol
i-stam: Yeap, I thought about this. the above implementation makes it more explicit but adds the extra code. Removing it makis it more versatile but we have to make sure that we pass the right address everytime.

package.json
```diff
@@ -22,6 +22,7 @@
 		"test:prod": "hardhat test:prod --network localhost --optimizer",
 		"test:prod:gas": "hardhat test:prod --gas --network localhost --optimizer --gas-output-file test-gas-used-prod.log",
 		"test:prod:ovm": "concurrently --kill-others --success first \"wait-port 8545 && wait-port 9545 && mocha test/optimism --timeout 900000\"",
+		"test:migrate:bridge": "concurrently --kill-others --success first \"wait-port 8545 && wait-port 9545 && mocha test/migrate-bridge --timeout 900000\"",
```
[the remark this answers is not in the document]
i-stam: yeap, we should remove it

contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -230,20 +197,23 @@ contract SynthetixBridgeToOptimism is Owned, MixinSystemSettings, ISynthetixBrid
...
     event ExportedVestingEntries(
         address indexed account,
         uint256 escrowedAccountBalance,
         VestingEntries.VestingEntry[] vestingEntries
     );
+
     event RewardDeposit(address indexed account, uint256 amount);
-    event WithdrawalCompleted(address indexed account, uint256 amount);
+
+    event WithdrawalFinalized(address indexed to, uint256 amount);
```
justinjmoses, line 218: As above - I don't think we need to redefine these since the interface brings them in

contracts/SynthetixBridgeToBase.sol
i-stam: yeap, my only concern is that someone checking the contract will be confused trying to find where this event is declared originally.

contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -78,103 +54,90 @@ contract SynthetixBridgeToOptimism is Owned, MixinSystemSettings, ISynthetixBrid
...
-        uint256 contractBalance = ERC20Synthetix.balanceOf(address(this));
-        ERC20Synthetix.transfer(newBridge, contractBalance);
-
-        emit BridgeMigrated(address(this), newBridge, contractBalance);
+        emit WithdrawalFinalized(to, amount);
     }
 
     // invoked by RewardsDistribution on L1 (takes SNX)
-    function notifyRewardAmount(uint256 amount) external requireActive {
+    function notifyRewardAmount(uint256 amount) external requireInitiationActive {
```
justinjmoses, line 111: Note: due to the way `RewardsDistribution.distributeRewards` works, even if this reverts the SNX will still be sent to it. So if this is disabled and `Synthetix.mint` is called with this as an inflationary supply recipient, the SNX would come here but nothing marked on L2. As such, I don't think we should require initiation active - but allow it in all cases. At least the SNX will go into the escrow contract in that case - rather than being stuck in here.

https://github.com/Synthetixio/synthetix/blob/master/contracts/RewardsDistribution.sol#L162-L173

cc @jacko125

publish/src/commands/deploy.js
```diff
@@ -1213,6 +1214,11 @@ const deploy = async ({
 			deps: ['AddressResolver'],
 			args: [account, addressOf(readProxyForResolver)],
 		});
+		await deployer.deployContract({
```
justinjmoses, line 1217: technically this stuff can all be controlled via config.json files and then it doesn't need to be put in if blocks, but it's fine cc @ajsantander

publish/src/commands/deploy.js
```diff
@@ -68,6 +68,7 @@ const deploy = async ({
 	ignoreSafetyChecks,
 	ignoreCustomParameters,
 	concurrency,
+	specifyContracts,
```
justinjmoses, line 71: Did the code for this get lost in a merge?

justinjmoses (commented):
Changes inline

contracts/SynthetixBridgeToOptimism.sol
i-stam: Won't everything just revert?

contracts/SynthetixBridgeEscrow.sol
justinjmoses: I think it's totally worth the trade-off. It's not something that should happen very often anyhow.

contracts/SynthetixBridgeToBase.sol
justinjmoses: Well if you emit them using the interface prefix: `emit iOVM_L2DepositedToken.WithdrawalFInalized` - then it should be clear enough, no?

contracts/SynthetixBridgeEscrow.sol
i-stam: done

contracts/SynthetixBridgeToBase.sol
i-stam: done

package.json
i-stam: done

contracts/SynthetixBridgeToOptimism.sol
i-stam: done

package.json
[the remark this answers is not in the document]
i-stam: fixed

publish/src/commands/deploy.js
i-stam: ah right, must have been removed in the recent merge, I faced the same thing when I merged from develop in the other branch, added it back

contracts/SynthetixBridgeEscrow.sol
i-stam: Updated

contracts/SynthetixBridgeEscrow.sol
i-stam: Updated

contracts/SynthetixBridgeToBase.sol
```diff
@@ -126,7 +112,8 @@ contract SynthetixBridgeToBase is Owned, MixinSystemSettings, ISynthetixBridgeTo
         uint256 escrowedAmount,
         VestingEntries.VestingEntry[] vestingEntries
     );
-    event MintedSecondary(address indexed account, uint256 amount);
+
+    event DepositFinalized(address indexed to, uint256 amount);
```
zyzek, line 116: Is this event not already declared within `iOVM_L2DepositedToken.sol`?

contracts/SynthetixBridgeToBase.sol
```diff
@@ -126,7 +112,8 @@ contract SynthetixBridgeToBase is Owned, MixinSystemSettings, ISynthetixBridgeTo
...
+
     event MintedSecondaryRewards(uint256 amount);
```
zyzek, line 118: I'm wondering if the nomenclature here should be made consistent with the "Finalized" language; e.g. `MintedSecondaryRewards` -> `RewardDepositFinalized` or the like, and similarly with `ImportedVestingEntries`.

contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -230,20 +197,21 @@ contract SynthetixBridgeToOptimism is Owned, MixinSystemSettings, ISynthetixBrid
...
                 emit ExportedVestingEntries(msg.sender, escrowedAccountBalance, vestingEntries);
             }
         }
     }
 
     // ========== EVENTS ==========
 
-    event BridgeMigrated(address oldBridge, address newBridge, uint256 amount);
-    event Deposit(address indexed account, uint256 amount);
+    event DepositInitiated(address indexed from, address to, uint256 amount);
```
zyzek, line 208: This appears to already exist in the `iOVM_L1TokenGateway` interface.

contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -230,20 +197,21 @@ contract SynthetixBridgeToOptimism is Owned, MixinSystemSettings, ISynthetixBrid
...
+
     event ExportedVestingEntries(
         address indexed account,
         uint256 escrowedAccountBalance,
         VestingEntries.VestingEntry[] vestingEntries
     );
+
     event RewardDeposit(address indexed account, uint256 amount);
```
zyzek, line 216: And again as in the opposing bridge, I'm wondering if the event naming here might be a bit more sensible if it were to use the "Initiated" style language of the interfaces, which would also be consistent with the "Finalized" language that exists on the other side.

contracts/SynthetixBridgeToBase.sol
```diff
@@ -60,27 +38,32 @@ contract SynthetixBridgeToBase is Owned, MixinSystemSettings, ISynthetixBridgeTo
...
-    function initiateWithdrawal(uint amount) external {
+    function withdraw(uint amount) external requireInitiationActive {
+        _initiateWithdraw(msg.sender, amount);
+    }
+
+    function withdrawTo(address to, uint amount) external requireInitiationActive {
+        _initiateWithdraw(to, amount);
+    }
+
+    function _initiateWithdraw(address to, uint amount) private {
```
zyzek, line 58: I assume we don't care too much if people call this (and the deposit function on L1) with 0 amount? Just wondering if that incurs a cost for us on the other end.

zyzek (commented):
The structure looks pretty good to me. I only have a couple of small observations about naming and declaration sites.

contracts/SynthetixBridgeToOptimism.sol
jacko125: @i-stam  So `Synthetix.mint` and `RewardsDistribution.distributeRewards` will use a low level call that will not revert if calling `notifyRewardAmount` fails here. What's the plan on recovering the SNX from this contract if it's transferred here but the bridge is not active?

contracts/SynthetixBridgeEscrow.sol
```diff
@@ -0,0 +1,26 @@
+pragma solidity ^0.5.16;
+
+// Inheritance
+import "./Owned.sol";
+import "./MixinResolver.sol";
+
+// External references.
+import "openzeppelin-solidity-2.3.0/contracts/token/ERC20/SafeERC20.sol";
+
+contract SynthetixBridgeEscrow is Owned {
```
jacko125, line 10: So the BridgeEscrow will hold all the SNX tokens as a vault / escrow so that other bridge implementations can spend the tokens on withdrawals from L2?

contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -213,13 +177,16 @@ contract SynthetixBridgeToOptimism is Owned, MixinSystemSettings, ISynthetixBrid
             uint256 escrowedAccountBalance;
             VestingEntries.VestingEntry[] memory vestingEntries;
             (escrowedAccountBalance, vestingEntries) = rewardEscrowV2().burnForMigration(msg.sender, _entryIDs[i]);
+
             // if there is an escrow amount to be migrated
             if (escrowedAccountBalance > 0) {
+                // NOTE: transfer SNX to synthetixBridgeEscrow because burnForMigration() transfers them to this contract.
+                synthetixERC20().transfer(synthetixBridgeEscrow(), escrowedAccountBalance);
```
jacko125, line 184: Ok with this, as RewardEscrowV2 still sends the SNX to the `synthetixBridgeToOptimism` bridge first and doesn't know about the BridgeEscrow.

contracts/SynthetixBridgeToBase.sol
jacko125: Good point, and have we looked into what will the gas costs be for withdrawals on L1 for transferring the SNX out of the `BridgeEscrow` on withdrawals. 

These are covered by the messenger / relayer currently, would it be needed that some of the gas cost is repaid with SNX.

contracts/SynthetixBridgeToBase.sol
i-stam: Good catch! we can emit it via the `iOVM_L2DepositedToken` interface, I missed this one

contracts/SynthetixBridgeToBase.sol
i-stam: it should be, we just didn't change it in the past (we just changed the ones specified in the ovm interfaces to be consistent) and got stuck with it. If on time, I will change these too

contracts/SynthetixBridgeToBase.sol
i-stam: The thing is that withdrawals will be expensive and users will have to pay. we can definitely disallow 0 amount withdrawals though.

contracts/SynthetixBridgeEscrow.sol
i-stam: yes, in theory, you can have multiple bridge implementations being live concurrently.

contracts/SynthetixBridgeToOptimism.sol
i-stam: yes, we did it this way because we didn't want to change the `RewardEscrowV2` implementation.

contracts/SynthetixBridgeToOptimism.sol
i-stam: ah right,` (bool success, ) = distributions[i].destination.call(payload);`, `success` does not trigger a revert, I forgot about this part. It all boild down to whether we want to include in the bridge the SNX recovery mechanism or not.

contracts/BaseSynthetixBridge.sol
```diff
@@ -0,0 +1,84 @@
...
+
+    function synthetix() internal view returns (ISynthetix) {
+        return ISynthetix(requireAndGetAddress(CONTRACT_SYNTHETIX));
+    }
+
+    function rewardEscrowV2() internal view returns (IRewardEscrowV2) {
+        return IRewardEscrowV2(requireAndGetAddress(CONTRACT_REWARDESCROW));
+    }
+
+    function initiatingActive() internal view {
```
ajsantander, line 43: nit: `initiationActive()` is more consistent with other naming.

test/contracts/SynthetixBridgeToBase.spec.js
```diff
@@ -11,8 +11,8 @@ const {
 	},
 } = require('../../');
 
-contract('SynthetixBridgeToBase (spec tests)', accounts => {
-	const [, owner, user] = accounts;
+contract('SynthetixBridgeToBase (spec tests) @ovm-skip', accounts => {
```
ajsantander, line 14: `@ovm-skip` is no longer used. See: https://github.com/Synthetixio/synthetix/pull/1160

They should be removed, but I do wonder why you added these @i-stam

ajsantander:
@i-stam I couldn't find any tests for `initiationActive` actually stopping things like a deposit or a withdrawal, and not affecting other things. We shouldn't go into prod without these.

ajsantander (commented):
Requesting more tests pls 🙏 

As commented above, we should add more tests for the functionality that is radically new:
- Initiation modifiers stopping things starting
- Initiation modifiers not stopping things finalizing
- Initiation modifiers affecting some things but not others
- Tokens held in the escrow instead of the bridge

contracts/SynthetixBridgeToBase.sol
justinjmoses: I don't think it's necessary to prevent this - Yannis is correct as ultimately the user will pay to initiate a withdrawal

justinjmoses (approved):

i-stam:
> @i-stam I couldn't find any tests for `initiationActive` actually stopping things like a deposit or a withdrawal, and not affecting other things. We shouldn't go into prod without these.

They have been added in the new prod tests and there are also unit tests for this.

PR #1110, opened 2021-02-26
SIP-118 Conditional Fee Reclamation

i-stam:
implementation of https://github.com/Synthetixio/SIPs/pull/403


publish/src/commands/deploy-ovm-pair.js
```diff
@@ -58,7 +58,7 @@ const deployInstance = async ({ useOvm, privateKey, l1ProviderUrl, l2ProviderUrl
 		methodCallGasLimit: '3500000',
 		contractDeploymentGasLimit: useOvm ? OVM_MAX_GAS_LIMIT : '9500000',
 		privateKey,
-		ignoreCustomParameters: true,
+		ignoreCustomParameters: false,
```
justinjmoses, line 61: @i-stam is this necessary if instead we have the local-ovm tests simply check what `waitingPeriodSecs` is on-chain and test based on that?

i-stam: not necessary in this specific PR, we will address this when we update the ovm prod tests

test/contracts/Exchanger.spec.js
```diff
@@ -192,6 +194,141 @@ contract('Exchanger (spec tests)', async accounts => {
 		});
 	});
 
+	describe('When the waiting period is set to 0', () => {
+		let initialWaitingPeriod;
+
+		beforeEach(async () => {
+			initialWaitingPeriod = await systemSettings.waitingPeriodSecs();
+			await systemSettings.setWaitingPeriodSecs('0', { from: owner });
```
justinjmoses, line 202: this isn't necessary to checj here but rather in `SystemSettings.js`

test/contracts/Exchanger.spec.js
```diff
@@ -192,6 +194,141 @@ contract('Exchanger (spec tests)', async accounts => {
...
+				assert.bnEqual(await sETHContract.balanceOf(owner), '0');
+			});
+
+			describe('When the waiting period is switched on again', () => {
+				beforeEach(async () => {
+					await systemSettings.setWaitingPeriodSecs(initialWaitingPeriod, { from: owner });
+				});
+
+				it('is set correctly', async () => {
+					assert.bnEqual(await systemSettings.waitingPeriodSecs(), initialWaitingPeriod);
```
justinjmoses, line 241: as before, this isn't necessary to check as the logic belongs to SystemSettings

justinjmoses (approved):
💯 Nice work - just a note about a test or two.

test/contracts/Exchanger.spec.js
i-stam: Hmmm, I see, I just added it to make it more explicit that the change has taken effect.

contracts/Exchanger.sol
```diff
@@ -587,15 +587,18 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
-        // persist the exchange information for the dest key
-        appendExchange(
-            destinationAddress,
-            sourceCurrencyKey,
-            sourceAmountAfterSettlement,
-            destinationCurrencyKey,
-            amountReceived,
-            exchangeFeeRate
-        );
+        // iff the waiting period is gt 0
```
gberg1, line 590: @i-stam extra f in iff

i-stam: ah, this is an abbreviation of `if and only if` i.e.` <=>`

PR #1119, opened 2021-03-02
ci: add L2 ovm tests

drptbl:


.circleci/config.yml
```diff
@@ -128,6 +136,38 @@ jobs:
...
+            cd optimism-integration
+            docker-compose pull
+      - run:
+          name: Start chains
+          background: true
+          command: |
+            cd optimism-integration
+            ./up.sh
+      - cmd-wait-for-rpc
+      - cmd-wait-for-hardhat
```
justinjmoses, line 162: can we change the port to a parameter and pass it in and then we don't have two separate commands?

justinjmoses (changes requested):
One change with a parameter for a ci command

.circleci/config.yml
drptbl: @justinjmoses done!

justinjmoses (approved):
🙏🏼

PR #1121, opened 2021-03-02
Explicitly control compilation in CI diff prod local tests for ovm

ajsantander:
These are a bit tricky because we need to compile for EVM, but run prod tests for an OVM configuration (not artifacts).

With the latest hardhat, the task runner's decision to compile internally could break the delicate balance described above. This PR makes it explicit, by force compiling before, and force NOT compiling during tests.


i-stam (approved):

hardhat/tasks/task-test-prod.js
```diff
@@ -5,6 +5,7 @@ task('test:prod', 'run production tests against a running fork')
 	.addFlag('gas', 'Compile gas usage')
 	.addFlag('patchFreshDeployment', 'Patches up some things in production tests for new deployments')
 	.addFlag('useOvm', 'Uses an Optimism configuration')
+	.addFlag('noCompile', 'Avoid auto compilation')
```
justinjmoses, line 8: does this do anything if it's just a flag?

ajsantander: What we want is --no-compile to reach the test task. By declaring the flag here, we can call test-ovm with the flag, and make it reach the test task.

PR #1122, opened 2021-03-02
Enable CI ovm unit tests

ajsantander:
Re enables the call to `npm run test:ovm` on CI, but adds `@ovm-skip` to all contracts.

The idea is to use this as a base to slowly remove those skips, and incrementally support ovm unit tests. The reasoning behind this is that we've been disabling this type of testing for months because its just too hard to fix it all. So, even having one of those tests running would be better than our current situation.


K-Ho:
@ajsantander - can you make some indication of which contracts we would like to support for L2 phase 1 (e.g. Exchanger) vs. contracts which we don't need to support for L2 phase 1 (e.g. BinaryOptionMarket)

ajsantander:
> @ajsantander - can you make some indication of which contracts we would like to support for L2 phase 1 (e.g. Exchanger) vs. contracts which we don't need to support for L2 phase 1 (e.g. BinaryOptionMarket, EtherCollateral)

@K-Ho I'd look at `publish/deployed/mainnet-ovm/deployments.json` for anything that's already deployed.

justinjmoses (approved):

PR #1127, opened 2021-03-04
SIP-120: TWAP Exchange Function

sohkai:
See [SIP-120](https://sips.synthetix.io/sips/sip-120).

Exposes a new `Exchanger.exchangeAtomically()` function for L1 that allows for atomic exchanges priced via both Chainlink and DEX-sourced TWAP rates.

Requires SIP-127 in order to be deployed (due to codesize limitations solved with SIP-127).

Todo:

- [x] ~Reduce codesize in `ExchangerWithVirtualSynth`~ (see SIP-127)
- [x] Spec tests
- [x] Exchanger unit tests (~requires investigation into smock mocks not reporting expected calls~)
- [x] Update for Uniswap v3 oracles
- [x] Split overweight contracts for L1/L2
- [x] Move `DexPriceAggregator` configuration to using `FlexibleStorage`
- [x] Add volatility breaker
- [x] Dangling questions (see `TODO`s)
- [x] Finalize deployer and initial configuration (see `TODO`s)


contracts/ExchangeRates.sol
```diff
@@ -340,6 +369,50 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
...
+
+        (systemValue, systemSourceRate, systemDestinationRate) =
+            _effectiveValueAndRates(sourceCurrencyKey, sourceAmount, destinationCurrencyKey);
+        uint pClbufValue = systemValue.multiplyDecimal(SafeDecimalMath.unit().sub(getAtomicPriceBuffer()));
+
+        // Normalize decimals in case equivalent asset uses different decimals from internal unit
+        // TODO: prefer SafeDecimalMath.multiplyDecimal() and SafeDecimalMath.divideDecimal() instead?
+        uint sourceAmountInEquivalent = sourceAmount * 10 ** uint(sourceEquivalent.decimals()) / SafeDecimalMath.unit();
+        // TODO: add sanity check here to make sure the price window isn't 0?
+        IDexTwapAggregator.QuoteParams memory dexTwapQuote = dexTwapAggregator.assetToAsset(
```
sohkai, line 403: This seems to be the problem for native solc (throwing an error pointing to https://github.com/ethereum/solidity/blob/v0.5.16/libsolidity/codegen/CompilerUtils.cpp#L1367).

The custom `publish build` script using solcjs seems to compile this fine, and I'm not sure exactly what's up with the returned struct that's causing native solc to throw.

contracts/Exchanger.sol
```diff
@@ -597,6 +631,122 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
+
+        // Update and check against per-block volume limit
+        /*
+        uint currentVolume = uint(lastAtomicVolume.time) == block.timestamp
+            ? uint(lastAtomicVolume.volume).add(sourceSusdValue)
+            : sourceSusdValue;
+        require(currentVolume <= getAtomicMaxVolumePerBlock(), "Surpassed per-block atomic exchange volume limit");
+        lastAtomicVolume.time = uint64(block.timestamp);
+        lastAtomicVolume.volume = uint192(currentVolume); // Protected by volume limit check above
+        */
```
sohkai, line 702: The couple of stack items needed for the block limit update above are causing stack too deeps.

contracts/Exchanger.sol
```diff
@@ -597,6 +631,122 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
+            _suspendIfRateInvalid(sourceCurrencyKey, systemSourceRate) ||
+            _suspendIfRateInvalid(destinationCurrencyKey, systemDestinationRate)
+        ) {
+            return (0, 0);
+        }
+
+        // Sanity check observed output amount against current system value
+        // TODO: maybe this shouldn't revert?  Is the (0, 0) return enough to let a trader know
+        // their exchange inputs were incorrect?
+        require(!_isDeviationAboveThreshold(systemAmountReceived, amountReceived), "Atomic rate too low");
```
sohkai, line 681: This is a deviation from the SIP but I thought the extra sanity check here to disqualify extremely bad trades (via bad rates, manipulation, etc) could be useful.

contracts/ExchangeRates.sol
sohkai: Fixed with 6c599d1; it was just an issue with a subclass also needing to use ABIEncoderV2

contracts/Exchanger.sol
sohkai: Fixed with eb434bc; encapsulating it into its own function with a internal JUMP seems to let the compiler better know that some stack vars can be let go

contracts/ExchangeRates.sol
```diff
@@ -17,6 +20,28 @@ import "@chainlink/contracts-0.0.10/src/v0.5/interfaces/FlagsInterface.sol";
 import "./interfaces/IExchanger.sol";
 
 
+// TODO: where best to put this?
+interface IDexTwapAggregator {
```
justinjmoses, line 24: just another file in interfaces I reckon

contracts/ExchangeRates.sol
```diff
@@ -340,6 +386,54 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
...
+        require(address(destEquivalent) != address(0), "No atomic equivalent for dest");
+
+        (systemValue, systemSourceRate, systemDestinationRate) = _effectiveValueAndRates(
+            sourceCurrencyKey,
+            sourceAmount,
+            destinationCurrencyKey
+        );
+        // Derive P_CLBUF from highest configured buffer between source and destination synth
+        uint priceBuffer = Math.max(getAtomicPriceBuffer(sourceCurrencyKey), getAtomicPriceBuffer(destinationCurrencyKey));
+        uint pClbufValue = systemValue.multiplyDecimal(SafeDecimalMath.unit().sub(priceBuffer));
```
sohkai, line 419: Based on discussions with @kaleb-keny, we decided to allow this chainlink price buffer to be set per-synth rather than as a global setting. Oracle performance and market conditions may dictate different values for each synth in the future.

Note that not having a value set for a synth defaults the buffer to 0bps (no adjustment).

In theory this parameter would be best configured on each possible exchange pair (e.g. sBTC:sUSD, sETH:sUSD, sBTC:sETH, etc), but:

- Configuring each possible exchange pair would be onerous
- I didn't want to constrain this part of the code to assume USD would always be in an atomic exchange pair (so far, only `Exchanger` enforces this constraint)
- Taking the maximum buffer between the source/destination synths is probably OK for most situations; to be absolutely safe, we could instead add the two buffers together (currently an exchange between sBTC:sETH requires two "hops" between BTC:USD and USD:ETH rates).

contracts/Exchanger.sol
```diff
@@ -808,25 +972,47 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
+        internal
+        view
+        returns (uint)
+    {
+        // Get the exchange fee rate as per destination currencyKey
+        uint baseRate = getAtomicExchangeFeeRate(destinationCurrencyKey);
+        if (baseRate == 0) {
+            // If no atomic rate was set, fallback to the regular exchange rate
+            baseRate = getExchangeFeeRate(destinationCurrencyKey);
         }
```
sohkai, line 999: From @kaleb-keny's backtesting efforts, we found that exchange fees played a large part in protecting the debt pool from CEX-based momentum trading strategies. As market conditions—and the required amount of protection—may evolve over time, we decided to allow atomic exchanges to be configured with a different fee model from normal exchanges.

Falling back to the regular exchange fee rate costs a bit more gas, but my thinking/hope here is that most synths won't need a different value.

contracts/Exchanger.sol
```diff
@@ -808,25 +972,47 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
+    }
+
+    function _calculateFeeRateFromExchangeSynths(
+        uint exchangeFeeRate,
+        bytes32 sourceCurrencyKey,
+        bytes32 destinationCurrencyKey
+    ) internal pure returns (uint) {
         // Is this a swing trade? long to short or short to long skipping sUSD.
         if (
             (sourceCurrencyKey[0] == 0x73 && destinationCurrencyKey[0] == 0x69) ||
```
sohkai, line 1011: Note: I took out the initial check for src/dest being sUSD to simplify this function. It might be a worth putting back in as an optimization if most exchanges involve sUSD.

sohkai: Nevermind, I see why that check was needed now 🤦‍♂️

kaleb-keny: I believe it's fine to keep the simplification 😅 , `triple undo`, the reason being is that is that iSynth are in deperecation stage and are unlikely to be included in atomic transactions... This can be confirmed by @justinjmoses thought

kaleb-keny:
Hey @sohkai one question please... Does the atomic contract check if the synth being exchanged has a circuit breaker triggered on it before proceeding? I know it's an edge case, but it would preferable to ensure that the synth is indeed tradeable on the exchange or not... before proceeding with the exchange.

sohkai:
@kaleb-keny 

> Does the atomic contract check if the synth being exchanged has a circuit breaker triggered on it before proceeding

Yes; the current implementation includes two circuit breakers (checked in this order):

1. SIP-65: checks the synth using the current chainlink rate (same as in the normal exchange flow)
2. Sanity check the received atomic rate is not too far off from the current chainlink rate, protecting cases where TWAP/DEX spot pricing is way off (uses the same threshold as the SIP-65 configuration)

kaleb-keny:
> @kaleb-keny
> 
> > Does the atomic contract check if the synth being exchanged has a circuit breaker triggered on it before proceeding
> 
> Yes; the current implementation includes two circuit breakers (checked in this order):
> 
> 1. SIP-65: checks the synth using the current chainlink rate (same as in the normal exchange flow)
> 2. Sanity check the received atomic rate is not too far off from the current chainlink rate, protecting cases where TWAP/DEX spot pricing is way off (uses the same threshold as the SIP-65 configuration)

ohh alright yeah, I was referring to SIP-65, perfect, thank you 🙏🏼

test/contracts/Exchanger.unit.js
```diff
@@ -0,0 +1,437 @@
...
+												{
+													setting: setAsOverrideRate ? 'atomicExchangeFeeRate' : 'exchangeFeeRate',
+													synth: sETH,
+													value: exchangeFeeRate,
+												},
+												() => {
+													beforeEach('attempt exchange', async () => {
+														await this.instance.exchangeAtomically(...defaultExchangeArgs);
+													});
+													// TODO: smock doesn't seem to be reporting that the mocked methods are called
```
sohkai, line 399: @justinjmoses I'm having an issue here with smock, where the mocked version of some contracts (sUSD, sETH, etc) should be having some functions called (e.g. `issue()` and `burn()`), but it's not being reported on `smock.<function>.calls`.

sohkai: It turns out each time I tried to mock a system settings value, it overrode the previous settings rather than compounding them.

A rudimentary solution to address this is in 8b15ea7.

contracts/Exchanger.sol
```diff
@@ -507,6 +552,126 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
+            // Remit the fee in sUSDs
+            issuer().synths(sUSD).issue(feePool().FEE_ADDRESS(), fee);
+
+            // Tell the fee pool about this
+            feePool().recordFeePaid(fee);
+        }
+
+        // Note: As of this point, `fee` is denominated in sUSD.
+
+        // TODO: sanity checking, does this statement still hold true for atomic exchanges since the execution rate may differ from the system's current rate?
```
sohkai, line 643: Wanted to explicitly call out this update to the debt cache for atomic exchanges and sanity check the accounting.

Since the atomic execution rates will be different than the system's, the actual amounts of debt changed will be non-zero in the system's eyes.

Taking an extreme example:

*Trading sUSD -> sETH*:
- Atomic rate: 1/50 (50 sUSD = 1 ETH)
- System rate: 1/100
- Input: 100 sUSD
- Atomic output: 2 sETH
- System expected output: 1 sETH

So now, from the system's eyes, the debt value should not be the same as before. The debt pool contains 1 sETH more in value than before.

contracts/Exchanger.sol
```diff
@@ -395,6 +439,7 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
         sourceAmountAfterSettlement = sourceAmount;
 
         // when settlement was required
+        // TODO: are refunds immediately added to the current exchange?
```
sohkai, line 442: Sanity checking my understanding of the behaviour, if you had a pending refund from fee reclamation, this amount would be silently added to what you inputed as `sourceAmount` rather minted separately?

E.g. I want to exchange 100 sUSD for sETH, with a 10 sUSD refund pending. My actual exchange would then consist of inputting 110 sUSD?

contracts/Exchanger.sol
```diff
@@ -507,6 +552,126 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
+        // Sanity check atomic output's value against current system value (checking atomic rates)
+        require(
+            !_isDeviationAboveThreshold(systemConvertedAmount, amountReceived.add(fee)),
+            "Atomic rate deviates too much"
+        );
+
+        // Ensure src/dest synth is sUSD and determine sUSD value of exchange
+        uint sourceSusdValue;
+        if (sourceCurrencyKey == sUSD) {
+            sourceSusdValue = sourceAmountAfterSettlement;
```
sohkai, line 604: Similar to above, this line was previously `sourceSusdValue = sourceAmount` but I've changed it to reflect the amount after fee reclamation. Should be more correct to use the actually burned sUSD amount to track volume than the input value provided by the user.

contracts/Exchanger.sol
```diff
@@ -340,17 +351,50 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
+            _emitTrackingEvent(trackingCode, destinationCurrencyKey, amountReceived, fee);
         }
+    }
+
+    function exchangeAtomically(
+        address from,
+        bytes32 sourceCurrencyKey,
+        uint sourceAmount,
+        bytes32 destinationCurrencyKey,
+        // TODO: this destinationAddress variable doesn't seem to be needed since it's always the same as from
```
sohkai, line 366: `exchangeAtomically()`'s API mostly follows from the previous `exchange*()` APIs called by `Synthetix`.

Do we need to make any adjustments (or simplifications) with the new refactor? For example, this `destinationAddress` argument is always filled in with the same value as `from` since there's no `exchangeAtomicallyOnBehalfOf()` or `exchangeAtomicallyWithTrackingForInitiator`.

contracts/Exchanger.sol
```diff
@@ -340,17 +351,50 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
-                trackingCode,
-                destinationCurrencyKey,
-                amountReceived,
-                fee
-            );
+            _emitTrackingEvent(trackingCode, destinationCurrencyKey, amountReceived, fee);
+        }
+    }
+
+    function _emitTrackingEvent(
```
sohkai, line 386: Note that I've opted to reverse #1300; those internal functions are used by `exchangeAtomically()` 😄

contracts/ExchangeRates.sol
```diff
@@ -339,6 +363,56 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
...
+            uint systemDestinationRate
+        )
+    {
+        IERC20 sourceEquivalent = IERC20(getAtomicEquivalentForDexPricing(sourceCurrencyKey));
+        require(address(sourceEquivalent) != address(0), "No atomic equivalent for src");
+
+        IERC20 destEquivalent = IERC20(getAtomicEquivalentForDexPricing(destinationCurrencyKey));
+        require(address(destEquivalent) != address(0), "No atomic equivalent for dest");
+
+        // TODO: this may return 0s if the CL aggregator reverts on latestRoundData()--should it revert?
```
sohkai, line 389: At the moment, nothing bad happens if 0s are returned for these "system" rates; the circuit breaker will trip and trades will be cancelled. This mimics how the normal exchange function works.

Judging by how this contract queries CL aggregators, this choice to return zeros seems explicit but I wanted to sanity check this.

contracts/ExchangeRates.sol
```diff
@@ -339,6 +363,56 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
...
+            sourceAmount,
+            destinationCurrencyKey
+        );
+        // Derive P_CLBUF from highest configured buffer between source and destination synth
+        uint priceBuffer = Math.max(getAtomicPriceBuffer(sourceCurrencyKey), getAtomicPriceBuffer(destinationCurrencyKey));
+        uint pClbufValue = systemValue.multiplyDecimal(SafeDecimalMath.unit().sub(priceBuffer));
+
+        // Normalize decimals in case equivalent asset uses different decimals from internal unit
+        uint sourceAmountInEquivalent = (sourceAmount * 10**uint(sourceEquivalent.decimals())) / SafeDecimalMath.unit();
+        // TODO: add sanity check here to make sure the price window isn't 0?
```
sohkai, line 401: This could happen if the twap price window setting wasn't initialized (system settings requires > 0 on updates), and a window of 0 will cause a revert in the oracle call (the univ3 twap oracle requires window > 0).

So it's not likely anything bad will happen, but would we prefer a strong guarantee through an explicit `require()`?

contracts/BaseSynthetix.sol
```diff
@@ -389,6 +398,7 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
     }
 
     // ========== EVENTS ==========
+    // TODO: another event for atomic exchanges?
```
sohkai, line 401: Sanity checking, would it be helpful to demarcate atomic exchanges from normal exchanges with a new event?

index.js
```diff
@@ -204,6 +205,24 @@ const defaults = {
 	ETHER_WRAPPER_MAX_ETH: w3utils.toWei('5000'),
 	ETHER_WRAPPER_MINT_FEE_RATE: w3utils.toWei('0.02'), // 200 bps
 	ETHER_WRAPPER_BURN_FEE_RATE: w3utils.toWei('0.0005'), // 5 bps
+
+	/***********************************
+	 * TODO: spec atomic configuration *
```
sohkai, line 210: I've stubbed these configuration values for now; wasn't sure how much should be specified here

publish/src/commands/deploy.js
```diff
@@ -1182,6 +1182,8 @@ const deploy = async ({
 				writeArg: [currencyKeyInBytes, feed],
 			});
 		}
+
+		// TODO: set up DEX TWAP oracle for exchange rates
```
sohkai, line 1186: Similar to above, I've assumed this is the correct place to set the configuration and left stubs

test/contracts/helpers.js
```diff
@@ -281,6 +284,50 @@ module.exports = {
 		return { mocks, resolver };
 	},
 
+	prepareFlexibleStorageSmock(flexibleStorage) {
```
sohkai, line 287: From previous confusion in https://github.com/Synthetixio/synthetix/pull/1127#discussion_r609717040, I made this utility to allow system settings / flexible storage mocks to be compounded over a test.

It's a bit of magic, but I think it's more intuitive than the normal behaviour :).

Alternative to this, we could deploy a real instance of `FlexibleStorage` for any unit tests requiring system settings mocks.

package.json
```diff
@@ -78,9 +78,10 @@
 		"@codechecks/client": "0.1.10",
 		"@eth-optimism/contracts": "0.2.10",
 		"@eth-optimism/plugins": "0.0.16",
-		"@eth-optimism/smock": "0.2.1-alpha.0",
+		"@eth-optimism/smock": "~1.0.2",
```
sohkai, line 81: The previous version of smock threw errors when trying to access the `.calls` property of a mock that wasn't yet called (e.g. when checking `calls.length === 0`).

`smock@1.0.2` is the latest version of smock that works with `hardhat@2.1.x` (later versions require updating to `hardhat@2.2.x`).

package.json
```diff
@@ -78,9 +78,10 @@
...
 		"@eth-optimism/solc": "0.5.16-alpha.7",
 		"@eth-optimism/watcher": "0.0.1-alpha.8",
+		"@nomiclabs/hardhat-ethers": "^2.0.2",
```
sohkai, line 84: This is not strictly necessary, but it hasn't conflicted with the `hardhat-truffle` plugin and makes it easier to use smock.

The newer versions of smock assume an ethers provider is made available through the hardhat environment (smock internally uses ethers). Technically we could wrap the web3 provider from truffle into an ethers provider, but we'd have to do this at each invocation of smock. So it seemed easier to use the `hardhat-ethers` plugin to provide the ethers provider.

sohkai (commented):
@justinjmoses I've left a couple comments to highlight areas worth looking at or explaining further. They're ordered by priority; the debt cache is my only implementation concern outside of splitting the contracts for L2.

publish/releases.json
```diff
@@ -243,8 +243,8 @@
 			"major": 2,
 			"minor": 46
 		},
-		"sources": ["Exchanger", "Synthetix"],
-		"sips": [138, 139, 140]
+		"sources": ["ExchangeRates", "Exchanger", "Synthetix", "SystemSettings"],
+		"sips": [120, 138, 139, 140]
```
sohkai, line 247: I've added to the next release to get the prod tests running again

package.json
sohkai: This was upgraded separately in #1308.

package.json
sohkai: This was added separately in #1308.

contracts/MixinSystemSettingsL1.sol
```diff
@@ -0,0 +1,48 @@
+pragma solidity ^0.5.16;
+
+// Inheritance
+import "./MixinSystemSettings.sol";
+
+// https://docs.synthetix.io/contracts/source/contracts/mixinsystemsettingsl1
+contract MixinSystemSettingsL1 is MixinSystemSettings {
```
sohkai, line 7: The commit that splits `SystemSettings` into two contracts is d864225.

TBH the multi-inheritance diamond from `MixinSystemSettings` and `MixinSystemSettingsL1` is pretty ugly due to the constructor hack (see l14 below).

test/contracts/setup.js
```diff
@@ -80,7 +80,7 @@ const mockGenericContractFnc = async ({ instance, fncName, mock, returns = [] })
 	const abiEntryForFnc = artifacts.require(mock).abi.find(({ name }) => name === fncName);
```
sohkai, line 80: To cope with a number of new contracts that are polymorph'ed between multiple versions, I've refactored how this setup script works.

Contracts are now specified via their contract name (i.e. solidity name), and include a `resolverAlias` to map them to their `AddressResolver` name. The list of a contract's `deps` is listed based on the `AddressResolver` name.

This lets users more concretely specify a particular version of a contract to use without any clashes with dependents, best seen in the `Exchanger`'s spec tests.

test/contracts/setup.js
```diff
@@ -101,7 +101,6 @@ const mockGenericContractFnc = async ({ instance, fncName, mock, returns = [] })
 const setupContract = async ({
 	accounts,
 	contract,
-	source = undefined, // if a separate source file should be used
```
sohkai, line 102: `source` has been been redundant as `contract` is now the artifact name

test/contracts/setup.js
```diff
@@ -824,6 +870,31 @@ const setupAllContracts = async ({
 		},
 	];
 
+	// check contract list for contracts with the same address resolver name
+	const checkConflictsInDeclaredContracts = ({ contractList }) => {
```
sohkai, line 874: This is just a sanity check in case a user directly specifies two contracts whose resolver aliases conflict with each other, e.g. `setupAllContracts({ contracts: ['Exchanger', 'ExchangerWithVirtualSynth'] })`.

I don't expect it to be tripped too often, but the odd case it does will likely be very helpful as silently overwriting an `AddressResolver` entry is pretty painful to debug.

test/contracts/Exchanger.spec.js
```diff
@@ -3922,7 +3962,7 @@ contract('Exchanger (spec tests)', async accounts => {
 		});
 	};
 
-	describe('When using Synthetix', () => {
+	describe('With L1 configuration (Synthetix, ExchangerWithVirtualSynth, ExchangeRatesWithDexPricing, SystemSettingsL1)', () => {
```
sohkai, line 3965: I've followed this testing pattern for polymorph'ed contracts that have needed to be split so they could share most of the same tests.

contracts/ExchangeRates.sol
```diff
@@ -260,16 +255,20 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
         return getAggregatorWarningFlags();
     }
 
+    // TODO: should these be included in IExchangeRates?
```
sohkai, line 258: In general I'm not sure about if these getters 1. should exist and 2. should be added to the interface, especially with the l1/l2 split.

Is there a particular rule or heuristic to follow?

contracts/ExchangerWithVirtualSynth.sol
```diff
@@ -6,10 +6,20 @@ import "./Exchanger.sol";
 // Internal references
```
sohkai, line 6: We can find a better name for `ExchangerWithVirtualSynth`

sohkai (commented):
The last few commits (starting from 2dfeff9) have split `Exchanger`, `ExchangeRates`, and `SystemSettings` into two variants each, as the atomic functionality will only be existing on L1.

The comments from the previous review still hold.

contracts/BaseSynthetix.sol
justinjmoses: perhaps an additional event that denotes that it is atomic?

contracts/ExchangeRates.sol
justinjmoses: If it's required in the base IExchangeRates interface then worth using these empty functions. I think the getters can be ignored but the atomic rate function can be put in ExchangeRates as not implemented as it'll be in the interface

contracts/ExchangeRatesWithoutInvPricing.sol
```diff
@@ -43,8 +43,4 @@ contract ExchangeRatesWithoutInvPricing is ExchangeRates {
     function _rateIsFrozen(bytes32) internal view returns (bool) {
         return false;
     }
-
```
justinjmoses, line 46: I don't think this file is required right?

contracts/Exchanger.sol
justinjmoses: Correct.

contracts/Exchanger.sol
justinjmoses: Note: 1300 got Exchanger under size for the OVM (contract size and also gas deployment under 9m gas), so we may need to move this logic to ExchangerWithVirtualSynth renaming it to ExchangerWithFeeReclamationAlternatives ?

contracts/ExchangerWithVirtualSynth.sol
justinjmoses: See above

contracts/Exchanger.sol
```diff
@@ -722,14 +742,22 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
         exchangeFeeRate = _feeRateForExchange(sourceCurrencyKey, destinationCurrencyKey);
     }
 
-    function _feeRateForExchange(bytes32 sourceCurrencyKey, bytes32 destinationCurrencyKey)
-        internal
-        view
-        returns (uint exchangeFeeRate)
-    {
+    function feeRateForAtomicExchange(bytes32, bytes32) external view returns (uint exchangeFeeRate) {
```
justinjmoses, line 745: Again if this isn't needed in the interface (and I don't think it really is) then we can remove it from this base class

contracts/Exchanger.sol
```diff
@@ -788,11 +816,28 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
             destinationCurrencyKey
         );
         exchangeFeeRate = _feeRateForExchange(sourceCurrencyKey, destinationCurrencyKey);
-        amountReceived = _getAmountReceivedForExchange(destinationAmount, exchangeFeeRate);
+        amountReceived = _deductFeesFromAmount(destinationAmount, exchangeFeeRate);
         fee = destinationAmount.sub(amountReceived);
     }
 
-    function _getAmountReceivedForExchange(uint destinationAmount, uint exchangeFeeRate)
+    function getAmountsForAtomicExchange(
```
justinjmoses, line 823: Ibid (as above)

contracts/MixinSystemSettingsL1.sol
justinjmoses: Do you think it's really necessary? Can't we just have system settings on both layers even if they aren't relevant? We can just keep them as 0 on the unneeded layer. It will add a bit of contract size to Exchanger but I think after my refactor we will have enough space

publish/src/commands/deploy/configure-standalone-price-feeds.js
```diff
@@ -25,4 +25,18 @@ module.exports = async ({ deployer, runStep, standaloneFeeds }) => {
 			});
 		}
 	}
+
+	if (!useOvm && ExchangeRates) {
+		/*
+			TODO: set up DEX TWAP oracle for exchange rates
```
justinjmoses, line 31: Yeah I like this. I guess the address for the twAp pricing goes in the getUsers function at the top level index filled, as there is one pet environment.

publish/src/commands/deploy/configure-system-settings.js
```diff
@@ -283,4 +283,24 @@ module.exports = async ({
...
 		});
 	}
+
+	if (!useOvm) {
+		// TODO: finish configuring new atomic exchange system settings
+		await runStep({
+			contract: 'SystemSettings',
+			target: SystemSettings,
+			read: 'atomicMaxVolumePerBlock',
+			expected: input => input !== '0', // only change if zero
```
justinjmoses, line 293: We may need to skip this if the value to set is actually 0. We do this with waitingPeriodSecs on L2 for example.

test/contracts/setup.js
```diff
@@ -152,6 +153,13 @@ const setupContract = async ({
 			[toBytes32('SNX')],
 			[toWei('0.2', 'ether')],
 		],
+		ExchangeRatesWithDexPricing: [
+			owner,
+			oracle,
+			tryGetAddressOf('AddressResolver'),
+			[toBytes32('SNX')], // TODO: can we remove these initial rates?
```
justinjmoses, line 160: Yeah for this version that should be fine / hopefully all the tests so priced before testing anything, but it should be pretty obvious from CI once you remove it.

contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -0,0 +1,95 @@
...
+            uint systemDestinationRate
+        )
+    {
+        IERC20 sourceEquivalent = IERC20(getAtomicEquivalentForDexPricing(sourceCurrencyKey));
+        require(address(sourceEquivalent) != address(0), "No atomic equivalent for src");
+
+        IERC20 destEquivalent = IERC20(getAtomicEquivalentForDexPricing(destinationCurrencyKey));
+        require(address(destEquivalent) != address(0), "No atomic equivalent for dest");
+
+        // TODO: this may return 0s if the CL aggregator reverts on latestRoundData()--should it revert?
```
sohkai, line 65: At the moment, nothing bad happens if 0s are returned for these "system" rates; the circuit breaker will trip and trades will be cancelled. This mimics how the normal exchange function works.

Judging by how this contract queries CL aggregators, this choice to return zeros seems explicit but I wanted to sanity check this.

contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -0,0 +1,95 @@
...
+            sourceAmount,
+            destinationCurrencyKey
+        );
+        // Derive P_CLBUF from highest configured buffer between source and destination synth
+        uint priceBuffer = Math.max(getAtomicPriceBuffer(sourceCurrencyKey), getAtomicPriceBuffer(destinationCurrencyKey));
+        uint pClbufValue = systemValue.multiplyDecimal(SafeDecimalMath.unit().sub(priceBuffer));
+
+        // Normalize decimals in case equivalent asset uses different decimals from internal unit
+        uint sourceAmountInEquivalent = (sourceAmount * 10**uint(sourceEquivalent.decimals())) / SafeDecimalMath.unit();
+        // TODO: add sanity check here to make sure the price window isn't 0?
```
sohkai, line 77: This could happen if the twap price window setting wasn't initialized (system settings requires > 0 on updates), and a window of 0 will cause a revert in the oracle call (the univ3 twap oracle requires window > 0).

So it's not likely anything bad will happen, but would we prefer a strong guarantee through an explicit `require()`?

contracts/ExchangeRatesWithoutInvPricing.sol
sohkai: Ah, this change was to move the `_notImplemented()` internal function up to the base `ExchangeRates` contract

sohkai (commented):
@justinjmoses I've push commits addressing the prior comments (see new commits: https://github.com/Synthetixio/synthetix/pull/1127/files/c0271f2..72f682d). I think some of the original comments related to TODOs were pushed below the fold and might've been missed, so I re-created them in this thread for visibility.

For visibility, there were just two comments I didn't mark as resolved yet and left comments in those threads:

- Moving `_emitTracking()` out of `Exchanger.sol`: https://github.com/Synthetixio/synthetix/pull/1127#discussion_r653123287
- A new event for tracking atomic exchanges: https://github.com/Synthetixio/synthetix/pull/1127#discussion_r653121141

contracts/MixinSystemSettingsL1.sol
sohkai: Agreed; the multiple inheritance structure was more unwieldy than it helped. It was reverted in 6c3c960.

contracts/ExchangeRates.sol
sohkai: Done in 42a004a and 5f80eed

contracts/ExchangerWithVirtualSynth.sol
sohkai: Done in 0039dc8

contracts/Exchanger.sol
sohkai: The debt pool contract is meant to be deprecated soon and we've decided this inaccuracy is acceptable. I've added a note in 6bf88f2.

contracts/Exchanger.sol
sohkai: Add a note in 5ba2cca to clarify this.

publish/releases.json
sohkai: Since 2.46 was already released, I've stubbed out a 2.47 release in cdcbc24.

contracts/ExchangerWithFeeReclamationAlternatives.sol
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    /* ========== MUTATIVE FUNCTIONS ========== */
+
+    function exchangeAtomically(
+        address from,
+        bytes32 sourceCurrencyKey,
+        uint sourceAmount,
+        bytes32 destinationCurrencyKey,
+        // TODO: this destinationAddress variable doesn't seem to be needed since it's always the same as from
```
sohkai, line 76: `exchangeAtomically()`'s API mostly follows from the previous `exchange*()` APIs called by `Synthetix`.

Do we need to make any adjustments (or simplifications) with the new refactor? For example, this `destinationAddress` argument is always filled in with the same value as `from` since there's no `exchangeAtomicallyOnBehalfOf()` or `exchangeAtomicallyWithTrackingForInitiator`.

contracts/Exchanger.sol
sohkai: I didn't understand this comment; the base contract will need to use `_emitTrackingEvent()` as well. Was there another part of the functionality you meant by this?

contracts/BaseSynthetix.sol
sohkai: Wasn't sure what event format would be best. I assume we wouldn't want to change the existing event's signature by adding a new parameter.

So that leaves two options (maybe more?):

- Also emit a new event just saying an atomic exchange happened: `event AtomicSynthExchange()`
  - Difficult to group with the `SynthExchange` event due to lack of a canonical id for each synth exchange
- Only emit a new event with the same parameters as `SynthExchange` (and don't emit `SynthExchange`): `event AtomicSynthExchange(address,bytes32,uint256,bytes32,uint256,address)`
  - This makes it more difficult to catch all synth exchanges executed by the system

Thoughts?

contracts/Exchanger.sol
sohkai: Done in 42a004a

contracts/Exchanger.sol
sohkai: Done in 42a004a

publish/src/commands/deploy/configure-standalone-price-feeds.js
sohkai: Actually I've just added it in the default args section as that also allows for separate networks: 431b9c0

publish/src/commands/deploy/configure-system-settings.js
sohkai: Done in 9f1c537

test/contracts/setup.js
sohkai: I've opted to keep this config for now to keep the non-exchange tests as-is. If it were perfect, I think we would avoid adding this rate to the default configuration (specifying it only when needed) and include a new `ExchangeRates` constructor test with these initial rates. But that sounds like it could be part of the larger `ExchangeRates` refactor you've been having in mind.

contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -92,6 +100,57 @@ contract ExchangeRatesWithDexPricing is ExchangeRates {
...
+            if (updateThreshold == 0) {
+                // Oracle updates reached allowed threshold in consideration window
+                // Consider the asset volatile
+                return true;
+            }
+
+            // TODO: this assumes chainlink may be unreliable in reporting rates, but **must** be
+            // reliable in reporting times (always incrementing, roughly similar to reported
+            // block.timestamps). Is this a good assumption?
+            (uint rate, uint time) = _getRateAndTimestampAtRound(currencyKey, roundId);
```
sohkai, line 131: I'm pretty sure we can trust chainlink's reported times (will be easy to check from the BigQuery results), but just in case, I wanted to point this out.

I'm making two assumptions here:

1. An aggregator's entire round or a rate in a round may be unavailable (resulting in either a revert in the call to `getRoundData()` or a 0 rate), and
2. If a round is available, its reported `updateTime` is sane: always sequential (increasing in value) and always `<= block.timestamp`.

1 is handled by a case (`rate == 0 || time == 0`), but 2. is not considered—if 2. is violated (i.e. the times are not sequential or are in the future), this loop may behave funny.

sohkai (commented):
@justinjmoses I've pushed a few more commits. Sequentially,

- 43d820f and ae49dbb: cosmetic rewordings and renamings of concepts; will be updated in the SIP as well
- 1aa47be: moves the `dexPriceOracle` setting to `FlexibleStorage`, previously stored in `ExchangeRates`'s own storage
- 5368328 and 7d4e868: adds the volatility breaker, based on an oracle update threshold

contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -21,12 +19,17 @@ contract ExchangeRatesWithDexPricing is ExchangeRates {
...
-        dexPriceAggregator = _dexPriceAggregator;
+        flexibleStorage().setAddressValue(CONTRACT_NAME, SETTING_DEX_PRICE_AGGREGATOR, address(_dexPriceAggregator));
         emit DexPriceAggregatorUpdated(address(_dexPriceAggregator));
     }
 
     /* ========== VIEWS ========== */
 
+    function dexPriceAggregator() public view returns (IDexPriceAggregator) {
+        // TODO: this fetch could be made cheaper with an internal cache that gets wiped on `setDexPriceAggregator()` calls
+        return IDexPriceAggregator(flexibleStorage().getAddressValue(CONTRACT_NAME, SETTING_DEX_PRICE_AGGREGATOR));
```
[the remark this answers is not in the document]
justinjmoses: Agreed - let's keep it simple for now.

contracts/ExchangeRatesWithDexPricing.sol
justinjmoses: Yes I think an extra require for the price window is a good call

contracts/ExchangeRatesWithDexPricing.sol
justinjmoses: 1. Any potential revert is handled in `_getRateAndTimestampAtRound` so testing for `(rate == 0 || time == 0)` is perfect. 
2. Yes, I've made that same assumption in `getLastRoundIdBeforeElapsedSecs()`

contracts/ExchangerWithFeeReclamationAlternatives.sol
```diff
@@ -123,6 +123,12 @@ contract ExchangerWithFeeReclamationAlternatives is MinimalProxyFactory, Exchang
         address destinationAddress
     ) internal returns (uint amountReceived, uint fee) {
         _ensureCanExchange(sourceCurrencyKey, sourceAmount, destinationCurrencyKey);
+        require(
```
justinjmoses, line 126: what about ETH to BTC? Isn't that a possibility that needs to check both rates?

publish/releases.json
```diff
@@ -304,5 +299,15 @@
 			"SynthsETH"
 		],
 		"sips": [142]
+	},
+	{
```
justinjmoses, line 303: This PR should help with this - https://github.com/Synthetixio/synthetix/pull/1464 - hopefully it's merged soon and you can bring it into your PR via develop, and then please put this work into *Sargas*

contracts/ExchangerWithFeeReclamationAlternatives.sol
sohkai: Ah, at the moment we're enforcing that either the source or destination synth is sUSD (it's checked later down as an small gas optimization with accounting for the per-block exchange volume).

TBH at the moment an extra call to explicitly check both source/destination synth would not be too costly (~3-5k gas) and could be safer if the sUSD requirement is ever upgraded out? Or I can leave a comment explaining this.

contracts/BaseSynthetix.sol
```diff
@@ -390,6 +399,7 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
```
sohkai, line 402: Re-surfacing an older comment that's been lost in the collapse (https://github.com/Synthetixio/synthetix/pull/1127#discussion_r653121141):

> Q: Would it be helpful to demarcate atomic exchanges from normal exchanges with a new event?

> A: perhaps an additional event that denotes that it is atomic?

Wasn't sure what event format would be best. I assume we wouldn't want to change the existing event's signature by adding a new parameter.

So that leaves two options (maybe more?):

- Also emit a new event just saying an atomic exchange happened: `event AtomicSynthExchange()`
  - Difficult to group with the `SynthExchange` event due to lack of a canonical id for each synth exchange
- Only emit a new event with the same parameters as `SynthExchange` (and don't emit `SynthExchange`): `event AtomicSynthExchange(address,bytes32,uint256,bytes32,uint256,address)`
  - This makes it more difficult to catch all synth exchanges executed by the system

Thoughts?

------

I've noticed a some other events having their signatures changed in recent past SIPs though; is the frontend/subgraph/etc. infrastructure set up to easily consume new event signatures?

contracts/BaseSynthetix.sol
sohkai: Thread replaced with https://github.com/Synthetixio/synthetix/pull/1127/files#r695201309.

publish/src/commands/deploy/configure-standalone-price-feeds.js
```diff
@@ -28,16 +28,15 @@ module.exports = async ({ deployer, runStep, standaloneFeeds, useOvm }) => {
...
-			});
-			*/
+		/* TODO (SIP-120): set up DEX price oracle for exchange rates
+		await runStep({
+			contract: `ExchangeRates`,
+			target: ExchangeRates,
+			read: 'dexPriceAggregator',
+			expected: input => input === dexPriceAggregator,
+			write: 'setDexPriceAggregator',
+			writeArg: [dexPriceAggregator],
```
sohkai, line 38: @justinjmoses Wasn't sure where to add this as a parameter; `publish/deployed/mainnet/params.json` seems to be empty so I'm not sure how `getDeployParameter()` works.

sohkai:
Updated for the latest production oracle (see https://github.com/sohkai/uniswap-v3-spot-twap-oracle/pull/1).

contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -0,0 +1,150 @@
...
+        external
+        view
+        returns (
+            uint value,
+            uint systemValue,
+            uint systemSourceRate,
+            uint systemDestinationRate
+        )
+    {
+        IERC20 sourceEquivalent = IERC20(getAtomicEquivalentForDexPricing(sourceCurrencyKey));
```
dbeal-eth, line 69: np consider it would be kind of slick to split majority this function into 2 internal functions, leaving the function body like:

```
{
   uint clValue = calculateClRate();
   uint dexValue = calculateTwapRate();
   
   value = clValue < dexValue ? clValue : dexValue; // min
}
```

contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -0,0 +1,150 @@
...
+        require(address(sourceEquivalent) != address(0), "No atomic equivalent for src");
+
+        IERC20 destEquivalent = IERC20(getAtomicEquivalentForDexPricing(destinationCurrencyKey));
```
dbeal-eth, line 72: equivalent seems like misleading terminology here, maybe better to just say `sourceTokenAddress` `destTokenAddress`

contracts/BaseSynthetix.sol
```diff
@@ -409,7 +418,36 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+            addressToBytes32(account),
+            0,
+            0
+        );
+    }
+
+    event AtomicSynthExchange(
+        address indexed account,
+        bytes32 fromCurrencyKey,
+        uint256 fromAmount,
```
dbeal-eth, line 431: since the data is available/easily calculated in the caller, can you also include `fromUSDAmount` and `toUSDAmount` fields? Would seriously be helpful for the subgraphs.

dbeal-eth (approved):

contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -0,0 +1,152 @@
...
+        if (considerationWindow == 0 || updateThreshold == 0) {
+            // If either volatility setting is not set, never judge an asset to be volatile
+            return false;
+        }
+
+        // Go back through the historical oracle update rounds to see if there have been more
+        // updates in the consideration window than the allowed threshold.
+        // If there have, consider the asset volatile--by assumption that many close-by oracle
+        // updates is a good proxy for price volatility.
+        uint considerationWindowStart = block.timestamp.sub(considerationWindow);
```
artdgn, line 125: Is this heuristic too coupled to current CL implementation details? Would be easily broken improvements or differences between L1 & L2, requiring extra configuration overhead. It might be better to have a volatility heuristic that doesn't rely on an implementation quirk, but on the price values themselves.

Volatility itself can be aggregated quite cheaply by aggregating the average price and the deviation from average price (https://www.investopedia.com/terms/v/volatility.asp) as EMA with L1 metrics (to avoid square and sqrt). This would also require just one call per round / per tx on average, and not `O(updateThreshold)` of calls for each tx.

Something like:
```solidity
// system settings
uint decayParam; // moving average decay
uint volatilityTreshold; // threshold for checking volatility

...
// persistent state variables
mapping(bytes32 => uint) lastRoundsSeen;
mapping(bytes32 => uint) priceEMAs; // mean aggregator
mapping(bytes32 => uint) deviationEMAs; // herustic volatility aggregator

...
// in the main method, check the volatility value vs. some threshold after updateing the aggregators
return _updatedVolatilityEMA(currencyKey) > volatilityTreshold;


...
function _updatedVolatilityEMA(currencyKey) internal returns (uint) {
  uint roundId = _getCurrentRoundId(currencyKey);
  if (roundId > lastRoundsSeen[currencyKey]) { // if round seen we can just use the up to date value
    uint lastRoundSeen = lastRoundsSeen[currencyKey];
    uint priceEMA = priceEMAs[currencyKey];
    uint deviationEMA = deviationEMAs[currencyKey];

    // for each round that was not see yet
    for (lastRoundSeen;  lastRoundSeen <= roundId; lastRoundSeen++) {
      (uint price, uint time) = _getRateAndTimestampAtRound(currencyKey, roundId);
      // update mean:  priceEMA = priceEMA * decayParam + (1 - decayParam) * price
      priceEMA = priceEMA.multiplyDecimal(decayParam).add(SafeDecimal.unit().sub(decayParam).multiplyDecimal(price));
       // current deviation: curDeviation = abs(price - priceEMA)
       // we may use abs (L1) difference to avoid squaring and squarerooting because 
       // we just need a value that can be used as threshold for heuristic and 
       // we don't need precise definition of volatility
      uint curDeviation =  price > priceEMA ? price - priceEMA : priceEMA - price;
      deviationEMA = deviationEMA.multiplyDecimal(decayParam).add(SafeDecimal.unit().sub(decayParam).multiplyDecimal(curDeviation));
    }
    lastRoundsSeen[currencyKey] = lastRoundSeen;
    priceEMAs[currencyKey] = priceEMA;
    deviationEMAs[currencyKey] = deviationEMA;
  }
  return deviationEMAs[currencyKey];
}

**Review by artdgn — CHANGES_REQUESTED** [review]
In addition to the questions I've added on specific lines it would also be good to see a diff of gas usage report (by running relevant tests with `--gas` flag) for the relevant external methods affected.

**Comment by artdgn** [comment]
Gas measurement - `exchangeAtomically` is roughly 600K gas in tests.

## After (current version)
```
····································|····································|··············|·············|···················|···············|··············
|  Contract                         ·  Method                            ·  Min         ·  Max        ·  Avg              ·  # calls      ·  usd (avg)  │
....
····································|····································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchange                          ·      164627  ·     766894  ·           505766  ·          392  ·          -  │
····································|····································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchangeAtomically                ·      589640  ·     599906  ·           598563  ·           17  ·          -  │
····································|····································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchangeOnBehalf                  ·      716538  ·     724294  ·           717508  ·            8  ·          -  │
····································|····································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchangeOnBehalfWithTracking      ·      723153  ·     730921  ·           724124  ·            8  ·          -  │
····································|····································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchangeWithTracking              ·      705559  ·     713312  ·           709436  ·            2  ·          -  │
····································|····································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchangeWithVirtual               ·      911751  ·     917927  ·           917602  ·           38  ·          -  │
```

## Before (develop):
```
|  Synthetix                  ·  exchange                          ·      164469  ·     766581  ·           505403  ·          392  ·          -  │
······························|····································|··············|·············|···················|···············|··············
|  Synthetix                  ·  exchangeOnBehalf                  ·      716203  ·     723959  ·           717173  ·            8  ·          -  │
······························|····································|··············|·············|···················|···············|··············
|  Synthetix                  ·  exchangeOnBehalfWithTracking      ·      722796  ·     730564  ·           723767  ·            8  ·          -  │
······························|····································|··············|·············|···················|···············|··············
|  Synthetix                  ·  exchangeWithTracking              ·      705180  ·     712933  ·           709057  ·            2  ·          -  │
······························|····································|··············|·············|···················|···············|··············
|  Synthetix                  ·  exchangeWithVirtual               ·      911438  ·     917570  ·           917247  ·           38  ·          -  │
```

**Comment by artdgn** [comment]
## Gas usage for `exchangeAtomically`: 670K with current params.
The previous test setup was not really using the volatility breaker (because config values were not set). After adding the configuration into the test we get for:

`> npx hardhat test test/contracts/Exchanger* --grep "exchangeAtomically" --gas`

```
····································|··········································|··············|·············|···················|···············|··············
|  Contract                         ·  Method                                  ·  Min         ·  Max        ·  Avg              ·  # calls      ·  usd (avg)  │
····································|··········································|··············|·············|···················|···············|··············
...
····································|··········································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchangeAtomically                      ·      667576  ·     677838  ·           676495  ·           17  ·          -  │
····································|··········································|··············|·············|···················|···············|··············

```

artdgn:
Volatility check gas usage with current parameters (`updateThreshold` = 2) cannot be optimised by removing the loop. However if the parameters used cause a longer loop (e.g. `updateThreshold` = 6, and consideration window of 30 minutes). The gas usage will increase by about 50K. The version without the loop will only increase by around 10K gas.

Assuming, the current parameters aren't expected to change, there probably is no urgency to update the calculation.

artdgn:
I've opened https://github.com/Synthetixio/synthetix/pull/1596 to propose fixes for some of the unaddressed issues in this PR.

Issues in contract changes:
- Not using safemath for price calculations in `ExchangeRatesWithDexPricing`
- Allowing setting dex oracle to zero address in `SystemSettings`
- Not checking dex oracle returned price to at least not be 0 in `effectiveAtomicValueAndRates`
- Comments missing about the the dex oracle `IDexPriceAggregator` impelementation source, reference interface source.

Issues in tests:
- Tests for `exchangeAtomically` not being correctly set up to trigger volatility checks (no configuration for the check which causes the check to be skipped completely). Which makes the test skip an important part of functionality it's testing, and also makes gas measurement inaccurate.
- Setting minutes to 3600 seconds in tests instead of 60 seconds, which makes the test use numbers that are too different from the values used in real world config.
- Using `.sort()` for numeric times incorrectly (using alphabetic sort for numbers) - and so setting up the mock correctly only when minute is 3600 seconds but not 60 seconds.

justinjmoses (approved):

PR #1129, opened 2021-03-04
Run Exchanger tests with both Synthetix and MintableSynthetix

i-stam:
Ths PR refactors the exchanger tests to run both against Syntehtix as well as against MintableSynthetix.
This approach detects cases that code that should be run (and used by Exchanger) on both layers exists only on one of them.


test/contracts/Exchanger.spec.js
```diff
@@ -3637,54 +3649,257 @@ contract('Exchanger (spec tests) @ovm-skip', async accounts => {
...
-				// Read all rates
-				const sAUDRate = await exchanger.feeRateForExchange(empty, sAUD);
-				assert.bnEqual(sAUDRate, newFxBIPS);
-				const sUSDRate = await exchanger.feeRateForExchange(empty, sUSD);
-				assert.bnEqual(sUSDRate, newFxBIPS);
-				const sBTCRate = await exchanger.feeRateForExchange(empty, sBTC);
-				assert.bnEqual(sBTCRate, newCryptoBIPS);
-				const sETHRate = await exchanger.feeRateForExchange(empty, sETH);
-				assert.bnEqual(sETHRate, newCryptoBIPS);
+	describe('When using Synthetix', () => {
```
justinjmoses, line 3706: looking great!

one way to make it even simpler:

```javascript
['Synthetix','MintableSynthetix'].forEach(snxContract => {
  describe(`When using ${snxContract}`, () => {
    ... 
  });
});
```

See here for instance: https://github.com/Synthetixio/synthetix/blob/develop/test/contracts/Exchanger.spec.js#L749-L766

justinjmoses (approved):
Just one suggestion, otherwise :shipit:

test/contracts/Exchanger.spec.js
ajsantander: @justinjmoses we are explicitly trying to avoid this kind of structure. I consider it a bad practice, but lets discuss.

ajsantander (approved):
This is awesome @i-stam 💯

test/contracts/Exchanger.spec.js
ajsantander: A bit of history... When @i-stam initially wanted to modify exchanger to be tested with Synthetix and MintableSynthetix, the file was mostly based around this array/forEach structure. We tried to reason about it, but it turned out to be very hard to do so.

Also, from my experience, these structure very often produce this scenario; they don't scale well.

So we decided to go for this behavior/utility function based approach instead, and try to stay away from array/forEach structures.

In general I recommend to avoid them in testing. It makes tests very hard to read and understand and may work for simple cases, but as tests become more complex, the cognitive overhead they require makes the code a bit obscure.

PR #1134, opened 2021-03-09
fix: added support for local hardhat forks from dapp side

JChiaramonte7:


index.js
```diff
@@ -166,7 +168,7 @@ const getPathToNetwork = ({ network = 'mainnet', file = '', useOvm = false, path
 
 // Pass in fs and path to avoid webpack wrapping those
 const loadDeploymentFile = ({ network, path, fs, deploymentPath, useOvm = false }) => {
-	if (!deploymentPath && network !== 'local' && (!path || !fs)) {
```
JChiaramonte7, line 169: Had to remove this conditional because `path` is undefined in our js library so the else case here doesn't work for local. Let me know if this causes other issues @ajsantander

justinjmoses:
Hey @JChiaramonte7 thanks for the PR. So you're intending a mainnet fork here yes? I think instead of mapping `local` to `mainnet` in the `data` we provide a way for you to fetch mainnet but with a separate chain ID. That's what you want, right?

JChiaramonte7:
> Hey @JChiaramonte7 thanks for the PR. So you're intending a mainnet fork here yes? I think instead of mapping `local` to `mainnet` in the `data` we provide a way for you to fetch mainnet but with a separate chain ID. That's what you want, right?

Yeah that should work better. If that works on https://github.com/Synthetixio/synthetix/pull/1141 lets close this PR.

PR #1141, opened 2021-03-11
Adding chainid lookup to module

justinjmoses:
This will be used by the js library which needs to know the network and the ovm status from a chain ID.

cc @Synthetixio/frontend-engineers


ajsantander (approved):

justinjmoses:
Also closes #1134

JChiaramonte7 (approved):

PR #1143, opened 2021-03-12
Adds utility contracts for debt pool mirror

gberg1:


contracts/Issuer.sol
```diff
@@ -165,6 +165,16 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
         return getIssuanceRatio();
     }
 
+    function currentOwnershipDebt(address issuer) external view returns (uint) {
```
justinjmoses, line 168: as discussed offline - this can and should live completely in `SynthUtil`

contracts/SynthUtil.sol
```diff
@@ -28,6 +31,10 @@ contract SynthUtil {
         return IExchangeRates(addressResolverProxy.requireAndGetAddress(CONTRACT_EXRATES, "Missing ExchangeRates address"));
     }
 
+    function _currentOwnershipDebt(address issuer) internal view returns (uint) {
+        return Issuer.currentOwnershipDebt(issuer);
```
justinjmoses, line 35: we should use `IIssuer(addressResolverProxy.requireAndGetAddress("Issuer"))...`

justinjmoses (changes requested):

PR #1145, opened 2021-03-15
Change waitingPeriodSecs if not default

ajsantander:


i-stam (approved):
lgtm, but maybe we should also discuss about what is the expected behavior e.g. if (!freshDeploy) then ignore params

ajsantander:
Even though its not completely correct, the !== 0 check is expected behavior, and is tested for. This change is causing tests to fail and would require a redesign of a lot of functionality in the deploy script.

PR #1146, opened 2021-03-15
Fixing caches of Binary Option Markets

justinjmoses:
Each `BinaryOptionMarket` is created by a user via the `BinaryOptionMarketFactory` contract. These contracts rely on `MixinResolver` however they aren't connected to the system in any way other than via the `BinaryOptionMarketManager`. This fix ensures that whenever core contracts are upgraded, these markets have their caches rebuilt as necessary.


jacko125 (approved):
LGTM. Not sure that we have more than 100+ active / matured markets so shouldn't be issue.

i-stam (approved):
lgtm, I wasn't so familiar with binaryOptionMarketManager so I had to go through the contracts too. It would be nice if there was a way to easily test this.

publish/src/commands/deploy.js
```diff
@@ -1567,6 +1567,131 @@ const deploy = async ({
...
+			);
+			console.log(
+				gray('Found'),
+				yellow(numBinaryOptionMarkets),
+				gray(marketType, 'binary option markets')
+			);
+
+			if (numBinaryOptionMarkets > binaryOptionsFetchPageSize) {
+				console.log(
+					redBright(
```
ajsantander, line 1588: Show we throw here instead? Should probably never happen, but if it ever does, it'd be a shame to have it fail silently and not notice this red log.

publish/src/commands/deploy.js
```diff
@@ -1567,6 +1567,131 @@ const deploy = async ({
...
+			} else {
+				// fetch the list of markets
+				const marketAddresses = await binaryOptionMarketManager.methods[
+					`${marketType.toLowerCase()}Markets`
+				](0, binaryOptionsFetchPageSize).call();
+
+				// wrap them in a contract via the deployer
+				const markets = marketAddresses.map(
+					binaryOptionMarket =>
+						new deployer.web3.eth.Contract(compiled['BinaryOptionMarket'].abi, binaryOptionMarket)
```
ajsantander, line 1606: Instead of the try/catch statement below, another alternative would be to query the code and see if it has the old `isResolverCached` signature, then do the appropriate wrapping here.

publish/src/commands/deploy.js
```diff
@@ -1567,6 +1567,131 @@ const deploy = async ({
...
+				}
+				console.log(
+					gray('Binary option market'),
+					yellow(addressOf(market)),
+					gray('is newer and cache status'),
+					yellow(isCached)
+				);
+			} catch (err) {
+				// the challenge being that some used an older MixinResolver API
+				const oldBinaryOptionMarketABI = [
```
ajsantander, line 1629: Would it be better to get this ABI from artifacts?

ajsantander (approved):
Looks good, and should work as is. Just left some comments with some possible improvements.

ajsantander:
Also @justinjmoses, does this mean that a market created by a user will not be properly connected to the system until this script is run again? If so, the deploy script would have to be run periodically just for this.

Wouldn't it be better to have the manager cache child markets upon creation?

publish/src/commands/deploy.js
justinjmoses: Hmmmm. Probably better yeah.

publish/src/commands/deploy.js
justinjmoses: What do you mean by query the code?

publish/src/commands/deploy.js
justinjmoses: We don't store historical ABIs...

justinjmoses:
> Also @justinjmoses, does this mean that a market created by a user will not be properly connected to the system until this script is run again? If so, the deploy script would have to be run periodically just for this.

All binary option markets are disconnected from the system - this is the mechanism to ensure they stay in sync until they expire. As long as we ensure their resolvers are cached each time we deploy new contracts, there isn't a problem.

publish/src/commands/deploy.js
ajsantander: provider.getCode

publish/src/commands/deploy.js
ajsantander: Hmm. Well yeah. If this file was compact, I'd insist to move this out since its data, but we're way beyond that point with this file anyway. Up to you!

PR #1160, opened 2021-03-16
Remove optimistic unit tests

ajsantander:
We've been trying to run our unit tests in "ovm mode" for some time. This involved using an `@ovm-skip` marker in tests that are not applicable in the context of OE.

Thing is, trying to do so is hard, and maybe pointless. Hard because the jsvm for optimism does not exactly mimic optimisitic geth, and pointless because producing ovm artifacts and running unit tests with them only validates the ovm compiler? This is certainly not our responsibility.

We might as well remove optimistic unit tests then.


justinjmoses (approved):
:shipit:

PR #1162, opened 2021-03-19
SIP-113 Add sKRW

hav-noms:
https://sips.synthetix.io/sips/sip-113


publish/deployed/mainnet/config.json
```diff
@@ -661,5 +661,14 @@
 	},
 	"SynthsTSLA": {
 		"deploy": false
+	},
```
justinjmoses, line 664: recall this isn't necessary when adding new synths @hav-noms

PR #1177, opened 2021-03-26
Update gnosis safe utils off chain signing

jacko125:


publish/src/commands/owner.js
```diff
@@ -257,17 +274,33 @@ const owner = async ({
 
 			try {
 				if (isContract) {
-					const newNonce = await createAndSaveApprovalTransaction({
+					const { txHash, newNonce } = await getNewTransactionHash({
```
justinjmoses, line 277: Don't you think we could de-dupe all this logic with the section above?

justinjmoses (approved):
Just one comment about de-duping, but otherwise LGTM

i-stam (approved):

publish/src/commands/owner.js
jacko125: Only difference is the section above is iterating the `owner-actions` list of items and the 2nd section is checking for any `acceptOwnership` that are pending on the contracts. 

I had a look at how it can be combined, would be mainly refactoring the logic out into a separate func

PR #1180, opened 2021-03-26
Adds read only debt ownershiip and shorts for a given synth for debt pool hedging

gberg1:


gberg1:
@justinjmoses updated PR

contracts/SynthUtil.sol
```diff
@@ -43,6 +63,22 @@ contract SynthUtil {
         return total;
     }
 
+    function currentOwnershipDebt(address issuer) external view returns (uint) {
```
justinjmoses, line 66: So the debt ownership is a % value. You want this rather than the actual debt amount they have? `Issuer.debtBalanceOf(address, "sUSD")`?

justinjmoses (approved):
Looks good, just a question however

justinjmoses (commented):
It's also missing test coverage, which should be easy enough to add into `SynthUtil.js`

contracts/SynthUtil.sol
jacko125: I discussed with @gberg1  that it would be helpful is to also return a minter's debt % for the dapp to show this and provide information on how much % of each synth (net skew) they would should be buying to hedge their debt.

test/contracts/SynthUtil.js
```diff
@@ -90,6 +90,11 @@ contract('SynthUtil', accounts => {
 				]);
 			});
 		});
+		describe('currentDebtOwnership', () => {
+			it('should return the debt ownership % of the total debt pool for a given address', async () => {
```
gberg1, line 94: @jacko125 the error I get is Error: Number can only safely store up to 53 bits when I run the test. Any thoughts?

PR #1181, opened 2021-03-26
SIP-131: toggle off "all" approval if set, when withdrawing other approvals

vbstreetz:
Depends on [this draft SIP](https://github.com/Synthetixio/SIPs/blob/sip-fix-delegate-approval-withdrawal/sips/sip-draft_fix_delegate_approval_withdrawal.md).

The issue at the moment is one has to perform 2 transactions in order to downgrade the powers of a delegated account from all to a single one.

Part of: https://github.com/Synthetixio/staking/issues/267

Screenshot 2021-03-26 at 21 17 28


package.json
```diff
@@ -124,6 +124,8 @@
 		"table": "^5.0.2",
 		"wait-port": "^0.2.2",
 		"web3": "1.3.3",
+		"web3-core-helpers": "^1.3.4",
```
vbstreetz, line 127: Had to install these for some reason 🤔

test/contracts/DelegateApprovals.js
```diff
@@ -275,4 +275,32 @@ contract('DelegateApprovals', async accounts => {
 			);
 		});
 	});
+	describe('revoking non-ALL power', () => {
```
vbstreetz, line 278: @i-stam, I've added the test case for the change...

vbstreetz:
- [ ] `approveAllDelegatePowers` probably also needs to be updated in this PR i.e., should explicitly turn on all the powers.

i-stam (approved):

package.json
drptbl: @vbstreetz please lock versions 🙏

drptbl (changes requested):

KillerByte (changes requested):
talked privately with Mitch, and we agreed this PR does not quite meet the specification of the SIP. The "Approve All" flag uses a separate flag, and so if it is set, the other flags remain unset. Therefore, if a user approves all, and then tries to disable a specific permission, they would just be removing all permissions and would have to re-approve all the remaining ones.

PR #1182, opened 2021-03-27
Fix compile warnings for mutability restrictions on BaseSynthetix

sohkai:
Believe these warnings might've been carried over from `Synthetix.sol` during the move in https://github.com/Synthetixio/synthetix/pull/1086.

Nothing critical, since these are all internal jumps.


justinjmoses (approved):

barrasso:
Fixed in #1647

PR #1183, opened 2021-03-29
Fix deploy script overriding waitingPeriodSecs from zero

ajsantander:
Although we recently addressed the setting in of `waitingPeriodSecs` in the deploy script, it still not behaving as expected. Them problem is that, in Optimism configurations, we actually want the value to be zero. The deploy script considers something "unset" when it is zero. Thus, if it finds `waitingPeriodSecs` to be zero, it will set it back to 300s when redeploying to an ovm network, effectively turning fee reclamation back on!

Since the deploy script would require a complete redesign regarding setting parameters to fix this, the modification proposed here merely always considers whatever is set to be valid when the value specified in params.json is zero, and maintains the old behavior otherwise.

### DIFF
--- a/publish/src/commands/deploy.js
+++ b/publish/src/commands/deploy.js
@@ -1985,7 +1985,7 @@ const deploy = async ({
 			contract: 'SystemSettings',
 			target: systemSettings,
 			read: 'waitingPeriodSecs',
-			expected: input => input !== waitingPeriodSecs,
+			expected: input => (waitingPeriodSecs === '0' ? true : input !== '0'),
 			write: 'setWaitingPeriodSecs',
 			writeArg: waitingPeriodSecs,
 		});
--- a/publish/src/commands/deploy.js
+++ b/publish/src/commands/deploy.js
@@ -1979,13 +1979,15 @@ const deploy = async ({
 		}
 
 		// setup initial values if they are unset
+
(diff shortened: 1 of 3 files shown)


publish/src/commands/deploy.js
```diff
@@ -1979,13 +1979,15 @@ const deploy = async ({
...
 
 		// setup initial values if they are unset
+
+		const waitingPeriodSecs = await getDeployParameter('WAITING_PERIOD_SECS');
 		await runStep({
 			contract: 'SystemSettings',
 			target: systemSettings,
 			read: 'waitingPeriodSecs',
-			expected: input => input !== '0',
+			expected: input => (waitingPeriodSecs === '0' ? true : input !== '0'),
```
i-stam, line 1988: I got my head around it for quite a bit, it is a bit confusing. Why don't we just make it more explicit by adding a new ovm specific param in `defaults` i.e. `WAITING_PERIOD_SECS_OVM`

i-stam (commented):
In order for this to work with the current setup, we must add WAITING_PERIOD_SECS = 0 in the relevant `params.json` file.

publish/src/commands/deploy.js
ajsantander: As discussed offline, yes, it is confusing, but the proposed solution is not ideal. Nor is the current params.json in the deployer.

i-stam (approved):

PR #1188, opened 2021-03-31
Use Gnosis Multisig for managing contract ownership on L2

i-stam:
Tested on a Kovan-ovm deployment that can be found here: https://github.com/Synthetixio/synthetix/tree/multisig-simulation-kova-ovm-test


ajsantander (changes requested):
Left a couple of comments and think we should probably go on another path. Something that doesn't duplicate the owner.js code as much. Perhaps split owner.js into owner-eoa.js, owner-safe.js, and implement this as owner-multisig.js.

ajsantander:
Deprecated in favor of https://github.com/Synthetixio/synthetix/pull/1402

PR #1191, opened 2021-03-31
SIP-127: deploy ERC-1167 proxies for virtual synths

sohkai:
See related [SIP](https://github.com/Synthetixio/SIPs/pull/434).

----

Results seem pretty good so far (slightly more now after acc54b6)!

Codesize breather:
Screen Shot 2021-03-31 at 4 22 27 PM

Gas on vSynth exchanges (now, local prod test):
Screen Shot 2021-03-31 at 5 24 40 PM

Gas (previous):
Screen Shot 2021-03-31 at 5 19 29 PM


contracts/VirtualSynth.sol
```diff
@@ -36,13 +36,25 @@ contract VirtualSynth is ERC20, IVirtualSynth {
...
 
-    constructor(
+    // TODO: move this state elsewhere (no gas advantage)?
+    bool public initialized = false;
+
+    // TODO: alternatively a subclass could add this behaviour; might be cleaner for tests?
+    constructor() public ERC20() {
+        // Freeze base copy on deployment so it can never be initialized with real arguments
+        initialized = true;
+    }
```
sohkai, line 46: This is the only design area we could discuss:

1. Should we split this "petrify" behaviour of the base into a separate contract? It'd perhaps simplify the tests so they don't need to be done via proxy.
2. Some ERC-1167 deployers (e.g. [Erasure's](https://github.com/0age/Spawner/blob/master/contracts/Spawner.sol#L10-L37)) have the ability to delegate call into the base contract at deploy time, giving a better guarantee the proxy can't be messed up. I didn't think this was necessary for our use case since synths need to be sent into the proxy for it to work.

contracts/MinimalProxyFactory.sol
```diff
@@ -0,0 +1,33 @@
+pragma solidity ^0.5.16;
+
+
+// TODO: how is this page generated?
+// https://docs.synthetix.io/contracts/source/contracts/minimalproxyfactory
```
justinjmoses, line 5: auto-generated from the AST so you can ignore this for now.

contracts/ExchangerWithVirtualSynth.sol
```diff
@@ -5,24 +5,36 @@ import "./Exchanger.sol";
...
 import "./interfaces/IVirtualSynth.sol";
+import "./MinimalProxyFactory.sol";
 import "./VirtualSynth.sol";
 
 
 // https://docs.synthetix.io/contracts/source/contracts/exchangerwithvirtualsynth
-contract ExchangerWithVirtualSynth is Exchanger {
-    constructor(address _owner, address _resolver) public Exchanger(_owner, _resolver) {}
+contract ExchangerWithVirtualSynth is MinimalProxyFactory, Exchanger {
+    address public baseVirtualSynth;
```
justinjmoses, line 14: We can just take this from the `AddressResolver` - we just have to make sure it's set in tests and in the deployment script (I can assist)

contracts/VirtualSynth.sol
justinjmoses: 1. Hmm. Unless we dupe the initializer args into a constructor then it's not useful for testing cause we can't set it up right? I think removing the constructor altogether and just testing with init only is enough, no?
2. Yeah I think it's fine as-is given our use-case

sohkai: > I think removing the constructor altogether and just testing with init only is enough, no?

Yes, I didn't explain myself very well earlier but this is an option to split the `VirtualSynth` contract into two:

1. `VirtualSynth`: containing all the current logic minus constructor (usable without a proxy)
2. `PetrifiedVirtualSynth` (or something): just adds the constructor (necessitating use with a proxy)

So the `VirtualSynth`'s units tests could be run against `VirtualSynth`, simplifying them to not require a proxy, and other tests and production deploy would use the `PetrifiedVirtualSynth`.

contracts/ExchangerWithVirtualSynth.sol
sohkai: Done in acc54b6.

test/contracts/ExchangerWithVirtualSynth.unit.js
```diff
@@ -44,6 +49,7 @@ contract('ExchangerWithVirtualSynth (unit tests)', async accounts => {
 					const args = [owner, toBytes32('sUSD'), '100', toBytes32('sETH'), owner, toBytes32()];
 
 					behaviors.whenInstantiated({ owner }, () => {
+						// TODO: not sure I get this test--why does it need the behaviour mocked to allow checks?
```
sohkai, line 52: This question is more related to SIP-120 since I changed the related auth check there to save gas/bytecode, but is this `whenMockedToAllowChecks` behaviour only to avoid the mocked synths from causing the test to fail?

contracts/VirtualSynth.sol
justinjmoses: I appreciate the detail. Yeah OK that makes sense to me. Let's split it out then to make testing a bit easier.

sohkai: Done in e1005ba.

test/contracts/ExchangerWithVirtualSynth.unit.js
sohkai: I tried removing the mocking in #1127 (https://github.com/Synthetixio/synthetix/pull/1127/commits/2977e5b78291ffdfab468a027b88e2f15cd99bfa) and it seems to not be necessary anymore with the changes I made to the auth modifier there.

test/publish/index.js
```diff
@@ -1584,6 +1584,8 @@ describe('publish scripts', () => {
 									// address here we should look up all required contracts and ignore any that have
 									// ':' in it
 									.filter(([contract]) => !/^SynthetixBridge/.test(contract))
+									// Note: the VirtualSynth mastercopy is null-initialized and shouldn't be checkd
+									.filter(([contract]) => !/^VirtualSynthMastercopy/.test(contract))
```
justinjmoses, line 1588: 👍🏼

ajsantander (approved):
Awesome 🙏

test/publish/index.js
```diff
@@ -1584,6 +1584,8 @@ describe('publish scripts', () => {
 									// address here we should look up all required contracts and ignore any that have
 									// ':' in it
 									.filter(([contract]) => !/^SynthetixBridge/.test(contract))
+									// Note: the VirtualSynth mastercopy is null-initialized and shouldn't be checkd
```
ajsantander, line 1587: Typo

sohkai: Fixed in 8c834b7 🙏

justinjmoses (approved):

PR #1199, opened 2021-04-09
BO minimal proxies pattern

dgornjakovic:
Following Brett's approach to implementing minimal proxies for BOs.
Tests show a reduction of gas to create markets of 50%.
Differences to Brett's implemention:
 - Mastercopy for BinaryOptionsMarket inherits Owned and MixinResolver so those two now have the init methods to be called from the BinaryOptionsMarketFactory. The init method will only work when a contract has been copied through the MinimalProxyFactory and thus had its state reset (Brett advised this approach)
 - Used a different approach for initializing BinaryOptionsMarketMastercopy so that if adheres to the pattern established in setup.js

### DIFF
--- a/contracts/BinaryOptionMarket.sol
+++ b/contracts/BinaryOptionMarket.sol
@@ -14,7 +14,7 @@ import "./BinaryOption.sol";
 import "./interfaces/IExchangeRates.sol";
 import "./interfaces/IERC20.sol";
 import "./interfaces/IFeePool.sol";
-
+import "./interfaces/IAddressResolver.sol";
 
 // https://docs.synthetix.io/contracts/source/contracts/binaryoptionmarket
 contract BinaryOptionMarket is Owned, MixinResolver, IBinaryOptionMarket {
@@ -74,10 +74,10 @@ contract BinaryOptionMarket is Owned, MixinResolver, IBinaryOptionMarket {
 
     /* ========== CONSTRUCTOR ========== */
 
-    constructor(
-        address _owner,
+    bool public initialized = false;
+
+    function initialize(
         address _creator,
-        address _resolver,
         uint[2] memory _creatorLimits, // [capitalRequirement, skewLimit]
         bytes32 _oracleKey,
         uint _strikePrice,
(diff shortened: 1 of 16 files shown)


contracts/BinaryOptionMarket.sol
```diff
@@ -74,18 +74,20 @@ contract BinaryOptionMarket is Owned, MixinResolver, IBinaryOptionMarket {
...
         uint[2] memory _creatorLimits, // [capitalRequirement, skewLimit]
         bytes32 _oracleKey,
         uint _strikePrice,
         bool _refundsEnabled,
         uint[3] memory _times, // [biddingEnd, maturity, expiry]
         uint[2] memory _bids, // [longBid, shortBid]
         uint[3] memory _fees // [poolFee, creatorFee, refundFee]
-    ) public Owned(_owner) MixinResolver(_resolver) {
+    ) public {
+        require(!initialized, "vSynth already initialized");
```
justinjmoses, line 89: Check the message here

contracts/BinaryOptionMarket.sol
```diff
@@ -74,18 +74,20 @@ contract BinaryOptionMarket is Owned, MixinResolver, IBinaryOptionMarket {
 
     /* ========== CONSTRUCTOR ========== */
 
-    constructor(
-        address _owner,
+    bool public initialized = false;
+
+    function initialize(
         address _creator,
-        address _resolver,
```
justinjmoses, line 80: Things won't work without the address resolver. It's necessary to pass it in (see #1191). Though frankly I think we should remove the `MixinResolver` from the `BinaryOptionMarket` altogether and do what `VirtualSynth` does - see here: https://github.com/Synthetixio/synthetix/blob/sip-minimal-proxies/contracts/VirtualSynth.sol#L70-L72.  That way you can also remove `resolverAddressesRequired`

contracts/MixinResolver.sol
```diff
@@ -16,6 +15,15 @@ contract MixinResolver {
 
     constructor(address _resolver) internal {
         resolver = AddressResolver(_resolver);
+        initialized = true;
+    }
+
+    bool public initialized = false;
+
+    function initResolver(address _resolver) public {
```
justinjmoses, line 23: no don't add anything to this file. it's used everywhere.

contracts/Owned.sol
```diff
@@ -9,14 +10,23 @@ contract Owned {
...
     }
 
+    bool public initialized = false;
+
     function nominateNewOwner(address _owner) external onlyOwner {
         nominatedOwner = _owner;
         emit OwnerNominated(_owner);
     }
 
+    function initOwner(address _owner) public {
```
justinjmoses, line 24: and don't add anything to this either.

contracts/BinaryOptionMarketFactory.sol
```diff
@@ -48,18 +56,11 @@ contract BinaryOptionMarketFactory is Owned, MixinResolver {
...
-                oracleKey,
-                strikePrice,
-                refundsEnabled,
-                times,
-                bids,
-                fees
-            );
+        BinaryOptionMarket bom =
+            BinaryOptionMarket(_cloneAsMinimalProxy(_binaryOptionMastercopy(), "Could not create a Binary Option Market"));
+        bom.initOwner(manager);
```
justinjmoses, line 61: you don't want to be doing this - only one initialize.

PR #1200, opened 2021-04-09
add accrued interest utility contract and add a test for it

dvd-schwrtz:


contracts/AccruedInterestUtil.sol
```diff
@@ -0,0 +1,62 @@
...
+        (uint entryRate, uint lastRate, uint lastUpdated, ) =
+            loan.short
+                ? _manager.getShortRatesAndTime(loan.currency, loan.interestIndex)
+                : _manager.getRatesAndTime(loan.interestIndex);
+
+        // 2. Get the instantaneous rate.
+        (uint rate, bool invalid) =
+            loan.short ? _manager.getShortRate(collateral.synthsByKey(loan.currency)) : _manager.getBorrowRate();
+
+        require(!invalid, "Rates are invalid");
```
i-stam, line 48: This is controversial, usually, in the Synthetix codebase reverting inside view functions is avoided. Better to return instead, even if this is a standalone function.

i-stam (changes requested):

contracts/AccruedInterestUtil.sol
dvd-schwrtz: done! @i-stam

i-stam (approved):

contracts/AccruedInterestUtil.sol
```diff
@@ -0,0 +1,64 @@
...
+            return 0;
+        }
+
+        // 3. Get the time since we last updated the rate.
+        uint timeDelta = block.timestamp.sub(lastUpdated).mul(SafeDecimalMath.unit());
+
+        // 4. Get the latest cumulative rate. F_n+1 = F_n + F_last
+        uint latestCumulative = lastRate.add(rate.multiplyDecimal(timeDelta));
+
+        // 5. If the loan was just opened, don't record any interest. Otherwise multiple by the amount outstanding.
```
justinjmoses, line 58: this comment isn't strictly correct - there's no recording happening here.

contracts/AccruedInterestUtil.sol
```diff
@@ -0,0 +1,64 @@
...
+
+    function state() external view returns (ICollateralState);
+}
+
+contract AccruedInterestUtil is ICollateralLoan {
+    /* ========== LIBRARIES ========== */
+    using SafeMath for uint;
+    using SafeDecimalMath for uint;
+
+    // define and emit events
```
justinjmoses, line 27: there is no event emission here

PR #1202, opened 2021-04-15
Adding a prod test for binary options

dgornjakovic:
Added a prod test which runs the following BOM scenarios and prints reports for them:
- Create a market
- Bid
- Resolve


justinjmoses:
Tests are failing as expected as forks don't have CI secrets.

justinjmoses (approved):
Nice work - now we have BinaryOption gas reports! :100:

PR #1205, opened 2021-04-16
Minimal Proxies for Binary options

dgornjakovic:
- removed Mixin resolver from BinaryOptionsMarket
- Owner public interface not changed, instead a private init method is invoked from BOM initialize
- Merged the prod test for BOM to get gas reports


dgornjakovic:
Moving to main repo

PR #1216, opened 2021-04-21
Minimal Proxy pattern for Binary Options

dgornjakovic:
Minimal Proxy pattern applied to Binary Options:

- BinaryOptionMarketMastercopy added as in VirtualSynths
- BinaryOptionMarket now uses IAddressResolver as implemented in VirtualSynths
- Adapted tests accordingly
- Owned interface extended with an internal method which is called from BOM initialize logic


contracts/OwnedWithInit.sol
```diff
@@ -0,0 +1,42 @@
+pragma solidity ^0.5.16;
+
+// https://docs.synthetix.io/contracts/source/contracts/owned
+contract OwnedWithInit {
+    address public owner;
+    address public nominatedOwner;
+
+    constructor(address _owner) public {
+        require(_owner != address(0), "Owner address cannot be 0");
+        initOwner(_owner);
```
zyzek, line 10: Could this constructor not be empty?

contracts/BinaryOptionMarket.sol
```diff
@@ -114,34 +123,27 @@ contract BinaryOptionMarket is Owned, MixinResolver, IBinaryOptionMarket {
         // Instantiate the options themselves
         options.long = new BinaryOption(_creator, longBid);
```
zyzek, line 124: Should the Binary Option instances themselves not use minimal proxies?

contracts/BinaryOptionMarketMastercopy.sol
```diff
@@ -0,0 +1,13 @@
...
+
+// Inheritance
+import "./OwnedWithInit.sol";
+import "./MixinResolver.sol";
+
+// Internal references
+import "./BinaryOptionMarket.sol";
+
+// https://docs.synthetix.io/contracts/source/contracts/binaryoptionmarket
+contract BinaryOptionMarketMastercopy is BinaryOptionMarket {
```
zyzek, line 11: Not sure why we need this master copy; can't we just stub out the constructor of the BOM contract itself?

zyzek (changes requested):

contracts/OwnedWithInit.sol
dgornjakovic: Goog point, it could now that a separate owned contract is used. Will adapt.

contracts/BinaryOptionMarket.sol
dgornjakovic: Good hint, will investigate.

contracts/BinaryOptionMarketMastercopy.sol
dgornjakovic: That I wouldnt know, I followed Brett's approach from the VirtualSynths MinimalProxy impl.

contracts/OwnedWithInit.sol
dgornjakovic: Constructors is now empty

contracts/BinaryOptionMarket.sol
dgornjakovic: Implemented

dgornjakovic: Gas costs to create a new market dropped 5.5x with this change!

contracts/BinaryOptionMarketMastercopy.sol
zyzek: Doesn't this need to set `initialized = true`?

contracts/OwnedWithInit.sol
```diff
@@ -0,0 +1,39 @@
+pragma solidity ^0.5.16;
+
+// https://docs.synthetix.io/contracts/source/contracts/owned
+contract OwnedWithInit {
```
justinjmoses, line 4: Let's try to de-dupe code here. I think a `BaseOwned.sol` with the shared code including `initOwner` but no constructor and `Owned.sol` which extends from the base and adds a constructor that simply calls `initOwner`

publish/src/commands/deploy.js
```diff
@@ -1689,28 +1681,76 @@ const deploy = async ({
...
+				const isCached = await oldBinaryOptionMarket.methods.isResolverCached().call();
 				if (!isCached) {
 					binaryOptionMarketsToRebuildCacheOn.push(addressOf(market));
 				}
 
 				console.log(
 					gray('Binary option market'),
 					yellow(addressOf(market)),
-					gray('is older and cache status'),
+					gray('is newer and cache status'),
```
justinjmoses, line 1701: technically it's newer but not as new as the latest that doesn't even need cache rebuilding.

publish/src/commands/deploy.js
```diff
@@ -1689,28 +1681,76 @@ const deploy = async ({
...
+					}
+
+					console.log(
+						gray('Binary option market'),
+						yellow(addressOf(market)),
+						gray('is older and cache status'),
+						yellow(isCached)
+					);
+				} catch (err) {
+					console.error(red(err));
```
justinjmoses, line 1752: Won't all of your new markets appear in here? Because they don't have a `isResolverCached()` function anymore and thus they don't need cache rebuilding. You can safely skip them for this part.

justinjmoses:
After some discussion, it's been agreed that Thales will clone these contracts and redeploy them going forward, rather than modifying these directly.

PR #1226, opened 2021-04-26
Remove bridge migration script

ajsantander:
Removing this script because it was too fine tuned to the v1 bridge mechanics.

The incoming bridge 2 is going to look a little different, and it seems safer to act without a script with the multisig, being carefiul.


justinjmoses (approved):

justinjmoses (changes requested):
You need to remove the other references to this file as well - grep thru and you'll see them

PR #1237, opened 2021-05-03
Adding forwarding functionality to optimistic bridge

justinjmoses:
Proposing this instead of #1230. Note however this is just for L1.

### DIFF
--- a/contracts/SynthetixBridgeToOptimism.sol
+++ b/contracts/SynthetixBridgeToOptimism.sol
@@ -7,12 +7,14 @@ import "./interfaces/ISynthetixBridgeToOptimism.sol";
 import "@eth-optimism/contracts/iOVM/bridge/tokens/iOVM_L1TokenGateway.sol";
 
 // Internal references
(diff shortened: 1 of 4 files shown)


contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -92,6 +92,12 @@ contract SynthetixBridgeToOptimism is BaseSynthetixBridge, ISynthetixBridgeToOpt
         _depositReward(msg.sender, amount);
     }
 
+    // forward any accidental tokens sent here to the escrow
+    function forwardTokensToEscrow(address token) external {
+        IERC20 erc20 = IERC20(token);
+        erc20.transfer(synthetixBridgeEscrow(), erc20.balanceOf(address(this)));
```
i-stam, line 98: Should we use `safeTransfer` instead?

i-stam (commented):
I like this approach, I left a comment regarding transfer()

i-stam (approved):

PR #1238, opened 2021-05-03
Implements a contract for safer bridge migration

ajsantander:
Since the migration of SNX from the existing SynthetixBridgeToOptimism (v1) to the new SynthetixBridgeEscrow (v2) is a delicate operation, this PR proposes a way to do so with a script and a migration contract that will perform and validate the migration in a single atomic transaction, thus completely failing or succeeding.

### DIFF
--- a/contracts/BridgeMigrator.sol
+++ b/contracts/BridgeMigrator.sol
@@ -15,7 +15,7 @@ interface IOwned {
 }
 
 interface IOldSynthetixBridgeToOptimism {
-    function active() external view returns (bool);
+    function activated() external view returns (bool);
 
     function migrateBridge(address newBridge) external;
 }
@@ -98,22 +98,21 @@ contract BridgeMigrator {
     }
 
     function _provideAllowance() internal {
-        require(snx.allowance(newEscrow, newBridge) == 0, "Unexpected initial new bridge allowance");
-
+        ISynthetixBridgeEscrow(newEscrow).approveBridge(address(snx), newBridge, 0);
         ISynthetixBridgeEscrow(newEscrow).approveBridge(address(snx), newBridge, uint256(-1));
         require(snx.allowance(newEscrow, newBridge) == uint256(-1), "Unexpected final new bridge allowance");
     }
 
     function _validateStateBefore() internal {
-        require(IOldSynthetixBridgeToOptimism(oldBridge).active() == true, "Unexpected initial old bridge state");
+        require(IOldSynthetixBridgeToOptimism(oldBridge).activated() == true, "Unexpected initial old bridge state");
     }
 
     function _migrateSNX() internal {
-        IOldSynthetixBridgeToOptimism(oldBridge).migrateBridge(newBridge);
+        IOldSynthetixBridgeToOptimism(oldBridge).migrateBridge(newEscrow);
     }
 
     function _validateStateAfter() internal {
-        require(IOldSynthetixBridgeToOptimism(oldBridge).active() == false, "Unexpected final old bridge state");
+        require(IOldSynthetixBridgeToOptimism(oldBridge).activated() == false, "Unexpected final old bridge state");
     }
 
     function _validateBalancesAfter() internal {
--- a/contracts/BridgeMigrator.sol
+++ b/contracts/BridgeMigrator.sol
@@ -81,7 +81,7 @@ contract BridgeMigrator {
 
     function _takeOwnership() internal {
         require(IOwned(oldBridge).owner() == pdao, "Unexpected old bridge owner");
-        require(IOwned(newEscrow).owner() == deployer, "Unexpected new escrow owner");
+        require(IOwned(newEscrow).owner() == deployer || IOwned(newEscrow).owner() == pdao, "Unexpected new escrow owner");
 
         IOwned(oldBridge).acceptOwnership();
         IOwned(newEscrow).acceptOwnership();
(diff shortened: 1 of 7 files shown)


contracts/BridgeMigrator.sol
```diff
@@ -0,0 +1,132 @@
...
+        string memory _network
+    ) public {
+        newBridge = _newBridge;
+        newEscrow = _newEscrow;
+
+        if (keccak256(abi.encodePacked(_network)) == keccak256(abi.encodePacked("mainnet"))) {
+            oldBridge = 0x045e507925d2e05D114534D0810a1abD94aca8d6;
+            pdao = 0xEb3107117FEAd7de89Cd14D463D340A2E6917769;
+            deployer = 0xDe910777C787903F78C89e7a0bf7F4C435cBB1Fe;
+            snx = IERC20(0x97767D7D04Fd0dB0A1a2478DCd4BA85290556B48);
```
justinjmoses, line 45: why not use the proxy?

contracts/BridgeMigrator.sol
```diff
@@ -0,0 +1,132 @@
...
+
+    // ----------------------------------------
+    // INTERNAL
+    // ----------------------------------------
+
+    function _takeOwnership() internal {
+        require(IOwned(oldBridge).owner() == pdao, "Unexpected old bridge owner");
+        require(IOwned(newEscrow).owner() == deployer, "Unexpected new escrow owner");
+
+        IOwned(oldBridge).acceptOwnership();
```
justinjmoses, line 87: worth doing a check to make sure the owners have been nominated?

publish/src/commands/migrate-bridge.js
```diff
@@ -0,0 +1,177 @@
...
+
+	snx = getTarget({ network, contract: 'Synthetix' }).address;
+	newBridge = getTarget({ network, contract: 'SynthetixBridgeToOptimism' }).address;
+	newEscrow = getTarget({ network, contract: 'SynthetixBridgeEscrow' }).address;
+	console.log(chalk.gray(`Synthetix: ${snx}`));
+	console.log(chalk.gray(`New bridge: ${newBridge}`));
+	console.log(chalk.gray(`New escrow: ${newEscrow}`));
+
+	if (network === 'mainnet') {
+		oldBridge = '0x045e507925d2e05D114534D0810a1abD94aca8d6';
```
justinjmoses, line 72: you could use getVersions to find the last `SynthetixBridgeToOptimism` but I guess you want to be super explicit here.

publish/src/commands/migrate-bridge.js
```diff
@@ -0,0 +1,177 @@
+const chalk = require('chalk');
+const fs = require('fs');
+const path = require('path');
+const ethers = require('ethers');
+const assert = require('assert');
+const { ensureNetwork, loadConnections, confirmAction } = require('../util');
+const { wrap } = require('../../..');
+
+let signer;
```
justinjmoses, line 9: I don't love this pattern - using the closure of the module. it means if it's somehow run again in the same node instance all these variables are already set.

publish/src/commands/migrate-bridge.js
```diff
@@ -0,0 +1,177 @@
...
+	if (!useMigrator) {
+		await confirmAction(chalk.yellow('Type "y" to deploy the migrator contract'));
+
+		console.log(chalk.gray('Deploying BridgeMigrator...'));
+
+		const Migrator = new ethers.ContractFactory(artifacts.abi, artifacts.bytecode, signer);
+		migrator = await Migrator.deploy(newBridge, newEscrow, network, txParams);
+
+		const tx = migrator.deployTransaction;
+		console.log(chalk.gray(tx.hash));
```
justinjmoses, line 97: what about outputting the new migrator contract address?

publish/src/commands/migrate-bridge.js
```diff
@@ -0,0 +1,177 @@
...
+	assert(_newEscrow === newEscrow, 'Unexpected new escrow address');
+
+	console.log(chalk.gray(`Contract's snx: ${_snx} OK ✓`));
+	console.log(chalk.gray(`Contract's old bridge: ${_oldBridge} OK ✓`));
+	console.log(chalk.gray(`Contract's new bridge: ${_newBridge} OK ✓`));
+	console.log(chalk.gray(`Contract's new escrow: ${_newEscrow} OK ✓`));
+}
+
+async function _nominate() {
+	await confirmAction(
```
justinjmoses, line 127: i would check these and only output if not yet nominated (I suspect you'll re-run this a few times)

contracts/BridgeMigrator.sol
ajsantander: Sure, will change

contracts/BridgeMigrator.sol
ajsantander: Accept ownership already does that I believe:

```
    function acceptOwnership() external {
        require(msg.sender == nominatedOwner, "You must be nominated before you can accept ownership");
        emit OwnerChanged(owner, nominatedOwner);
        owner = nominatedOwner;
        nominatedOwner = address(0);
    }
```

publish/src/commands/migrate-bridge.js
ajsantander: Yes exactly. My intention was to replace some automation with hardcoded data to be explicit.

publish/src/commands/migrate-bridge.js
ajsantander: Sure, I can change it to pass variables around and not define anything in the global scope if you prefer.

publish/src/commands/migrate-bridge.js
ajsantander: Sure, will implement

justinjmoses (approved):

justinjmoses (approved):

i-stam (approved):

PR #1239, opened 2021-05-04
Add etherwrapperRewards and tracking to Etherwrapper

jacko125:


contracts/EtherWrapper.sol
```diff
@@ -46,6 +47,9 @@ contract EtherWrapper is Owned, Pausable, MixinResolver, MixinSystemSettings, IE
     // ========== STATE VARIABLES ==========
     IWETH internal _weth;
 
+    // EtherWrapper rewards
+    IEtherWrapperRewards public wrapperRewards;
```
justinjmoses, line 51: I suggest using the `AddressResolver` for this, unless you wanted to make it optional?

contracts/EtherWrapperRewards.sol
```diff
@@ -0,0 +1,175 @@
...
+        if (reward > 0) {
+            rewards[account] = 0;
+            rewardsToken.safeTransfer(account, reward);
+            emit RewardPaid(account, reward);
+        }
+    }
+
+    /* ========== RESTRICTED FUNCTIONS ========== */
+
+    function notifyRewardAmount(uint256 reward) external onlyRewardsDistribution updateReward(address(0)) {
```
justinjmoses, line 122: so this can only be funded via inflation? is that intentional?

PR #1243, opened 2021-05-05
SIP-136: MultiCollateral/CollateralEth debt is not excluded correctly from debt calculations

liamzebedee:
Synths issued by the MultiCollateral and v1 Collateral ETH and sUSD contracts is incorrectly excluded by the debt cache. The debt is currently subtracted from the respective totalSupply's of each token, however, if the synth is exchanged into another synth, the debt may not be excluded here. 

This PR includes:

 * changes according to [SIP 136](https://sips.synthetix.io/sips/sip-136)
 * test cases for the exclusion of debt
 * after discussion with Jackson, the update of excluded debt will only occur within the debt snapshotting functionality. Otherwise, exchanges and other code will rely on the cached excluded debt. This was a tradeoff of correctness for user cost (gas), as the excluded debt cost another 200K in gas.

Some other changes:

 * additional edge case: clamping the debt to 0 if excludedDebt is larger than SNX debt
 *  excluding MC short debt from the _debtBalanceOfAndTotalDebt calculation


contracts/BaseDebtCache.sol
```diff
@@ -191,19 +166,44 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
-        // subtract the USD value of all shorts.
+        // 1. Subtract the USD value of all shorts.
+        uint excludedDebt = 0;
         (uint susdValue, bool shortInvalid) = collateralManager().totalShort();
+        isInvalid = isInvalid || shortInvalid;
+        excludedDebt = excludedDebt.add(susdValue);
 
-        total = total.sub(susdValue);
+        // 2. Subtract the USD debt of the CollateralManager and v1 Collateral contracts.
+        for (uint i = 0; i < numValues; i++) {
```
zyzek, line 187: If we're concerned about inefficiency due to this extra loop, we could compute `excludedDebt` inside `_issuedSynthValues` and return it as another value e.g. the return signature of `_currentSynthDebts` would become `(uint[] memory snxIssuedDebts, uint collateralIssuedDebt, bool anyRateIsInvalid)`.

contracts/BaseDebtCache.sol
```diff
@@ -191,19 +166,44 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
         for (uint i; i < numValues; i++) {
             total = total.add(values[i]);
         }
 
-        // subtract the USD value of all shorts.
+        // 1. Subtract the USD value of all shorts.
+        uint excludedDebt = 0;
         (uint susdValue, bool shortInvalid) = collateralManager().totalShort();
+        isInvalid = isInvalid || shortInvalid;
+        excludedDebt = excludedDebt.add(susdValue);
```
zyzek, line 184: Just do `uint excludedDebt = susdValue;`, no need for the declaration `uint excludedDebt = 0;` above.

contracts/BaseDebtCache.sol
```diff
@@ -191,19 +166,44 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
 
-        total = total.sub(susdValue);
+        // 2. Subtract the USD debt of the CollateralManager and v1 Collateral contracts.
+        for (uint i = 0; i < numValues; i++) {
+            bytes32 key = currencyKeys[i];
 
-        isInvalid = isInvalid || shortInvalid;
+            if (collateralManager().isSynthManaged(key)) {
```
zyzek, line 190: Not for this PR, but it strikes me that an external function call to the collateral manager for every synth is pretty inefficient. Surely it would be more effective to just return an array of the synths with non-snx collateral and iterate those.

contracts/BaseDebtCache.sol
```diff
@@ -191,19 +166,44 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
+                uint collateralIssued = collateralManager().long(key);
+                excludedDebt = excludedDebt.add(collateralIssued.multiplyDecimalRound(rates[i]));
+            }
+
+            if (key == sUSD) {
```
zyzek, line 195: Given that the code here is identical, and code size might be an actual issue on L2 so I wonder whether collapsing the `sUSD` and `sETH` cases back into one conditional might actually be preferable here. Not too fussed either way, especially if we're well below the L2 contract size limit.

contracts/BaseDebtCache.sol
```diff
@@ -191,19 +166,44 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
+                IEtherCollateral etherCollateralContract = IEtherCollateral(address(etherCollateralsUSD()));
+                excludedDebt = excludedDebt.add(etherCollateralContract.totalIssuedSynths().multiplyDecimalRound(rates[i]));
+            }
+
+            if (key == sETH) {
```
zyzek, line 200: Tiny nitpick: this could be an `else if`.

contracts/BaseDebtCache.sol
```diff
@@ -191,19 +166,44 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
+                IEtherCollateral etherCollateralContract = etherCollateral();
+                excludedDebt = excludedDebt.add(etherCollateralContract.totalIssuedSynths().multiplyDecimalRound(rates[i]));
+            }
+        }
+
+        total = total.sub(excludedDebt);
```
zyzek, line 206: Worried that we might be able to break the whole system if `excludedDebt > total`.

Feels like we should either set the total system debt to zero in this case, or throw a more informative error message.

This seems like an odd case, but it could happen if, for example, everyone issued sUSD on the multicollateral contracts, converted to sBTC, then the price of sBTC collapsed, meaning there was not enough value left in the system to pay back the loans. This is just a top of the head thing so maybe I'm missing something here? Seems like a pretty fundamental problem if correct though.

test/contracts/DebtCache_bug.js
```diff
@@ -0,0 +1,262 @@
...
+			const { debtValues } = await debtCache.currentSynthDebts(synths.map(toBytes32));
+			console.log(
+				synths.map((synth, i) => `${synth} debt = ${debtValues[i].toString()}`).join('\n')
+			);
+			const { debt } = await debtCache.currentDebt();
+			console.log(`currentDebt: ${debt.toString()}`);
+			console.log('');
+		}
+
+		describe('after minting sETH issued by non-SNX collateral, and the full supply of sETH is exchanged for another synth', async () => {
```
zyzek, line 224: Do you think it's worth moving this test case into the main `Collateral.sol` test file?

zyzek (changes requested):

contracts/BaseDebtCache.sol
zyzek: Hm, I guess they could issue against SNX to pay back the loans. So then we have:

1. Issue 50000 sUSD against ETH.
2. Convert sUSD into 1 sBTC
3. sBTC goes to zero
4. Issue 50000 more sUSD against SNX
5. repay the original debt
6. Now the circulating supply of synths is zero again, so the system debt is zero
7. Your SNX is free, so you repaid the debt for no cost.

Maybe there is something wrong with this analysis, but I feel like the debt pool ramifications of the multicollat system may not have been completely characterised.

contracts/BaseDebtCache.sol
jacko125: @zyzek  Good point, should have on the manager `managedSynths() return (bytes32[] keys)` to save gas. Checked that the CollateralManager doesn't have anything like that.

contracts/BaseDebtCache.sol
jacko125: Could reuse this block of code here: 

```            bool isSUSD = key == sUSD;
            if (isSUSD || key == sETH) {
                IEtherCollateral etherCollateralContract =
                    isSUSD ? IEtherCollateral(address(etherCollateralsUSD())) : etherCollateral();
     ```

File: contracts/BaseDebtCache.sol (hunk 2 above)
Comment (jacko125) [inline contracts/BaseDebtCache.sol:184]: @liamzebedee  

In the takeDebtSnapshot function it also removes the `collateralManager().totalShort()` from the total debt so it should be removed from the `takeSnapshot()` function. 

https://github.com/Synthetixio/synthetix/blob/master/contracts/DebtCache.sol#L23-L24

File: test/contracts/DebtCache_bug.js [hunk 8]
```diff
@@ -0,0 +1,262 @@
...
+const { setupAllContracts, setupContract } = require('./setup');
+
+const { currentTime, toUnit, fastForward } = require('../utils')();
+
+const {
+	toBytes32,
+	constants: { ZERO_ADDRESS },
+} = require('../..');
+
+contract('DebtCache', async accounts => {
```
justinjmoses, line 16: Why not combine this as a new describe block in the DebtCache tests? This would introduce a new pattern `_bug` to test files which isn't very scalable.

contracts/BaseDebtCache.sol
liamzebedee: good point

liamzebedee: ugh, looks like some code is repeated. I'll try address this too so that we only calculate debt and issued synth values in one place, while maintaining the same external API.

contracts/BaseDebtCache.sol
liamzebedee: Agreed, let's not add another O(n).

contracts/BaseDebtCache.sol
liamzebedee: Nice spot! @zyzek

contracts/BaseDebtCache.sol
liamzebedee: > Worried that we might be able to break the whole system if excludedDebt > total.

Isn't `excludedDebts` always smaller than `total`, since `total` is based in the total supply of each synth?

> Maybe there is something wrong with this analysis, but I feel like the debt pool ramifications of the multicollat system may not have been completely characterised. Do we think this is possible to abuse without the debt load going completely all the way to zero?

I think you're right. Basically, if synths created by MC loans go to $0, then the debt excluded will decrease, and thus the global debt will increase. Conversely, if the synths created by MC loans go to $infinity, then the debt excluded will increase, and the global debt will decrease. However, as I mentioned before, the global debt is calculated as the total supply of all synths, so it can't go negative. 

So you have:

 1. 5,000 sBTC MC debt, worth $250M
 2. 5,000 totalSupply of MC debt, worth $250M
 3. sBTC goes up 10x
 4. sBTC MC debt is $2,500M, totalSupply of MC debt is $2,500M

So the debt can never go negative. However, the MC synths can certainly begin to outweight SNX debt, such that the MC loans can pay off SNX debt. ie. there is arbitrage between the different loan systems

> Do we think this is possible to abuse without the debt load going completely all the way to zero?

I did some back of the spreadsheet calculations here - https://docs.google.com/spreadsheets/d/1yn7xDie2b3D3n0M2PUPc4piV3u0FkNjSZ_X_ISVIUYY/edit?usp=sharing

Maybe you can go into it and clarify it. I think what you're identifying is that cost of MC debt and SNX debt can be arbed? But I don't think it poses a risk to the total debt calculations.

test/contracts/DebtCache_bug.js
liamzebedee: Yeah nah this was definitely temp. I just wanted to write a quick repro - what would be better, instead of this test, is some proper testing for the debt calculations on how it excludes various collateral subsystems (multi, ethercollateralv1).

contracts/BaseDebtCache.sol
zyzek: Left a scenario in the spreadsheet. This still seems like it's a problem. If the debt is all ETH-backed sUSD for example, then moves into sBTC, the total value of circulating synths can fall below the debt in the multicollateral contract.

The answer here is to check if `excludedDebt` exceeds the total synth value and clamp to zero if so. This still means that SNX issuers will be able to issue more sUSD for free in this case, I suppose. Another possibility is, for example, if the MC-issued debt is all denominated in sETH, if sETH appreciates after those synths have been converted to something else that did not appreciate so quickly, then this will decrease the system debt, allowing more issuance by SNX stakers.

contracts/BaseDebtCache.sol
```diff
@@ -120,59 +120,66 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
+                    key == sUSD ? IEtherCollateral(address(etherCollateralsUSD())) : etherCollateral();
+                excludedDebt = excludedDebt.add(etherCollateralContract.totalIssuedSynths().multiplyDecimalRound(rates[i]));
             }
-
-            values[i] = supply.multiplyDecimalRound(rates[i]);
         }
-        return values;
+
+        // 3. Short debt.
+        (uint shortValue, ) = collateralManager().totalShort();
```
zyzek, line 155: I was going to say we could do this inside the loop as well rather than adding an additional function call, but I guess `collateralManager().isSynthManaged(key)` doesn't actually guarantee anything about whether the synth is or is not shortable.

contracts/BaseDebtCache.sol
```diff
@@ -120,59 +120,66 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
-                // so we check explicitly and 0 it out to prevent
-                // a safesub overflow.
-
-                if (collateralIssued > supply) {
-                    supply = 0;
-                } else {
-                    supply = supply.sub(collateralIssued);
-                }
+                uint issuedSynths = collateralManager().long(key);
+                excludedDebt = excludedDebt.add(issuedSynths.multiplyDecimalRound(rates[i]));
```
zyzek, line 144: Was going to suggest we could pull this out of the loop and put it next to the `totalShort` call below, but I realised internally that function fetches the rates individually rather than all at once, so I'm not clear on whether that would be more or less efficient, gas-wise. My hunch is that it would be better since `CollateralManager` won't iterate over synths it isn't managing.

zyzek (changes requested):

contracts/BaseDebtCache.sol
liamzebedee: I did start having a go on this - https://github.com/Synthetixio/synthetix/tree/debt-cache-bug@CollateralManager.managedSynths

I'll see if there's enough time to work through a good change by EOD. Still waiting for hardhat fork to cache properly so I'm not slown down by this anymore.

contracts/BaseDebtCache.sol
jacko125: Interesting scenario, I guess in this case the SNX stakers would be able to mint sUSD up to the excess `excludedDebt` amount and still be able to withdraw their SNX as they won't incur any debt. Which would give them a free sUSD that is fully backed by other collateral.

contracts/BaseDebtCache.sol
jacko125: @zyzek  I guess including it inside the loop iterating over all synths was programatic way of ensuring that if we added a new synth type to the Multicollateral loans it would be automatically picked up.

contracts/DebtCache.sol
```diff
@@ -82,14 +79,15 @@ contract DebtCache is BaseDebtCache {
         // Update the cached values for each synth, saving the sums as we go.
         uint cachedSum;
         uint currentSum;
-        uint[] memory currentValues = _issuedSynthValues(currencyKeys, currentRates);
+        (uint[] memory currentValues, uint excludedDebt) = _issuedSynthValues(currencyKeys, currentRates);
```
jacko125, line 82: @liamzebedee  @zyzek  How would this impact the partial debt cache updates for synths that only change? Would increase the gas usage for partial debt cache updates as it needs to load all of the excluded Debt?

liamzebedee: @jacko125 good question. There might be an issue in how we do the partial debt updates anyways right now:

 * note that `_updateCachedSynthDebtsWithRates` is called with `currencyKeys` which we will update, which is later passed down into `_issuedSynthValues`. 
 * in `_issuedSynthValues`, we exclude MultiCollateral debt during the loop over `currencyKeys`. This is supposed to save on gas by doing it in the `currencyKeys` loop, instead of doing a separate loop for all MC-issued synths. However, in this example, `currencyKeys` may not include all the MC synths.

> Would increase the gas usage for partial debt cache updates as it needs to load all of the excluded Debt?

It doesn't "load" all excluded debt atm. This is a problem with the current deployment and the code in the PR. The impact might not be huge, since we already do the regular debt snapshot.

jacko125: @liamzebedee  

Got it, only when we pass in the `sETH, sUSD, sBTC` as keys to `_issuedSynthValues` would the excludeDebt include the MC-issued synths.

test/contracts/DebtCache.js
```diff
@@ -1152,14 +1236,37 @@ contract('DebtCache', async accounts => {
...
+
+					// Take out a short position on sETH.
+					// sUSD collateral = 1.5 * rate_eth
+					amount = multiplyDecimalRound(rate, toUnit('1.5'));
+					await sUSDContract.issue(account1, amount, { from: owner });
+					await short.open(amount, oneETH, sETH, { from: account1 });
+				});
+
+				it('increases', async () => {
+					assert.bnEqual(totalNonSnxBackedDebt.add(rate), await debtCache.totalNonSnxBackedDebt());
```
jacko125: @liamzebedee  this test should be adding `oneETH` as that is 1 ETH short opened instead of checking `rate`.

contracts/DebtCache.sol
jacko125: @zyzek 

Would be good to get your thoughts on the partial synth cache update as now we calculate the `excludedDebt` as nominal USD value instead of subtracting the `sETH supply` and `sUSD supply` for MC-loans before they are returned by `_totalIssuedSynths` for each synth caching. 

 This would mean that the `_cachedSynthDebt[key]` would be the `synth.totalSupply()` as returned by `_totalIssuedSynths` and would be inaccurately adding the excluded debt on partial debt cache updates for MC synths.

contracts/BaseDebtCache.sol
zyzek: I was just thinking one single call to `totalLong` would sweep away this complexity.

contracts/DebtCache.sol
zyzek: I think the excluded debt component should be treated as its own 'Synth'. At any rate a debt component which can be updated independently by passing in a currency key corresponding to it, and with its own delta logic etc. I tend to think this should also be how cross-chain debt communication should be handled. I'm also wondering here, whether issuance and burning by the multicollat contracts are properly handled. As I recall, the mint function just adds sUSD debt, but if the mint is coming from an MC contract, it should not do this. This needs to be checked, but if it's correct, then multicollateral may be creating false volatility in the debt cache as loans opening and closing could be erroneously affecting the cached debt value.

justinjmoses:
@liamzebedee can you please add an entry for *Aloith* in `releases.json` in this PR and add the changed files?

contracts/BaseDebtCache.sol
```diff
@@ -190,22 +179,43 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
 
-        // subtract the USD value of all shorts.
-        (uint susdValue, bool shortInvalid) = collateralManager().totalShort();
-
-        total = total.sub(susdValue);
-
-        isInvalid = isInvalid || shortInvalid;
-
-        return (total, isInvalid);
+        return (total, isInvalid && isAnyNonSnxDebtRateInvalid);
```
jacko125, line 218: @liamzebedee  `isAnyNonSnxDebtRateInvalid` would be overlapping with the same `invalid` flag because if one of the rates are in invalid already for `exchangeRates().ratesAndInvalidForCurrencies(currencyKeys)` , it would also be invalid with `isAnyNonSnxDebtRateInvalid`. Not sure if this is needed?

contracts/DebtCache.sol
```diff
@@ -91,6 +91,13 @@ contract DebtCache is BaseDebtCache {
             _cachedSynthDebt[key] = currentSynthDebt;
         }
 
+        // Always update the cached value of the excluded debt -- it's computed anyway.
```
jacko125, line 94: @liamzebedee @zyzek 

Would be interesting to see how much extra gas this adds to the `exchange()` function when it calls down into here to update the cache for the 2x pairs. 

We could have a separate function for `synthetix.exchange()` that doesn't need to recompute the `excludedDebt` and leave the cached value for recalculating the ``cachedSum && currentSum`` so not to recompute the excludedDebt as that shouldn't have changed during the exchange function.

contracts/BaseDebtCache.sol
```diff
@@ -190,22 +179,43 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
         return _cachedSynthDebts(currencyKeys);
     }
 
+    // Returns the total sUSD debt backed by non-SNX collateral.
+    function totalNonSnxBackedDebt() external view returns (uint excludedDebt, bool isInvalid) {
```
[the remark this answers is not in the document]
liamzebedee: Nice catch, I missed this

contracts/DebtCache.sol
liamzebedee: Yeah I will do some gas estimations tomorrow.

contracts/BaseDebtCache.sol
liamzebedee: Hmmm - I think it's better to be defensive, even if we know `availableCurrencyKeys` is a superset of the synths that  `_totalNonSnxBackedDebt` will receive as input. We currently don't have tests for these `isInvalid` calculations re: the excluded debt, and the gas saving here is negligible. What do you reckon?

contracts/DebtCache.sol
```diff
@@ -74,15 +74,18 @@ contract DebtCache is BaseDebtCache {
...
+        bool anyRateIsInvalid,
+        bool recomputeExcludedDebt
     ) internal {
         uint numKeys = currencyKeys.length;
         require(numKeys == currentRates.length, "Input array lengths differ");
 
         // Update the cached values for each synth, saving the sums as we go.
         uint cachedSum;
         uint currentSum;
+        uint excludedDebtSum;
```
jacko125: @liamzebedee  excludedDebtSum should be set to the _cachedDebtSum when it's not recomputing the excluded Debt

jacko125 (approved):
LGTM

test/prod/MultiCollateral.prod.js
```diff
@@ -134,7 +134,8 @@ contract('MultiCollateral (prod tests)', accounts => {
 		});
 	});
 
-	describe('when using multiple types of loans', () => {
+	// TODO: disabled while debtLimit is reached.
+	xdescribe('when using multiple types of loans', () => {
```
justinjmoses, line 138: This should no longer be needed once #1259 is merged. cc @ajsantander

contracts/BaseDebtCache.sol
```diff
@@ -190,22 +179,51 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
+    }
+
+    function _totalNonSnxBackedDebt() internal view returns (uint excludedDebt, bool isInvalid) {
+        // Calculate excluded debt.
+        // 1. Ether Collateral.
+        if (resolver.getAddress(CONTRACT_ETHERCOLLATERAL_SUSD) != address(0)) {
+            excludedDebt = etherCollateralsUSD().totalIssuedSynths(); // Ether-backed sUSD
+        }
+
+        if (address(issuer().synths(sETH)) != address(0) && resolver.getAddress(CONTRACT_ETHERCOLLATERAL) != address(0)) {
```
justinjmoses, line 194: do you think it's necessary to spend the extra gas to check that `sETH` exists? I can't imagine how EtherCollateral could exist in the resolver but no sETH? And further, even if so, the `sETH` rate would be invalid.

contracts/BaseDebtCache.sol
```diff
@@ -190,22 +179,51 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
+    // Returns the total sUSD debt backed by non-SNX collateral.
+    function totalNonSnxBackedDebt() external view returns (uint excludedDebt, bool isInvalid) {
+        return _totalNonSnxBackedDebt();
+    }
+
+    function _totalNonSnxBackedDebt() internal view returns (uint excludedDebt, bool isInvalid) {
+        // Calculate excluded debt.
+        // 1. Ether Collateral.
+        if (resolver.getAddress(CONTRACT_ETHERCOLLATERAL_SUSD) != address(0)) {
+            excludedDebt = etherCollateralsUSD().totalIssuedSynths(); // Ether-backed sUSD
```
justinjmoses, line 191: [np] might be better to add this to `excludedDebt` as it's order-dependent right now and if the order was ever switched this would override the existing value of `excludedDebt`

contracts/SafeDecimalMath.sol
```diff
@@ -184,4 +184,9 @@ library SafeDecimalMath {
 
         return quotientTimesTen / 10;
     }
+
+    // Computes `a - b`, setting the value to 0 if b > a.
+    function floorsub(uint a, uint b) internal pure returns (uint) {
```
justinjmoses, line 189: Because this is internal, it simply gets imported into the source right? So no need to worry about new deploys.

publish/deployed/kovan/config.json
```diff
@@ -45,7 +45,7 @@
 		"deploy": false
 	},
 	"DebtCache": {
-		"deploy": false
```
justinjmoses, line 48: Please no need to set these. Invoking `prepare-deploy` does this for you via `releases.json`

publish/deployed/mainnet/config.json
```diff
@@ -15,7 +15,7 @@
 		"deploy": false
 	},
 	"DebtCache": {
-		"deploy": false
```
justinjmoses, line 18: same as here

justinjmoses (approved):
A few comments and suggestions here and there.

contracts/BaseDebtCache.sol
```diff
@@ -190,22 +179,51 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
 
+    // Returns the total sUSD debt backed by non-SNX collateral.
+    function totalNonSnxBackedDebt() external view returns (uint excludedDebt, bool isInvalid) {
+        return _totalNonSnxBackedDebt();
+    }
+
+    function _totalNonSnxBackedDebt() internal view returns (uint excludedDebt, bool isInvalid) {
+        // Calculate excluded debt.
+        // 1. Ether Collateral.
+        if (resolver.getAddress(CONTRACT_ETHERCOLLATERAL_SUSD) != address(0)) {
```
justinjmoses, line 190: Because `DebtCache` uses `MixinResolver`, it ultimately runs `AddressResolver.requireAndGetAddress()` meaning that this can't happen. On L2, this contract equates to an empty contract: `EmptyEtherCollateral`. Thus, this check is unnecessary (as it defaults to returning `0` for total supply when empty)

contracts/BaseDebtCache.sol
justinjmoses: Same point about the `EtherCollateral` contract.

contracts/BaseDebtCache.sol
```diff
@@ -190,22 +179,51 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
+        }
+
+        if (address(issuer().synths(sETH)) != address(0) && resolver.getAddress(CONTRACT_ETHERCOLLATERAL) != address(0)) {
+            (uint sETHRate, bool sETHRateIsInvalid) = exchangeRates().rateAndInvalid(sETH);
+            isInvalid = sETHRateIsInvalid;
+            excludedDebt = excludedDebt.add(etherCollateral().totalIssuedSynths().multiplyDecimalRound(sETHRate)); // Ether-backed sETH
+        }
+
+        // 2. MultiCollateral long debt + short debt.
+        if (resolver.getAddress(CONTRACT_COLLATERALMANAGER) != address(0)) {
```
justinjmoses, line 201: and again here - on L2 we deploy `EmptyCollateralManager`

justinjmoses (changes requested):
A few suggestions, nothing major but should be considered.

contracts/BaseDebtCache.sol
liamzebedee: Good point

justinjmoses (approved):

jacko125 (approved):

PR #1245, opened 2021-05-05
Approve bridge on deploy

ajsantander:


publish/src/commands/deploy.js
```diff
@@ -1227,16 +1227,32 @@ const deploy = async ({
...
+			.call();
+		if (allowance.toString() === '0') {
+			await runStep({
+				contract: `SynthetixBridgeEscrow`,
+				target: SynthetixBridgeEscrow,
+				write: 'approveBridge',
+				writeArg: [
+					addressOf(proxyERC20Synthetix),
+					addressOf(SynthetixBridgeToOptimism),
+					w3utils.toWei('100000000'),
```
i-stam, line 1252: isn't this a bit arbitrary, it works for the tests but we should make sure that we don't deploy an escrow with an approval less than the  total withdrawal demand

i-stam: my bad, i was confused,

i-stam (approved):

PR #1247, opened 2021-05-05
Moving eth-optimism/contracts to devDep to reduce size

justinjmoses:
`eth-optimism/contracts` is pretty huge: 

image

Unfortunately, by having it as a dependency it bloats the synthetix dependency and causes downstream systems that have size limits to complain.

This PR moves the `@eth-optimism/contracts` dep to a dev dep so it doesn't add to the production load. Once Optimism can reduce their package size, we can put this back as a prod dep. (Though arguably this isn't a prod dep - it's only needed for contract code which is compiled at build time, though if a user were to grab the synthetix contracts via an npm dependency and compile them locally, then they'd need this. 🤔 )


justinjmoses:
Note: this should be fixed in an upcoming version of `@eth-optimism/contracts` - see https://github.com/ethereum-optimism/optimism/issues/777#issuecomment-833212511

i-stam (approved):

PR #1248, opened 2021-05-05
SIP 135: Adds CollateralUtil view contract to decrease size of Collateral.sol and adds getters

gberg1:
TODO will need to start eliminating things from Collateral.sol that should be moved to CollateralUtil


contracts/CollateralManager.sol
```diff
@@ -485,6 +479,35 @@ contract CollateralManager is ICollateralManager, Owned, Pausable, MixinResolver
         state.decrementShorts(synth, amount);
     }
 
+    function accrueInterest(
```
gberg1, line 482: What is the reasoning for adding accrueInterest?

PR #1251, opened 2021-05-07
Fixes Ci tests

ajsantander:
* Disables Optimism prod tests until the ops tool is fixed
* Fixes EtherCollateral prod tests to not use the Depot anymore

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,34 +146,7 @@ jobs:
       - checkout
       - attach_workspace:
           at: .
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -146,34 +146,7 @@ jobs:
       - checkout
       - attach_workspace:
           at: .
-      - run:
```
justinjmoses, line 149: perhaps comment this instead as it's temporary?

PR #1258, opened 2021-05-11
change to correct block number for v2.0-19 release

KillerByte:
The `synthetix-subgraphs` repo will be relying on this `versions.json` file in order to include relevant block numbers for all upgrades on all networks. on recent testing, we discovered that the `synthetix` subgraph was lacking a large number of stakers in its aggregate statistics.

after investigation, it was discovered that if we made the block number adjustment included herein, the number of stakers was corrected to the higher, expected number.

Please note that the block number will not match the displayed date. Should the date be updated to match the new block number, or somethign else?

/hours 1


justinjmoses:
This isn't the right way to do this - the `versions.json` is incomplete and only starts from `2.0-19`. The block number that is in there is actually correct for that version.

PR #1260, opened 2021-05-13
Deploy 2.45 to kovan

liamzebedee:


jacko125 (approved):
lgtm the fix for the burn fee rounding & checked with Iosiro.

jacko125 (approved):

PR #1267, opened 2021-05-17
update collateralState to be internal to Collateral as mapping

jacko125:
- Remove `CollateralState` for each Collateral 
- Loans are stored within each Collateral contract as a mapping now. `mapping(uint => Loan) public loans;` 
- Removed `repayAndWithdraw` and `depositAndDraw` from `Collateral.sol`
- Add interaction delay back in for functions that modify the short position.

### DIFF
--- a/contracts/Exchanger.sol
+++ b/contracts/Exchanger.sol
@@ -661,9 +661,12 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
 
     // SIP-139
     function resetLastExchangeRate(bytes32[] calldata currencyKeys) external onlyOwner {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -83,7 +83,7 @@ const constants = {
 	AST_FOLDER: 'ast',
 
 	CONFIG_FILENAME: 'config.json',
-	PARAMS_FILENAME: 'params.json',
+	PARAMS_FILENAME: 'params.js',
```
justinjmoses, line 86: this parameter `PARAMS_FILENAME` won't be needed if you use `.js`

PR #1269, opened 2021-05-17
Prepare next release entry for L2

ajsantander:


publish/releases.json
```diff
@@ -236,5 +236,15 @@
...
 		"sips": [112, 136]
+	},
+	{
+		"name": "TBD",
+		"version": {
+			"major": 2,
+			"minor": 46
+		},
+		"ovm": true,
+		"sources": ["DebtCache"],
```
justinjmoses, line 247: It was actually more than this IIRC - it's everything that SIP-136 affected, which also includes an empty `EtherWrapper`. So I think you ned `DebtCache`, `Issuer` and `EtherWrapper` (which needs to be flipped to use `EmptyEtherCollatereal` (cause it just needs `totalIssuedSynths` to be `0` for https://github.com/Synthetixio/synthetix/blob/master/contracts/BaseDebtCache.sol#L215

publish/releases.json
```diff
@@ -236,5 +236,15 @@
...
+		"sips": []
```
justinjmoses, line 248: please include `136` in here

publish/releases.json
ajsantander: Done

publish/releases.json
ajsantander: Done

justinjmoses (approved):
Let's just not merge this yet, as an L1 release may supersede it.

publish/releases.json
```diff
@@ -236,5 +236,15 @@
...
 		"sips": [112, 136]
+	},
+	{
+		"name": "TBD",
+		"version": {
+			"major": 2,
+			"minor": 46
+		},
+		"ovm": true,
+		"sources": ["DebtCache", "Issuer", "EtherWrapper"],
```
justinjmoses, line 247: ah we also need a new `SystemSettings`- as it's a property that is checked in deployment

justinjmoses: see https://github.com/Synthetixio/synthetix/blob/master/publish/src/commands/deploy.js#L2269-L2292

ajsantander:
deprecated

PR #1271, opened 2021-05-18
Replace web3 with ethers in deployment

leomassazza:
Add a wrapper object in Deployer.js to start the migration process from web3 to ethers.js.
Deployer.web3 is replaced with an intermediate object **provider** that will hold also ethers as a provider option. (Deployer.provider.web3, Deployer.provider.ethers).

In this PR also some direct internal usage of web3 or web3_utils is replaced by its ethers counterpart in deploy.js


publish/src/Deployer.js
```diff
@@ -48,15 +48,23 @@ class Deployer {
 		this.ignoreSafetyChecks = ignoreSafetyChecks;
 
 		// Configure Web3 so we can sign transactions and connect to the network.
```
ajsantander, line 50: Nit: Perhaps update the comment here to clarify why we're setting up two different providers here?

publish/src/commands/deploy.js
```diff
@@ -76,7 +76,7 @@ const deploy = async ({
 
 	// OVM uses a gas price of 0 (unless --gas explicitely defined).
 	if (useOvm && gasPrice === DEFAULTS.gasPrice) {
-		gasPrice = w3utils.toBN('0');
+		gasPrice = ethers.constants.Zero;
```
ajsantander, line 79: 💯

ajsantander (approved):
LGTM @leomassazza 

I would document the process a bit in the PR description. Why we're doing this, and what would the next steps/PRs be.

publish/src/commands/deploy.js
```diff
@@ -311,16 +311,16 @@ const deploy = async ({
 		currentSynthetixSupply = await oldSynthetix.methods.totalSupply().call();
 
 		// inflationSupplyToDate = total supply - 100m
-		const inflationSupplyToDate = w3utils
-			.toBN(currentSynthetixSupply)
-			.sub(w3utils.toBN(w3utils.toWei((100e6).toString())));
+		const inflationSupplyToDate = ethers.utils
```
justinjmoses, line 314: [np (nitpick)] i'd probably destruct these utils at the top of the file

publish/src/commands/replace-synths.js
```diff
@@ -132,7 +132,8 @@ const replaceSynths = async ({
 		},
 	};
 
-	const { web3, account } = deployer;
+	const { account } = deployer;
+	const web3 = deployer.provider.web3;
```
justinjmoses, line 136: are you sure there are no other instances like this around? would be tricky to find them using search...

justinjmoses (approved):
Just a nitpick and a seconding of Ale's comments

publish/src/commands/replace-synths.js
leomassazza: I did a search not only of mentions but also checked which instances included Deployer to see if there was any direct or indirect reference. AFAIK there's no other usage.

PR #1274, opened 2021-05-19
Introduces local L1 and L2 integration tests

ajsantander:
### General
* Deprecates `npm run test:prod:ovm`
* Deprecates local ovm CI tests
* Introduces a new type of "integration" tests that can be run granularly on top of multiple local setups

### L2 local integration tests
Runs all tests under `test/integration/l2`, which test an L2 instance in isolation.

* Start the Optimism ops tool as described here: https://github.com/ethereum-optimism/optimism#integration-tests (checkout `master`)
* Compile the artifacts / contracts, run `npx hardhat test:integration:l2 --compile`
* The first time, run `npx hardhat test:integration:l2 --deploy`
* After that, while contracts do not change, the `--deploy` flag can be avoided.

### L1 local integration tests
Idem L2 local integration tests, but targeting an isolated L1 instance.

### Dual local integration tests
As before, but deploys both L1 and L2 instances and connects them. These tests mainly targets the bridges.


jacko125 (approved):
lgtm

PR #1282, opened 2021-05-24
Updated kovan L2 feeds

justinjmoses:
Newly deployed aggregators on L2 Kovan from Chainlink


i-stam (approved):

justinjmoses:
This is not needed. Kovan L2 is already using proxies and these are the underlying aggregators.

PR #1283, opened 2021-05-24
Changes to shrink deploy size

DominicRomanowski:


contracts/Collateral.sol
```diff
@@ -629,6 +568,14 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
         }
     }
 
+    function _getLoan(uint id, address owner) internal returns (Loan storage loan) {
```
jacko125, line 571: Would this be better named `_getLoanAndAccrueInterest` as it stands currently reads as a view but actually modifies the loan.

contracts/Collateral.sol
```diff
@@ -79,18 +76,22 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
...
-        address _manager,
-        address _resolver,
+        ICollateralManager _manager,
         bytes32 _collateralKey,
         uint _minCratio,
         uint _minCollateral
-    ) public Owned(_owner) MixinSystemSettings(_resolver) {
+    ) external onlyOwner {
+        require(!initialized);
+        owner = _owner;
```
jacko125, line 89: owner is set to msg.sender already

jacko125 (approved):
one comment in initialize for the owner.

PR #1291, opened 2021-05-25
Refactoring of Exchanger.sol to reduce size on OVM

justinjmoses:
#1285 caused `Exchanger.sol` to be oversize on OVM during deployment. 

This PR overhauls `Exchanger.sol` by deduping the restricted `exchange*()` functions into one mighty handler 💪🏼 .

### DIFF
--- a/contracts/Exchanger.sol
+++ b/contracts/Exchanger.sol
@@ -661,12 +661,11 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
 
     // SIP-139
     function resetLastExchangeRate(bytes32[] calldata currencyKeys) external onlyOwner {
-        IExchangeRates exRates = exchangeRates();
-        require(!exRates.anyRateIsInvalid(currencyKeys), "Rates for given synths not valid");
+        require(!exchangeRates().anyRateIsInvalid(currencyKeys), "Rates for given synths not valid");
 
         for (uint i = 0; i < currencyKeys.length; i++) {
             bytes32 currencyKey = currencyKeys[i];
-            lastExchangeRate[currencyKey] = exRates.rateForCurrency((currencyKey));
+            lastExchangeRate[currencyKey] = exchangeRates().rateForCurrency((currencyKey));
         }
     }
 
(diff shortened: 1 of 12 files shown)


contracts/Exchanger.sol
```diff
@@ -661,12 +578,12 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
 
     // SIP-139
     function resetLastExchangeRate(bytes32[] calldata currencyKeys) external onlyOwner {
-        IExchangeRates exRates = exchangeRates();
-        require(!exRates.anyRateIsInvalid(currencyKeys), "Rates for given synths not valid");
+        (uint[] memory rates, bool anyRateInvalid) = exchangeRates().ratesAndInvalidForCurrencies(currencyKeys);
```
justinjmoses, line 581: Note: drive by cleanup - using one function instead of two.

zyzek (approved):
Looks good to me.
Only a couple of idle observations / nitpicks:

1. I note that the behaviour has changed not to process fee rewards inside `Synth._transferToFeeAddress` and `PurgeableSynth.purge`. This makes sense, just  flagging it and wondering whether it's worth a comment inline.
2. Given that `originator` argument to the exchange function is sometimes zero, and its only purpose is to process trading rewards, and it could in principle in the future be different than the actual originator, I'm wondering if this name might be a little misleading. For example in the context of SIP-140 being invoked by non-EoA addresses. I might rename it to `rewardAddress` or something.

jacko125 (approved):
lgtm. Think Anton's comment on the `originator` address could be updated to make it clearer given we'll be having support for `tx.orgin` for. 1inch

contracts/Exchanger.sol
ajsantander: 💯

ajsantander (approved):
LGTM

This is a diagram I made to check the routing, which looks good.
https://app.diagrams.net/#G1BO13UOo3nlHYjkPBV31sr8w2vECCFXHB

justinjmoses:
@zyzek @jacko125 

1. Noted. As it's just for Purging and sending to the FeePool, and it's now labelled as trading rewards, I think it's fine without a comment.
2. Great point. I renamed it all to `rewardAddress`.

PR #1294, opened 2021-05-26
Hook up futures to system debt.

zyzek:


contracts/BaseDebtCache.sol
```diff
@@ -205,8 +215,10 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
         }
 
         // 2. MultiCollateral long debt + short debt.
-        (uint longValue, bool anyTotalLongRateIsInvalid) = collateralManager().totalLong();
-        (uint shortValue, bool anyTotalShortRateIsInvalid) = collateralManager().totalShort();
+        ICollateralManager _collateralManager = collateralManager();
```
liamzebedee, line 218: Oh, nice. 🙏

contracts/BaseDebtCache.sol
```diff
@@ -151,21 +158,24 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
         view
         returns (
             uint[] memory snxIssuedDebts,
+            uint _futuresDebt,
```
liamzebedee, line 161: I can see that you're separating the futuresDebt from the excludedDebt, presumably in order to better specify what parts of the debt we're recomputing (below with recomputeFuturesDebt) - is that a fair appraisal?

contracts/DebtCache.sol
```diff
@@ -99,17 +104,24 @@ contract DebtCache is BaseDebtCache {
             _cachedSynthDebt[key] = currentSynthDebt;
         }
 
-        // Always update the cached value of the excluded debt -- it's computed anyway.
+        // Unused code until we circle back to refactor the debt system
```
liamzebedee, line 107: When do we wanna do this?

liamzebedee: Given that we're launching futures on L2 only (IIRC), should we be feature flagging this debt logic so it only runs on L2? I don't think we're using the two debt caches anymore.

liamzebedee: I suppose this ties into the broader question of where are we hosting the source of truth for the debt pool, in context of launching the futures platform? Will there be two, one for L1 and L2, and then eventually moving to the off-chain pool?

contracts/DebtCache.sol
```diff
@@ -99,17 +104,24 @@ contract DebtCache is BaseDebtCache {
...
             (uint excludedDebt, bool anyNonSnxDebtRateIsInvalid) = _totalNonSnxBackedDebt();
             anyRateIsInvalid = anyRateIsInvalid || anyNonSnxDebtRateIsInvalid;
-            excludedDebtSum = excludedDebt;
+            // TODO: This feels incorrect -- the diff needs to be computed against the total cached debt, not just the delta
+            cachedSum = cachedSum.floorsub(_cachedSynthDebt[EXCLUDED_DEBT_KEY]);
+            currentSum = currentSum.floorsub(excludedDebt);
+            _cachedSynthDebt[EXCLUDED_DEBT_KEY] = excludedDebt;
         }
 
-        cachedSum = cachedSum.floorsub(_cachedSynthDebt[EXCLUDED_DEBT_KEY]);
```
liamzebedee, line 109: How come you're moving this into the `if(recomputeExcludedDebt)` block above? Even if we don't update the excluded debt, I'm fairly sure we still want to subtract it?

liamzebedee (changes requested):

contracts/BaseDebtCache.sol
zyzek: Futures debt is a positive contribution to the debt pool, so it can't be part of the excluded debt number. But it doesn't belong in the Synth debt values list; it's a distinct category. In principle I think both the futures debt and the excluded debt should just be in their own separate functions, but overhauling this interface isn't the purpose of this PR so I wanted to keep it as light touch as possible for now.

contracts/DebtCache.sol
zyzek: In the end, we compute `debt = debt - (cachedSum - cachedExcludedDebt) + (currentSum - excludedDebtSum)` (`cachedExcludedDebt` is shorthand for `_cachedSynthDebt[EXCLUDED_DEBT_KEY]`).
But if `recomputeExcludedDebt` is false, then `excludedDebtSum` is initialised to `cachedExcludedDebt`, so we're really computing the following:

`debt - (cachedSum - cachedExcludedDebt) + (currentSum - cachedExcludedDebt)`
`debt - cachedSum + currentSum + cachedExcludedDebt - cachedExcludedDebt`
`debt - cachedSum + currentSum`

In other words, there is no effect if the we aren't recomputing the excluded debt, so we should just save the computation.

One caveat here: the floorsub screws up this logic. I haven't got a clear grasp on what it means logically for the cached value to hit the floor and the current value not to, or vice versa. One reason I've annotated that this logic doesn't feel quite right.

contracts/DebtCache.sol
zyzek: Yes, since futures will only be on L2, we will eventually as part of the port-futures-to-L2 work have to generate a stubbed-out L1 version of the futures manager that returns 0 for the debt. Or we could generate different versions of the debt cache for each layer which only takes into account the debt components it needs.

liamzebedee: In terms of the tradeoff of engineering cost, option (1) with stubs seems easier, with the tradeoff being that you have to keep up-to-date with what's deployed on L1 vs. L2 to understand the gas costs of the contracts (as some calls will be stubbed and others won't).

contracts/BaseDebtCache.sol
liamzebedee: Nice, that makes sense.

contracts/DebtCache.sol
liamzebedee: Right - this makes sense to me. Are we gonna get Iosiro to look at it too? 

If you feel like going the extra mile, I think we'd do well with a test for this logic block. Could be good to ease your mind re: the floorsub conundrum too.

@jacko125 feeding you in ☕, you should be aware of this change.

liamzebedee (approved):

contracts/DebtCache.sol
zyzek: I assume iosiro will look at all code changes; not sure how to test this without adding a public function in an inheriting test contract, and even then I don't think this behaviour will actually be used until the debt cache is refactored.

contracts/DebtCache.sol
zyzek: Agreed.

PR #1302, opened 2021-05-28
L2 shorts base collateral manager

jacko125:
Draft PR - WIP

Refactoring CollateralManager to support L1 Inverse synths for short rates calculation and L2. 

CollateralManagerWithInverse will be able to support L1 for Inverse synths 

- BaseCollateralManager
- CollateralManagerWithInverse
- CollateralManager (L2 shorts)

### DIFF
--- a/publish/deployed/local-ovm/params.js
+++ b/publish/deployed/local-ovm/params.js
@@ -2,7 +2,7 @@
 const w3utils = require('web3-utils');
 
 const params = {
-	INITIAL_ISSUANCE: '0',
+	INITIAL_ISSUANCE: w3utils.toWei('10000000'), // owner gets the SNX
 	COLLATERAL_MANAGER: {
 		SYNTHS: ['sUSD', 'sETH'],
 		SHORTS: [{ long: 'sETH' }],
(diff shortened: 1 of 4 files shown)


publish/deployed/local-ovm/params.js
```diff
@@ -2,7 +2,7 @@
 const w3utils = require('web3-utils');
 
 const params = {
-	INITIAL_ISSUANCE: '0',
+	INITIAL_ISSUANCE: w3utils.toWei('10000000'), // owner gets the SNX
```
i-stam, line 5: Why did you change this? I think that with the new testing infra this is handled by either depositing to L2 or via hack-minting.

i-stam (commented):
lgtm,  just one comment

publish/deployed/local-ovm/params.js
jacko125: @i-stam  That's right, will remove this initial issuance hack to issue owner snx as the new Infra will allow `ensureBalances()` to give the SNX to owner instead.

jacko125:
We will be deprecating L1 inverse synths shortly. We probably won't need this PR to split the CollateralManager for L1 and L2 to be handle inverse synths.

PR #1303, opened 2021-05-28
Better way to get SNX in integration tests

ajsantander:
Previously, to ensure that the owner has SNX in a standalone L2 instance (remember there is no initial supply in L2), the deploy was forced to ignore deployment parameters, which is where 0 issuance was set.

This PR removes that, because we actually need deployment parameters for other things, not just issuance. Instead, it tries different methods to get SNX for the owner in tests. In the particular standalone L2 case, it "hacks" the minting process and prints SNX for the user in `balances.js::_getSNXForOwnerOnL2ByHackMinting()`
https://github.com/Synthetixio/synthetix/pull/1303/files#diff-f4eeff274fd8a21e1f7fae9052616e130c9c52e7642d45ac742ba310881ad1caR55


i-stam (approved):
nice hack!

PR #1304, opened 2021-05-28
Reordering the deploy script for sanity

justinjmoses:
It's time to address this bad boy - the deploy script has become a nightmare. 

## Work completed:
1. Move all the deployments up top and the transactions to configure after the address resolver (this should have been done with the address resolver change; it makes sense to push ahead with all the deployments first and then move to the possibly permissioned steps afterwards). This will also make a solidity version of the deployment easier to produce. 
2. Remove flag checks of `if (useOvm)` for deployments and relying on the `config.json` of the network along with an `if (contract)` check (as if the contract isn't in the corresponding `config.json` the contract will be `undefined`)
3. Added a `--skip-feed-check` flag to speed up deployments (this check of all feeds on mainnet was painfully slow).
4. Refactor the deploy script into functional sections for better management and maintenance.


ajsantander (approved):
💯 

Yeah awesome. Let's start reducing tech debt in this area.

i-stam (approved):
:)

publish/src/commands/deploy/configure-loans.js
```diff
@@ -0,0 +1,275 @@
...
+			read: 'associatedContract',
+			expected: input => input === addressOf(CollateralManager),
+			write: 'setAssociatedContract',
+			writeArg: addressOf(CollateralManager),
+		});
+	}
+
+	console.log(gray(`\n------ INITIALISING MULTI COLLATERAL ------\n`));
+
+	if (CollateralEth && CollateralErc20 && CollateralShort) {
```
jacko125, line 72: @justinjmoses  In the L2 shorts, I've updated this part to filter undefined collateral contracts and not add them to the `collateralsArg`.

https://github.com/Synthetixio/synthetix/blob/l2-shorts-baseCollateralManager/publish/src/commands/deploy.js#L2252-L2254

publish/src/commands/deploy/configure-loans.js
```diff
@@ -0,0 +1,275 @@
...
+			],
+			expected: input => input,
+			write: 'addSynths',
+			writeArg: [
+				CollateralManagerSynths.map(key => toBytes32(`Synth${key}`)),
+				CollateralManagerSynths.map(toBytes32),
+			],
+		});
+
+		const CollateralManagerShorts = collateralManagerDefaults['SHORTS'];
```
jacko125, line 255: Note: @jacko125 Reminder to update these in the L2 shorts branch to handle different CollateralManager implementation after merging in develop.

jacko125 (approved):
Splitting out each function of the deploy script is great! 

Just left some reminders for myself to update the configure-loans and deploy-loans once merged.

PR #1305, opened 2021-05-31
Add gas reporting job parallelization on CI

mjlescano:
The objective of this PR is to be able to generate gas reports on the parallelized unit tests job. To achieve that we have to be able to generate multiple gas reports and merge them later on a separate job.

Right now, gas reporting is done by 2 steps:
1. Run `npx hardhat test --gas`, which will generate the file `gasReporterOutput.json` if running on CI
2. Then `npx codechecks codechecks.unit.yml`, which will generate a `test-gas-used.log` and upload it to Codechecks

And, for implementing parallelization, this PR generates different `gasReporterOutput-N.json` files, and then merges them into a single file to be able to upload it to codechecks using the `npx hardhat test:merge-gas-reports gasReporterOutput-*.json` command.

## Pendings
- [ ] The merge of the `gasReporterOutput.json` files has different results compared when we generate it in a single run. This has to be fixed before merging this PR


ajsantander (approved):
Wow! This is fantastic 🥰

test/contracts/AddressResolver.js
```diff
@@ -142,7 +142,7 @@ contract('AddressResolver', accounts => {
 	});
 
 	describe('rebuildCaches()', () => {
-		describe('when some MixinResolver contracts exist', () => {
+		describe('when some MixinResolver contracts exist @gas-skip', () => {
```
ajsantander, line 145: @mjlescano for some reason, this particular test fails with a smock error. I'll disable it on gas reports for now, but we should check with them.

https://github.com/ethereum-optimism/optimism/tree/develop/packages/smock
Error:
```
1) Contract: AddressResolver
       rebuildCaches()
         when some MixinResolver contracts exist
           "before each" hook: smock some MixinResolver contracts for "shows that rebuildCache() was called on imported addresses":
     TypeError: provider._init is not a function
      at exports.smockit (node_modules/@eth-optimism/smock/src/smockit/smockit.ts:22:35)
      at Context.<anonymous> (test/contracts/AddressResolver.js:149:28)
      at processImmediate (internal/timers.js:461:21)
```

ajsantander:
@mjlescano I thought it was a good idea to not duplicate running unit tests and unit tests with gas, and instead just run unit tests with gas as you suggested. Its inefficient indeed. However, some unit tests are skipped during gas reports, so I created a new job that only runs the unit tests as before.

.circleci/config.yml
```diff
@@ -359,18 +363,30 @@ jobs:
...
+          name: Pre compile
+          command: npx hardhat compile --showsize --optimizer
       - run:
           name: Test and output gas used
           command: |
             set +e
             circleci tests glob 'test/contracts/*.js' |
             circleci tests split |
-            xargs npm test
+            xargs npm test -- --gas
```
ajsantander, line 378: I had to remove the --optimize option for now because of a problem with hardhat that should be fixed in 2.4.0 soon.

Added a ticket to re-add when done.

PR #1307, opened 2021-05-31
Introduces forking via integration tests

ajsantander:
This PR introduces the ability to simulate an L1 deploy and run integration tests against it, with:
1. `npm run fork`
2. `npx hardhat test:integration:l1 --fork --compile --deploy`

Which runs tests in `test/integration/l1` against a fork of mainnet. The PR also introduces a CI job for the test. Note that not specifying `--deploy` will simply run the integration tests against the current mainnet deployment.

PS: This also fixes a problem with the old ovm prod tests.

### DIFF
--- a/test/integration/utils/users.js
+++ b/test/integration/utils/users.js
@@ -1,11 +1,34 @@
 const ethers = require('ethers');
+const { getUsers } = require('../../../index');
 
 async function loadUsers({ ctx, network }) {
-	const wallets = _createWallets({ provider: ctx.provider });
+	let wallets;
+
+	if (!ctx.fork) {
+		wallets = _createWallets({ provider: ctx.provider });
+	} else {
+		wallets = _getWallets({ provider: ctx.provider });
+	}
 
 	ctx.users = {};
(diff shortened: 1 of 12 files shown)


test/integration/utils/users.js
```diff
@@ -1,13 +1,36 @@
 const ethers = require('ethers');
+const { getUsers } = require('../../../index');
 
 async function loadUsers({ ctx, network }) {
-	const wallets = _createWallets({ provider: ctx.provider });
+	let wallets;
```
liamzebedee, line 5: btw, it'd be useful to have these wallets in `ctx` at some point 😃 - since OVM doesn't have snapshots, I'm resorting to using different accounts right now in order to test things in isolation.

liamzebedee (approved):
Nice work!

test/integration/utils/users.js
ajsantander: They are in ctx.users

liamzebedee: @ajsantander we only have two users there right now? The rest of the wallets would be useful

PR #1314, opened 2021-06-04
Futures order fee charged from margin

zyzek:


contracts/FuturesMarket.sol
```diff
@@ -567,27 +574,40 @@ contract FuturesMarket is Owned, Proxyable, MixinSystemSettings, IFuturesMarket
...
-            !invalid &&
+            !invalid && // Price is valid
             price != 0 &&
-            _orderPending(order) &&
-            order.roundId < _currentRoundId(exRates) &&
-            !_canLiquidate(positions[account], _liquidationFee(), fundingSequence.length, price, invalid);
+            _orderPending(order) && // There is actually an order
+            order.roundId < _currentRoundId(exRates) && // A new price has arrived
+            !_canLiquidate(position, _liquidationFee(), fundingSequenceIndex, price, invalid) && // No existing position can be liquidated
+            0 <= _marginPlusProfitFunding(position, fundingSequenceIndex, price).sub(int(order.fee)); // Margin has not dipped negative due to fees, profit, funding
```
liamzebedee, line 610: We can confirm an order where the `margin + profit - funding` is 0. Cool.

contracts/FuturesMarket.sol
```diff
@@ -928,8 +936,11 @@ contract FuturesMarket is Owned, Proxyable, MixinSystemSettings, IFuturesMarket
         require(!_canLiquidate(position, _liquidationFee(), fundingIndex, price, invalid), "Position can be liquidated");
 
         // We weren't liquidated, so we realise the margin to compute the new position size.
-        uint margin = _realiseMargin(position, fundingIndex, price, 0);
-        int newSize = int(margin).multiplyDecimalRound(order.leverage).divideDecimalRound(int(price));
+        // The fee is deducted at this stage; the transaction will revert if the realised margin minus the fee is negative.
+        uint margin = _realiseMargin(position, fundingIndex, price, -int(order.fee));
```
liamzebedee, line 940: So the fee is charged at time of confirmation, though it is set according to the skew at the time the order is submitted. 

I read in SIP-80 that the fee is an incentive to reduce skew:

> Users pay a fee whenever they open or increase a position. However, we wish to incentivise reduction of skew, so we distinguish between maker and taker fees. A maker is someone reducing skew and a taker is someone increasing it, and so we charge makers less than takers, possibly even zero insofar as this is possible in the presence of front-running

What would be the effect on the incentive if the market skew flips (eg. from long to short) in the tx before a user is confirms their order, potentially as a frontrun?

contracts/FuturesMarket.sol
```diff
@@ -928,8 +936,11 @@ contract FuturesMarket is Owned, Proxyable, MixinSystemSettings, IFuturesMarket
...
+        // The order size is computed pre-fee, so the order size is accurate, though their leverage may
+        // be slightly higher than what was requested if the fee is nonzero.
+        int newSize = int(margin.add(order.fee)).multiplyDecimalRound(order.leverage).divideDecimalRound(int(price));
```
liamzebedee, line 943: Not sure if this applies in futures - but would it be beneficial to introduce some slippage parameter that bounds the user's margin in this case? Like Uniswap where your order is cancelled if the price slips too much.

liamzebedee (changes requested):

contracts/FuturesMarket.sol
zyzek: We will potentially over- or under-charge in some cases, so there is a slight incentive for the market to overcorrect. Ultimately I'm not to concerned, though, because the real lever is supposed to be the funding rate. From this perspective, the fee being computed ahead of time is intended to make sure the cost of entering a position is known ahead of time, so that the funding rate is more reliable to arbitrage.

contracts/FuturesMarket.sol
zyzek: Yes, I agree this would be a great feature to have. It's actually one of the stretch tasks listed here: https://github.com/Synthetixio/issues/issues/102
If you want to implement it, feel free the break the stretch task out into a ticket of its own.

contracts/FuturesMarket.sol
liamzebedee: Okay. We're comfortable with saying that the fee is a fixed fee, set ahead of time, paid to stakers. I'm okay with that.

test/contracts/FuturesMarket.js
```diff
@@ -288,6 +288,34 @@ contract('FuturesMarket', accounts => {
 	describe('Order fees', () => {
 		const margin = toUnit('1000');
 
+		// In this section, where inscrutable numbers are being compared, we're
```
liamzebedee, line 249: 🙏

PR #1315, opened 2021-06-04
Completely deprecates prod tests

ajsantander:
These are to be replaced with the new integration tests.


justinjmoses (approved):
The 👑  is dead. Long live the 👑 .

PR #1316, opened 2021-06-06
replace web3 with ethers in NounceManager and others in Deployer

leomassazza:
This PR is to continue replacing web3 references with ethers.js. In this case the usage of of web3 was removed from NonceManager and it's usages and almost all references of web3 were removed from Deployer (missing only one)


leomassazza:
Closed. Covered here: https://github.com/Synthetixio/synthetix/pull/1418

PR #1317, opened 2021-06-07
Futures record fees properly

zyzek:


contracts/FeePool.sol
```diff
@@ -725,17 +732,19 @@ contract FeePool is Owned, Proxyable, LimitedSetup, MixinSystemSettings, IFeePoo
...
-        bool isCollateral = collateralManager().hasCollateral(msg.sender);
-        bool isEtherWrapper = msg.sender == address(etherWrapper());
 
-        require(
-            isExchanger || isSynth || isEtherCollateralsUSD || isCollateral || isEtherWrapper,
-            "Only Internal Contracts"
-        );
+    function _isInternalContract(address account) internal view returns (bool) {
+        return
+            account == address(exchanger()) ||
```
liamzebedee, line 738: The previous is more readable imo, was this a gas optimization? Maybe you could reorder the special case lines 739 and 741 to the top/bottom and see how that looks?

contracts/FuturesMarket.sol
```diff
@@ -955,7 +951,7 @@ contract FuturesMarket is Owned, Proxyable, MixinSystemSettings, IFuturesMarket
 
         // Send the fee to the fee pool
         if (0 < order.fee) {
-            _manager().issueSUSD(_feePool().FEE_ADDRESS(), order.fee);
+            _manager().payFee(order.fee);
```
liamzebedee, line 954: Fee payment flow for orders:

 1. Order is submitted, fee (sUSD) is burnt
 2. Order is confirmed, manager.payFee(), fee (sUSD) is issued to fee pool

Instead of burning the fee, why don't we escrow it? It seems like it would add some noise to the debt composition of sUSD, and I can't really see the benefit by burning it and then reissuing it?

contracts/FuturesMarketManager.sol
```diff
@@ -131,6 +138,12 @@ contract FuturesMarketManager is Owned, MixinResolver, Proxyable, IFuturesMarket
         _sUSD().burn(account, amount);
     }
 
+    function payFee(uint amount) external onlyMarkets {
+        IFeePool pool = _feePool();
+        _sUSD().issue(pool.FEE_ADDRESS(), amount);
+        pool.recordFeePaid(amount);
```
liamzebedee, line 144: Side note - we have a nice pattern whereby if you transfer directly to the fee address, recordFeePaid is called as part of it. This means you don't need separate permissioning in order to call `recordFeePaid`, and you might be able to bundle this function inside of the FuturesMarket, avoiding an external call and making the relationships simpler.

https://github.com/Synthetixio/synthetix/blob/0e1d8d5e302d2beeead45569773e3b47e1f6319c/contracts/Synth.sol#L129-L145

liamzebedee (changes requested):

contracts/FeePool.sol
zyzek: Yeah, it's so that we short circuit without executing these things. The order is supposed to be based on the frequency that they are executed with. Though perhaps futures and ether wrapper should be higher.

contracts/FuturesMarket.sol
zyzek: The fee will ultimately not be burnt, but deducted from their margin in #1314.

contracts/FuturesMarketManager.sol
zyzek: Hm, I am concerned that this would require extra accounting and debt management logic to actually keep the sUSD around, particularly if the market has to pay out more sUSD than it has, in which case it'll have to issue anyway.

liamzebedee: > if the market has to pay out more sUSD than it has

liamzebedee: Or does the funding rate payment fall under the guise of this `payFees` logic?

liamzebedee: Okay, after a nice long discussion with Anton, I've gotten a better understanding of the funding rate mechanism. IN particular, I wanted to understand how the fees are distributed to stakers. In the current design we have, the margin is deposited, and upon submission/confirmation, the debt composition is changed by the issuing and burning of an sUSD amount of fees. There is also additional sUSD paid to the fee pool due to the funding rate scheme, which occurs when a user realises their margin. 

We agreed it would be nice to explicitly see both of these fees sent, rather than the debt composition being implicitly affected, though it requires a bit more accounting complexity in the logic. So I'm happy to resolve this for now, and keep a keen eye on https://github.com/Synthetixio/issues/issues/87.

liamzebedee: Just an example for anyone who comes across this - 

```
when the heavier side is losing money, the market is reducing debt.
100 btc [heavy] * funding rate (1%/day)
heavy side (100btc) losing 1BTC per day
light side (50btc)  gaining 0.5BC (1% of 50) per day
net funding flow out the market of 0.5BTC
```

liamzebedee (approved):

PR #1319, opened 2021-06-07
Adds ether wrapper integration tests

ajsantander:
Adds the most basic EtherWrapper tests to L1 (local and mainnet forks), by minting and burning sETH via the wrapper.

Fixed a couple of things along the way, and added the ability to guarantee ETH and WETH balances for users. The `balances.js` is definitely getting a little complex, so a future PR could perhaps break it up in files.

@liamzebedee I abstracted as much as I could to a behavior so we can reuse most of it when (and I guess _if_) we take this to L2. I don't make a distinction between native/non-native, not sure if this is correct.


hardhat.config.js
```diff
@@ -50,14 +50,12 @@ module.exports = {
 			allowUnlimitedContractSize: true,
 			gasPrice: GAS_PRICE,
 			initialDate: new Date(inflationStartTimestampInSecs * 1000).toISOString(),
-			timeout: 600000,
```
ajsantander, line 53: This is something not directly associated to EtherWrapper, which I've been meaning to fix for a while: Properly extending timeouts when using integration tests (especially forking).

test/integration/abis/WETH.json
```diff
@@ -0,0 +1,279 @@
+[
```
ajsantander, line 1: I need WETH's ABI in the context of a mainnet fork, in which case I won't be able to find it in deployment.json.

test/integration/behaviors/stake.behavior.js
```diff
@@ -106,7 +106,7 @@ function itCanStake({ ctx }) {
 				const newDebtsUSD = await Synthetix.debtBalanceOf(user.address, toBytes32('sUSD'));
 				const debtReduction = debtsUSD.sub(newDebtsUSD);
 
-				const tolerance = ethers.utils.parseUnits('40', 'gwei');
+				const tolerance = ethers.utils.parseUnits('0.01', 'ether');
```
ajsantander, line 109: 1 cent is ok. 40gwei was way too fragile.

test/integration/utils/balances.js
```diff
@@ -3,24 +3,77 @@ const { deposit } = require('./bridge');
 const { toBytes32 } = require('../../..');
 
 async function ensureBalance({ ctx, symbol, user, balance }) {
-	const token = _getTokenFromSymbol({ ctx, symbol });
-	const currentBalance = await token.balanceOf(user.address);
+	const currentBalance = await _readBalance({ ctx, symbol, user });
```
ajsantander, line 6: Revamped this util to support ETH.

test/integration/utils/bootstrap.js
```diff
@@ -16,6 +16,11 @@ function bootstrapL1({ ctx }) {
 		ctx.provider = _setupProvider({ url: `${hre.config.providerUrl}:${hre.config.providerPort}` });
 
 		await loadUsers({ ctx });
+		if (ctx.fork) {
```
ajsantander, line 19: Ensures all users have ETH in a fork.

test/integration/utils/users.js
```diff
@@ -2,18 +2,22 @@ const ethers = require('ethers');
 const { getUsers } = require('../../../index');
 
 async function loadUsers({ ctx, network }) {
-	let wallets;
+	let wallets = [];
```
ajsantander, line 5: This now combines system users and users created in the local node (which have ETH).

hardhat/tasks/task-test-integration-dual.js
```diff
@@ -14,7 +14,7 @@ task('test:integration:dual', 'run integrated layer 1 and layer 2 production tes
 		const providerPortL1 = (hre.config.providerPortL1 = '9545');
 		const providerPortL2 = (hre.config.providerPortL2 = '8545');
 
-		const timeout = 5 * 60 * 1000;
```
leomassazza, line 17: np: add an inline comment to clarify it's 10'

test/integration/utils/balances.js
```diff
@@ -3,24 +3,77 @@ const { deposit } = require('./bridge');
...
 	if (currentBalance.lt(balance)) {
 		const amount = balance.sub(currentBalance);
 
-		await _getTokens({ ctx, symbol, user, amount });
+		await _getAmount({ ctx, symbol, user, amount });
 	}
 }
 
-async function _getTokens({ ctx, symbol, user, amount }) {
+async function _readBalance({ ctx, symbol, user }) {
```
leomassazza, line 15: To make it easier to read I'd get the token for all cases and do the if (symbol === 'ETH') just for getting the balance in ERC20 mode or native for ETH

test/integration/utils/balances.js
```diff
@@ -3,24 +3,77 @@ const { deposit } = require('./bridge');
...
 		await _getSNX({ ctx, user, amount });
+	} else if (symbol === 'WETH') {
+		await _getWETH({ ctx, user, amount });
 	} else if (symbol === 'sUSD') {
 		await _getsUSD({ ctx, user, amount });
+	} else if (symbol === 'ETH') {
+		await _getETHFromOtherUsers({ ctx, user, amount });
 	} else {
-		// TODO: will need to get SNX and then exchange
+		throw new Error('TODO: Ability to ensure non sUSD synth balance not yet implemented');
```
leomassazza, line 35: np: throw error Symbol not yet supported (and maybe your TODO)

test/integration/utils/balances.js
```diff
@@ -3,24 +3,77 @@ const { deposit } = require('./bridge');
...
+}
+
+async function _getETHFromOtherUsers({ ctx, user, amount }) {
+	for (const otherUser of Object.values(ctx.users)) {
+		if (otherUser.address === user.address) {
+			continue;
+		}
+
+		const otherUserBalance = await ctx.provider.getBalance(otherUser.address);
+		if (otherUserBalance.gte(ethers.utils.parseEther('1000'))) {
```
leomassazza, line 46: it can fail if the amount to get is > 1000

leomassazza:
LGTM. Some np, but nothing big

test/integration/utils/contracts.js
```diff
@@ -17,6 +17,24 @@ function connectContracts({ ctx }) {
...
 	});
+
+	_ensureWETH({ ctx });
+}
+
+function _ensureWETH({ ctx }) {
+	if (!ctx.contracts.WETH) {
+		if (ctx.fork) {
+			ctx.contracts.WETH = new ethers.Contract(
+				'0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2', // WETH on mainnet L1
```
justinjmoses, line 28: should we put this in `index.js` instead?

justinjmoses (approved):
💯  - just a couple of comments.

test/integration/utils/contracts.js
ajsantander: I thought about it a bit... this is something needed for tests right? So we might prefer to store these sort of things at the test/integration layer? Like this PR does with the WETH ABI?

test/integration/utils/balances.js
ajsantander: Yep, I'm aware. This should be ok for now IMO.

test/integration/utils/contracts.js
liamzebedee: @ajsantander I put the addresses in `index.js`, since we use them in the deploy script too.  https://github.com/Synthetixio/synthetix/blob/develop/index.js#L159

PR #1321, opened 2021-06-08
Updates Optimism's watcher dependency

ajsantander:
This version is supposed to report tx finalization reversions instead of just hanging when they occur.


justinjmoses (approved):

ajsantander:
Deprecated - newer version already merged to develop

PR #1322, opened 2021-06-08
Enables separate folder compilation for integration tests

ajsantander:
This allows integration tests to reuse compilation artifacts in `./build` and `./build-ovm` folders, greatly increasing the flexibility of integration tests.

For example, previously you could run `npx hardhat test:integration:l1 --compile --deploy`, and then re-run without `--compile` or even without `--deploy`.

Then you could do the same with the l2 tests. However, if you went back to l1, or dual, you'd be forced to compile again every time.

Holding the artifacts in different folders, allows you to only have to compile when contracts change, and not when switching between l1, l2, or dual integration tests.

### DIFF
--- a/hardhat/tasks/task-test-integration-dual.js
+++ b/hardhat/tasks/task-test-integration-dual.js
@@ -1,3 +1,7 @@
+const path = require('path');
+const {
+	constants: { BUILD_FOLDER },
+} = require('../..');
 const { task } = require('hardhat/config');
 const {
 	compileInstance,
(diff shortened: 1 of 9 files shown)


hardhat/tasks/task-test-integration-dual.js
```diff
@@ -6,28 +10,58 @@ const {
...
 		const providerPortL2 = (hre.config.providerPortL2 = '8545');
+		const buildPathEvm =
+			taskArguments.buildPathEvm || path.join(__dirname, '..', '..', BUILD_FOLDER);
+		const buildPathOvm =
+			taskArguments.buildPathOvm || path.join(__dirname, '..', '..', `${BUILD_FOLDER}-ovm`);
 
 		const timeout = 600000; // 10m
 		hre.config.mocha.timeout = timeout;
-		hre.config.mocha.bail = false;
+		hre.config.mocha.bail = true;
```
justinjmoses, line 33: fail on error?

justinjmoses (approved):
Nice. Just one question

hardhat/tasks/task-test-integration-dual.js
ajsantander: Yes, since these tests are so slow, I'd rather see the error quickly than have to wait 30m to see the same error four time.
Bail makes it so that any error cancels the test suite.

PR #1323, opened 2021-06-08
Merge all integration test tasks into one

ajsantander:
Also simplifies parameter usage and fixes a couple of things in integration tests along the way.


leomassazza:
LGTM

PR #1324, opened 2021-06-09
Add optimism scripts to run Optmism via hardhat

leomassazza:
This PR adds a new task to hardhat to easily build and start Optimism network using the verified commit.

There are 2 main scripts:
- `npx hardhat ops --start` to start a optimism chain.  
- `npx hardhat ops --stop` to stop the optimism chain

Start without any flag or parameter will start the docker chain in ~/optimism. If the path or the docker images doesn't exist it will create the path, clone the optimism repo, checkout a verified commit, build the code and images and start the chain.

The path and commit hash can be override via `-optimism-path `or `--optimism-commit`.

There are also some useful options to get a fresh copy of the repository, just build the code, the images or do all the steps together.
- `--fresh` will prune docker images and remove the optimism-path and clone optimism repository
- `--build` will get the get the verified commit (or the one specified with --optimism-commit) and build the code (yarn && yarn build)
- `--build-ops` will build the docker images


ajsantander (changes requested):
It does look good, and I do wanna merge it, but I'm a bit hesitant to the idea of adding 6 new npm scripts for something that is used only in a particular type of tests. Devs looking into this will probably be confused with all the different scripts.

IMO, we should have a single entry point (probably a hardhat task) that by default figures out what needs to be done. All these different flavors can be flags in the task. If you do go that way, much of this would have to be re-implemented via javascript, which should be fine. You can look at .circleci/pack.js for usage of the "execa" library, to see how to call shell commands from javascript.

Finally, I would suggest that we modify .circleci/src/jobs/job-integration-tests.yml to use these commands. Don't forget to run `npm run build:ci` afterwards to regenerate the master .circleci/config.yml file from the .circleci/src folder.

leomassazza:
> It does look good, and I do wanna merge it, but I'm a bit hesitant to the idea of adding 6 new npm scripts for something that is used only as a pre-step of running a particular type of tests. Devs looking into this will probably be confused as hell with all the different scripts.
> 
> IMO, we should have a single entry point (probably a hardhat task `npx hardhat ops`) that by default figures out what needs to be done. All these different flavors can be flags in the task that expose fine tuned functionality for advanced users. If you do go that way, much of this would have to be re-implemented via javascript, which should be fine. You can look at .circleci/pack.js for usage of the "execa" library, which allows you to call shell commands from javascript.
> 
> Finally, I would suggest that we modify .circleci/src/jobs/job-integration-tests.yml to use these commands. Don't forget to run `npm run build:ci` afterwards to regenerate the master .circleci/config.yml file from the .circleci/src folder.

@ajsantander changes applied

hardhat/tasks/task-ops.js
```diff
@@ -0,0 +1,154 @@
...
+	{ service: 'l2geth', image: 'ethereumoptimism/l2geth' },
+	{ service: 'relayer', image: 'ethereumoptimism/message-relayer' },
+];
+
+task('ops', 'Run Optimism chain')
+	.addFlag('fresh', 'Clean up docker and get a fresh clone of the optimism repository')
+	.addFlag('build', 'Get the right commit and builds the repository')
+	.addFlag('buildOps', 'Build fresh docker images for the chain')
+	.addFlag('start', 'Start the latest build')
+	.addFlag('stop', 'Deploy an l1 instance before running the tests')
```
ajsantander, line 20: Wrong description?

leomassazza: yup. copy/paste failed

hardhat/tasks/task-ops.js
```diff
@@ -0,0 +1,154 @@
...
+	.setAction(async (taskArguments, hre, runSuper) => {
+		const opsPath = taskArguments.optimismPath.replace('~', homedir);
+		const opsCommit = taskArguments.optimismCommit;
+
+		console.log(gray('optimism folder:', opsPath));
+
+		if (taskArguments.stop) {
+			console.log(yellow('stoping'));
+			if (fs.existsSync(opsPath)) {
+				await _stop({ opsPath });
```
ajsantander, line 36: Why is this `_stop` awaited? Also other `_stop`'s don't seem to be awaited.

leomassazza: Good catch! I forgot to remove the await when moved from asynch to synch in execa

ajsantander (changes requested):
This is awesome. Super clean and easy to read.

However, I'd love to see the CLI output in the start process and keep the process open.
This should be possible with `docker-compose up` (with no `-d` option).

I tried a few things but couldn't get something reliable. Honestly, I've always found execa weird and hard to get right if you're not doing basic things. And I feel like it's gotten weirder. Thus, I recommend that we go low level to node's child_process api (https://nodejs.org/api/child_process.html#child_process_class_childprocess), and gain more control. Something like this should work:

```
async function _start({ opsPath }) {
	console.log(gray('  start ops'));
	spawn('sh', ['-c', `cd ${opsPath}/ops && docker-compose up`], { stdio: 'inherit' });
	await new Promise(() => {}); // Keeps the process open
}
```

ajsantander (approved):
Thank you!

PR #1328, opened 2021-06-09
Remove web3 from the deploy script test

ajsantander:
Superficially removes web3 from `test/publish/index.js` in favor of ethers.

I don't expect anyone to fully check the diff in this file, but I don't think its necessary. It may not be the prettiest thing but it works. Removing web3 is not easy!

Something interesting in this PR is how utils are starting to surface from `test/integration/utils` into `test/test-utils`. My plan is that debt reduction will end up with 1 or probably 2 util folders, based on ethers, for the whole repo.


leomassazza:
LGTM

leomassazza (approved):
LGTM

PR #1331, opened 2021-06-11
Unpins the usage of the Optimism ops tool and starts managing L2 gas

ajsantander:
Allows us to run integration tests and deploy with the latest Optimism ops tool, which begins to use the new L2 gas rules:
https://www.notion.so/How-to-pay-Fees-in-Optimistic-Ethereum-f706f4e5b13e460fa5671af48ce9a695

This is achieved by:
* Not specifying a gas limit and using estimateGas in web3 instead.
* Removing the default ops commit in `npx hardhat ops`
* Removing web3 in `connect-bridge.js` in favour of ethers, because ethers auto estimates gas.

### DIFF
--- a/index.js
+++ b/index.js
@@ -97,7 +97,7 @@ const constants = {
 	ZERO_ADDRESS: '0x' + '0'.repeat(40),
 	ZERO_BYTES32: '0x' + '0'.repeat(64),
 
-	OVM_MAX_GAS_LIMIT: '8999999',
+	OVM_GAS_PRICE_GWEI: '0.0',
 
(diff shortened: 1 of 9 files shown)


index.js
```diff
@@ -97,7 +97,7 @@ const constants = {
 	ZERO_ADDRESS: '0x' + '0'.repeat(40),
 	ZERO_BYTES32: '0x' + '0'.repeat(64),
 
-	OVM_MAX_GAS_LIMIT: '8999999',
+	OVM_GAS_PRICE_GWEI: '0.0',
```
ajsantander, line 100: The current ops tool tolerates this not to be 0.015 gwei, but it will soon. So that's why I'm introducing the constant here, for easy change.

PR #1332, opened 2021-06-14
Removes web3 from nominate script

ajsantander:
Web3 is easy to remove from admin scripts like nominate.js. This PR adds an integration test for it, then replaces usage of web3 to ethers.

### DIFF
--- a/publish/src/commands/deploy/index.js
+++ b/publish/src/commands/deploy/index.js
@@ -224,6 +224,7 @@ const deploy = async ({
 		oracleAddress,
 	} = await systemAndParameterCheck({
 		account,
+		buildPath,
 		addNewSynths,
 		concurrency,
 		config,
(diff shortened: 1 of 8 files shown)


publish/src/commands/deploy/index.js
```diff
@@ -224,6 +224,7 @@ const deploy = async ({
 		oracleAddress,
 	} = await systemAndParameterCheck({
 		account,
+		buildPath,
```
ajsantander, line 227: On the go mini-improvement... deploy script now prints the build path that will be used. Useful for integration tests where buildPath can be `./build` or `./build-ovm`.

publish/src/commands/nominate.js
```diff
@@ -27,8 +26,15 @@ const nominate = async ({
 	gasPrice,
 	gasLimit,
 	useOvm,
+	privateKey,
```
ajsantander, line 29: Adds a few new options required for programatic use and testing.

test/integration/behaviors/nominate.behavior.js
```diff
@@ -0,0 +1,82 @@
+const { getLocalPrivateKey } = require('../../test-utils/wallets');
```
ajsantander, line 1: Test first. Change after.
Then test again :D

leomassazza:
LGTM

PR #1336, opened 2021-06-15
Removes web3 from the owner script

ajsantander:
Also,
* Allows the owner script to be run on a fork.
* Adds an integration test for the owner script.
* Removes auto detection of isContract so that the script can be used in L2.


publish/src/commands/nominate.js
```diff
@@ -29,12 +29,7 @@ const nominate = async ({
 	privateKey,
 	providerUrl,
 	yes,
-	quiet,
 }) => {
-	if (quiet) {
```
ajsantander, line 34: Decided to get rid of this feature I recently introduced, because it mutes logs in tests and can lead to bizarre behavior.

publish/src/commands/owner.js
```diff
@@ -45,6 +44,7 @@ const owner = async ({
 	yes,
 	useOvm,
 	providerUrl,
+	useFork,
```
ajsantander, line 47: Integration tests also target an L1 mainnet fork

publish/src/commands/owner.js
```diff
@@ -297,24 +304,24 @@ const owner = async ({
...
 				} else {
-					const tx = await web3.eth.sendTransaction({
-						from: account,
-						to: deployedContract.options.address,
-						gasPrice: w3utils.toWei(gasPrice, 'gwei'),
-						gas: gasLimit,
+					const tx = await wallet.sendTransaction({
+						to: deployedContract.address,
+						gasPrice: ethers.utils.parseUnits(gasPrice, 'gwei'),
+						gasLimit: ethers.BigNumber.from(gasLimit),
```
ajsantander, line 319: Not entirely sure why I had to big numberify gasLimit here. Spent a while trying to use a string as usual but it would fail with "invalid hexlify value". Maybe Ethers needs a big number for gas limit when sending a generic transaction?

publish/src/safe-utils.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const Web3 = require('web3');
```
ajsantander, line 3: Since owner.js is no longer passing a build web3 instance to this util, this util is now in charge of building it for itself.

test/integration/behaviors/owned.behavior.js
```diff
@@ -0,0 +1,61 @@
+const {
+	acceptOwnership,
+	nominateOwnership,
+	verifyNomination,
+	verifyOwnership,
+} = require('../utils/owned');
```
ajsantander, line 6: Moved the calls to the owner and nominate scripts to a util so that the test is clean and easier to read.

test/test-utils/wallets.js
```diff
@@ -6,7 +6,7 @@ function loadLocalWallets({ provider }) {
 	for (let i = 0; i < 10; i++) {
 		const privateKey = getLocalPrivateKey({ index: i });
 		const wallet = new ethers.Wallet(privateKey, provider);
-		wallet.privateKey = privateKey;
+		wallet.pk = privateKey;
```
ajsantander, line 9: Also, not sure why attempting to write `wallet.privateKey` instead of `wallet.pk` failed.

publish/src/commands/owner.js
```diff
@@ -85,10 +88,8 @@ const owner = async ({
 		privateKey = envPrivateKey;
 	}
 
-	const web3 = new Web3(new Web3.providers.HttpProvider(providerUrl));
+	const provider = new ethers.providers.JsonRpcProvider(providerUrl);
 
-	const code = await web3.eth.getCode(newOwner);
```
ajsantander, line 90: On evm, we could determine if the new owner is an EOA or a multisig wallet by checking if there was code in the address. However, this doesn't work for L2 since all EOA's in Optimism have code. Instead, we manually pass isContract now.

test/test-utils/wallets.js
leomassazza: in ethers Wallet.ts code you can see: 
    `get privateKey(): string { return this._signingKey().privateKey; }`
    That means its immutable

leomassazza:
LGTM

leomassazza (approved):

PR #1338, opened 2021-06-15
Fix for SupplySchedule - should use the ERC20 proxy

justinjmoses:
There is a current bug in the deploy script - SupplySchedule is configured to use the legacy proxy. This has two issues:

1. When the legacy proxy is removed and longer connected to the system, `Synthetix.mint()` (called each week to mint inflation) will fail and;
2. Once terminal inflation hits, this line will fail: https://github.com/Synthetixio/synthetix/blob/master/contracts/SupplySchedule.sol#L99 - this is because the old proxy mutates `messageSender` in every function call, including views like `totalSupply` which break the `IERC20` interface. 

The solution is to both update the deploy script and for the pDAO to change this parameter on-chain (Council have agreed there is no need for a SCCP).


publish/src/commands/deploy/configure-legacy-settings.js
```diff
@@ -248,7 +248,7 @@ module.exports = async ({
 			read: 'synthetixProxy',
 			expected: input => input === addressOf(ProxySynthetix),
```
zyzek, line 249: Does the expected address not also need to be updated in order for this to actually trigger?

liamzebedee: Yep, same question as Anton.

justinjmoses: Derrrr. Thanks guys.

i-stam (approved):

PR #1340, opened 2021-06-16
SIP-150 Fix excluded debt calculation for partial snapshots.

zyzek:
Someone at yEARN found that if you issue some sUSD, your debt balance is lower than the sUSD you actually issued.
The problem was that the issued amount was not being properly added to the cached system debt value, so the new issuer's share of debt was too low by exactly the factor accounted for by their fraction of the total debt pool.

This issue has been latent for some time, but was not revealed until the ether wrapper cap got high enough that the value of non-SNX-backed debt exceeded the value of circulating sUSD. Once this occurred, the delta for partial debt snapshot updates due to sUSD issuance always ended up being set to zero, so the newly-issued synths were not recorded.

There is an accompanying SIP: https://github.com/Synthetixio/SIPs/pull/547


contracts/DebtCache.sol
```diff
@@ -99,24 +98,22 @@ contract DebtCache is BaseDebtCache {
...
-
         // Compute the difference and apply it to the snapshot
         if (cachedSum != currentSum) {
             uint debt = _cachedDebt;
             // This requirement should never fail, as the total debt snapshot is the sum of the individual synth
             // debt snapshots.
             require(cachedSum <= debt, "Cached synth sum exceeds total debt");
             debt = debt.sub(cachedSum).add(currentSum);
+
+            // Unused code until we circle back to refactor the debt system
```
justinjmoses, line 109: a touch more comments here please - could you label this SIP-150 and explain why this is dead? (why this is never hit)

justinjmoses (approved):
Approved but please add more commentary around the section that is dead code :pray:

contracts/DebtCache.sol
zyzek: Done

PR #1341, opened 2021-06-16
Removes web3 from the purge script

ajsantander:
Pretty much a simple port like the other ones going on, except that this one uses `publish/src/util.js::performTransactionalStep`, which is a function that will only transact if needed, and append to `owner-actions.json` if the available signer is not the owner of the contract being interacted with.

Since we don't want to port `performTransactionalStep` right now and break every other script using it (i.e. the deploy script), I created a new `publish/src/command-utils/` folder with a `transact.js` file that contains a direct port of `performTransactionalStep` which uses ethers.js. The "old" function is renamed to `performTransactionalStepWeb3` for distinction.

The interface of `performTransactionalStep` is identical to `performTransactionalStepWeb3`. It could have been improved, but keeping it the same should facilitate the removal process of web3.

Finally, I could have continued to split `publish/src/util.js` into `publish/src/command-utils` but I didn't want to stray away from the main topic of the PR too much, for ease of reviewing.


publish/src/commands/purge-synths.js
```diff
@@ -75,17 +77,27 @@ const purgeSynths = async ({
...
+
+	let wallet;
+	if (useFork) {
+		const account = getUsers({ network, user: 'owner' }).address; // protocolDAO
+		wallet = provider.getSigner(account);
+	} else {
+		wallet = new ethers.Wallet(privateKey, provider);
+	}
+	console.log(wallet);
+	if (!wallet.address) wallet.address = wallet._address;
```
ajsantander, line 99: @leomassazza we should probably think a little more about this.
In one case, wallet will have `.address` and be read only, in the other, it will have `._address`. Not sure if we should raise this to ethers.js, and/or make sure we always do this so that `wallet.address` is available with a util or something.

publish/src/commands/purge-synths.js
```diff
@@ -50,7 +52,7 @@ const purgeSynths = async ({
 	});
 
 	if (synthsToPurge.length < 1) {
-		console.log(gray('No synths provided. Please use --synths-to-remove option'));
+		console.log(gray('No synths provided. Please use --synths-to-purge option'));
```
leomassazza, line 55: 👍

leomassazza (approved):
LGTM

PR #1342, opened 2021-06-17
Remove web3 from remove-fee-periods

leomassazza:
Another step in the web3 removal process. 

It removes web3 from import-fee-period script and adds the capability of stringify BigNumber from ethers in a human readable format. 
Also small additional fix: remove non-deleted development log from purge-synths

### DIFF
--- a/publish/src/commands/import-fee-periods.js
+++ b/publish/src/commands/import-fee-periods.js
@@ -138,19 +138,19 @@ const importFeePeriods = async ({
 			}
 		}
 
-		console.log(period);
-		console.log(Object.keys(period));
-
 		// remove redundant index keys (returned from struct calls)
+		const filteredPeriod = {};
 		Object.keys(period)
-			.filter(key => /^[0-9]+$/.test(key))
-			.forEach(key => {
-				console.log(key);
-				delete period[key];
-			});
-		feePeriods.push(period);
+			.filter(key => /^[0-9]+$/.test(key) === false)
+			.forEach(key => (filteredPeriod[key] = period[key]));
+
+		feePeriods.push(filteredPeriod);
 		console.log(
-			gray(`loaded feePeriod ${i} from FeePool (startTime: ${new Date(period.startTime * 1000)})`)
+			gray(
+				`loaded feePeriod ${i} from FeePool (startTime: ${new Date(
+					filteredPeriod.startTime * 1000
+				)})`
+			)
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


publish/src/commands/import-fee-periods.js
```diff
@@ -93,11 +93,11 @@ const importFeePeriods = async ({
 		if (lastEntry.address !== targetContractAddress) {
 			sourceContractAddress = lastEntry.address;
 		} else if (secondLastEntry.address !== targetContractAddress) {
-			sourceContractAddress = targetContractAddress.address;
+			sourceContractAddress = secondLastEntry.address;
```
leomassazza, line 96: Small fix found from the original code. If secondLastEntry was the one to check it will fail.

publish/src/commands/import-fee-periods.js
```diff
@@ -138,21 +139,27 @@ const importFeePeriods = async ({
 		}
 
 		// remove redundant index keys (returned from struct calls)
+		const filteredPeriod = {};
```
leomassazza, line 142: Adding an intermediate object to collect the "non-filtered" keys instead of deleting the filtered ones since ethers will return an object with non-configurable properties and delete will fail.

publish/src/commands/purge-synths.js
```diff
@@ -95,7 +95,7 @@ const purgeSynths = async ({
 	} else {
 		wallet = new ethers.Wallet(privateKey, provider);
 	}
-	console.log(wallet);
+
```
leomassazza, line 98: This little guy was missing here

publish/src/util.js
```diff
@@ -35,7 +36,14 @@ const {
 });
 
 const { networks } = require('../..');
-const stringify = input => JSON.stringify(input, null, '\t') + '\n';
+const JSONreplacer = (key, value) => {
```
leomassazza, line 39: Adding a JSONReplacer function to stringify BigNumber objects to the value.

publish/src/commands/import-fee-periods.js
ajsantander: Yeah good catch!

publish/src/commands/purge-synths.js
ajsantander: 🙏

publish/src/util.js
```diff
@@ -35,7 +36,14 @@ const {
...
+	if (typeof value === 'object' && value.type && value.type === 'BigNumber') {
+		//
```
ajsantander, line 41: nit: what's up with this empty comment?

ajsantander (approved):
LGTM!
(mini nit)

PR #1178, opened 2021-03-26
SIP-112: EtherWrapper

liamzebedee:
Implements the EtherWrapper as described in [SIP-112](https://sips.synthetix.io/sips/sip-112). The WETH changes to the SIP were discussed [here](https://research.synthetix.io/t/sip-112-wrappr/318/4?u=liamz) and [here](https://github.com/Synthetixio/SIPs/pull/432), though the PR's still need updating.

#### New contracts

1. `EtherWrapper`. Implements the SIP for wrapping/unwrapping WETH into sETH, compatible with L1 and L2.
2. `NativeEtherWrapper`. A helper contract which allows users to use native ETH with the EtherWrapper. This contract is **not** deployed for L2, as using native ETH opcodes (`callvalue`, `balance`) in the OVM will revert.
3. `WETH` and `MockWETH`.

#### Permission changes

EtherWrapper is permissioned to call internal functions on the following:

 - `FeePool` - to record the sUSD fees paid to the pool.
 - `MultiCollateralSynth` - in order to issue sETH and sUSD.

#### Upgraded contracts

 - FeePool
 - SynthsETH and SynthsUSD
 - DebtCache: these are to incorporate changes to `totalIssuedSynths`

#### Audit issues

 - [x] Should verify whether the rate is valid in remitFee before using effectiveValue https://github.com/Synthetixio/synthetix/blob/ac7dafbf461695ad10f41cdb2bcc32d3b9fc9caf/contracts/EtherWrapper.sol#L228(e.g. require(!exchangeRates().rateIsInvalid(ETH))
 - [x] Rounding issue in burn as divideDecimalRound should be divideDecimal (https://github.com/Synthetixio/synthetix/blob/ac7dafbf461695ad10f41cdb2bcc32d3b9fc9caf/contracts/EtherWrapper.sol#L166) The impact of it is fairly small. It's only noticeable because if you only mint sETH through Wrapper the maxBurnAmount will require 1 unit more than is available (i.e. will need to burn sETH minted from a different source).
- [x] address.transfer should not be used (use call.value() instead) https://github.com/Synthetixio/synthetix/blob/ac7dafbf461695ad10f41cdb2bcc32d3b9fc9caf/contracts/NativeEtherWrapper.sol#L81 further reading: https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now/
[...]

PR #1349, opened 2021-06-21
Deploy EmptyEtherWrapper

liamzebedee:
https://github.com/Synthetixio/synthetix/pull/1178 introduced the EtherWrapper, however it was never deployed to L2. Some context:

> To add some more context here for why there is no 'shell' contract currently - originally my rough idea was we would deploy EtherWrapper to L1 and L2. Implicitly I guess we came to the conclusion the L2 deploy wasn't needed, as the 1wk withdrawal delay would shelve any benefits of the arbitrage for wrappr users

This can cause the contracts, DebtCache and FeePool, to revert on L2, as the AddressResolver resolves to 0x0 for the contract. For now, I've added an EmptyEtherWrapper contract.


ajsantander (changes requested):
Thanks @liamzebedee 

Looks good! A few of things are needed to wrap this up tho:

1) Coverage is failing. I think adding the new contract to .solcover ignore should be enought right?
2) To work around the problem, some of the ovm deployments set sUSD to be a MultiCollateralSynth. I think they should be set back to a regular Synth.
3) Why PR against `ops-time-shifting`? I don't think the EtherWrapper integration tests need any time-shifting, so you could PR against develop directly.

ajsantander (approved):

PR #1352, opened 2021-06-21
Remove web3 from persist-tokens.

leomassazza:
For this PR I removed the web3 dependency since SetContenthash is not working on web3 and people using this script actually need to do that step manually using ENS dapp. 
Ethers.js doesn't have a direct replacement for interacting with ens contract, but has an interesting CLI command to do it via command line. On a next PR we can add the in-script interaction with ENS to set the content hash (ens.setContenthash).


ajsantander:
@leomassazza can we not leave commented out code in the script? Not just the main commented out block, but also the smaller ones. The script should do what it can do. If there's functionality to be added in the future, we can just leave a comment, but just throw away code that is currently not used.

ajsantander (approved):

PR #1361, opened 2021-06-24
Remove web3 from getExistingContract (Deployer)

leomassazza:
Remove web3 from getExistingContract at Deployer.js and users. 
This PR is part of an ongoing process and adds some intermediate duplicate functions (web3 and ethers) like makeContractWeb3() to maintain compatibility that will be removed once all references to it are migrated.


publish/src/Deployer.js
```diff
@@ -427,10 +427,14 @@ class Deployer {
 		return deployedContract;
 	}
 
-	makeContract({ abi, address }) {
+	makeContractWeb3({ abi, address }) {
```
leomassazza, line 430: Duplicated makeContract to start migrating it. It's only used internally but the two pending references will be updated in a later PR to make the PRs manageable

publish/src/commands/deploy-shorting-rewards.js
```diff
@@ -258,7 +272,7 @@ const deployShortingRewards = async ({
 			});
 
 		// Rebuild the cache so it knows about CollateralShort
-		await runStep({
+		await runStepWeb3({
```
leomassazza, line 275: Uses a contract that comes from the caller and is still built with web3. Once migrated, runStepWeb3 can be removed.

ajsantander (commented):
This looks great @leomassazza.

Apart from the change to deploy-shorting-rewards.js, I understand this PR as the Deployer.js file now returning ethers contract instances instead of web3 contract instances, but keeping the ability to return the latter around for now. This means that all deploy script related contracts now use ethers contracts.

I like the strategy, but there seem to be big changes incoming to the deploy script (see: https://github.com/Synthetixio/synthetix/pull/1313) so I'd recommend we hold any task related to removing web3 from the deployer until that PR gets merged. Otherwise we might face bugs and merge conflicts.

If you need this change for scripts _other_ than the deploy script, I'd say you tweak the strategy a bit:
1) Don't touch any deploy related code to avoid upcoming merge conflicts for now.
2) Introduce a new `makeContractEthers` instead, and use that from non deploy scripts.
3) When the deploy script stabilizes, we continue to swap its usages to `makeContractEthers`, util we can deprecate all web3 usage and rename the function back to `makeContract`.

ajsantander:
Closing this since it will be included in https://github.com/Synthetixio/synthetix/pull/1418

PR #1363, opened 2021-06-24
Updating Alnitak release to include DebtCache from 145 and 150

justinjmoses:


publish/releases.json
```diff
@@ -253,7 +253,7 @@
 			"minor": 46
 		},
 		"ovm": true,
-		"sources": ["Exchanger", "MintableSynthetix"],
```
justinjmoses, line 256: Note @dvd-schwrtz that we use the label not the actual source of the contract here, so `Synthetix` suffices - the deploy script figures out which source should be mapped to `Synthetix`

publish/releases.json
```diff
@@ -253,7 +253,7 @@
...
-		"sips": [138, 139, 140]
+		"sources": ["DebtCache", "Exchanger", "Synthetix"],
+		"sips": [138, 139, 140, 145, 150]
```
zyzek, line 257: 151 is not being used for optimism?

zyzek (approved):

publish/releases.json
justinjmoses: No - not yet. Perhaps in the near future - we'd need to make sure the upgrade could fit into a single block, which could be challenging when transpiled for the OVM

PR #1366, opened 2021-06-25
Minor fix to connect-bridges script

ajsantander:


publish/src/commands/connect-bridge.js
```diff
@@ -149,8 +149,10 @@ const connectLayer = async ({
 
 	const params = {
 		gasPrice: ethers.utils.parseUnits(gasPrice.toString(), 'gwei'),
-		gas: gasLimit,
```
ajsantander, line 152: Ether overrides 'gasLimit' not 'gas'

PR #1367, opened 2021-06-25
Adds documentation for integration tests

ajsantander:


test/integration/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+1. Run `npx hardhat ops --start`
+2. Run `npx hardhat test:integration:dual --compile --deploy`
+
+## Advanced usage
+
+### Avoiding compilation and deployment
+
+All integration tests commands include a `--compile` and a `--deploy` flag.
+`--compile` should be used the first time you run the tests, and every time you make a change to a contract.
```
justinjmoses, line 56: just thinking about this - why not compile every time? hardhat is smart enough to not compile if no changes so it's a no-op.

test/integration/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+## L2 tests
+
+These tests deploy a standalone instance of the system in L2 or an ovm network. They are located in `test/integration/l2`.
+
+Atm, the only way to run a local L2/ovm chain is using Optimism's ops tool, built on docker.
+
+### Running standalone L1 tests against the ops tool
+
+1. Run `npx hardhat ops --start`
```
justinjmoses, line 32: maybe add a point that Docker needs to be installed locally and running for this step to work?

test/integration/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Please consider the following aspects when developing new integration tests.
+
+### What to test?
+
+Unlike unit tests, integration tests run against complex set ups (i.e. Optimism's ops tool) and are thus fragile and slow. Coverage should be kept to a bare minimum, and test files should be ~100 lines of code and be kept minimal and super easy to read. Anything that makes these tests slow or hard to maintain should be avoided.
+
+Small details of a contract should be tested in unit tests, and integration tests should be reserved to global emergent features of the system that required multiple parts of it integrated together. Always ask yourself if something can be tested in an unit test instead, and really needs to be tested at an integration level, and avoid writing tests at this level if possible.
+
+### Behaviors
```
liamzebedee, line 81: I would also refer to your test helpers here, as they are a new pattern and worth documenting. For example, `ensureBalance` is quite behavioural and abstracts away the key details of how to get balances, such as the "hacky method" on L2.

liamzebedee (commented):
Looking good!

i-stam (approved):
Thanks for this, really helpful!

test/integration/README.md
ajsantander: Right, but these tests don't use hardhat's compilation engine, they use our own `node publish build` process. We already have an issue to merge the two, which should auto fix this request. If you'd like it before, could you open a separate issue in this project pls: https://github.com/orgs/Synthetixio/projects/8

test/integration/README.md
ajsantander: Added

test/integration/README.md
ajsantander: Added

PR #1369, opened 2021-06-25
Upgrading etherscan links to optimistic explorer when required

justinjmoses:
Support for Optimism's explorer links

### DIFF
--- a/publish/src/util.js
+++ b/publish/src/util.js
@@ -136,11 +136,13 @@ const loadAndCheckRequiredSources = ({ deploymentPath, network, freshDeploy }) =
 	};
 };
 
-const getEtherscanLinkPrefix = network => {
-	return `https://${network !== 'mainnet' ? network + '.' : ''}etherscan.io`;
+const getExplorerLinkPrefix = ({ network, useOvm }) => {
+	return `https://${network !== 'mainnet' ? network + (useOvm ? '-' : '.') : ''}${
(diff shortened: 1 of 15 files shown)


publish/src/util.js
```diff
@@ -136,11 +136,13 @@ const loadAndCheckRequiredSources = ({ deploymentPath, network, freshDeploy }) =
 	};
 };
 
-const getEtherscanLinkPrefix = network => {
-	return `https://${network !== 'mainnet' ? network + '.' : ''}etherscan.io`;
+const getExplorerLinkPrefix = ({ network, useOvm }) => {
+	return `https://${network !== 'mainnet' ? network + (useOvm ? '-' : '.') : ''}${
+		useOvm ? 'explorer.optimism' : 'etherscan'
```
justinjmoses, line 141: this is the main change - everything else is mostly just a rename

leomassazza: `(*-)explorer.optimism.io` will redirect to `(*-)optimistic.etherscan.io`.
Is it not easier to just add "**optimistic.**" if useOvm is true?

leomassazza (approved):

publish/src/util.js
justinjmoses: it is, but Optimism created the abstraction in case they want to use other explorers down the line

PR #1375, opened 2021-06-28
Shorts integration tests for deposit and withdraw

gberg1:


test/integration/behaviors/short.behavior.js
```diff
@@ -38,10 +40,21 @@ function itCanOpenAndCloseShort({ ctx }) {
 			});
 		});
 
-		describe('open and close a short', async () => {
+		before('skip waiting period by setting interaction delay to zero', async () => {
+			CollateralShortAsOwner = CollateralShort.connect(ctx.users.owner);
+			interactionDelay = await CollateralShortAsOwner.interactionDelay();
+
+			await CollateralShortAsOwner.setInteractionDelay('0');
```
ajsantander, line 47: Why not use fast forwarding here?

test/integration/behaviors/short.behavior.js
```diff
@@ -38,10 +40,21 @@ function itCanOpenAndCloseShort({ ctx }) {
 			});
 		});
 
-		describe('open and close a short', async () => {
+		before('skip waiting period by setting interaction delay to zero', async () => {
+			CollateralShortAsOwner = CollateralShort.connect(ctx.users.owner);
```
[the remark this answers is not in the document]
justinjmoses: I believe this is required because we need to use it in two contexts - one as the owner to update a setting and one as the user.

ajsantander (approved):
Just one minor request

test/integration/behaviors/short.behavior.js
ajsantander: E.g. https://github.com/Synthetixio/synthetix/blob/develop/test/integration/behaviors/exchange.behavior.js#L61

ajsantander: If not available, add a new one here https://github.com/Synthetixio/synthetix/blob/develop/test/integration/utils/skip.js

gberg1: Setting the interaction delay on shorts is needed in addition to skipping tried just skipping but still ran into an error due to the interaction delay needing to be zero before re-interacting with the opened short

PR #1378, opened 2021-06-29
Simplify conditional logic of generate solidity

justinjmoses:
The logic for adding to `runSteps` was overly complicated from previous versions and not properly refactored.


publish/src/util.js
```diff
@@ -254,7 +254,7 @@ const performTransactionalStepWeb3 = async ({
 				`[GENERATE_SOLIDITY_SIMULATION] Successfully completed ${action} number ${++_dryRunCounter}.`
 			)
 		);
-		return { mined: true };
+		return {};
```
justinjmoses, line 257: returning `mined` here was a hack - so let's not rely on this.

liamzebedee (approved):

PR #1380, opened 2021-06-30
Adding EtherWrapper for L2 to deploy empty and SystemSettings for Eth…

justinjmoses:
1. L2 now needs an `EmptyEtherWrapper` in order for the `DebtCache` to re-use the same code as on L1.
2. `SystemSettings` has some extra settings on L1 around the `EtherWrapper` that aren't on L2, so deploy it here just to keep things in line as there's no concern with having extra empty settings. (Also makes it easier on deploys so we don't have to check which functions exist on the current layer before invoking them to read and check values)

### DIFF
--- a/publish/releases.json
+++ b/publish/releases.json
@@ -253,7 +253,13 @@
 			"minor": 46
 		},
 		"ovm": true,
-		"sources": ["DebtCache", "Exchanger", "Synthetix"],
+		"sources": [
+			"DebtCache",
+			"EtherWrapper",
+			"Exchanger",
+			"Synthetix",
+			"SystemSettings"
+		],
 		"sips": [138, 139, 140, 145, 150]
 	}
 ]


publish/releases.json
```diff
@@ -253,7 +253,13 @@
 			"minor": 46
 		},
 		"ovm": true,
-		"sources": ["DebtCache", "Exchanger", "Synthetix"],
+		"sources": [
+			"DebtCache",
+			"EtherWrapper",
```
liamzebedee, line 258: Sorry, I forgot to add this in https://github.com/Synthetixio/synthetix/pull/1349.

liamzebedee (approved):

publish/releases.json
justinjmoses: no worries

PR #1384, opened 2021-06-30
Fixing local dev to properly only deploy WETH the first time

justinjmoses:


publish/src/commands/deploy/deploy-loans.js
```diff
@@ -34,8 +34,8 @@ module.exports = async ({ account, addressOf, deployer, getDeployParameter, netw
 		// Hardhat for the local-ovm environment, we must deploy
 		// our own.
 		const weth = await deployer.deployContract({
-			name: useOvm ? 'MockWETH' : 'WETH',
-			force: true,
+			name: 'WETH',
```
justinjmoses, line 37: 1. the name should not be switched on but rather the source. 
2. force was used instead of the config.json, making it always deploy on local upgrades, which is incorrect.

ajsantander (approved):

PR #1391, opened 2021-07-01
Fix up futures deployment script and settings contract

zyzek:


publish/src/commands/deploy/deploy-futures.js
```diff
@@ -48,20 +48,26 @@ module.exports = async ({ account, addressOf, deployer, runStep, useOvm }) => {
...
-	const makerFee = w3utils.toWei('0.001');
-	const maxLeverage = w3utils.toWei('10');
-	const maxMarketDebt = w3utils.toWei('100000');
-	const minInitialMargin = w3utils.toWei('100');
-	const fundingParameters = [
-		w3utils.toWei('0.1'), // max funding rate per day
-		w3utils.toWei('1'), // max funding rate skew
-		w3utils.toWei('0.0125'), // max funding rate delta per hour
-	];
+	const futuresAssets = ['BTC', 'ETH', 'LINK'];
```
liamzebedee, line 60: Note: we've been talking about it, we should move these either into `index.js` or a separate per-network config like `{network}/futures-markets.json`. It'd be useful so I can `require('synthetix').futuresAssets` within the keeper-bot for example.

liamzebedee (approved):

publish/src/commands/deploy/deploy-futures.js
zyzek: Agreed. Definitely want the `{network}/futures-markets.json`, and the internal structure should also include the per-market settings. Issue here: https://github.com/Synthetixio/issues/issues/98

PR #1398, opened 2021-07-08
Removing exceptions from deploy configure code

justinjmoses:
These fees were added here as each new SCCP for rates were added but honestly, they are just extra complication. The deploy script doesn't need to know about the current rates - this code is simply here to set default rates when none are set. The pDAO will change the fee rates on-chain when required and these will stay inside `FlexibleStorage` for `SystemSettings`, even if a new SystemSettings contract is redeployed. Thus, in anticipation of SCCP-130, these exceptions are to be removed.


jacko125 (approved):
lgtm

leomassazza (approved):
LGTM

PR #1399, opened 2021-07-12
Futures merge develop

liamzebedee:


contracts/DebtCache.sol
```diff
@@ -131,6 +115,16 @@ contract DebtCache is BaseDebtCache {
             require(cachedSum <= debt, "Cached synth sum exceeds total debt");
             debt = debt.sub(cachedSum).add(currentSum);
 
+            // TODO: is this the right place to relocate this, @Anton?
```
zyzek, line 118: This needs to be where it was, outside the `if (cachedSum != currentSum)` conditional, as it is a positive contribution to the system debt which isn't bugged out like the excluded debt component

index.js
```diff
@@ -211,6 +210,10 @@ const defaults = {
 	ETHER_WRAPPER_MAX_ETH: w3utils.toWei('5000'),
 	ETHER_WRAPPER_MINT_FEE_RATE: w3utils.toWei('0.02'), // 200 bps
 	ETHER_WRAPPER_BURN_FEE_RATE: w3utils.toWei('0.0005'), // 5 bps
+
+	FUTURES_ASSETS: ['BTC', 'ETH', 'LINK'],
```
zyzek, line 214: Might as well use this if we're gunna include it here.

zyzek (changes requested):

index.js
zyzek: Relevant deleted line: https://github.com/Synthetixio/synthetix/pull/1399/commits/f4175a14ac2724357a385da661a09ea6381b3c89#diff-16319ff3ce6ffd9de269997039bf84eeb9070081570dd338bcb851101818be94L51

PR #1402, opened 2021-07-12
SIP-167 Introduces an L2 governance bridge

ajsantander:
## Overview

This is an implementation of [SIP 167 "L2 Bridged Governance"](https://sips.synthetix.io/sips/sip-167).

The basic idea is to own the system contracts on L2 via a contract, that is controlled from L1. This L2 owner contract only responds to transactions relayed from its L1 counterpart, which in turn only responds to the L1 owner.

L2 contract: `OwnerRelayOnOptimism.sol`
L1 counterpart: `OwnerRelayOnEthereum.sol`

Thus, L2 system contracts need to nominate `OwnerRelayOnOptimism` as their new owner. This effectively makes `OwnerRelayOnOptimism` the owner of the L2 instance 😮 

The L1 owner (the pDAO) will then be able to use `OwnerRelayOnEthereum` to relay arbitrary messages to `OwnerRelayOnOptimism`, which will accept the messages as long as the L1 tx originator was effectively the `OwnerRelayOnEthereum`. It will then forward these messages to any contract on L2, with owner access.

E.g. changing a SystemSetting on L2, via this owner relay:
* Optimistic etherscan is used to get the encoded calldata for changing the setting on L2 (or the CLI can be used as well).
* A call to OwnerRelayOnEthereum.initiateRelay(_SystemSettings address on L2_, _encoded calldata_) is staged in the pDAO
* The pDAO signs and executes the tx
* 5-8 minutes later, the tx will be relayed to L2
* OwnerRelayOnEthereum.finalizeRelay(...) will be called by Optimism's messenger, pass all checks, and change the owner protected setting. Yay 🎉 

## Batching

Similarly, txs can be relayed in batches via `OwnerRelayOnEtherem.initiateRelayBatch(...)` and `OwnerRelayOnOptimisim.finalizeRelayBatch(...)`. This is handy for when the pDAO needs to relay multiple txs at once, like for example when accepting ownership after all contracts on L2 have nominated `OwnerRelayOnOptimism` as their new owner.

## Gas limits on L1 and L2

Both `OwnerRelayOnEthereum.initiateRelay(...)` and `OwnerRelayOnEthereum.initiateBatchRelay(...)` accept a `crossDomainGasLimit` parameter, which ultimately determines the gas limit of the L2 relayed transaction. If the value used is zero, the default system setting will be used (currently 8m gas).

In practice, we've found that accepting ownership on more than 25 contracts runs out of gas on L2 if the `crossDomainGasLimit` is 8m.

## Temporary fallback mechanism

Since this is a novel mechanism controlling a very critical aspect of the system on L2, a temporary fallback mechanism on L2 is implemented: `OwnerRelayOnOptimism.directRelay(...)`.

This allows an EOA to directly control the contract owner on L2, but this functionality expires after a certain amount of time. The EOA that has access to this functionality and the amount of time it lasts for has to be set on contract deployment and cannot be changed.

Ownership control in the new contracts is as follows:
* `OwnerRelayOnEthereum` => `Owned` => The pDAO
* `OwnerRelayOnOptimism` => `TemporarilyOwned` => an EOA

## New settings and defaults

Adds `SystemSettings.CROSS_DOMAIN_RELAY_GAS_LIMIT`.

Adds default values in index.js for
* `TEMP_OWNER_DEFAULT_DURATION`: 60 days
* `CROSS_DOMAIN_RELAY_GAS_LIMIT`: 8000000

## New scripts

`connect-bridge.js` is updated to also connect the owner relay contracts.

`nominate-relay.js` is added, similar to `nominate.js` but specialized for the relay. This should NOT be used to nominate all contracts to `OwnerRelayOnOptimism` when the current owner is an EOA, but when the current owner is the relay, and a new relay or EOA is to be nominated.

`owner-relay.js` is added, similar to `owner.js` but specialized for the relay. This should be used to (a) accept owner ship for `OwnerRelayOnOptimism` after nomination, or to stage owner actions to the pDAO multisig.

## Tests

New unit tests:
* `OwnerRelayOnEthereum.js`
* `OwnerRelayOnOptimism.js`
* `TemporarilyOwned.js`
* `SystemSettings.js` (partial addition)

New integration tests:
* `relay.integration.js` - dual test changing an L2 SystemSetting from L1
* `relayBatch.integration.js` - dual test changing ownership of all L2 contracts and accepting it via L1
* `directRelay.l2.integration.js` - L2 test on the temporary fallback mechanism

## Deployment runsheet
To deploy this feature in production, contracts need to be deployed both on `mainnet` and `optimistic-mainnet` networks, and the contracts then need to be connected (i.e. each instance needs to know of the addresses of its counterparts in the opposite network).

L1 deployment:
* Deploy `OwnerRelayOnEthereum` and `SystemSettings`

L2 deployment
* Deploy `OwnerRelayOnOptimism`

Connection
* On L1, Stage `AddressResolver.importAddresses("ovm:OwnerRelayOnOptimism", address)` and execute it on the pDAO.
* On L1, call `OwnerRelayOnEthereum.rebuildCache()`
* On L2, run `AddressResolver.importAddresses("base:OwnerRelayOnEthereum", address)`.
* On L1, call `OwnerRelayOnOptimism.rebuildCache()`

Ownership transfer on L2
* Use `nominate.js` on L2 (with `--use-ovm`) to nominate all L2 contracts to the deployed address of `OwnerRelayOnOptimism`. Note that `nominate-relay.js` is only needed in the future, after ownership is the relay and not an EOA.
* Accept L2 relay ownership by `OwnerRelayOnOptimism` using the bridge. Use the new command `owner-relay`. It will use `OwnerRelayOnEthereum.initiateRelayBatch` and will be done via an EOA if you are on a test network or staging a transaction on Gnosis Safe if you are deploying for mainnet. This should be repeated at least twice (until you confirm there's no more contracts pending) since it's a batch execution with a limit in the number of contracts per batch to prevent a revert due to exceeding the gas limit. 
* Verify `OwnerRelayOnOptimism` is the owner of the contracts on L2. 

#### Connect the relay contracts (only works on instances owned by EOAs, i.e. testnets)
In this example we deployed a brand new system on kovan (https://github.com/Synthetixio/synthetix/pull/1467). 
```
node publish connect-bridge \
    --l1-network kovan \
    --l2-network kovan \
    --l1-provider-url https://kovan.infura.io/v3/1xxxxxxba1 \
    --l2-provider-url https://kovan.optimism.io/ \
    --l1-deployment-path ./publish/deployed/kovan-test \
    --l2-deployment-path ./publish/deployed/kovan-test-ovm \
    --l1-private-key 0x479xxxxxxxxxxxda0 \
    --l2-private-key 0x479xxxxxxxxxxxda0 \
    --l1-messenger 0x4361d0F75A0186C05f971c566dC6bEa5957483fD \
    --l2-messenger 0x4200000000000000000000000000000000000007 \
    --l1-gas-price 2 \
    --l1-gas-limit 8000000 --dry-run
```

#### Nominate `OwnerRelayOnOptimism`
Using the nominate command set the --new-owner to the OwnerRelayOnOptimism address.
```
node publish nominate --network kovan --deployment-path ./publish/deployed/kovan-test-o
vm  --new-owner 0x41e5175CD6eB75a96b50D9935a6E534A5B1559B2 --use-ovm --provider-url https://kovan.optimism.io
```

#### Accept L2 relay ownership by `OwnerRelayOnOptimism` using the bridge
In this case it was necessary to run this twice, to cover all the contracts. Note that before running the second time you should confirm the first batch succeeded.
You can use https://kovan-optimistic.etherscan.io/txsEnqueued (or https://optimistic.etherscan.io/txsEnqueued for mainnet) to check if the transaction was sent to L2 by Optimism network. 

```
node publish  owner-relay \
    --l1-network kovan \
    --l2-network kovan \
    --l1-deployment-path ./publish/deployed/kovan-test \
    --l2-deployment-path ./publish/deployed/kovan-test-ovm \
    --l1-provider-url https://kovan.infura.io/v3/1xxxxxxba1 \
    --l2-provider-url https://kovan.optimism.io/ \
    --l1-private-key 0x479xxxxxxxxxxxda0 \
    --gas-price 2 \
    --gas-limit 8000000
```
This example was run using a EOA on L1 to accept ownership via Relay on L2. In case a Gnosis Safe is used as owner on L1, the command should include the flag `--is-contract` in the same way as it's done with `node publish owner` command.

#### Change ownership on L2  (back to EOA or other OwnerRelayOnOptimism)
In case something went wrong, there are two alternatives to regain ownership of the L2 contracts. One using the relay itself or another, that is only available for a period of time, directly using OwnerRelayOnOptimism with the designated temporary owner.

> Using the Relay
Once the relay is configured, `OwnerRelayOnOptimism` will be the owner of all L2 contracts. You can change it to a new owner (a new contract or an EOA) using the new command nominate-relay to nominate a new owner in batches for all L2 contracts. 
An example of how to use this new contract can be seen here: 

```
node publish  nominate-relay \
    --l1-network kovan \
    --l2-network kovan \
    --l1-deployment-path ./publish/deployed/kovan-test \
    --l2-deployment-path ./publish/deployed/kovan-test-ovm \
    --l1-provider-url https://kovan.infura.io/v3/1xxxxxxba1 \
    --l2-provider-url https://kovan.optimism.io/ \
    --l1-private-key 0x479xxxxxxxxxxxda0 \
    --gas-price 2 \
    --gas-limit 8000000 \
    --new-owner 0x23e78AaC93254a858c1bA7F6fE92468fE367B3f1
```

The same note mentioned on owner-relay using an EOA or Gnosis Safe on L1 are valid here, you can set the same parameters like --is-contract to indicate the command to stage a transaction instead of executing it as an EOA. 

> Using the temporary owner
In order to nominate a new owner and there was something wrong setting up the relay, you can use the temporary owner to act on that situation using the function `OwnerRelayOnOptimism.directRelay(address target, bytes calldata payload)`


contracts/MixinSystemSettings.sol
```diff
@@ -27,13 +27,14 @@ contract MixinSystemSettings is MixinResolver {
     bytes32 internal constant SETTING_CROSS_DOMAIN_ESCROW_GAS_LIMIT = "crossDomainEscrowGasLimit";
     bytes32 internal constant SETTING_CROSS_DOMAIN_REWARD_GAS_LIMIT = "crossDomainRewardGasLimit";
     bytes32 internal constant SETTING_CROSS_DOMAIN_WITHDRAWAL_GAS_LIMIT = "crossDomainWithdrawalGasLimit";
+    bytes32 internal constant SETTING_CROSS_DOMAIN_RELAY_GAS_LIMIT = "crossDomainRelayGasLimit";
```
ajsantander, line 30: Since we're relaying a new kind of message to L2, it needs its own cross domain gas limit setting.

hardhat/tasks/task-interact.js
```diff
@@ -355,7 +355,6 @@ task('interact', 'Interact with a deployed Synthetix instance from the command l
 					const txPromise = contract[abiItemName](...inputs, overrides);
 					result = await _sendTx({
 						txPromise,
-						provider,
```
ajsantander, line 358: FYI: Fly by fix, this wasn't needed.

publish/deployed/goerli-ovm/config.json
```diff
@@ -115,5 +115,8 @@
 	},
 	"TokenStatesUSD": {
 		"deploy": false
+	},
```
ajsantander, line 118: FYI: Adding the new contracts in all the config.json's...
(Pls check that ovm deploys get OwnerRelayOnOptimism, and evm deploys get OwnerRelayOnEthereum)

publish/releases.json
```diff
@@ -255,5 +255,25 @@
...
+	},
+	{
+		"name": "Mirfak",
+		"version": {
+			"major": 2,
+			"minor": 999
+		},
+		"ovm": true,
+		"sources": ["OwnerRelayOnEthereum"],
+		"sips": [999]
```
ajsantander, line 267: TODO: Need SIP number here

publish/src/commands/connect-bridge.js
```diff
@@ -47,6 +47,7 @@ const connectBridge = async ({
 		wallet: walletL1,
 		AddressResolver: AddressResolverL1,
 		SynthetixBridge: SynthetixBridgeToOptimism,
+		OwnerRelay: OwnerRelayOnEthereum,
```
ajsantander, line 50: FYI: This script also connects the owner relays now. I could have renamed the file to connect-bridges, but didn't want to mess up the diff for you guys.

test/contracts/OwnerRelayOnEthereum.js
```diff
@@ -0,0 +1,173 @@
+const chalk = require('chalk');
```
ajsantander, line 1: I encourage you to look at the mocking in the unit tests, it's interesting, especially in OwnerRelayOnOptimism.js

test/contracts/SystemSettings.js
```diff
@@ -88,46 +88,24 @@ contract('SystemSettings', async accounts => {
...
-			const actual = await systemSettings.crossDomainMessageGasLimit(gasLimitType);
-			assert.equal(actual, newLimit, 'Configured cross domain gas limit is set correctly');
-			assert.eventEqual(txn, 'CrossDomainMessageGasLimitChanged', [gasLimitType, newLimit]);
-		});
-		it('the owner can invoke and replace with emitted event', async () => {
-			const newLimit = 4e6;
-			const gasLimitType = 3;
-			const txn = await systemSettings.setCrossDomainMessageGasLimit(gasLimitType, newLimit, {
-				from: owner,
+
```
ajsantander, line 91: Fly by tidy up: no code duplication pls 🙏

contracts/OwnerRelayOnOptimism.sol
```diff
@@ -0,0 +1,64 @@
...
+        addresses[1] = CONTRACT_BASE_OWNER_RELAY_ON_ETHEREUM;
+    }
+
+    /* ========== EXTERNAL ========== */
+
+    function acceptOwnershipOn(address target) external {
+        IOwned(target).acceptOwnership();
+    }
+
+    function finalizeRelay(address target, bytes calldata data) external {
```
ajsantander, line 47: This is the gist of the PR

contracts/OwnerRelayOnOptimism.sol
```diff
@@ -0,0 +1,64 @@
...
+
+    function resolverAddressesRequired() public view returns (bytes32[] memory addresses) {
+        addresses = new bytes32[](2);
+        addresses[0] = CONTRACT_EXT_MESSENGER;
+        addresses[1] = CONTRACT_BASE_OWNER_RELAY_ON_ETHEREUM;
+    }
+
+    /* ========== EXTERNAL ========== */
+
+    function acceptOwnershipOn(address target) external {
```
ajsantander, line 43: Notice this accessory function: It allows us to nominate ownership and accept it, all purely L2 transactions. The idea is to use this function when we change the owner of all the L2 contracts to OwnerRelayOnOptimism, instead of having to relay dozens of `acceptOwnership` calls.

test/integration/dual/relay.integration.js
```diff
@@ -0,0 +1,140 @@
...
+				await tx.wait();
+			});
+
+			it('shows that the L2 relay is the nominated owner', async () => {
+				assert.equal(await SystemSettingsL2.nominatedOwner(), OwnerRelayOnOptimism.address);
+			});
+
+			describe('when the L2 relay accepts ownership', () => {
+				before('call acceptOwnershipOn() directly on OwnerRelayOnOptimism', async () => {
+					const tx = await OwnerRelayOnOptimism.connect(ownerL2).acceptOwnershipOn(
```
ajsantander, line 50: Using that handy `acceptOwnershipOn` function here.

test/integration/dual/relay.integration.js
```diff
@@ -0,0 +1,140 @@
...
+			before('store minimumStakeTime', async () => {
+				originalMinimumStakeTime = await SystemSettingsL2.minimumStakeTime();
+			});
+
+			before('relay setMinimumStakeTime via the bridge', async () => {
+				const calldata = SystemSettingsL2.interface.encodeFunctionData('setMinimumStakeTime', [
+					newMinimumStakeTime,
+				]);
+
+				const tx = await OwnerRelayOnEthereum.connect(ownerL1).initiateRelay(
```
ajsantander, line 81: ✨ 
An L1 owner tx changes a system setting on L2.

test/integration/dual/relay.integration.js
```diff
@@ -0,0 +1,140 @@
...
+			});
+		});
+
+		describe('when the relay relinquishes ownership back to an EOA on L1', () => {
+			before('relay a tx to nominateNewOwner() from L1', async () => {
+				const calldata = SystemSettingsL2.interface.encodeFunctionData('nominateNewOwner', [
+					ownerL2.address,
+				]);
+
+				const tx = await OwnerRelayOnEthereum.connect(ownerL1).initiateRelay(
```
ajsantander, line 119: 🧙 
An L1 tx nominates a new owner on L2.

test/integration/utils/bootstrap.js
```diff
@@ -74,7 +74,7 @@ function bootstrapL2({ ctx }) {
 }
 
 function bootstrapDual({ ctx }) {
-	before('bootstrap layer 1 and layer 2 intances', async () => {
+	before('bootstrap layer 1 and layer 2 instances', async () => {
```
ajsantander, line 77: 🚸

ajsantander, line 77: come-on-maaate

publish/releases.json
leomassazza: SIP-167

publish/releases.json
```diff
@@ -255,5 +255,25 @@
...
 		"sources": ["EtherWrapper", "Exchanger", "Synthetix", "SystemSettings"],
 		"sips": [138, 139, 140]
+	},
+	{
+		"name": "Mirfak",
+		"version": {
+			"major": 2,
+			"minor": 999
+		},
+		"ovm": true,
```
leomassazza, line 265: ovm: false for OwnerRelayOnEthereum

publish/deployed/mainnet/config.json
```diff
@@ -793,5 +793,8 @@
 	},
 	"NativeEtherWrapper": {
 		"deploy": false
+	},
+	"OwnerRelayOnEthereum": {
```
justinjmoses, line 797: so you shouldn't commit these - recall the prepare script will enable it

test/contracts/OwnerRelayOnOptimism.js
```diff
@@ -0,0 +1,196 @@
...
+		before('mock the target contract acceptOwnership() function', async () => {
+			MockedOwnedL2.smocked.acceptOwnership.will.return();
+		});
+
+		before('call the target acceptOwnership() function via OwnerRelayOnOptimism', async () => {
+			const tx = await OwnerRelayOnOptimism.connect(owner).acceptOwnershipOn(MockedOwnedL2.address);
+			await tx.wait();
+		});
+
+		it('called the function on the target contract', async () => {
```
leomassazza, line 99: shouldn't we validate the ownership actually changed?

leomassazza (approved):
LGTM. Made a few comments. Please check those.

contracts/OwnerRelayOnEthereum.sol
```diff
@@ -0,0 +1,61 @@
...
+
+    /* ========== RESTRICTED ========== */
+
+    function initiateRelay(address target, bytes calldata data) external onlyOwner {
+        bytes memory messageData = abi.encodeWithSelector(IOwnerRelayOnOptimism(0).finalizeRelay.selector, target, data);
+
+        messenger().sendMessage(
+            ownerRelayOnOptimism(),
+            messageData,
+            uint32(getCrossDomainMessageGasLimit(CrossDomainMessageGasLimits.Relay))
```
K-Ho, line 52: Depends on how much you care about L1 gas costs, but you might want this to be variable - e.g. for a simple L2 tx you might want it to be set lower vs a more complicated L2 tx. Otherwise you could just default it to 11m gas

ajsantander: Will do! Thanks @K-Ho

publish/releases.json
ajsantander: Fixed

publish/deployed/mainnet/config.json
ajsantander: Right! Undoing...

test/contracts/OwnerRelayOnOptimism.js
ajsantander: Nope. What we're calling here is the mocked function `MockedOwnedL1.acceptOwnership()` which actually does nothing. In this test, we only want to assert that calling `OwnerRelayOnOptimism.acceptOwnershipOn(...)` ends up calling that mock.

What you are suggesting is tested here: https://github.com/Synthetixio/synthetix/pull/1402/files#diff-bace10c6169c2b1a2e3c7d69a63b775d5e11c3c5cba28636b2edf595e9bd80b7R50

ajsantander:
@leomassazza @justinjmoses 

I introduced a non-trivial change suggested by @K-Ho:
`OwnerRelayOnEthereum.initiateRelay(address target, bytes data)` becomes `OwnerRelayOnEthereum.initiateRelay(address target, bytes data, uint32 xGasLimit)`, so that we have finer control over L1 costs in the relay. If zero is used, the default in SystemSettings is taken.

contracts/OwnerRelayOnEthereum.sol
ajsantander: Implemented

publish/releases.json
```diff
@@ -275,5 +275,24 @@
 		"ovm": true,
 		"sources": ["EtherWrapper", "Exchanger", "Synthetix", "SystemSettings"],
 		"sips": [138, 139, 140]
+	},
+	{
+		"name": "Mirfak",
+		"version": {
+			"major": 2,
+			"minor": 999
```
justinjmoses, line 283: version number check before merge :pray: - Dubhe is first in #1407

publish/src/commands/deploy/deploy-core.js
```diff
@@ -241,4 +241,18 @@ module.exports = async ({
 		deps: ['AddressResolver'],
 		args: [account],
 	});
+
+	if (useOvm) {
```
justinjmoses, line 245: No - please don't switch on `useOvm` - you don't need to because the respective config.json files have the right flags - so just include both and trust the configs to figure it out.

test/contracts/OwnerRelayOnEthereum.js
```diff
@@ -0,0 +1,200 @@
...
+	const mockedOwnerRelayOnOptimismAddress = ethers.Wallet.createRandom().address;
+	const mockedContractAddressOnL2 = ethers.Wallet.createRandom().address;
+	const mockedRelayData = '0xdeadbeef';
+
+	before('initialize signers', async () => {
+		[owner, user] = await ethers.getSigners();
+	});
+
+	before('mock other contracts used by OwnerRelayOnEthereum', async () => {
+		MockedMessenger = await smockit(
```
justinjmoses, line 30: you don't want to use `prepareSmocks`?

publish/releases.json
ajsantander: Ok, will do!

publish/src/commands/deploy/deploy-core.js
ajsantander: Right!
Fixed.

test/contracts/OwnerRelayOnEthereum.js
ajsantander: My preference is to be more explicit, and show the reader exactly what is going on, without assuming that how `prepareSmocks` works. I also don't like the fact that you have to assign a set of addresses to the tool, and like to leave that to smockit.

But if you insist, I can update the PR to use the tool.

contracts/OwnerRelayOnEthereum.sol
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /* ========== RESTRICTED ========== */
+
+    function initiateRelay(
+        address target,
+        bytes calldata data,
+        uint32 crossDomainGasLimit // If zero, uses default value in SystemSettings
+    ) external onlyOwner {
+        bytes memory messageData = abi.encodeWithSelector(IOwnerRelayOnOptimism(0).finalizeRelay.selector, target, data);
```
i-stam, line 51: `IOwnerRelayOnOptimism.finalizeRelay.selector` should also work and it is a bit cleaner

ajsantander: Incoming in https://github.com/Synthetixio/synthetix/pull/1466

publish/src/commands/deploy/configure-system-settings.js
```diff
@@ -257,6 +258,17 @@ module.exports = async ({
 			comment: 'Set the gas limit for withdrawing from L2',
 		});
 
+		await runStep({
```
justinjmoses, line 261: Note that adding this here will revert if SystemSettings isn't deployed.

publish/src/commands/deploy/deploy-core.js
```diff
@@ -241,4 +242,16 @@ module.exports = async ({
...
+	await deployer.deployContract({
+		name: 'OwnerRelayOnOptimism',
+		deps: ['AddressResolver'],
+		args: [addressOf(readProxyForResolver), account, TEMP_OWNER_DEFAULT_DURATION],
+	});
+
+	await deployer.deployContract({
+		name: 'OwnerRelayOnEthereum',
+		deps: ['AddressResolver'],
+		args: [account, addressOf(readProxyForResolver)],
```
justinjmoses, line 255: The order here is flipped / let's make it consistent with the other contracts. (Owner is first then resolver IiRC)

justinjmoses: Or rather the contract above has the wrong order

justinjmoses (changes requested):
Mostly nitpicks but one important fix wrt constructor arg ordering

contracts/OwnerRelayOnEthereum.sol
```diff
@@ -0,0 +1,92 @@
...
+        // First check that the length of the arguments match
+        require(targets.length == payloads.length, "Argument length mismatch");
+
+        IOwnerRelayOnOptimism ownerRelayOnOptimism;
+        bytes memory messageData =
+            abi.encodeWithSelector(ownerRelayOnOptimism.finalizeRelayBatch.selector, targets, payloads);
+
+        _messenger().sendMessage(_ownerRelayOnOptimism(), messageData, _getxGasLimit(crossDomainGasLimit));
+
+        emit RelayBatchInitiated(targets, payloads);
```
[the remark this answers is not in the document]
ajsantander: We intentionally tried to avoid this, since it would significantly increase the gas cost of a batched relay.

contracts/TemporarilyOwned.sol
```diff
@@ -0,0 +1,23 @@
...
+
+contract TemporarilyOwned {
+    address public temporaryOwner;
+    uint public expiryTime;
+
+    constructor(address _temporaryOwner, uint _ownershipDuration) public {
+        require(_temporaryOwner != address(0), "Temp owner address cannot be 0");
+
+        temporaryOwner = _temporaryOwner;
+        expiryTime = now + _ownershipDuration;
```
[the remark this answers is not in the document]
ajsantander: Do you mean because of an arithmetic overflow?

publish/src/commands/deploy/deploy-core.js
justinjmoses: actually I see why this is - because owner is temp for OwnerRelayOnOptimism. I think we should reverse the order of these definitions - so `OwnerRelayOnEthereum` is deployed first - as it uses the standard mechanism

contracts/TemporarilyOwned.sol
```diff
@@ -0,0 +1,47 @@
...
+    constructor(address _temporaryOwner, uint _ownershipDuration) public {
+        require(_temporaryOwner != address(0), "Temp owner address cannot be 0");
+        require(_ownershipDuration > 0, "Duration cannot be 0");
+
+        temporaryOwner = _temporaryOwner;
+        expiryTime = block.timestamp + _ownershipDuration;
+    }
+
+    function setNewExpiryTime(uint _duration) external onlyTemporaryOwner {
+        require(_duration > 0, "Duration cannot be 0");
```
justinjmoses, line 17: this is not required - allow `0` so that the owner can effectively end this early

publish/src/commands/deploy/configure-system-settings.js
```diff
@@ -283,6 +284,17 @@ module.exports = async ({
 		comment: 'Set the gas limit for withdrawing from L2',
 	});
 
+	await runStep({
+		contract: 'SystemSettings',
+		target: SystemSettings,
+		read: 'crossDomainMessageGasLimit',
+		readArg: 4,
+		expected: input => input !== '0', // only change if zero
```
justinjmoses, line 292: please add `readTarget: previousSystemSettings` as well as above (changed since you copied the block here)

test/contracts/TemporarilyOwned.js
```diff
@@ -0,0 +1,233 @@
...
+			assert.bnClose(
+				expectedExpiry.toString(),
+				(await TestableTempOwned.expiryTime()).toString(),
+				'10'
+			);
+		});
+
+		describe('before expiration', () => {
+			it('only allows the owner to execute', async () => {
+				await onlyGivenAddressCanInvoke({
```
justinjmoses, line 65: please also add the `reason`

test/contracts/TemporarilyOwned.js
```diff
@@ -0,0 +1,233 @@
...
+			});
+
+			assert.bnClose(
+				expectedExpiry.toString(),
+				(await TestableTempOwned.expiryTime()).toString(),
+				'10'
+			);
+		});
+
+		it('should not allow nominated owner to accept ownership after expiration', async () => {
```
justinjmoses, line 135: 👍🏼

test/contracts/TemporarilyOwned.js
```diff
@@ -0,0 +1,233 @@
...
+			ownershipDuration = DAY;
+
+			expectedExpiry = (await currentTime()) + ownershipDuration;
+
+			TestableTempOwned = await TestableTempOwnedFactory.new(temporaryOwner, ownershipDuration, {
+				from: deployerAccount,
+			});
+		});
+
+		it('should not nominate new owner when not invoked by current contract owner', async () => {
```
justinjmoses, line 168: can you also check to be sure you can't nominate or accept after expiry? You've tested the modifier above but should assert here that the modifier is actually being used.

contracts/TemporarilyOwned.sol
barrasso: https://github.com/Synthetixio/synthetix/commit/b33a1cdefab43ac5368d064a731b78754d02c647

test/contracts/TemporarilyOwned.js
barrasso: https://github.com/Synthetixio/synthetix/commit/cb07d300cc1ce3aaa0cd211b1d5ac5c44cd1eafb

test/contracts/TemporarilyOwned.js
barrasso: https://github.com/Synthetixio/synthetix/commit/cb07d300cc1ce3aaa0cd211b1d5ac5c44cd1eafb

publish/src/commands/deploy/configure-system-settings.js
barrasso: https://github.com/Synthetixio/synthetix/commit/5b3624a1ef0c5a04829633fe97e90f9b11c963e2

justinjmoses (approved):

Issue #1403, opened 2021-07-13
`npx hardhat ops` doesn't detect if ops node running correctly


liamzebedee:
It seems the logic here is broken.

Steps to repro:

 1. Run a different docker-based project. In this circumstance, I ran graph-node, which starts 3 services in docker compose. https://github.com/graphprotocol/graph-node/tree/master/docker
 2. Attempt to run the ops instance using `npx hardhat ops --start`. It will give 
```
(base) ➜  synthetix git:(develop) ✗ npx hardhat ops --start --detached
optimism branch: develop
optimism commit: e61514437b945a0050521e909bf757f8e3d34e15
optimism folder: ./optimism
starting
  check if services are running
already running
```

Additional debugging:

It looks like statusCode isn't checked correctly.

```
# this shows the graph-node docker containers
(base) ➜  ops git:(e6151443) docker ps -q --no-trunc
ec07b148d7614ffdbde3320610466b5cbdb87b2316ec44b8f38a0c0e05601131
e9624ab2ee6218ea95219b7e2c867bcb47f958b669632171c473ad9a5294a48d
f73f0e8cde9741584fe74755c7c4c8b5534bbb3f9c34031920be0dafe730a4ee 

# ops node isn't running inside synthetix/optimism/ops
(base) ➜  ops git:(e6151443) docker-compose ps
Name   Command   State   Ports
------------------------------

# based on the output of this command, the status code is -1, but doesn't seem the script catches this
(base) ➜  ops git:(e6151443) docker-compose ps -q  ec07b148d7614ffdbde3320610466b5cbdb87b2316ec44b8f38a0c0e05601131
ERROR: No such service: ec07b148d7614ffdbde3320610466b5cbdb87b2316ec44b8f38a0c0e05601131
(base) ➜  ops git:(e6151443) echo $?
1
```

https://github.com/Synthetixio/synthetix/blob/dd40ad55d8cd969c03594cc91ceba20c58118ae8/hardhat/tasks/task-ops.js#L106

leomassazza:
Fix done in this PR: https://github.com/Synthetixio/synthetix/pull/1405

PR #1405, opened 2021-07-14
Fix issue 1403 - npx hardhat ops doesn't detect if ops node running correctly

leomassazza:
This PR fixes the issue found by Liam [Issue-1403](https://github.com/Synthetixio/synthetix/issues/1403) where ops is not detecting correctly if ops is running when other docker-compose based image is running.

### DIFF
--- a/hardhat/tasks/task-ops.js
+++ b/hardhat/tasks/task-ops.js
@@ -114,6 +114,10 @@ function _isRunning({ opsPath }) {
 			if (err.exitCode === 1) {
 				result = false;
 			}
+			if (err.exitCode === 2) {
+				// Usually the grep cmd failed because there's no output from docker-compose ps -q ...
+				result = false;
+			}
 		}
 	});
 	return result;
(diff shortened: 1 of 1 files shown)


hardhat/tasks/task-ops.js
```diff
@@ -114,6 +114,10 @@ function _isRunning({ opsPath }) {
 			if (err.exitCode === 1) {
 				result = false;
 			}
+			if (err.exitCode === 2) {
```
leomassazza, line 117: The shell command will return error code 2 if the grep command failed. That case wasn't covered.

liamzebedee: Wouldn't it be more robust to check if the command succeeded at all, rather than specific error cases?  https://tldp.org/LDP/abs/html/exitcodes.html

leomassazza: makes sense. And is simpler. Thanks for the comment. Fixed

ajsantander (approved):
Cool!

liamzebedee:
@leomassazza I think it'd be more robust if `result = false` initially and you flagged it true only if you could prove it was running. Just a suggestion.

liamzebedee (approved):

PR #1410, opened 2021-07-16
Add a released flag to releases list

mjlescano:
This PR will add a flag to mark when we make releases, and avoid unnecessary compilation of forks.


ajsantander (approved):

bin.js
```diff
@@ -182,6 +182,18 @@ program
 		console.log(JSON.stringify(versions, null, 2));
 	});
 
+program
+	.command('releases')
+	.description('Get the list of releases')
+	.option('--no-released', 'Only retreive the unreleased ones.')
```
ajsantander, line 188: Could we rename this to `--unreleased`?

mjlescano: I'm using this flag because by default on commander if you add `--no-*` it will add the `--*` with `true` value. So, on the action the `released` argument will be `true` by default: https://www.npmjs.com/package/commander#default-option-value

ajsantander: Ok, sounds good 👍

PR #1414, opened 2021-07-20
SIP-142 deprecate ethercollateral loans

jacko125:


package.json
```diff
@@ -78,6 +78,7 @@
 		"@eth-optimism/hardhat-ovm": "0.2.2",
 		"@eth-optimism/smock": "1.1.4",
 		"@eth-optimism/solc": "0.5.16-alpha.7",
+		"@nomiclabs/ethereumjs-vm": "^4.2.2",
```
justinjmoses, line 81: Why this?

justinjmoses (changes requested):
Also we need an entry in releases.JSON for these changes

package.json
jacko125: image

publish/releases.json
```diff
@@ -275,5 +275,23 @@
 		"ovm": true,
 		"sources": ["EtherWrapper", "Exchanger", "Synthetix", "SystemSettings"],
 		"sips": [138, 139, 140]
+	},
+	{
+		"name": "Mirfak",
```
justinjmoses, line 280: Note: in terms of timing, this will likely come after the exchanging release on L2 this week and thus need to use the next release.

justinjmoses (approved):
Thanks! But let's HOLD on merging until after the L2 release.

package.json
justinjmoses: According to Kelvin, this is no longer needed with hardhat 2.5.0 - so if you've merged in develop you can remove this

publish/releases.json
jacko125: That's fine, will move it to next release

PR #1415, opened 2021-07-20
Add acceptRelayOwner script

leomassazza:
Add new script to bulk accept ownership by OwnerRelayOnOptimism. 
This script checks all contracts in L2 to see if OwnerRelayOnOptimism is nominated as new owner and in that case accepts ownership using the acceptOwnershipOn function.

### DIFF
--- a/publish/src/commands/accept-relay-owner.js
+++ b/publish/src/commands/accept-relay-owner.js
@@ -1,0 +1,177 @@
+'use strict';
+
+const ethers = require('ethers');
+const { gray, yellow, red, cyan } = require('chalk');
+
+const {
+	constants: { CONFIG_FILENAME, DEPLOYMENT_FILENAME },
+} = require('../../..');
+
+const {
+	ensureNetwork,
+	ensureDeploymentPath,
+	getDeploymentPathForNetwork,
+	loadAndCheckRequiredSources,
+	loadConnections,
+	confirmAction,
+} = require('../util');
+
+const DEFAULTS = {
+	gasPrice: '15',
+};
+
+const acceptRelayOwner = async ({
+	network,
+	deploymentPath,
+	gasPrice,
+	yes,
+	useOvm,
+	useFork,
+	providerUrl,
+}) => {
+	ensureNetwork(network);
+	deploymentPath = deploymentPath || getDeploymentPathForNetwork({ network, useOvm });
+	ensureDeploymentPath(deploymentPath);
+
+	function logTx(tx) {
+		console.log(gray(`  > tx hash: ${tx.transactionHash}`));
+	}
+
+	if (!useOvm) {
+		console.error(red('Only runnable in ovm'));
+		process.exit(1);
+	}
+
+	// ensure all nominated owners are accepted
+	const { config, deployment } = loadAndCheckRequiredSources({
+		deploymentPath,
+		network,
+	});
+
+	const { providerUrl: envProviderUrl } = loadConnections({
+		network,
+		useFork,
+	});
+
+	if (!providerUrl) {
+		if (!envProviderUrl) {
+			throw new Error('Missing .env key of PROVIDER_URL. Please add and retry.');
+		}
+
+		providerUrl = envProviderUrl;
+	}
+
+	const provider = new ethers.providers.JsonRpcProvider(providerUrl);
+
+	const wallet = provider.getSigner();
+	console.log(gray(`Using account with public key ${await wallet.getAddress()}`));
+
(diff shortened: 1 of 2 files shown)


publish/src/commands/accept-relay-owner.js
```diff
@@ -0,0 +1,177 @@
...
+	ensureNetwork,
+	ensureDeploymentPath,
+	getDeploymentPathForNetwork,
+	loadAndCheckRequiredSources,
+	loadConnections,
+	confirmAction,
+} = require('../util');
+
+const DEFAULTS = {
+	gasPrice: '15',
```
ajsantander, line 20: There is a default gas price for this in the main index.js. New gas rules by optimism require that it is always 0.015 gwei.

publish/src/commands/accept-relay-owner.js
```diff
@@ -0,0 +1,177 @@
...
+};
+
+const acceptRelayOwner = async ({
+	network,
+	deploymentPath,
+	gasPrice,
+	yes,
+	useOvm,
```
ajsantander, line 28: Should we remove this as an option and hardcode it to true?

publish/src/commands/accept-relay-owner.js
```diff
@@ -0,0 +1,177 @@
...
+	function logTx(tx) {
+		console.log(gray(`  > tx hash: ${tx.transactionHash}`));
+	}
+
+	if (!useOvm) {
+		console.error(red('Only runnable in ovm'));
+		process.exit(1);
+	}
+
+	// ensure all nominated owners are accepted
```
ajsantander, line 45: Rogue comment?

publish/src/commands/accept-relay-owner.js
```diff
@@ -0,0 +1,177 @@
...
+
+		providerUrl = envProviderUrl;
+	}
+
+	const provider = new ethers.providers.JsonRpcProvider(providerUrl);
+
+	const wallet = provider.getSigner();
+	console.log(gray(`Using account with public key ${await wallet.getAddress()}`));
+
+	console.log(gray(`Gas Price: ${gasPrice} gwei`));
```
ajsantander, line 69: Not much point on logging this since its supposed to be fixed in Optimism anyways.

publish/src/commands/accept-relay-owner.js
```diff
@@ -0,0 +1,177 @@
...
+
+	// Get OwnerRelayOnOptimism contract and connect to wallet
+	const { address: relayAddress, source: relaySource } = deployment.targets['OwnerRelayOnOptimism'];
+	let OwnerRelayOnOptimism = new ethers.Contract(
+		relayAddress,
+		deployment.sources[relaySource],
+		provider
+	);
+	OwnerRelayOnOptimism = OwnerRelayOnOptimism.connect(wallet);
+
```
ajsantander, line 79: Do we want to check the "health" of OwnerRelayOnOptimism here? I.e. that it is set up with the right resolver, etc.

publish/src/commands/accept-relay-owner.js
```diff
@@ -0,0 +1,177 @@
...
+			console.log(gray(`${relayAddress} is already the owner of ${contract}`));
+		} else if (nominatedOwner === relayAddress) {
+			// continue if no pending tx found
+			await confirmOrEnd(yellow(`Confirm: Accept ownership on ${contract}?`));
+
+			console.log(yellow(`Accepting ownership on ${contract}...`));
+
+			try {
+				const tx = await OwnerRelayOnOptimism.acceptOwnershipOn(address, {
+					gasLimit: undefined, // Only run on optimism
```
ajsantander, line 132: I'd delete this line. Not specifying it is enough to state that you want Ethers.js to query the gasLimit for you.

publish/src/commands/accept-relay-owner.js
```diff
@@ -0,0 +1,177 @@
...
+					gasPrice: ethers.utils.parseUnits(gasPrice, 'gwei'),
+				});
+				const receipt = await tx.wait();
+				logTx(receipt);
```
ajsantander, line 136: Not sure we need an actual `logTx` function, since it's only called from here.

ajsantander (approved):
Looks good! Added a few minor comments.
Perhaps one not so minor, the health check of OwnerRelayOnOptimism.

publish/src/commands/accept-relay-owner.js
leomassazza: ok

PR #1418, opened 2021-07-21
Removes Web3 from the deploy script

ajsantander:


leomassazza (approved):
LGTM

publish/src/Deployer.js
```diff
@@ -121,41 +106,38 @@ class Deployer {
...
 			value: 0,
 			gasPrice: ethers.utils.parseUnits(this.gasPrice.toString(), 'gwei'),
 		};
 
-		if (this.useOvm) {
-			tx.gas = await this.provider.web3.eth.estimateGas(tx);
-		} else {
-			tx.gas = 1000000;
+		if (!this.useOvm) {
+			tx.gasLimit = 1000000;
```
justinjmoses, line 121: why remove this?

justinjmoses: ah because of below with the override gas limit

justinjmoses: then perhaps this isn't even needed?

publish/src/command-utils/transact.js
```diff
@@ -21,6 +21,7 @@ const performTransactionalStep = async ({
 	writeArg, // none, 1 or an array of args, array will be spread into params
 	gasLimit,
 	gasPrice,
+	generateSolidity,
```
justinjmoses, line 24: was this lost before somehow in a merge?

justinjmoses (approved):
just a couple of comments and a question

publish/src/command-utils/transact.js
ajsantander: Yeah, we renamed `performTransactionalStep` to `performTransactionalStepWeb3` and introduced a new `performTransactionalStep` in transact.js, which some non-deploy scripts (already ported to ethers) where using.

The two functions came out of sync, and I manually compared them and introduced these changes.

publish/src/Deployer.js
ajsantander: Will review and handle in a separate PR.

PR #1423, opened 2021-07-25
delete useless readProxy

zgfzgf:
rm readProxy.sol


justinjmoses:
Ahhh, the ReadProxy is used for the `ReadProxyAddressResolver`

PR #1430, opened 2021-07-28
canConfirmOrder respects max market size constraints + error management.

zyzek:


contracts/FuturesMarket.sol
```diff
@@ -576,6 +585,10 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         return 0 <= a * b;
     }
 
+    function _error(Error error) internal view {
+        require(error == Error.Ok, _errorMessages[uint(error)]);
```
liamzebedee, line 712: I like the refactoring into error types. Ok being a "subtype" of Error doesn't make sense IMO. Maybe something like `require(error == Error.Null, _errorMessages[uint(error)]);`, or maybe `require(result == 0, _errorMessages[uint(result)]`? Wish we had Rust's Result<T> here.

contracts/FuturesMarket.sol
```diff
@@ -80,6 +84,9 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
 
     uint internal _nextOrderId = 1; // Zero reflects an order that does not exist
 
+    // Holds the revert message for each type of error.
+    mapping(uint => string) internal _errorMessages;
```
liamzebedee, line 84: Potentially worth using a small uint type here to save on gas?

contracts/FuturesMarket.sol
```diff
@@ -743,7 +755,20 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+        return Error.Ok;
+    }
+
+    function _requireOrderSizeSmallEnough(
+        int size,
+        int newSize,
+        bool sameSide,
+        uint play
+    ) internal view {
+        _error(_orderSizeSmallEnough(size, newSize, sameSide, play));
```
liamzebedee: What do you think of giving it an more imperative name - something like `checkError` or `try` etc?

contracts/FuturesMarket.sol
```diff
@@ -821,57 +846,95 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+        }
 
         // We weren't liquidated, so we realise the margin to compute the new position size.
-        // The fee is deducted at this stage; the transaction will revert if the realised margin minus the fee is negative.
-        uint margin = _realiseMargin(position, fundingIndex, price, -int(order.fee));
-        // The order size is computed pre-fee, so the order size is accurate, though their leverage may
+        // The fee is deducted at this stage; it is an error if the realised margin minus the fee is negative or subject to liquidation.
+        uint fee = order.fee;
+        (uint margin, Error marginError) = _realisedMargin(position, fundingIndex, price, -int(fee));
+        if (marginError != Error.Ok) {
```
liamzebedee: Yeah this usage feels like Go's error semantics. `Error.Nil` could be another idea for a name.

contracts/FuturesMarket.sol
```diff
@@ -304,19 +319,139 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+        if (maxSize < _abs(newSideSize.div(2))) {
+            return Error.MaxMarketSizeExceeded;
+        }
+
+        return Error.Ok;
+    }
+
+    // TODO: Ensure that this is fine if the position is swapping sides
+    // TODO: Check that everything is fine if a position already exists.
+    function _orderStatusDetails(
```
liamzebedee, line 379: 👏  Love it.

test/contracts/FuturesMarket.js
```diff
@@ -18,6 +18,19 @@ const {
 	ensureOnlyExpectedMutativeFunctions,
 } = require('./helpers');
 
+const orderStatuses = {
```
liamzebedee, line 21: At this point I'm nitpicking naming haha - though is there a reason you don't refer to this as "futures errors"?

liamzebedee (changes requested):
Great PR

contracts/FuturesMarket.sol
```diff
@@ -304,19 +316,139 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+        if (0 < newSize) {
+            // long case: marketSize + skew = 2 * longSize
+            newSideSize = newMarketSize.add(newSkew);
+        } else {
+            // short case: marketSize - skew = 2 * shortSize
+            newSideSize = newMarketSize.sub(newSkew);
+        }
+
+        // newSideSize still includes an extra factor of 2 here, so we will divide by 2 in the actual condition
+        if (maxSize < _abs(newSideSize.div(2))) {
```
jacko125, line 370: @zyzek Where does the factor of 2 in the `newSideSize` gets introduced in the calculation in `newSideSize = newMarketSize.add(newSkew)` ?

jacko125 (approved):
looks good to me the changes. I had some questions on the max market size constraint and how the newSideSize introduces a factor of 2.

contracts/FuturesMarket.sol
zyzek: Hm, I'll have to sleep on this. `Error.Null` feels like it could indicate that the error is that something is null, like Java's null pointer exception. I wanted a name that indicated with positivity that the result was acceptable.

contracts/FuturesMarket.sol
zyzek: Hm, is hashing a uint8 cheaper? I guess I'll change it to uint8 just so it implies we're handling the enum type.

test/contracts/FuturesMarket.js
zyzek: Fair, I'll change this.

contracts/FuturesMarket.sol
zyzek: I've added a bit more explanation in the comments.

contracts/FuturesMarket.sol
```diff
@@ -910,27 +917,46 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+        uint minPrice,
+        uint maxPrice
+    ) public optionalProxy {
         uint price = _assetPriceRequireNotInvalid();
         uint fundingIndex = _recomputeFunding(price);
-        _submitOrder(leverage, price, fundingIndex, messageSender);
+        _submitOrder(leverage, price, [minPrice, maxPrice], fundingIndex, messageSender);
+    }
+
+    function submitOrder(int leverage) external optionalProxy {
```
jacko125, line 949: looks good.

contracts/FuturesMarket.sol
liamzebedee: How about something language agnostic, like Error.None or Error.NoError (kinda like [Compound](https://sourcegraph.com/github.com/compound-finance/compound-protocol/-/blob/contracts/ErrorReporter.sol?L5))?

contracts/FuturesMarket.sol
liamzebedee: > Hm, is hashing a uint8 cheaper?

Just checked, no it's not.

contracts/FuturesMarket.sol
zyzek: Ok, let's go with `Error.None`.

liamzebedee (approved):

PR #1432, opened 2021-07-28
Fixes to the deploy script regarding ethers.js

ajsantander:


publish/src/commands/deploy/rebuild-resolver-caches.js
```diff
@@ -77,27 +77,34 @@ module.exports = async ({
 	const resolverAddressesRequired = (
 		await Promise.all(
 			contractsWithRebuildableCache.map(([id, contract]) => {
-				return limitPromise(() => contract.resolverAddressesRequired()).then(result => [
-					contract.address,
-					result,
-				]);
+				return limitPromise(() => contract.resolverAddressesRequired())
+					.then(result => [contract.address, result])
+					.catch(() => {
```
ajsantander, line 82: Not sure why this wasn't happening with Web3. Depot.sol doesn't seem to have this function and it obviously throws when its called.

publish/src/commands/deploy/rebuild-resolver-caches.js
```diff
@@ -135,7 +142,16 @@ module.exports = async ({
 		contractsToRebuildCache = Array.from(contractsToRebuildCacheSet);
 	} else {
 		for (const [name, target] of contractsWithRebuildableCache) {
-			const isCached = await target.isResolverCached();
+			let isCached = true;
+
+			try {
+				isCached = await target.isResolverCached();
```
ajsantander, line 148: Depot also doesn't have this function.

publish/src/commands/deploy/system-and-parameter-check.js
```diff
@@ -159,14 +159,7 @@ module.exports = async ({
 		formatUnits(await deployer.provider.getBalance(account), 'ether'),
 		10
 	);
-	if (useFork) {
```
ajsantander, line 162: Decided to remove this. With web3 it worked because account 0 had ETH. An implementation in Ethers.js would be more involved, and should probably live in a separate tool.

test/integration/utils/bootstrap.js
```diff
@@ -115,6 +115,12 @@ function bootstrapDual({ ctx }) {
 	});
 }
 
+// Note: Currently not aware of a way to set a default gas price globally
+// on standalone ethers (used directly without hardhat).
+function _setDefaultGasPrice({ provider, gasPrice }) {
+	provider.getGasPrice = async () => ethers.utils.parseUnits(OVM_GAS_PRICE_GWEI, 'gwei');
```
leomassazza, line 121: Does it need to be async for any obscure requirement?

ajsantander: Ethers.js expects it to be async. If not defined like this, it will throw complaining that it doesn't have a `catch` definition.

justinjmoses (approved):

PR #1435, opened 2021-07-28
Remove some solidity warnings

leomassazza:
This PR reduces the number of warnings shown when compiling the contracts. 

Some compiler warnings are still present. 

-  **Experimental features are turned on...**: because we use `pragma experimental ABIEncoderV2;`. Those will remain unless we migrate to latest solidity version
- contracts/Exchanger.sol:536:5: Warning: **Function state mutability can be restricted to pure**: that function is overriding in another contract where is not pure.
- contracts/NativeEtherWrapper.sol:82:9: Warning: **Return value of low-level calls not used.**: we can assign the result (and check if the call failed)

### DIFF
--- a/contracts/BaseSynthetix.sol
+++ b/contracts/BaseSynthetix.sol
@@ -314,7 +314,7 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         bytes32,
         address,
         bytes32
-    ) external returns (uint amountReceive) {
+    ) external returns (uint amountReceived) {
         // FIXME remove parameter name in returns. Just (uint)
         _notImplemented();
     }
--- a/contracts/BaseSynthetix.sol
+++ b/contracts/BaseSynthetix.sol
@@ -314,7 +314,7 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         bytes32,
         address,
         bytes32
-    ) external returns (uint amountReceived) {
+    ) external returns (uint) {
         _notImplemented();
     }
 
@@ -358,7 +358,7 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         _;
     }
 
-    function _systemActive() private {
+    function _systemActive() private view {
         systemStatus().requireSystemActive();
     }
 
@@ -367,7 +367,7 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         _;
     }
 
-    function _issuanceActive() private {
+    function _issuanceActive() private view {
         systemStatus().requireIssuanceActive();
     }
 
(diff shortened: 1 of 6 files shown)


contracts/NativeEtherWrapper.sol
```diff
@@ -65,7 +65,7 @@ contract NativeEtherWrapper is Owned, MixinResolver {
 
     function burn(uint amount) public {
         require(amount > 0, "amount must be greater than 0");
-        IWETH weth = weth();
+        IWETH _weth = weth();
```
leomassazza, line 68: renamed to prevent naming collision with weth()

contracts/test-helpers/TestableAddressSet.sol
```diff
@@ -31,7 +31,7 @@ contract TestableAddressSet {
         return set.elements[index];
     }
 
-    function index(address element) public view returns (uint) {
-        return set.indices[element];
+    function index(address _element) public view returns (uint) {
```
leomassazza, line 34: renamed to prevent naming collision with element()

contracts/test-helpers/TestableBytes32Set.sol
```diff
@@ -31,7 +31,7 @@ contract TestableBytes32Set {
         return set.elements[index];
     }
 
-    function index(bytes32 element) public view returns (uint) {
-        return set.indices[element];
+    function index(bytes32 _element) public view returns (uint) {
```
leomassazza, line 34: renamed to prevent naming collision with element()

contracts/BaseSynthetix.sol
```diff
@@ -358,7 +358,7 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         _;
     }
 
-    function _systemActive() private {
+    function _systemActive() private view {
```
ajsantander, line 361: FYI: There is an ongoing argument within Solidity devs on wether a function that can revert be considered a view function. I think that what you did is fine, I just want you to be aware of this subtlety. These are strange view functions, cause they can cause the execution thread to throw.

contracts/NativeEtherWrapper.sol
```diff
@@ -23,7 +23,7 @@ contract NativeEtherWrapper is Owned, MixinResolver {
     /* ========== PUBLIC FUNCTIONS ========== */
 
     /* ========== VIEWS ========== */
-    function resolverAddressesRequired() public view returns (bytes32[] memory addresses) {
+    function resolverAddressesRequired() public view returns (bytes32[] memory) {
```
ajsantander, line 26: I would check what other contracts do and be consistent. The options are:
1) Not name the return arg (like its done here)
2) Name the return arg and avoid the return statement

ajsantander (approved):
LGTM

contracts/NativeEtherWrapper.sol
```diff
@@ -24,10 +24,9 @@ contract NativeEtherWrapper is Owned, MixinResolver {
 
     /* ========== VIEWS ========== */
     function resolverAddressesRequired() public view returns (bytes32[] memory addresses) {
-        bytes32[] memory addresses = new bytes32[](2);
```
leomassazza, line 27: updated to align with other contracts use pattern

ajsantander:
@leomassazza can we also tell solhint and/or solc to ignore the warnings?

PR #1436, opened 2021-07-29
Futures markets respect fee reclamation.

zyzek:


contracts/FuturesMarketManager.sol
```diff
@@ -131,11 +140,30 @@ contract FuturesMarketManager is Owned, MixinResolver, Proxyable, IFuturesMarket
...
+        // We'll settle first, in order to ensure the user has sufficient balance.
+        // If the settlement reduces the user's balance below the requested amount,
+        // the settled remainder will be the resulting deposit.
+
+        // Exchanger.settle ensures synth is active
+        ISynth sUSD = _sUSD();
+        (, , uint numEntriesSettled) = _exchanger().settle(account, SUSD);
+
+        uint balanceAfter = amount;
+        if (numEntriesSettled > 0) {
```
jacko125, line 157: @zyzek It'll be better to check for the `reclaimed` amount if that's > 0, otherwise we can skip updating the balanceAfter. As `numEntriesSettled` will be more always 1+ even if theres no reclaim or refund amount (it indicates that the entry was settled).

jacko125 (approved):
lgtm!

PR #1437, opened 2021-07-29
Add status task to hardhat

leomassazza:
This PR adds a new script to run from hardhat. Based (copied) on `synthetix-scripts`'s status.js. It allows the user to see state of the system.

Adds two capabilities:
- get ExchangeRates of all symbols (default) or specify which ones to retrieve.
- run a subset of actions to get specific system state.

i.e.
`npx hardhat status --currency-keys SNX,iCOMP `
will get only exchangeRate of SNX and iCOMP instead of the whole list

`npx hardhat status --execute-actions Synthetix,SystemSettings,FeePool`
will only get status information of Synthetix, SystemSettings and FeePool

or combining both
`npx hardhat status --currency-keys SNX,iCOMP --execute-actions ExchangeRates`
will only get ExchangeRates for SNX and iCOMP

The complete list of available actions/contracts to query up to now is:
- Synthetix
- DebtCache
- SynthetixState
- SupplySchedule
- FeePool
- FeePoolState
- AddressResolver
- SystemSettings
- ExchangeRates


hardhat/tasks/task-status.js
```diff
@@ -0,0 +1,70 @@
...
+	.addFlag('useOvm', 'Use an Optimism chain')
+	.addFlag('useFork', 'Use a local fork')
+	.addOptionalParam('targetNetwork', 'The network to run off', 'mainnet')
+	.addOptionalParam('addresses', 'Addresses to perform particular checks on')
+	.addOptionalParam('currencyKeys', 'Keys to get exchange rate on')
+	.addOptionalParam('executeActions', 'Areas to get state')
+	.addOptionalParam('block', 'Block number to check again')
+	.addOptionalParam('providerUrl', 'The http provider to use for communicating with the blockchain')
+	.addOptionalParam('deploymentPath', 'Specify the path to the deployment data directory')
+	.setAction(async (taskArguments, hre, runSuper) => {
```
mjlescano, line 28:
```suggestion
	.setAction(async (taskArguments, hre) => {
```

ajsantander (approved):
Looks awesome @leomassazza !!

A few comments:
* Can we remove the "get" from actions. Even though including it is more rational, I think it will get in the way in terms of UX.
* I'd prefer if we did not introduce hardhat task specific utils. We're spreading utils like getContract all over the place - in fact, we will probably need to unify them all soon. I'd rather putting them in place in the script or using things like the interact does, via synthetix.
* I'd also rather moving statusActions out of utils and instead put this task under a folder "status", where the main file is index.js, next to "actions.js" or something. Like the deploy script declares a folder because its a multi-file script.
* The latter two opinions would imply the deletion of hardhat/util.

PR #1438, opened 2021-07-29
Add L2 governance bridge relay temp circumvention

mjlescano:
This PR adds the method `directRelay` to the `OwnerRelayOnOptimism` contract so we can directly execute any relayed method directly on L2, circumventing the governance bridge functionality. The `OwnerRelayOnOptimism` contract now has to be instantiated with a TempOwner, which is the only address allowed to execute the `directRelay` method, and an EOL date for this permission. This is just a temporary precaution in case we encounter a bug on the new governance bridge and avoid a potential loss ownership of all the relayed contracts.

## Notes
* ⚠️  The temporary owner and EOL date can only be assigned when calling the constructor of `OwnerRelayOnOptimism` and cannot be modified after that. ⚠️
* Not assigning an address or a EOL will make the check fail, in case we don't want to add the backdoor at all.

## TODO
- [ ] Update deploy configuration


contracts/OwnerRelayOnOptimism.sol
```diff
@@ -58,6 +63,15 @@ contract OwnerRelayOnOptimism is MixinResolver {
         emit RelayFinalized(target, data);
     }
 
+    function directRelay(address target, bytes calldata data) external onlyTemporaryOwner {
+        // solhint-disable avoid-low-level-calls
+        (bool success, bytes memory result) = target.call(data);
+
+        require(success, string(abi.encode("xChain call failed:", result)));
+
+        emit RelayFinalized(target, data);
```
ajsantander, line 72: Perhaps this logic could be contained in an internal `_relay(...)` function that both call?

contracts/TempOwned.sol
```diff
@@ -0,0 +1,21 @@
...
+        tempOwnerEOL = _tempOwnerEOL;
+    }
+
+    modifier onlyTemporaryOwner {
+        _onlyTemporaryOwner();
+        _;
+    }
+
+    function _onlyTemporaryOwner() private view {
+        require(tempOwnerEOL >= now, "The EOL date for executing this function already reached");
```
ajsantander, line 18: It's a good idea to keep errors < 32 chars so that it fits in a 32 byte word, which is kinda like the Ethereum storage unit size.

leomassazza (approved):
LGTM

contracts/OwnerRelayOnOptimism.sol
```diff
@@ -30,6 +35,15 @@ contract OwnerRelayOnOptimism is MixinResolver {
         return requireAndGetAddress(CONTRACT_BASE_OWNER_RELAY_ON_ETHEREUM);
     }
 
+    function _relayCall(address target, bytes memory data) private {
+        // solhint-disable avoid-low-level-calls
+        (bool success, bytes memory result) = target.call(data);
+
+        require(success, string(abi.encode("xChain call failed:", result)));
+
+        emit RelayFinalized(target, data);
```
ajsantander, line 44: This event should probably not be here but in `finalizeRelay(...)`

contracts/test-helpers/TestableTempOwned.sol
```diff
@@ -0,0 +1,33 @@
+pragma solidity ^0.5.16;
+
+import "../TempOwned.sol";
+
+contract TestableTempOwned is TempOwned {
+    uint private meaningOfLife = 42;
+
+    constructor(address _tempOwner, uint _tempOwnerEOL) public TempOwned(_tempOwner, _tempOwnerEOL) {}
+
+    function getMeaningOfLife() external view onlyTemporaryOwner returns (uint) {
```
ajsantander, line 10: It's kinda weird to use an accessibility modifier on a view only function, because anything is public at a low level in Ethereum.

contracts/TempOwned.sol
```diff
@@ -0,0 +1,21 @@
+pragma solidity ^0.5.16;
+
+contract TempOwned {
+    address internal tempOwner;
+    uint internal tempOwnerEOL;
```
ajsantander, line 5: This can be a uint64

ajsantander (approved):

contracts/TempOwned.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity ^0.5.16;
+
+contract TempOwned {
```
justinjmoses, line 3: Maybe call this `TemporaryOwnership`? It's more explicit.

contracts/OwnerRelayOnOptimism.sol
```diff
@@ -10,26 +11,39 @@ interface IOwned {
...
     bytes32 private constant CONTRACT_EXT_MESSENGER = "ext:Messenger";
     bytes32 private constant CONTRACT_BASE_OWNER_RELAY_ON_ETHEREUM = "base:OwnerRelayOnEthereum";
 
     /* ========== CONSTRUCTOR ============ */
 
-    constructor(address _resolver) public MixinResolver(_resolver) {}
+    constructor(
+        address _resolver,
+        address _tempOwner,
+        uint _tempOwnerEOL
```
justinjmoses, line 25: Maybe instead of passing the EOL time, we pass the number of seconds it should be mutatable for? Like `LimitedSetup.sol` does?

justinjmoses: That way we avoid things like this in the deploy script:

```javascript
const EOL = Math.floor(new Date().getTime() / 1000) + TEMP_OWNER_DEFAULT_EOL;
```

contracts/TempOwned.sol
ajsantander: IMO your suggestion is prettier, but Mati's choice is more consistent with `Owned`

contracts/OwnerRelayOnOptimism.sol
ajsantander: What would be the advantage of this vs specifying a date in the future that will no longer allow ownership, as it is currently implemented?

contracts/TempOwned.sol
ajsantander: Ok you win. Renaming to TemporarilyOwned!

contracts/OwnerRelayOnOptimism.sol
ajsantander: Yep, it's less prone to mistakes. Changing.

i-stam: I guess the EOL would be set in the constructor (`block.timestamp.add(duration)`) instead of passing the EOL. The basic sanity check is in `TempOwned() `constructor so it should be fine. I am wondering if we can make it more resilient to error.

i-stam (approved):
LGTM: left a minor comment (edit: let's use `block.timestamp` since `now` is deprecated in newer versions of solidity)

contracts/TemporarilyOwned.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity ^0.5.16;
+
+contract TemporarilyOwned {
+    address public tempOwner;
+    uint public initialTimestamp;
+    uint public duration;
```
justinjmoses, line 6: Can we use the same pattern that LimitedSetup does and store the timestamp but calculate it on construction?  https://github.com/Synthetixio/synthetix/blob/master/contracts/LimitedSetup.sol#L12  Then we simplify L22 below.

index.js
```diff
@@ -125,6 +125,7 @@ const knownAccounts = {
 // The solidity defaults are managed here in the same format they will be stored, hence all
 // numbers are converted to strings and those with 18 decimals are also converted to wei amounts
 const defaults = {
+	TEMP_OWNER_DEFAULT_DURATION: 60 * 60 * 24 * 30 * 2, // 2 months
```
justinjmoses, line 128: [np] technically there is no official "month" duration, so it's really just 60 days.

contracts/TemporarilyOwned.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity ^0.5.16;
+
+contract TemporarilyOwned {
+    address public tempOwner;
```
justinjmoses, line 4: [np] shouldn't we call this `temporaryOwner` to be consistent?

test/contracts/TemporarilyOwned.js
```diff
@@ -0,0 +1,74 @@
...
+			assert.equal(tempOwner, await TestableTempOwned.tempOwner());
+		});
+
+		it('properly set duration', async () => {
+			assert.equal(duration, await TestableTempOwned.duration());
+		});
+
+		describe('before duration expires', () => {
+			it('does not allow any address to change the value', async () => {
+				await assert.revert(
```
justinjmoses, line 48: It seems to have been missed from last round - I recommend using `onlyGivenAddressCanInvoke` rather than just trying a random account (it cycles through all given accounts)

test/integration/dual/relay.integration.js
```diff
@@ -61,18 +86,55 @@ describe('owner relay integration tests (L1, L2)', () => {
 	});
 
 	describe('when SystemSettings on L2 is owned by the relay', () => {
+		let originalMinimumStakeTime;
```
justinjmoses, line 89: These dual tests are 🔥🔥🔥

test/integration/dual/relay.integration.js
```diff
@@ -61,18 +86,55 @@ describe('owner relay integration tests (L1, L2)', () => {
...
 		});
 
-		describe('when changing an L2 system setting with an L1 tx', () => {
-			let originalMinimumStakeTime;
-			const newMinimumStakeTime = '42';
+		before('store minimumStakeTime', async () => {
+			originalMinimumStakeTime = await SystemSettingsL2.minimumStakeTime();
+		});
+
+		describe('when changing an L2 system setting with directRelay', () => {
```
justinjmoses, line 100: Technically this is an L2 isolated test rather than dual, but I get it's easier to just connect it up here as we have the structure already.

justinjmoses (commented):
A few items left - nothing major

test/contracts/TemporarilyOwned.js
ajsantander: Unfortunately, this function is web3 only, and this test is using ethers. Will other other manual checks tho.

contracts/TemporarilyOwned.sol
```diff
@@ -0,0 +1,25 @@
...
+    uint public initialTimestamp;
+    uint public duration;
+
+    constructor(address _tempOwner, uint _duration) public {
+        require(_tempOwner != address(0), "Owner address cannot be 0");
+
+        tempOwner = _tempOwner;
+        initialTimestamp = now;
```
i-stam, line 12: I think that a more elegant way would be this: `expiration = block.timestamp.add(duration);`

i-stam: no need for extra variables

contracts/TemporarilyOwned.sol
```diff
@@ -0,0 +1,25 @@
...
+        duration = _duration;
+    }
+
+    modifier onlyTemporaryOwner {
+        _onlyTemporaryOwner();
+        _;
+    }
+
+    function _onlyTemporaryOwner() private view {
+        require(now - initialTimestamp < duration, "Ownership expired");
```
i-stam, line 22: and here the check would be simple: `block.timestamp < expiration`

i-stam (changes requested):

test/integration/dual/relay.integration.js
ajsantander: Yeah, agreed! Let's keep it here for now because it is easier to correlate with the rest of the tests.

contracts/TemporarilyOwned.sol
ajsantander: Yep! Implemented. Thank you.

contracts/TemporarilyOwned.sol
ajsantander: Implemented!

i-stam (approved):

PR #1440, opened 2021-08-02
Futures: Market deployment and management scripts

liamzebedee:
* moves futures config into JSON file, added default for all networks
 * updates deploy script to use this config
 * interact CLI tested with the FuturesMarketSettings
 * refactors setup unit test to work with testing aspects of L2 deployment. Adds new test for futures markets.

The L2 publish script tests are currently skipped. I think it might be best to open a separate PR for properly adding support for L2 publish script testing, in order to unblock some other work. Thoughts @zyzek?

Refs: https://github.com/Synthetixio/issues/issues/98


test/publish/index.js
```diff
@@ -1593,3 +1594,243 @@ describe('publish scripts', () => {
 		});
 	});
 });
+
+describe.skip('publish scripts (l2)', async () => {
```
zyzek, line 1598: This has a lot of commented-out pieces. If we're skipping it anyway, then maybe just throw it in a gist to attach to one of the tickets so it can be revisited. Not sure this should be committed if unused and messy.

zyzek (changes requested):

test/publish/index.js
liamzebedee: Yeah that's a fair solution actually.

liamzebedee: https://gist.github.com/liamzebedee/64f372465d9ef9aeb01174bdec998ac7

zyzek (approved):
ell gee tee em

PR #1441, opened 2021-08-02
Add integration tests for diverting inflation to L2

i-stam:


justinjmoses (approved):
Looks good once the tests pass

test/integration/dual/notifyRewardAmount.integration.js
```diff
@@ -0,0 +1,105 @@
+const ethers = require('ethers');
```
ajsantander, line 1: Would suggest to rename the file to `inflation.integration.js` to be more consistent with the other names.

test/integration/dual/notifyRewardAmount.integration.js
```diff
@@ -0,0 +1,105 @@
...
+		});
+
+		describe('when a new distribution is added (bridge)', () => {
+			let escrowBalance;
+
+			before('record values', async () => {
+				escrowBalance = await Synthetix.balanceOf(SynthetixBridgeEscrow.address);
+			});
+
+			before('add a new distribution and mint', async () => {
```
ajsantander, line 52: nit: split this before in 2:
1) describe "when a new distribution is added", then before "add a new distribution", then test that its there
2) describe "when mint is called", then before "mint", then test

test/integration/dual/notifyRewardAmount.integration.js
```diff
@@ -0,0 +1,105 @@
...
+						currentFeePeriodRewards.add(rewardsToDeposit)
+					);
+				});
+
+				it('increases the RewardEscrowV2 balance', async () => {
+					assert.bnEqual(
+						await Synthetix.balanceOf(RewardEscrowV2.address),
+						rewardEscrowBalanceL2.add(rewardsToDeposit)
+					);
+				});
```
ajsantander, line 101: We should probably check what `rewards.integration.js` does here right?
If so, we should probably try to abstract that into a behavior, so that we don't duplicate test code.

test/integration/dual/notifyRewardAmount.integration.js
```diff
@@ -0,0 +1,105 @@
...
+				RewardsDistribution,
+				Synthetix,
+				SynthetixBridgeEscrow,
+				SynthetixBridgeToOptimism,
+			} = ctx.l1.contracts);
+
+			owner = ctx.l1.users.owner;
+		});
+
+		before('approve if needed', async () => {
```
ajsantander, line 36: Why is this needed?

i-stam: not needed, good catch!

test/integration/dual/notifyRewardAmount.integration.js
i-stam: this is a good point,  functionality is shared

test/integration/dual/notifyRewardAmount.integration.js
i-stam: done

ajsantander (approved):

PR #1443, opened 2021-08-03
Extra integration tests for adding a new reward distribution on L2

i-stam:
- Add `TradingRewards` as a new reward distribution entry on L2.

- Tests whether the rewards as distributed as expected via `L2.RewardsDistribution.distributeRewards() `on L2.


index.js
```diff
@@ -169,7 +169,7 @@ const defaults = {
 	INITIAL_ISSUANCE: w3utils.toWei(`${100e6}`),
 	CROSS_DOMAIN_DEPOSIT_GAS_LIMIT: `${3e6}`,
 	CROSS_DOMAIN_ESCROW_GAS_LIMIT: `${8e6}`,
-	CROSS_DOMAIN_REWARD_GAS_LIMIT: `${3e6}`,
```
justinjmoses, line 172: why was this change needed?

justinjmoses (approved):
Just one question

index.js
i-stam: This is what was causing the tx to fail. I realised it after everything else that I tried failed.

PR #1445, opened 2021-08-05
Fix merge history for futures-implementation

liamzebedee:
I had an issue with my git setup, where the merge commits I had been creating were not referencing the correct parents in the git history. Some context:

> For a while now, my PR diffs have been displaying very strangely on Github - lots of noise, including diffs of files that were already existing in the target branch. I did some debugging with a mate, Anton and Jacko. It seems that the commits I created by running git merge xyz were missing the parent commit hashes that are usually part of the merge commit, which are instrumental for rendering the right diff in the PR view. The culprit? It seems like something was awry with my .gitconfig. Checking it, I realised there was a mergetool configured (Sourcetree if anyone knows it lol) - I ran the path specified and turns out it couldn't run because of some macOS perms issue. I removed the mergetool config line and then reran git merge xyz and voila, the merge commit had the parent commit hashes. 

I have merged `develop` into `futures-implementation` in multiple commits, however the merge history is incomplete. For the most recent merge in PR https://github.com/Synthetixio/synthetix/pull/1399, it does not have a parent which references the `develop` tip at the time of merge (this commit - https://github.com/Synthetixio/synthetix/pull/1399/commits/2ff91d9860e0eeee8bec63f5f1c750515aef908d) - as such, this PR contains a change which re-does that merge commit, with the correct parent reference. This fixes up the diff view:

 - Before: https://github.com/Synthetixio/synthetix/compare/develop...futures-implementation
 - After: https://github.com/Synthetixio/synthetix/compare/develop...fix-merge


zyzek (approved):

liamzebedee:
Follow-up - this didn't actually fix it, so I pushed 08057995f78ea17a05c6eb0a315f6b8b4bac1bd0 and it did.

I pushed the merge commit directly to futures-implementation - it doesn't contain any working tree changes, the only difference is that it includes the parents (see the Merge field) in the commit object. This seems to have fixed the history, though idk why the GitHub PR didn't work - I assume it has something to do with how it squashes commits

PR #1446, opened 2021-08-05
Futures order slippage

liamzebedee:
Refs: https://github.com/Synthetixio/issues/issues/211

### DIFF
--- a/contracts/FuturesMarket.sol
+++ b/contracts/FuturesMarket.sol
@@ -759,6 +759,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
     function _submitOrder(
         int leverage,
         uint price,
+        uint maxSlippage,
         uint fundingIndex,
         address sender
     ) internal {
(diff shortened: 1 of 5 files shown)


contracts/FuturesMarket.sol
```diff
@@ -759,6 +761,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
     function _submitOrder(
         int leverage,
         uint price,
+        uint[2] memory priceBounds,
```
zyzek, line 764: Why's this in an array? Is this a stack too deep thing?

test/contracts/FuturesMarket.js
```diff
@@ -1368,13 +1385,13 @@ contract('FuturesMarket', accounts => {
 		});
 	});
 
-	describe('Confirming orders', () => {
+	describe.only('Confirming orders', () => {
```
zyzek, line 1388: And there I thought the precommit constraints were supposed to catch .onlys

test/contracts/FuturesMarket.js
```diff
@@ -1456,6 +1473,30 @@ contract('FuturesMarket', accounts => {
...
 
+		it.only('Cannot confirm an order if the price slippage since order submission price exceeds tolerance', async () => {
+			const startPrice = toUnit('200');
+			await setPrice(baseAsset, startPrice);
+
+			const margin = toUnit('1000');
+			await futuresMarket.modifyMargin(margin, { from: trader });
+			const leverage = toUnit('10');
+			const maxSlippage = toUnit('0.01'); // 1%
+			const minPrice = startPrice.mul(toUnit(1).add(maxSlippage));
```
zyzek, line 1484: It looks to me like `minPrice` is greater than `maxPrice`, and under these conditions the order can never confirm, so much of the logic below isn't really testing anything.

Interestingly I guess if you submit an order with `currentPrice < minPrice`, the order won't confirm until the price rises above the specified point, which I guess means we have effectively created a limit order. Not sure whether we want to check at order submission time that the current price is within the bands or not.

test/contracts/FuturesMarketData.js
```diff
@@ -25,6 +25,7 @@ contract('FuturesMarketData', accounts => {
 	const trader2 = accounts[3];
 	const trader3 = accounts[4];
 	const traderInitialBalance = toUnit(1000000);
+	const defaultPriceBounds = [ethers.constants.Zero, ethers.constants.MaxUint256];
```
zyzek, line 28: We should be testing in the position details test below that these are present in the returned order struct and that they have accurate values.

zyzek (changes requested):

contracts/FuturesMarket.sol
liamzebedee: Yepp.

test/contracts/FuturesMarket.js
liamzebedee: @zyzek that's interesting. I guess we have limit orders now. 

Sorry, I messed up - fixing these now.

test/contracts/FuturesMarketData.js
liamzebedee: I've tested this here in order submission - is that sufficient? These params are only part of the order, not the position?

https://github.com/Synthetixio/synthetix/pull/1446/files#diff-ea3e2c01da4d3ffe85d68d81159997ff1b2a8ca320fdddf8cf5c2ff3ca165d98R1092

zyzek: The test is to ensure that the data contract is surfacing all data that it should be surfacing.

jacko125:
@liamzebedee 
Looks good the contract changes, 

So what happens is the order can't be confirmed when the slippage params are exceeding the max / min? But the order would still be valid and can be confirmed once the price moves into that range?

jacko125 (approved):

liamzebedee:
> So what happens is the order can't be confirmed when the slippage params are exceeding the max / min? But the order would still be valid and can be confirmed once the price moves into that range?

Exactly. Maybe at a later point we can choose to add order expiry. More higher priority items for now I guess.

test/contracts/FuturesMarketData.js
liamzebedee: Oh, I was looking at the wrong test. Yeah added this now.

PR #1447, opened 2021-08-05
Fix fork-test failures

leomassazza:
After some research I found the issue is that we are getting a timeout on ethers.js while using a fork. 

There's an issue filled in ethers project [Ethers Issue 1862]( https://github.com/ethers-io/ethers.js/issues/1862).

In the meantime I'm adding a workaround that implies pre-fetching the same data ethers will pull from the underlying network so that is already cached and ethers won't see a timeout.


publish/src/commands/deploy/index.js
```diff
@@ -416,6 +416,7 @@ const deploy = async ({
 		generateSolidity,
 		runStep,
 		useOvm,
+		useFork,
```
leomassazza, line 419: I need to send the useFork flag to apply the workarond ONLY if we are in a forked network

publish/src/commands/deploy/take-debt-snapshot-when-required.js
```diff
@@ -39,11 +41,38 @@ module.exports = async ({
...
+					method: 'eth_call',
+					params: [
+						{
+							from: contract.signer.address,
+							to: contract.address,
+							data: '0x759076e5',
+						},
+						'latest',
+					],
+					id: contract.provider._nextId,
```
leomassazza, line 62: I have a doubt here if I that nextId value needs to be updated. The tests work but may be just pure luck.

publish/src/commands/deploy/take-debt-snapshot-when-required.js
```diff
@@ -39,11 +41,38 @@ module.exports = async ({
...
-	const checkSnapshot = async () => {
-		const [cacheInfo, currentDebt] = await Promise.all([
-			DebtCache.cacheInfo(),
-			DebtCache.currentDebt(),
-		]);
+	// This function is a workaround to pre-cache the CurrentDebt data if running in a fork.
+	// The original issue is that, in a forked environment, DebtCache.currentDebt() takes more than 120 secs which is the hardcoded timeout set in ether's web component
+	// So in order to prevent the call to take that long, we previously execute the call to cache the results and then let ethers do their job
+	// NOTE: This function should be removed once https://github.com/ethers-io/ethers.js/issues/1862 is resolved
+	const cacheCurrentDebtResults = async ({ contract, useFork }) => {
```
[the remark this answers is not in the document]
justinjmoses: The inline comment is 💯  - thank you!

justinjmoses (approved):
:shipit:

PR #1455, opened 2021-08-13
Add default private key for local-ovm deploys

liamzebedee:
Previously we had to specify this on the command-line as `--private-key`.


barrasso (approved):
LGTM

liamzebedee:
Hmmm, weird, job-fork-tests is still broken. Gonna check again.

PR #1456, opened 2021-08-13
Richer Futures position and margin events

zyzek:
- Adds a `PositionModified` event, to keep track of any changes made to a position, including by (involuntary) alterations to the margin balance when it is realised.
- Adds a `MarginTransferred` event, that tracks when a user has voluntarily deposited or withdrawn margin.
- Simplifies the `OrderConfirmed` event, moving most of its contents into the `PositionModified` event.
- 
- Deletes the redundant `ParameterUpdated` event, which was not properly cleaned up when settings were moved into a separate contract.
- Adds a `uint fee` field to the `PositionLiquidated` event, so we can track how much the liquidator was paid.
- Adds the `FundingRecomputed` event, to track all modifications of the funding sequence.
- Fixes a small bug where PnL and funding fields in a position were being updated when margin was being transferred, even if the user had no position open.
- Renames `modifyMargin` and related functions to `transferMargin`, in order to highlight that it was a deposit or a withdrawal, and not a modification due to profit or funding. The latter can now be inspected by examining `PositionModified` events.

### DIFF
--- a/contracts/FuturesMarket.sol
+++ b/contracts/FuturesMarket.sol
@@ -791,7 +791,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         return (uMargin, Status.Ok);
     }
 
-    function _modifyMargin(
+    function _transferMargin(
         int marginDelta,
         uint price,
         uint fundingIndex,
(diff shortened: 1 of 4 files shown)


jacko125 (approved):

liamzebedee (approved):
looks good to me!

clementbalestrat (approved):

PR #1458, opened 2021-08-13
Ensure the fork tests are correctly running the upcoming release

justinjmoses:
A number of problems are fixed in this PR:

1. The `releases.json` had two entries with `released: false` and *Mirfak* should have been marked as `ovm: false` (the latter prevented the CI `fork-tests` from properly deploying the changed contracts
2. SIP-150 had been merged into `develop` and was causing `Cached synth sum exceeds total debt` errors (which is why 150 was rejected in the first place). It had to be reverted along with the changed test.
3. There was a bug in `Deployer` when it was migrated to `ethers` that `deployed` was showing as `truthy` and hence the deploy script was showing that every contract was redeployed when they weren't.


publish/releases.json
```diff
@@ -278,7 +278,7 @@
 	},
 	{
 		"name": "Dubhe (Optimism)",
-		"released": false,
```
justinjmoses, line 281: This was indeed released. I think the `prepublishOnly` script should amend the last entry here to `true`. Thoughts @ajsantander ?

publish/releases.json
```diff
@@ -292,17 +292,16 @@
 		"released": false,
 		"version": {
 			"major": 2,
-			"minor": 47
```
justinjmoses, line 295: Mirfak should increment this number

publish/releases.json
```diff
@@ -292,17 +292,16 @@
...
+			"minor": 48
 		},
-		"ovm": true,
```
justinjmoses, line 297: This was incorrect

publish/releases.json
```diff
@@ -292,17 +292,16 @@
...
 		"sources": [
-			"Synthetix",
```
justinjmoses, line 299: sort ascending for easy reading

publish/releases.json
```diff
@@ -292,17 +292,16 @@
...
 			"DebtCache",
-			"SynthsUSD",
-			"SynthsETH"
+			"FeePool",
+			"Issuer",
+			"Synthetix",
+			"SynthsETH",
+			"SynthsUSD"
 		],
-		"sips": [142]
```
justinjmoses, line 306: adding 145 (already merged into DebtCache a while back and thus will be released with 142, and 170 for inflation diversion to L2)

publish/src/Deployer.js
```diff
@@ -409,7 +409,7 @@ class Deployer {
 		await this._updateResults({
 			name,
 			source: deployedContract.source,
-			deployed: deployedContract.deployed,
```
justinjmoses, line 412: `deployed` is a function on an ethers contract and hence it was coercing to a truthy value.

publish/src/commands/deploy/system-and-parameter-check.js
```diff
@@ -215,20 +215,20 @@ module.exports = async ({
 			: yellow('⚠ NO'),
 	});
 
+	console.log(
```
justinjmoses, line 218: Moving this above the confirm so even when run with `--yes` we can see which contracts the deploy script is set to update

leomassazza (approved):
LGTM

PR #1459, opened 2021-08-16
deprecate ethercollateral loans

jacko125:
Minor audit feedback


justinjmoses (approved):
Just a note, even if we remove `EmptyEtherCollateral` it will still be referenced in `sources` for the contracts it was deployed for on L2...  Did we want to do this? I'm OK with it, just being explicit about this decision

contracts/BaseSynthetix.sol
```diff
@@ -87,8 +87,6 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         return issuer().totalIssuedSynths(currencyKey, false);
     }
 
-    // TODO: refactor the name of this function. It also incorporates the exclusion of
```
liamzebedee, line 90: The auditors told you to remove this comment? 😄 @jacko125

liamzebedee (approved):
This looks good.

liamzebedee:
> Just a note, even if we remove `EmptyEtherCollateral` it will still be referenced in `sources` for the contracts it was deployed for on L2... Did we want to do this? I'm OK with it, just being explicit about this decision

What happens usually with these?

contracts/BaseSynthetix.sol
jacko125: Yep the name of the function has been updated

PR #1461, opened 2021-08-16
Position ID, PositionOpened, PositionClosed added

clementbalestrat:


contracts/FuturesMarket.sol
```diff
@@ -1229,14 +1232,21 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+
         uint oldMargin = position.margin;
+
         int oldSize = position.size;
         position.margin = margin;
 
+        if (oldSize == 0) {
+            position.id = _nextPositionId;
+            _nextPositionId += 1;
+            emitPositionOpened(position.id, account);
```
clementbalestrat, line 1244: @zyzek I'm wondering if it's ok to emit an event this early in the code?

contracts/FuturesMarket.sol
```diff
@@ -135,6 +135,9 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
     // This increments for each order; zero reflects an order that does not exist.
     uint internal _nextOrderId = 1;
 
+    // This increments for each position; zero reflects an order that does not exist.
```
zyzek, line 138: `zero reflects an order` -> `zero reflects a position`

contracts/FuturesMarket.sol
```diff
@@ -1018,8 +1021,8 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         // Update the debt correction.
         int positionSize = position.size;
         _applyDebtCorrection(
```
zyzek, line 1023: In all these `_applyDebtCorrection` calls, it should be fine to construct these position objects the id argument equal to zero, as the ids aren't used. Would save a little gas in state queries.

contracts/FuturesMarket.sol
zyzek: It should be fine, but let's put it below in the `else` clause of the conditional where you're doing `emitPositionClosed`. That way we guarantee both that `newSize != 0` and `oldSize == 0`, and the opening and closing events are right next to one another.

While we're here, I noticed there's nothing stopping a user closing an already-closed position. I think we should probably revert inside `_submitOrder` if `leverage == position.size == 0`. Right now if a zero leverage order is submitted and the position size is already zero, then on confirmation it'll emit both a position opened and a position closed event. Probably undesirable.

contracts/FuturesMarket.sol
```diff
@@ -1256,6 +1266,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
             delete position.lastPrice;
             delete position.fundingIndex;
             emitPositionModified(account, margin, 0, 0, 0);
```
zyzek, line 1268: Question: shouldn't we also put the position id inside `PositionModified`? We could also detect if a position is it's opening or closing by looking at how the position size changes.

zyzek (changes requested):

contracts/FuturesMarket.sol
clementbalestrat: Yeah I just realised that. Adding it now

contracts/FuturesMarket.sol
```diff
@@ -1255,12 +1265,18 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
             delete position.size;
             delete position.lastPrice;
             delete position.fundingIndex;
-            emitPositionModified(account, margin, 0, 0, 0);
+            emitPositionModified(position.id, account, margin, 0, 0, 0);
```
clementbalestrat, line 1268: @zyzek do we even need to emitPositionModified when we know it's to close a position?

contracts/FuturesMarket.sol
```diff
@@ -1301,8 +1318,9 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         // Issue the reward to the liquidator.
         _manager().issueSUSD(liquidator, liquidationFee);
 
-        emitPositionModified(account, 0, 0, 0, 0);
-        emitPositionLiquidated(account, liquidator, positionSize, lPrice, liquidationFee);
+        emitPositionModified(positionId, account, 0, 0, 0, 0);
```
clementbalestrat, line 1321: Same question here

contracts/FuturesMarket.sol
```diff
@@ -1124,6 +1128,10 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         // We know that the price is not invalid now that we're in this function
         _revertIfError(_canLiquidate(position, _liquidationFee(), fundingIndex, price), Status.CanLiquidate);
 
+        if (leverage == position.size && position.size == 0) {
```
zyzek, line 1131: This should use the `revertIfError` structure so that we're not compiling these strings into the runtime bytecode and the reverts all follow the same pattern.

zyzek (changes requested):

zyzek (approved):

liamzebedee (approved):
Looks really good Clem, approved

jacko125 (approved):
Clem, lgtm the new PositionId for each FuturesMarket

PR #1464, opened 2021-08-17
Adding the next few releases

justinjmoses:
In order to prevent confusion of adding and selecting releases, this PR preps a handful of upcoming releases. It also adds entries for each release in both L1 and OVM. 

The trick is ensuring that the release is marked as `true` once complete. So when a release is kicked off, either it starts `node publish release .... --layer ovm` or `--layer base` or `--layer both`.

## Tasks
1. Added `finalize-release` command to be run by releases pipeline (which both updates the version files and the releases.json) (the updating of the version files [was previously run by the releases pipeline](https://github.com/Synthetixio/releases/blob/master/.circleci/config.yml#L163) so this will be managed here going forward)
1. Updated the `releases` binary command to a) use `--unreleased` and b) use `--with-sources` and configured `job-fork-tests` to use this - so it only compiles and deploys when there are new sources to deploy
3. Upgraded `commander` to latest in order to get required choiced options
4. Added an additional parameter to the `releases` command to be sent to the releases CI pipeline.

## TODO Once merged

- [ ] Add step in releases CI pipeline to mark the deploy done (https://github.com/Synthetixio/releases/blob/master/.circleci/config.yml#L163)

### DIFF
--- a/publish/releases.json
+++ b/publish/releases.json
@@ -306,6 +306,7 @@
 	},
 	{
 		"name": "Mirfak (Optimism)",
+		"ovm": true,
 		"released": false,
 		"version": {
 			"major": 2,
--- a/publish/releases.json
+++ b/publish/releases.json
@@ -389,7 +389,7 @@
 		"sips": []
 	},
 	{
-		"name": "Avoir (Optimism)",
+		"name": "Avior (Optimism)",
 		"ovm": true,
 		"released": false,
 		"version": {
--- a/publish/releases.json
+++ b/publish/releases.json
@@ -303,5 +303,141 @@
 			"SynthsUSD"
 		],
 		"sips": [142, 145, 170]
+	},
+	{
+		"name": "Mirfak (Optimism)",
+		"released": false,
+		"version": {
+			"major": 2,
+			"minor": 48
+		},
+		"sources": [],
+		"sips": []
+	},
+	{
+		"name": "Wezen",
+		"released": false,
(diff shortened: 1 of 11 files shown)


ajsantander (approved):

publish/src/commands/finalize-release.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
+
+const path = require('path');
+const fs = require('fs');
+const semver = require('semver');
+
+const prettier = require('prettier');
+
+// const { gray, yellow } = require('chalk');
```
ajsantander, line 9: Remove comment?

jacko125 (approved):
lgtm

publish/releases.json
```diff
@@ -303,5 +303,142 @@
...
+		"released": false,
+		"version": {
+			"major": 2,
+			"minor": 52
+		},
+		"sources": [],
+		"sips": []
+	},
+	{
+		"name": "Avoir (Optimism)",
```
liamzebedee, line 392: This isn't the same name @justinjmoses

liamzebedee (dismissed):
Looking good

justinjmoses:
@liamzebedee @ajsantander PR feedback addressed

mjlescano (approved):
💯

barrasso (approved):
LGTM 🚀

i-stam (approved):

PR #1468, opened 2021-08-19
Remove price update constraint from order flow

liamzebedee:
We decided on Tuesday in Discord that in order to proceed with resolving https://github.com/Synthetixio/issues/issues/260, we would remove the price update constraint from order flow and use the spot price for order confirmation in futures.

To unblock the monsieur le frontend, this PR only removes logic, and doesn't refactor. I started on refactoring the submitOrder and confirmOrder into one func - there's a lot of work to be done in refactoring the code to work with instant confirms - eg. `canConfirmOrder` and `orderStatus` currently rely on the submitted order in storage, whereas now we should be passing the order as an argument. Same with the tests, moving price updates up before the submitOrder.

### DIFF
--- a/contracts/FuturesMarket.sol
+++ b/contracts/FuturesMarket.sol
@@ -519,12 +519,6 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
 
         Order memory order = orders[account];
 
-        // Has the price updated?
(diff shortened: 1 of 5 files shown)


liamzebedee:
Closed in favour of https://github.com/Synthetixio/synthetix/pull/1477

PR #1471, opened 2021-08-19
Proposal for new release logic

justinjmoses:
The current release logic is very brittle as can be seen by:

1. Combining releases and SIPs makes it hard to mix and match SIPs into releases when there are a number of in-flight SIPs - this makes it harder to manually move SIPs around releases 
2. The `fork-tests` need to know which SIPs are upcoming and need to be tested, along with which layer requires testing (ovm doesn't yet support fork tests but this will pave the way for that once it is ready)

This proposal introduces a few new features:

1. `releases.json` includes SIPs as well. These SIPs have a `layer` (`base|ovm|both`) and `released`. When a release is complete, `layer === released`. 
2. `sources` can either be an array or if `layer === 'both'` then an object denoting the targets to deploy in each environment (look at SIP-116 for instance). 

By doing this, we a) accept that some SIPs do require releasing to both layers and b) make it easier for developers to work on their changes by simply ensuring their SIPs are added to the file along with the necessary sources, they do NOT have to worry about which release a SIP belongs to - that will happen prior to a release going out.

## Todo

- [x] `bin.js releases` needs to be able to parse this new `releases.json` structure and determine which sources exist for an upcoming release (if any) (this will fix `fork-tests`)
- [x] The `finalize-release` command has to be updated to handle this new logic.
- [x] The `prepare-deploy` command has to be updated to handle this new logic.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -188,9 +188,14 @@ program
 	.option('--unreleased', 'Only retrieve the unreleased ones.')
 	.option('--with-sources', 'Only retrieve ones with files.')
 	.action(async ({ unreleased, withSources }) => {
-		const result = releases
+		const getSip = sipNumber => releases.sips.find(({ sip }) => sip === sipNumber);
+
+		const result = releases.releases
 			.filter(release => release.released === !unreleased)
-			.filter(({ sources }) => (withSources ? sources.length > 0 : true));
+			.filter(release => {
+				if (!withSources) return true;
+				return release.sips.some(s => !!getSip(s).sources);
+			});
 
 		if (result.length > 0) {
 			console.log(JSON.stringify(result, null, 2));
--- a/bin.js
+++ b/bin.js
@@ -202,6 +202,32 @@ program
 		}
 	});
 
+program
+	.command('sips')
+	.description('Get the list of released or unreleased SIPs.')
+	.option('--unreleased', 'Only retrieve the SIPs that are not released on the given layer.')
+	.option('--with-sources', 'Only retrieve ones with source files.')
+	.addOption(
+		new commander.Option('-l, --layer <value>', `The layer(s) corresponding to the SIPs`)
+			.choices(['base', 'ovm', 'both'])
+			.default('both')
+	)
+	.action(async ({ unreleased, withSources, layer }) => {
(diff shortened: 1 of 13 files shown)


bin.js
```diff
@@ -188,9 +188,14 @@ program
 	.option('--unreleased', 'Only retrieve the unreleased ones.')
 	.option('--with-sources', 'Only retrieve ones with files.')
 	.action(async ({ unreleased, withSources }) => {
-		const result = releases
+		const getSip = sipNumber => releases.sips.find(({ sip }) => sip === sipNumber);
```
mjlescano, line 191: Opted for leaving this helper on the same scope to not clutter all the commands. As soon as we need it everywhere else, we can move it.

leomassazza (approved):
LGTM

publish/src/commands/finalize-release.js
```diff
@@ -15,22 +15,25 @@ const {
 } = require('../../../.');
 
 const finalizeRelease = async ({ layer, release, versionTag, yes }) => {
+	const isBase = layer === 'base' || layer === 'both';
+	const isOvm = layer === 'ovm' || layer === 'both';
```
justinjmoses, line 19: yeah this is better logic

justinjmoses:
This is looking great, just one more thing - we want `fork-tests` to run all upcoming SIPs even if not assigned to a release (I said the wrong thing above). This way when someone works on a feature, CI will test their changes without them having to assign anything to a release. Thus, this PR should have SIPs 142, 145 and 170 all showing as unreleased. I think another command for the `bin.js` would work - something that returns all unreleased sips with the argument of which base (`base|ovm|both`) and `both` being the default.

package.json
```diff
@@ -16,7 +16,7 @@
 		"pack": "webpack --mode production",
 		"prepublishOnly": "npm run describe && npm run pack",
 		"fork": "node --max-old-space-size=4096 ./node_modules/.bin/hardhat node",
-		"fork:mainnet": "node --max-old-space-size=4096 ./node_modules/.bin/hardhat node --target-network mainnet",
+		"fork:mainnet": "node --max-old-space-size=4096 ./node_modules/.bin/hardhat node --target-network mainnet --hardfork london",
```
justinjmoses, line 19: :100:

PR #1472, opened 2021-08-19
Fix nominate and owner scripts

leomassazza:
Nominate and Owner scripts fails if any contract in config.json is not deployed (or present in deployment.json).
This PR adds a check on that condition and show a warning while trying to Nominate or Accept ownership.

### DIFF
--- a/publish/src/commands/connect-bridge.js
+++ b/publish/src/commands/connect-bridge.js
@@ -238,7 +238,7 @@ const setupInstance = async ({
 		useFork,
 		useOvm,
 	});
-	console.log(gray('  > provider:', provider.url));
+	console.log(gray('  > provider:', provider.connection.url));
 	console.log(gray('  > account:', wallet.address));
 
 	const AddressResolver = getContract({
--- a/publish/src/commands/nominate.js
+++ b/publish/src/commands/nominate.js
@@ -110,7 +110,14 @@ const nominate = async ({
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


publish/src/commands/nominate.js
```diff
@@ -110,7 +110,14 @@ const nominate = async ({
 		}
 	}
 
+	const warnings = [];
```
leomassazza, line 113: Adding this to be able to show a summary at the end of the process with all missing contracts found.

publish/src/commands/owner.js
```diff
@@ -242,7 +242,14 @@ const owner = async ({
 	}
 
 	console.log(gray('Looking for contracts whose ownership we should accept'));
+	const warnings = [];
```
leomassazza, line 245: dito

publish/src/commands/connect-bridge.js
```diff
@@ -238,7 +238,7 @@ const setupInstance = async ({
 		useFork,
 		useOvm,
 	});
-	console.log(gray('  > provider:', provider.url));
+	console.log(gray('  > provider:', provider.connection.url));
```
leomassazza, line 241: small fix. Provider was shown as undefined because of that missing intermediate object

ajsantander: This must be a change in the ethers interface

ajsantander (approved):

PR #1473, opened 2021-08-19
Updates batch relay tests to nominate ALL contracts

ajsantander:
Previous test was only testing batch relays by changing ownership to 2-3 contracts. This does all of them in a single call.


leomassazza:
LGTM

i-stam (approved):

PR #1475, opened 2021-08-20
OwnerRelayOnOptimism unit test simplification

i-stam:
This PR removes `triggerSendMessage()` from  `OwnerRelayOnOptimism.js` as it complicates the code without being needed.
Since smock gives us the possibility to use the contract itself as the `msg.sender` we can simplify the tests without having to simulate all the `sendMessage` complexity of the sequencer.
Similarly to `triggerSendMessage()`,  `finalizeRelay() `is being used for both `finalizeRelay` and `finalizeRelayBatch` functionality

### DIFF
--- a/test/contracts/OwnerRelayOnOptimism.js
+++ b/test/contracts/OwnerRelayOnOptimism.js
@@ -135,7 +135,7 @@ contract('OwnerRelayOnOptimism', () => {
 		let relayedMessageData;
 		let nominateNewOwnerCalldata;
 
-		async function finalizeRelay(isBatch, targets, calldata) {
+		async function triggerFinalizeRelay(isBatch, targets, calldata) {
 			let relayReceipt, sendMessageError;
 			const relayFnc = isBatch ? 'finalizeRelayBatch' : 'finalizeRelay';
 
@@ -181,11 +181,11 @@ contract('OwnerRelayOnOptimism', () => {
 				});
 
 				before('attempt to finalize the relay', async () => {
-					({ sendMessageError } = await finalizeRelay(
-						false,
+					({ sendMessageError } = await triggerFinalizeRelay(
+						false, // 1st param: isBatch == false then call finalizeRelay()
 						mockedContractAddressOnL2,
 						mockedRelayData
-					)); // isBatch == false i.e. finalizeRelay
+					));
 				});
 
 				it('reverts with the expected error', async () => {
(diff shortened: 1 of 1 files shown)


test/contracts/OwnerRelayOnOptimism.js
```diff
@@ -18,7 +18,7 @@ contract('OwnerRelayOnOptimism', () => {
 
 	// Mocked contracts
 	let MockedMessenger, MockedAddressResolver, MockedOwned1OnL2, MockedOwned2OnL2;
-
+	let MockedMessengerSigner;
```
ajsantander, line 21: ultra-nit: respect the spacing that was there before?

test/contracts/OwnerRelayOnOptimism.js
```diff
@@ -165,37 +171,21 @@ contract('OwnerRelayOnOptimism', () => {
...
 					);
 				});
 
 				before('attempt to finalize the relay', async () => {
-					await triggerSendMessage();
+					({ sendMessageError } = await finalizeRelay(
+						false,
+						mockedContractAddressOnL2,
+						mockedRelayData
+					)); // isBatch == false i.e. finalizeRelay
```
ajsantander, line 188: nit: a bit confused by this comment

test/contracts/OwnerRelayOnOptimism.js
```diff
@@ -165,37 +171,21 @@ contract('OwnerRelayOnOptimism', () => {
...
+
 				before('mock the Messenger to report some random account as the L1 initiator', async () => {
 					MockedMessenger.smocked.xDomainMessageSender.will.return.with(
 						ethers.Wallet.createRandom().address
 					);
 				});
 
 				before('attempt to finalize the relay', async () => {
-					await triggerSendMessage();
+					({ sendMessageError } = await finalizeRelay(
```
ajsantander, line 184: nit: maybe rename this to `simulateFinalizeRelay(...)` so that on code search it is clearly something different from the contract functions.

ajsantander (approved):
Great simplification. Just a couple of minor comments.

test/contracts/OwnerRelayOnOptimism.js
i-stam: updated

test/contracts/OwnerRelayOnOptimism.js
i-stam: renamed to `triggerFinalizeRelay()`.

PR #1476, opened 2021-08-20
SIP-174 Synth Redeemer

justinjmoses:


jacko125:
looks good to me the implementation. The total supply of deprecated synths are tracked with `sUSD` issued.

contracts/Issuer.sol
```diff
@@ -446,9 +454,21 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+        uint synthSupply = IERC20(synthToRemove).totalSupply();
+
+        if (synthSupply > 0) {
+            (uint amountOfsUSD, uint rateToRedeem, ) =
+                exchangeRates().effectiveValueAndRates(currencyKey, synthSupply, "sUSD");
+            require(rateToRedeem > 0, "Cannot remove synth to redeem without rate");
+            ISynthRedeemer _synthRedeemer = synthRedeemer();
+            synths[sUSD].issue(address(_synthRedeemer), amountOfsUSD);
+            // ensure the debt cache is aware of the new sUSD issued
+            debtCache().updateCachedSynthDebtWithRate(sUSD, SafeDecimalMath.unit());
```
jacko125, line 468: Looks good, this should update the sUSD amounts in the debt cache.

contracts/Issuer.sol
```diff
@@ -446,9 +458,21 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
         require(synthToRemove != address(0), "Synth does not exist");
-        require(IERC20(synthToRemove).totalSupply() == 0, "Synth supply exists");
         require(currencyKey != sUSD, "Cannot remove synth");
 
+        uint synthSupply = IERC20(synthToRemove).totalSupply();
+
+        if (synthSupply > 0) {
+            (uint amountOfsUSD, uint rateToRedeem, ) =
+                exchangeRates().effectiveValueAndRates(currencyKey, synthSupply, "sUSD");
+            require(rateToRedeem > 0, "Cannot remove synth to redeem without rate");
```
KillerByte, line 468: in the unlikely case we have to remove a synth during a time where its rate is defunct/stale, is there any escape hatch for specifying the redeem rate?

contracts/SynthRedeemer.sol
```diff
@@ -0,0 +1,86 @@
...
+
+    constructor(address _resolver) public MixinResolver(_resolver) {}
+
+    function resolverAddressesRequired() public view returns (bytes32[] memory addresses) {
+        addresses = new bytes32[](2);
+        addresses[0] = CONTRACT_ISSUER;
+        addresses[1] = CONTRACT_SYNTHSUSD;
+    }
+
+    function _issuer() internal view returns (IIssuer) {
```
KillerByte, line 30: np resolver view helper functions in other contracts do not use underscores to prefix

contracts/SynthRedeemer.sol
```diff
@@ -0,0 +1,86 @@
...
+        _issuer().burnForRedemption(address(synthProxy), msg.sender, amountOfSynth);
+        uint amountInsUSD = amountOfSynth.multiplyDecimal(rateToRedeem);
+        _sUSD().transfer(msg.sender, amountInsUSD);
+        emit SynthRedeemed(address(synthProxy), msg.sender, amountOfSynth, amountInsUSD);
+    }
+
+    function deprecate(
+        IERC20 synthProxy,
+        uint rateToRedeem,
+        uint totalSynthSupply
```
KillerByte, line 64: what does this parameter do? can't the `totalSynthSupply` be read from the synthProxy?

contracts/SynthRedeemer.sol
```diff
@@ -0,0 +1,86 @@
...
+        require(msg.sender == address(_issuer()), "Restricted to Issuer contract");
+    }
+
+    modifier onlyIssuer() {
+        requireOnlyIssuer();
+        _;
+    }
+
+    event SynthRedeemed(address synth, address account, uint amountOfSynth, uint amountInsUSD);
+    event SynthDeprecated(address synth, uint rateToRedeem, uint totalSynthSupply);
```
KillerByte, line 85: np might be helpful to include the supply value in sUSD as well, since its included in `SynthRedeemed` event

test/contracts/Issuer.js
```diff
@@ -700,42 +704,67 @@ contract('Issuer (via Synthetix)', async accounts => {
...
+										// 100 units of sBTC at a rate of 2:1
+										assert.bnEqual(amountOfsUSDIssued, toUnit('200'));
+									});
+									it('it invokes deprecate on the redeemer via the proxy', async () => {
+										const redeemRate = await synthRedeemer.redemptions(synthProxy.address);
+
+										assert.bnEqual(redeemRate, toUnit('2'));
+									});
+									it('and total debt remains unchanged', async () => {
+										// await debtCache.takeDebtSnapshot();
```
KillerByte, line 760: np remove comment

KillerByte (approved):
:+1:

contracts/Issuer.sol
justinjmoses: No, and I considered that as well but the owner can remove the aggregator behind the rate and add a custom rate using `ExchangeRates.updateRates` in this unlikely scenario, so I think it's better to keep it minimal here.

contracts/SynthRedeemer.sol
justinjmoses: Yeah good iea.

jacko125 (approved):
lgtm!

contracts/SynthRedeemer.sol
KillerByte: on line 71, the `totalIssuedSynths` value and `redeemed[synthProxy]` are both read from storage... one could simply divide by `rateToRedeem` to save even more gas by reducing calldata. Or change `totalSupply` to return both the sUSD and synth value as a tuple.

even so, the gas savings is pretty minimal (5000 units?) on a call which is executed only once for each synth that is deprecated... I comment because I doubt the code complexity of adding this extra argument is worth any of this.

justinjmoses: Yeah that's a fair point. Let me look into a small fix for this.

justinjmoses: https://github.com/Synthetixio/synthetix/pull/1476/commits/8433f12a185111611475baf0f3bebd61a4dac5c7

KillerByte:
👍

PR #1477, opened 2021-08-23
Futures spot trades

zyzek:
This does two major things:

1. Removes next-price orders entirely; positions are now modified immediately at spot.
2. Alters the position modification interface so that you provide a size delta instead of a desired leverage.

In consequence, the interface has been radically simplified. The main function of interest now is `modifyPosition`, which should receive the closest attention. It is essentially the result of merging the old `submitOrder` and `confirmOrder` functions.

Concludes https://github.com/Synthetixio/issues/issues/260 and probably supersedes https://github.com/Synthetixio/synthetix/pull/1468.


clementbalestrat (approved):

contracts/FuturesMarket.sol
```diff
@@ -1078,166 +923,54 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
-    /*
-     * Submit an order to adjust the position leverage to a target level.
-     * Reverts if the resulting position is too large, outside the max leverage, or if an existing position is liquidating.
-     */
-    function submitOrder(int leverage) external optionalProxy {
-        submitOrderWithPriceBounds(leverage, 0, uint(-1));
-    }
+        // Check that the maximum leverage is not exceeded (ignoring the fee)
+        int desiredLeverage = newSize.multiplyDecimalRound(int(price)).divideDecimalRound(int(margin.add(fee)));
+        _revertIfError(_maxLeverage(baseAsset) < _abs(desiredLeverage), Status.MaxLeverageExceeded);
```
jacko125, line 959: @zyzek Thoughts on capping the `newSize` here to cap at the max leverage or giving some buffer to reduce the amount of rejected tx's especially if close to the max leverage a position.

contracts/FuturesMarket.sol
```diff
@@ -1255,16 +988,38 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+    }
+
+    /*
+     * Submit an order to close a position.
+     */
+    function closePosition() external optionalProxy {
+        uint price = _assetPriceRequireNotInvalid();
+        uint fundingIndex = _recomputeFunding(price);
+        int size = positions[messageSender].size;
+        _revertIfError(size == 0, Status.NoPositionOpen);
```
jacko125, line 1021: Could check whether the position is open earlier?

zyzek: Good point, moving it up.

contracts/FuturesMarket.sol
zyzek: Good point, let's allow an extra 1% leverage above the maximum allowable.

contracts/FuturesMarket.sol
```diff
@@ -1078,166 +909,59 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
-        if (sameSide && _abs(currentLeverage_) <= _abs(desiredLeverage)) {
-            // minMargin + fee <= margin is equivalent to minMargin <= margin - fee
-            // except that we get a nicer error message if fee > margin, rather than arithmetic overflow.
-            _revertIfError(margin < _minInitialMargin().add(fee), Status.InsufficientMargin);
-        }
-    }
-
-    function _submitOrder(
-        int leverage,
+    function _modifyPosition(
```
liamzebedee, line 912: Hey, did you intend to remove the slippage protections altogether? Even though we fulfill at spot price, the user can still lose money if the price changes between tx submission and the tx being mined - eg. due to congestion, their tx is mined after a price update comes through. I don't think we want this, even if its low probability on Optimism.

contracts/FuturesMarket.sol
```diff
@@ -1338,80 +1075,50 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
-    event OrderSubmitted(
+    event PositionModified(
         uint indexed id,
         address indexed account,
-        int leverage,
-        uint fee,
-        uint indexed roundId,
-        uint minPrice,
-        uint maxPrice
+        uint margin,
```
liamzebedee, line 1081: @zyzek @clementbalestrat I noticed we removed `leverage` from the event - do we want to have to calculate this in the dapp and subgraphs for every position modification event? My intuition is that it's quicker to access this from the event. I'm thinking of a view like "position history" in the dapp, where we display the leverage for each modification.

contracts/FuturesMarket.sol
```diff
@@ -1078,166 +909,59 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
-            // minMargin + fee <= margin is equivalent to minMargin <= margin - fee
-            // except that we get a nicer error message if fee > margin, rather than arithmetic overflow.
-            _revertIfError(margin < _minInitialMargin().add(fee), Status.InsufficientMargin);
-        }
-    }
-
-    function _submitOrder(
-        int leverage,
+    function _modifyPosition(
+        int sizeDelta,
```
liamzebedee, line 913: Just thought of this, perhaps specifying the delta is more complex than the absolute size - now, in order to modify the position, we must check for pending modifications to the position in order to compute the correct delta for a given size. It introduces some additional state the dapp needs to grab. Did we come up with any benefits with using a delta? Do you think the absolute value is less complex?

liamzebedee (changes requested):
Some minor questions - what you've done here is solid 👍

contracts/FuturesMarket.sol
zyzek: On the dapp design stuff, It's not clear to me whether it's position size or leverage we really want to display here. I'll defer to @clementbalestrat on this one. 

On the technicals of actually including an explicit `leverage` field, I don't think that's necessary, as it's retrievable by doing `size * price / margin`.

contracts/FuturesMarket.sol
zyzek: I think the pending modifications question is probably solved on the front-end by disallowing any order to be submitted until any pending position modification transactions have confirmed.

I think extra complexity just depends on what the front end is actually doing. If we want to change the entire position to a specific size, then yeah, the current pattern would require calculating `desiredSize - currentSize`. On the other hand, if we change the interface to take an absolute value and someone wants to increase by `x` units, then they'll need to pass in `currentSize + x`. The internals of `_modifyPosition` do not really change mcuh either way.
The same question kind of applies to the `transferMargin` function, as presumably we want the interaction patterns to be the same?

I'm not particularly fussed either way, but perhaps @clementbalestrat has a front-end use case that will decide which one is best.

contracts/FuturesMarket.sol
zyzek: Hm, yeah I had assumed that slippage protection would be unnecessary with no confirmation delay, but I suppose your scenario is possible. I'll add a `_modifyPositionWithPriceBounds` function back in.

contracts/FuturesMarket.sol
jacko125: I feel that the delta here makes more sense where you would input the difference required, ie selling some units or increasing the units of the asset (buying) in the trade.

liamzebedee (approved):

jacko125 (approved):

PR #1484, opened 2021-08-29
Futures accessible margin

zyzek:
Does the following:

* Adds a function `accessibleMargin` that reports the amount of margin a user can actually withdraw given their current position.
* Updates `withdrawAllMargin` to withdraw this quantity.
* Fixes a bug in `transferMargin`, where the minimum initial margin check and some other pieces were not properly respecting short positions.

### DIFF
--- a/contracts/FuturesMarket.sol
+++ b/contracts/FuturesMarket.sol
@@ -543,8 +543,8 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         // Ugly solution to rounding safety: leave up to an extra tenth of a cent in the account/leverage
         // This should guarantee that the value returned here can always been withdrawn, but there may be
         // a little extra actually-accessible value left over, depending on the position size and margin.
-        uint mill = uint(_UNIT / 1000);
-        int maxLeverage = int(_maxLeverage(baseAsset).sub(mill));
+        uint milli = uint(_UNIT / 1000);
+        int maxLeverage = int(_maxLeverage(baseAsset).sub(milli));
         uint inaccessible = _abs(_notionalValue(position, price).divideDecimalRound(maxLeverage));
 
         // If the user has a position open, we'll enforce a min initial margin requirement.
@@ -553,7 +553,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
             if (inaccessible < minInitialMargin) {
                 inaccessible = minInitialMargin;
             }
-            inaccessible = inaccessible.add(mill);
+            inaccessible = inaccessible.add(milli);
         }
 
         uint remaining = _remainingMargin(position, fundingIndex, price);
--- a/contracts/FuturesMarket.sol
+++ b/contracts/FuturesMarket.sol
@@ -966,6 +966,9 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
             // A negative margin delta corresponds to a withdrawal, which will be minted into
             // their sUSD balance, and debited from their margin account.
             _manager().issueSUSD(sender, absDelta);
+        } else {
+            // Zero delta is a no-op
+            return;
         }
 
         Position storage position = positions[sender];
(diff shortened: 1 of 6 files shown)


contracts/FuturesMarket.sol
```diff
@@ -533,8 +532,45 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+
+    function _accessibleMargin(
+        Position storage position,
+        uint fundingIndex,
+        uint price
+    ) internal view returns (uint) {
+        // Ugly solution to rounding safety: leave up to an extra tenth of a cent in the account/leverage
+        // This should guarantee that the value returned here can always been withdrawn, but there may be
+        // a little extra actually-accessible value left over, depending on the position size and margin.
+        uint mill = uint(_UNIT / 1000);
```
jacko125, line 546: Interesting so this is 0.001 less on the max average

contracts/FuturesMarket.sol
```diff
@@ -533,8 +532,45 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+        int maxLeverage = int(_maxLeverage(baseAsset).sub(mill));
+        uint inaccessible = _abs(_notionalValue(position, price).divideDecimalRound(maxLeverage));
+
+        // If the user has a position open, we'll enforce a min initial margin requirement.
+        if (0 < inaccessible) {
+            uint minInitialMargin = _minInitialMargin();
+            if (inaccessible < minInitialMargin) {
+                inaccessible = minInitialMargin;
+            }
+            inaccessible = inaccessible.add(mill);
```
jacko125, line 556: @zyzek  Intention here is to add a `mill` buffer to the `inaccessible` amount even after the maxLeverage used had this included already?

contracts/FuturesMarket.sol
```diff
@@ -950,7 +988,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         // Update the account's position with the realised margin.
         position.margin = margin;
         // We only need to update their funding/PnL details if they actually have a position open
-        if (positionSize > 0) {
+        if (positionSize != 0) {
```
jacko125, line 991: Because positionSize can be -neg int representing a short position? good fix here.

contracts/FuturesMarket.sol
```diff
@@ -959,13 +997,11 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
         //     * the resulting margin would not be lower than the minimum margin
         //     * the resulting leverage is lower than the maximum leverage
-        if (0 < positionSize && marginDelta <= 0) {
-            if (margin < _minInitialMargin()) {
-                status = Status.InsufficientMargin;
-            } else if (_maxLeverage(baseAsset) < _abs(_currentLeverage(position, price, margin))) {
-                status = Status.MaxLeverageExceeded;
-            }
-            _revertIfError(status);
+        if (positionSize != 0 && marginDelta < 0) {
```
jacko125, line 1000: @zyzek  Could nest the `positionSize != 0` check here with the block above and add the check that there is sufficient margin / leverage if withdrawing margin.

contracts/FuturesMarket.sol
zyzek: Yes, in case the position is very small. It might not actually be necessary, I don't know.

contracts/FuturesMarket.sol
zyzek: Correct

contracts/FuturesMarket.sol
zyzek: Yeah; the idea being that if it reports that they can only withdraw slightly less than the contract will actually allow them to, then we won't have rounding errors.
If the user has no position open, then `_notionalValue` will return 0, and so the max leverage value will be irrelevant, as the inaccessible portion of their margin will be zero.

contracts/FuturesMarket.sol
zyzek: Good catch. Fixed.

jacko125 (approved):

liamzebedee (approved):

contracts/FuturesMarket.sol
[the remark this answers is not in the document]
zyzek: Good pickup, will fix this.

contracts/FuturesMarket.sol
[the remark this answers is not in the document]
zyzek: Yeah, I was kicking myself. I even came across this while writing some tests before and somehow it didn't twig what the base cause was.

contracts/FuturesMarket.sol
```diff
@@ -533,8 +532,45 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+    ) internal view returns (uint) {
+        // Ugly solution to rounding safety: leave up to an extra tenth of a cent in the account/leverage
+        // This should guarantee that the value returned here can always been withdrawn, but there may be
+        // a little extra actually-accessible value left over, depending on the position size and margin.
+        uint mill = uint(_UNIT / 1000);
+        int maxLeverage = int(_maxLeverage(baseAsset).sub(mill));
+        uint inaccessible = _abs(_notionalValue(position, price).divideDecimalRound(maxLeverage));
+
+        // If the user has a position open, we'll enforce a min initial margin requirement.
+        if (0 < inaccessible) {
```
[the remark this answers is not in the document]
zyzek: I guess it's just `max(inaccessible, minInitialMargin) + mill`. I had it this way before, but there was a lot of casting between `int` and `uint` to get the types to work out that I feel made it less legible in the end.

PR #1485, opened 2021-08-30
emit sizeDelta with PositionModified event

liamzebedee:


contracts/FuturesMarket.sol
```diff
@@ -1077,7 +1078,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
             delete position.size;
             delete position.lastPrice;
             delete position.fundingIndex;
-            emitPositionModified(newPosition.id, sender, newPosition.margin, 0, 0, 0, fee);
+            emitPositionModified(newPosition.id, sender, newPosition.margin, 0, 0, 0, 0, fee);
```
zyzek, line 1081: This still needs the `sizeDelta`, if the position went from nonzero to zero size, then it necessarily means the delta was nonzero.

zyzek (changes requested):

contracts/FuturesMarket.sol
```diff
@@ -1177,7 +1187,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         // Issue the reward to the liquidator.
         _manager().issueSUSD(liquidator, liquidationFee);
 
-        emitPositionModified(positionId, account, 0, 0, 0, 0, 0);
+        emitPositionModified(positionId, account, 0, 0, 0, 0, 0, 0);
```
zyzek, line 1190: Should the involuntary liquidation of a position include a sizeDelta that is the whole position size or not? I think probably it should, but maybe @clementbalestrat has some ideas on this too.

zyzek (changes requested):

zyzek (approved):

jacko125 (approved):

PR #1486, opened 2021-08-30
SIP-171 Allow transfer of non locked synths

jjgonecrypto:
POC


barrasso:
SIP rejected https://sips.synthetix.io/sips/sip-171/

PR #1487, opened 2021-08-31
Deploy kovan-ovm-futures v0.4.0 (Alpha)

liamzebedee:
Deployment of core contracts for [futures alpha](https://github.com/Synthetixio/issues/issues/273) on Kovan.

Two fixes included:
 - getFuturesMarkets (#1440) returns the wrong key
 - deployment.json is ignored in the local-ovm folder, which means require("synthetix") will fail. Fixes an issue from #1356 which broke require("synthetix")


publish/deployed/local-ovm/.gitignore
```diff
@@ -1,4 +1,3 @@
 # these files are irrelevant outside of a single developer's machine,
 # so no need to persist them in source control
-deployment.json
```
zyzek, line 3: I am not clear on what this fixes, since you aren't pushing `local-ovm/deployment.json`. In general isn't it sensible that updated versions of this aren't committed? Or are we just relying on developers to remember not to commit it?

liamzebedee: Sorry - just figured out that there was a better way of doing this. In https://github.com/Synthetixio/synthetix/pull/1356 I merged in the local-ovm support so that we could develop locally quickly with the futures-keepers. Turns out that NPM will ignore downloading any files specified by .gitignore - and since we need the deployment.json in a local-ovm development context, I tried to include it. Just made a change (and tested it) which overrides the gitignore using npmignore - which will allow us to achieve the above.

zyzek (approved):

jacko125 (approved):

PR #1489, opened 2021-08-31
[SIP PENDING] Wrapper Factory

KillerByte:
stubbed currently:
* totalIssuedSynths calculation
* distributeFees

tests to come


contracts/BaseDebtCache.sol
```diff
@@ -144,6 +145,11 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
         (uint[] memory rates, bool isInvalid) = exchangeRates().ratesAndInvalidForCurrencies(currencyKeys);
         uint[] memory values = _issuedSynthValues(currencyKeys, rates);
         (uint excludedDebt, bool isAnyNonSnxDebtRateInvalid) = _totalNonSnxBackedDebt();
+
+        for (uint i = 0; i < currencyKeys.length; i++) {
+            excludedDebt = excludedDebt.add(_excludedIssuedDebt[currencyKeys[i]].multiplyDecimalRound(rates[i]));
```
jacko125, line 150: why track the `_excludedIssuedDebt` in the Debt cache and also the `wrapperFactory.totalIssuedSynth()` ?

contracts/BaseDebtCache.sol
```diff
@@ -185,9 +204,9 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
         isInvalid = anyTotalLongRateIsInvalid || anyTotalShortRateIsInvalid;
         excludedDebt = longValue.add(shortValue);
 
-        // 2. EtherWrapper.
-        // Subtract sETH and sUSD issued by EtherWrapper.
-        excludedDebt = excludedDebt.add(etherWrapper().totalIssuedSynths());
+        // 2. Wrapper.
+        // Subtract totalIssuedSynths issued by all wrappers.
+        excludedDebt = excludedDebt.add(wrapperFactory().totalIssuedSynths());
```
jacko125, line 209: Do we want to use the `_excludedIssuedDebts` here or still read from the `wrapperFactory` ?

contracts/MixinSystemSettings.sol
```diff
@@ -27,9 +27,9 @@ contract MixinSystemSettings is MixinResolver {
     bytes32 internal constant SETTING_CROSS_DOMAIN_ESCROW_GAS_LIMIT = "crossDomainEscrowGasLimit";
     bytes32 internal constant SETTING_CROSS_DOMAIN_REWARD_GAS_LIMIT = "crossDomainRewardGasLimit";
     bytes32 internal constant SETTING_CROSS_DOMAIN_WITHDRAWAL_GAS_LIMIT = "crossDomainWithdrawalGasLimit";
-    bytes32 internal constant SETTING_ETHER_WRAPPER_MAX_ETH = "etherWrapperMaxETH";
-    bytes32 internal constant SETTING_ETHER_WRAPPER_MINT_FEE_RATE = "etherWrapperMintFeeRate";
-    bytes32 internal constant SETTING_ETHER_WRAPPER_BURN_FEE_RATE = "etherWrapperBurnFeeRate";
+    bytes32 internal constant SETTING_WRAPPER_MAX_TOKEN_AMOUNT = "wrapperMaxTokens";
```
jacko125, line 30: Could we look at these values being per `wrapper` asset and not set as a constant here rather during constructor or initialize().

contracts/SystemSettings.sol
```diff
@@ -288,21 +288,50 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
         emit AggregatorWarningFlagsUpdated(_flags);
     }
 
-    function setEtherWrapperMaxETH(uint _maxETH) external onlyOwner {
-        flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_ETHER_WRAPPER_MAX_ETH, _maxETH);
-        emit EtherWrapperMaxETHUpdated(_maxETH);
+    function setWrapperMaxTokenAmount(address _wrapper, uint _maxTokenAmount) external onlyOwnerOrSelf {
```
jacko125, line 291: I'd say it would be better to move these into an internal function `_setWrapperMaxTokenAmount` here for access control than the modifier.

contracts/Wrapper.sol
```diff
@@ -0,0 +1,238 @@
...
+// https://docs.synthetix.io/contracts/source/contracts/wrapper
+contract Wrapper is Owned, Pausable, MixinResolver, MixinSystemSettings, IWrapper {
+    using SafeMath for uint;
+    using SafeDecimalMath for uint;
+
+    /* ========== ENCODED NAMES ========== */
+
+    bytes32 internal constant sUSD = "sUSD";
+    bytes32 internal constant ETH = "ETH";
+    bytes32 internal constant SNX = "SNX";
```
jacko125, line 32: SNX constant is unused

contracts/WrapperFactory.sol
```diff
@@ -0,0 +1,106 @@
...
+import "./interfaces/IFlexibleStorage.sol";
+import "./interfaces/IWrapperFactory.sol";
+
+// https://docs.synthetix.io/contracts/source/contracts/wrapperfactory
+contract WrapperFactory is Owned, MixinResolver, IWrapperFactory {
+    bytes32 internal constant CONTRACT_FLEXIBLESTORAGE = "FlexibleStorage";
+    bytes32 internal constant CONTRACT_SYNTH_SUSD = "SynthsUSD";
+    bytes32 internal constant CONTRACT_FEEPOOL = "FeePool";
+
+    bytes32 internal constant WRAPPER_FACTORY_CONTRACT_NAME = "WrapperFactory";
```
justinjmoses, line 21: can you call this `CONTRACT_NAME`? It'll better align with what we have in the migrations (see https://github.com/Synthetixio/synthetix/blob/develop/publish/src/commands/deploy/generate-solidity-output.js#L163-L166)

contracts/Wrapper.sol
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    /* ========== PUBLIC FUNCTIONS ========== */
+
+    // ========== VIEWS ==========
+
+    function capacity() public view returns (uint _capacity) {
+        // capacity = max(maxETH - balance, 0)
+        uint balance = getReserves();
+        if (balance >= maxTokenAmount()) {
```
justinjmoses, line 109: you should get decent a gas improvement by assigning `maxTokenAmount()` to a local variable instead of loading it twice.

contracts/SystemSettings.sol
```diff
@@ -288,21 +288,50 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
...
+        this.setWrapperMintFeeRate(_wrapper, _mintFeeRate);
+        this.setWrapperBurnFeeRate(_wrapper, _burnFeeRate);
     }
 
-    function setEtherWrapperBurnFeeRate(uint _rate) external onlyOwner {
-        require(_rate <= MAX_ETHER_WRAPPER_BURN_FEE_RATE, "rate > MAX_ETHER_WRAPPER_BURN_FEE_RATE");
-        flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_ETHER_WRAPPER_BURN_FEE_RATE, _rate);
-        emit EtherWrapperBurnFeeRateUpdated(_rate);
+    modifier onlyOwnerOrSelf {
+        require(msg.sender == owner || msg.sender == address(this), "Only the contract owner may perform this action");
```
justinjmoses, line 333: if you stick with using a modifier and you use it more than once in a contract, it's better for contract size if you put the logic in a private function and call that in the modifier (see `Owned._onlyOwner()`)

contracts/SystemSettings.sol
```diff
@@ -288,21 +288,50 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
...
-        flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_ETHER_WRAPPER_MINT_FEE_RATE, _rate);
-        emit EtherWrapperMintFeeRateUpdated(_rate);
+    // Convenience function for setting all wrapper settings at once
+    function setWrapperSettings(
+        address _wrapper,
+        uint _maxTokenAmount,
+        uint _mintFeeRate,
+        uint _burnFeeRate
+    ) external onlyOwner {
+        this.setWrapperMaxTokenAmount(_wrapper, _maxTokenAmount);
```
justinjmoses, line 327: [np] while use `this` instead of making the other functions `public`?

contracts/WrapperFactory.sol
```diff
@@ -0,0 +1,106 @@
...
+    // ========== VIEWS ==========
+    // Returns the version of a wrapper created by this wrapper factory
+    // Used by MultiCollateralSynth to know if it should trust the wrapper contract
+    function isWrapper(address possibleWrapper) external view returns (bool) {
+        return flexibleStorage().getUIntValue(WRAPPER_FACTORY_CONTRACT_NAME, bytes32(uint(address(possibleWrapper)))) > 0;
+    }
+
+    // Returns sum of totalIssuedSynths for all wrappers deployed by this contract
+    function totalIssuedSynths() external view returns (uint) {
+        return 0; // stub
```
justinjmoses, line 56: missed this?

contracts/MultiCollateralSynth.sol
```diff
@@ -33,15 +33,15 @@ contract MultiCollateralSynth is Synth {
         return ICollateralManager(requireAndGetAddress(CONTRACT_COLLATERALMANAGER));
     }
 
-    function etherWrapper() internal view returns (IEtherWrapper) {
```
justinjmoses, line 36: If you remove this, how can the existing etherwrapper allow burning of sETH into ETH?

contracts/BaseDebtCache.sol
```diff
@@ -185,9 +204,9 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
         isInvalid = anyTotalLongRateIsInvalid || anyTotalShortRateIsInvalid;
         excludedDebt = longValue.add(shortValue);
 
-        // 2. EtherWrapper.
-        // Subtract sETH and sUSD issued by EtherWrapper.
-        excludedDebt = excludedDebt.add(etherWrapper().totalIssuedSynths());
```
justinjmoses, line 190: how do you plan on dealing with the fact that the wrapper has 170,000 ETH that need to be accounted for in the system? Will the `WrapperFactory.totalIssuedSynths()` take this into account?

justinjmoses (changes requested):

contracts/BaseDebtCache.sol
dbeal-eth: there is no reason, the `wrapperFactory.totalIssuedSynth()` line should be removed.

contracts/BaseDebtCache.sol
dbeal-eth: more discussion is needed on this on how it should work with l2 deployment, but this will probably be re-added

contracts/BaseDebtCache.sol
dbeal-eth: this line should have been removed

contracts/MixinSystemSettings.sol
dbeal-eth: could you explain exactly what you would like to see here? its not clear to me. I thought I was following the same exxample already set forth by `SETTING_EXCHANGE_FEE_RATE` with the kekkack concat

contracts/MultiCollateralSynth.sol
dbeal-eth: will probably be re-added

contracts/SystemSettings.sol
dbeal-eth: yea the problem is `SystemSettings` is right on the edge of the size limit for optimism contracts... if more internal functions are added could be a problem. Thats the main reason why it is the way it is right now.

contracts/SystemSettings.sol
dbeal-eth: good idea

contracts/SystemSettings.sol
dbeal-eth: is this because the modifier simply inlines code on the beginning of a function? interesting.

contracts/WrapperFactory.sol
dbeal-eth: yes

contracts/WrapperFactory.sol
```diff
@@ -0,0 +1,106 @@
...
+        bytes32 synthContractName
+    ) external onlyOwner returns (address) {
+        // Create the wrapper instance
+        Wrapper wrapper = new Wrapper(owner, address(resolver), token, currencyKey, synthContractName);
+
+        // Rebuild caches immediately since it will almost certainly need to be done
+        wrapper.rebuildCache();
+
+        // Register it so that MultiCollateralSynth knows to trust it
+        flexibleStorage().setUIntValue(WRAPPER_FACTORY_CONTRACT_NAME, bytes32(uint(address(wrapper))), WRAPPER_VERSION);
```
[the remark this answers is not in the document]
dbeal-eth: its not important. all that matters is to put a non-zero value in the storage slot, and the version is more potential use than just putting `1`. Who knows! maybe it will be helpful later on.

using the address as the value would not be very helpful since the usage of the `address` in the key does not glean any additional information that `isWrapper` didn't have before.

contracts/Wrapper.sol
```diff
@@ -0,0 +1,238 @@
...
+        require(amountIn <= token.balanceOf(msg.sender), "Balance is too low");
+        require(!exchangeRates().rateIsInvalid(currencyKey), "Currency rate is invalid");
+
+        uint currentCapacity = capacity();
+        require(currentCapacity > 0, "Contract has no spare capacity to mint");
+
+        uint actualAmountIn = currentCapacity < amountIn ? currentCapacity : amountIn;
+
+        uint feeAmountTarget = calculateMintFee(actualAmountIn);
+        uint mintAmount = actualAmountIn - feeAmountTarget;
```
liamzebedee, line 159: Can you swap the +/- out for the SafeMath ones please?

contracts/Wrapper.sol
```diff
@@ -0,0 +1,238 @@
...
+        // We use burn/issue instead of burning the principal and transferring the fee.
+        // This saves an approval and is cheaper.
+        // Escrow fee.
+        synthsUSD().issue(address(wrapperFactory()), excessAmountUsd);
+
+        _setTargetSynthIssued(targetSynthIssued + excessAmount - amount);
+    }
+
+    function _setTargetSynthIssued(uint _targetSynthIssued) internal {
+        debtCache().recordExcludedDebtChange(currencyKey, int256(_targetSynthIssued) - int256(targetSynthIssued));
```
liamzebedee, line 230: Target synth is a good choice of name, I like it.

contracts/Wrapper.sol
```diff
@@ -0,0 +1,238 @@
...
+        synth().issue(msg.sender, amount);
+
+        // Escrow fee.
+        synthsUSD().issue(address(wrapperFactory()), excessAmountUsd);
+
+        _setTargetSynthIssued(targetSynthIssued + excessAmount + amount);
+    }
+
+    function _burn(uint amount) internal {
+        uint excessAmount = getReserves() - (targetSynthIssued - amount);
```
liamzebedee, line 215: What does excess amount refer to? Excess of what? Excess capacity?

test/contracts/Wrapper.js
```diff
@@ -520,57 +559,73 @@ contract('EtherWrapper', async accounts => {
...
-			decodedEventEqual({
-				event: 'Issued',
-				emittedFrom: sUSDSynth.address,
-				args: [FEE_ADDRESS, multiplyDecimal(feesEscrowed, rate)],
-				log: logs
-					.reverse()
-					.filter(l => !!l)
-					.find(({ name }) => name === 'Issued'),
+
+		describe('mint after transfer without mint', () => {
```
liamzebedee, line 593: Uh, what does this test case do? I don't see any transfers / what does without mint refer to?

test/contracts/Wrapper.js
```diff
@@ -520,57 +559,73 @@ contract('EtherWrapper', async accounts => {
...
 	});
 
-	describe('distributeFees', async () => {
-		let tx;
-		let feesEscrowed;
-		let sETHIssued;
+	describe('transfer without mint', () => {
+		let preTargetIssuedSynths;
+
+		const extraTransferred = toUnit('2');
```
liamzebedee, line 565: extraTransferred? Extra of what? I think the naming here can more accurately reflect what's going on

test/contracts/helpers.js
```diff
@@ -30,12 +30,16 @@ module.exports = {
 	// Assert against decoded logs
 	decodedEventEqual({ event, emittedFrom, args, log, bnCloseVariance = '10' }) {
 		assert.equal(log.name, event);
-		assert.equal(log.address, emittedFrom);
+		assert.equal(log.address, emittedFrom, 'log emission address does not match');
```
liamzebedee, line 33: 🙏

liamzebedee (changes requested):
Overall, PR is looking good. Mostly I think the naming could be made better - as for the logic, tests are passing, so it's looking good. Nice work!

liamzebedee:
@dbeal-eth re: your failing test, the line of code in the test suite 

test/contracts/SynthetixState.js:166:4
`await assert.invalidOpcode(synthetixState.lastDebtLedgerEntry());`

indicates to me that there has been a debt snapshot or some addition to the debt ledger. Maybe your tests have a takeDebtSnapshot() which wasn't surrounded by a `addSnapshotBeforeRestoreAfterEach`?

liamzebedee:
"how do I represent MultiCollateralSynth change in releases file? by individual synths, or MultiCollateralSynth? just left it empty for right now."

See the EtherWrapper PR. You redeploy each individual synth, eg. `SynthsETH`, `SynthsUSD`  - https://github.com/Synthetixio/synthetix/pull/1178/files#diff-53c896eece47396479c4ce779cbf1e79043b5f53fde1c26c8555c46ee710438e

dbeal-eth:
> @dbeal-eth re: your failing test, the line of code in the test suite
> 
> test/contracts/SynthetixState.js:166:4
> `await assert.invalidOpcode(synthetixState.lastDebtLedgerEntry());`
> 
> indicates to me that there has been a debt snapshot or some addition to the debt ledger. Maybe your tests have a takeDebtSnapshot() which wasn't surrounded by a `addSnapshotBeforeRestoreAfterEach`?

I don't think that's the reason because:
* if I print `debtLedgerLength` prior within the test, it returns 0
* the function still fails with a revert, just not specifically `invalid opcode`
* the same test succeeds when I push code and have circleci run it

maybe a problem with my local setup or something

contracts/Wrapper.sol
dbeal-eth: so does that mean `token` should be `originToken`?

contracts/Wrapper.sol
dbeal-eth: this function is designed to take any reserves which are unaccounted for and send them to the sUSD fees (that way if someone sends WETH to the contract or something its not locked away forever, and it won't cause unaccounted issuance). Therefore, `excessAmount` is the amount of unaccounted reserves that haven't been issued, for which sUSD fees can be taken. Would `unaccountedReserves` be a better name?

test/contracts/Wrapper.js
dbeal-eth: how about `unaccountedTransfer`?

test/contracts/Wrapper.js
dbeal-eth: there is an edge case where a user could mess up the  `DebtCache` by using `ERC20.transfer` with the target to send funds to the wrapper, and effectively reduce the debt of the system without issuing any synths or fees. I didn't know the implications of this, so this test case was added to make sure proper handling of unaccounted transfers to the wrapper (aka `transfer` instead of calling `mint`).

contracts/MixinSystemSettings.sol
```diff
@@ -127,15 +127,27 @@ contract MixinSystemSettings is MixinResolver {
         return flexibleStorage().getUIntValue(SETTING_CONTRACT_NAME, SETTING_DEBT_SNAPSHOT_STALE_TIME);
     }
 
-    function getEtherWrapperMaxETH() internal view returns (uint) {
-        return flexibleStorage().getUIntValue(SETTING_CONTRACT_NAME, SETTING_ETHER_WRAPPER_MAX_ETH);
+    function getWrapperMaxTokenAmount(address wrapper) internal view returns (uint) {
```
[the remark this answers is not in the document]
dbeal-eth: within the code the naming has been consistently `token` as corresponding to WETH within the `EtherWrapper`, and `synth` corresponding to sETH. I agree it can be confusing at a glance. Since these terms are used repeatedly, it kind of became a game of brevity vs. clarity. If we want to be more verbose (ex. `originToken` and `targetSynth`, or perhaps `origin` and `target`, but these terms could easily be used for other purposes), let me know what would be most appropriate

In the case of this setting, usage of token or synth likely wouldn't cause confusion because whether your thinking it of WETH locked or sETH issued its the same...

contracts/BaseDebtCache.sol
dbeal-eth: re-added all `EtherWrapper` components in https://github.com/Synthetixio/synthetix/pull/1489/commits/7460d2113f9f4773110ddd297db7e94efd2d955a

contracts/MultiCollateralSynth.sol
dbeal-eth: readded with all the rest of EtherWrapper in https://github.com/Synthetixio/synthetix/pull/1489/commits/7460d2113f9f4773110ddd297db7e94efd2d955a

contracts/SystemSettings.sol
dbeal-eth: we might go over the contract limit with the SIP 135, and public functions cost more gas to deploy... and I don't think you can use public functions with interfaces

contracts/MixinSystemSettings.sol
dbeal-eth: resolved in private conversation

contracts/BaseDebtCache.sol
dbeal-eth: fixed with the EtherWrapper remerge https://github.com/Synthetixio/synthetix/pull/1489/commits/7460d2113f9f4773110ddd297db7e94efd2d955a

contracts/BaseDebtCache.sol
dbeal-eth: the mentioned line was removed https://github.com/Synthetixio/synthetix/pull/1489/commits/7460d2113f9f4773110ddd297db7e94efd2d955a

contracts/Wrapper.sol
dbeal-eth: https://github.com/Synthetixio/synthetix/pull/1489/commits/db1199479f2431da8b982304b8a868ccdf5b7953

contracts/WrapperFactory.sol
dbeal-eth: https://github.com/Synthetixio/synthetix/pull/1489/commits/586818dc6bc25bac90bac55f9d303de02facb60c

contracts/Wrapper.sol
dbeal-eth: https://github.com/Synthetixio/synthetix/pull/1489/commits/1e7c4b22b4452bd3f7edc2c9805c6cf1cb2d42b7

PR #1491, opened 2021-09-03
SIP-174 Fix for settlements to not cached removed synths

justinjmoses:
While `settle()` is smart enough to ignore any synth that is no longer in the system with a rate (see [this line](https://github.com/Synthetixio/synthetix/blob/master/contracts/Exchanger.sol#L225)) - and simply removes the entry in this case - there was an issue with `DebtCache` reverting. Thus, this change to 174 ensures that even after deprecation, if a user has unsettled exchanges into the synth being removed, settling them at a later date won't revert.

> Note: one requirement of this is that the aggregator for the synth is removed from `ExchangeRates` - which needs to be done during deprecation

Another way to solve this would be to guarantee all exchanges into the synth were settled prior to removing it, but this is an onerous task that would cost a lot of gas to process and even a single missed entry would mean that that account could no longer settle into that synth.

### DIFF
--- a/contracts/Exchanger.sol
+++ b/contracts/Exchanger.sol
@@ -660,9 +660,9 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
             refund(from, currencyKey, refunded);
         }
 
-        // we check the synth is still in the system before updating the cache
-        // (this protects against settlements for removed synths causing any issues)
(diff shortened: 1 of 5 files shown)


barrasso (approved):
LGTM

KillerByte (approved):

PR #1493, opened 2021-09-05
Adding simulated-release task and CI step

justinjmoses:
- `generateSolidity` now checks `CONTRACT_NAME` of the new contract using its `source` (to work with contracts like `SynthsUSD` and `SynthsETH` which require `MultiCollateralSynth` be checked (see https://github.com/Synthetixio/synthetix/pull/1493/commits/717bed025d2acaa2a612ff97681e054d71e03efb) 
- Migrates `import-fee-periods` into the deploy script, to ensure the steps are correctly handled by the `generateSolidity` command (see https://github.com/Synthetixio/synthetix/pull/1493/commits/db63d8efeed0b688a9827e289d0357eb63a0cccc)
- `generateSolidity` creates internal functions when any calls require arrays to prevent stack too deep errors.
- `generateSolidity` allows for custom solidity for the complex steps such as FeePool import and Synth totalSupply copy.
- Also adds a new hardhat task `simulate:migration` which can compile, prepare and deploy against a fork to generate a migration contract for the current release, and then compile and deploy the migration, nominate the appropriate contracts to it, run the migration, then run the owner script back again to make sure all contracts are returned and then run the integration tests. 
- Add a `CI` step `job-simulate-release` that does the above task only when there is an upcoming release (see https://github.com/Synthetixio/synthetix/pull/1493/commits/36933dd8b737b4969d356bded49b61638b821cae)
- Adds `--layer` and `--name-only` to `releases` in `bin.js` for use with CI 
- Fixes a bug with `prepare-deploy` which was coming up with the integrations tests as we recently added `prepareDeploy` for non-forked tests to ensure a new synth: `sREDEEMER` was prepped to help with the SynthRedeemer integration tests. See https://github.com/Synthetixio/synthetix/commit/05e39e0e474fb7d55680aca0c0180a332e277687 (cc @mjlescano)
- Ensures that `Synth.setTotalSupply()` is invoked when creating a new synth and generating solidity.


contracts/Issuer.sol
```diff
@@ -21,18 +21,14 @@ import "./interfaces/IHasBalance.sol";
...
 import "./interfaces/ICollateralManager.sol";
+import "./interfaces/IRewardEscrowV2.sol";
 import "./interfaces/ISynthRedeemer.sol";
 import "./Proxyable.sol";
 
 interface IProxy {
     function target() external view returns (address);
 }
 
-interface IRewardEscrowV2 {
```
justinjmoses, line 31: Note: having this redefined here was causing issues with the migration compilation, so I elected to remove it.

contracts/migrations/Migration_Mirfak.sol
```diff
@@ -0,0 +1,422 @@
...
+        contracts[15] = address(proxyseth_i);
+        contracts[16] = address(issuer_i);
+    }
+
+    function migrate(address currentOwner) external onlyDeployer {
+        require(owner == currentOwner, "Only the assigned owner can be re-assigned when complete");
+
+        // NEW CONTRACTS DEPLOYED TO BE ADDED TO PROTOCOL
+        // https://etherscan.io/address/0x4959851554b87D3f113d1e9383d601F27A9f8F71
+        address new_FeePool_contract = 0x4959851554b87D3f113d1e9383d601F27A9f8F71;
```
justinjmoses, line 97: Note: these are all temporary based on the fork, and this file will be regenerated once a) #1494 is merged and b) this is regenerated after the new contracts are deployed and verified on mainnet

contracts/BaseSynthetix.sol
```diff
@@ -18,7 +18,6 @@ import "./interfaces/IVirtualSynth.sol";
 
 contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
     // ========== STATE VARIABLES ==========
-    bytes32 public constant CONTRACT_NAME = "Synthetix";
```
justinjmoses, line 21: This is being moved to `Synthetix.sol` as the new migration tweaks require the source actually what's deployed, so this is future-proofing a situation where we have a migration on L2 and `Synthetix` migrating would mean `MintableSynthetix.sol` should expose `CONTRACT_NAME="MintableSynthetix"`

liamzebedee (commented):
I only got 80% through reviewing this today. Looks good so far.

hardhat/tasks/task-simulate-migration.js
```diff
@@ -0,0 +1,131 @@
...
+		const ownerAddress = getUsers({ network: 'mainnet', user: 'owner' }).address;
+
+		// but deploy this new migration contract using regular ethers onto the fork (using Migration.deploy won't deploy to the fork as needed)
+		const Factory = new ethers.ContractFactory(
+			Migration.interface,
+			Migration.bytecode,
+			provider.getSigner(getUsers({ network: 'mainnet', user: 'deployer' }).address)
+		);
+
+		const migration = await Factory.deploy({ gasPrice: ethers.utils.parseUnits('0') });
```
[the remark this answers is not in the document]
justinjmoses: so the migration is only configured for L1 right now cause L2 doesn't support forking - so this is L1.

jacko125 (approved):
moving the import fee period into deploy steps so the steps are included in `runSteps` looks good.

contracts/BaseSynthetix.sol
jacko125: How does the `CONTRACT_NAME` get used in the generate solidity / migration script to look up the source contract file?

justinjmoses: It’s simply a sanity check that the contract address being imported does indeed match the contract we expect

PR #1496, opened 2021-09-06
Add closeWithCollateral to SIP-135

barrasso:
This allows users to close their short when repaying with collateral.

- Added tests for the `closeWithCollateral` function that repays the loan in full, records it as closed, and transfers the remaining collateral back to the borrower.
- Removed the `payInterest` flag -> pay back the `accruedInterest` in all cases.


contracts/Collateral.sol
```diff
@@ -356,6 +356,20 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
         _recordLoanAsClosed(loan);
     }
 
+    function _closeLoanByRepayment(address borrower, uint id) internal returns (uint amount, uint collateral) {
+        // 0. Get the loan.
+        Loan storage loan = loans[id];
+
+        // 1. Repay the loan with its collateral.
+        (amount, collateral) = _repayWithCollateral(borrower, id, loan.amount);
```
jacko125, line 364: looks good, so this will repay all the accrued interest as well in `_repayWithCollateral`

contracts/CollateralShort.sol
```diff
@@ -58,17 +58,14 @@ contract CollateralShort is Collateral {
         (principal, collateral) = _repay(borrower, msg.sender, id, amount);
     }
 
-    // Needed for Lyra.
-    function repayWithCollateral(uint id, uint amount) external returns (uint principal, uint collateral) {
-        (principal, collateral) = _repayWithCollateral(msg.sender, id, amount, true);
+    function repayAndClose(uint id) external returns (uint amount, uint collateral) {
```
jacko125, line 61: Should this function be named `closeLoanWithCollateral` ?

barrasso: Agreed. That sounds better. Resolved

PR #1497, opened 2021-09-06
Fixes for imports to prevent issues with IERC20

justinjmoses:
This is a cleanup of the contract imports so that multiple definitions of `IERC20` - both internally and via Open Zeppelin, don't start conflicting with each other at compile time (msg: `DeclarationError: Identifier already declared.`). This started appearing in the #1493 PR as the migration needed to import both `Issuer` and `Exchanger` which led to `ERC20` being imported, along with IERC20 from Open Zeppelin, and then clashed with other imports of the internal `IERC20`.

The solution was a) to remove all references to `VirtualSynth` from `Exchanger` in the first place, relying instead on the interface, as the minimal proxy solution allows for this; then b) remove `IERC20` imported into  VirtualSynth, relying on the OZ one.

### DIFF
--- a/contracts/TradingRewards.sol
+++ b/contracts/TradingRewards.sol
@@ -6,7 +6,6 @@ import "./MixinResolver.sol";
 import "./Owned.sol";
 
 // External dependencies.
-import "openzeppelin-solidity-2.3.0/contracts/token/ERC20/ERC20Detailed.sol";
 import "openzeppelin-solidity-2.3.0/contracts/token/ERC20/SafeERC20.sol";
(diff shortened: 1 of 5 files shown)


barrasso (approved):

contracts/TradingRewards.sol
```diff
@@ -6,7 +6,6 @@ import "./MixinResolver.sol";
 import "./Owned.sol";
 
 // External dependencies.
-import "openzeppelin-solidity-2.3.0/contracts/token/ERC20/ERC20Detailed.sol";
```
justinjmoses, line 9: This is a no-op - just cleaning up an unused import.

PR #1499, opened 2021-09-07
Moving the new synth REDEEMER in the tests to the appropriate setup

justinjmoses:
Note: config.json no longer uses `true` for everything locally - but `freshDeploy` enables it by default.

This PR is designed to remove the adding of `sREDEEMER` synth for testing out of the necessary prep for all integration tests, and into the actual integration test that requires it (thereby removing it from the simulated migration code that runs in #1493 )


justinjmoses:
Closing for now. Alternative solution in #1493

PR #1501, opened 2021-09-09
Mirfak release to kovan

justinjmoses:


publish/releases.json
```diff
@@ -320,7 +320,8 @@
 				"Issuer",
 				"Synthetix",
 				"SynthsETH",
-				"SynthsUSD"
+				"SynthsUSD",
+				"SynthsBTC"
```
justinjmoses, line 324: Note: @jacko125 this was missing from SIP-142

contracts/migrations/Migration_Mirfak.sol
```diff
@@ -0,0 +1,611 @@
+pragma solidity ^0.5.16;
```
justinjmoses, line 1: Note: this is the kovan migration, but will be overridden by the mainnet one during the release next week.

barrasso (approved):

PR #1502, opened 2021-09-10
SIP-135: audit fixes

barrasso:
- [x]  contracts/CollateralState.sol
    - [x]  Contract should be removed as it is unused
- [x]  contracts/Collateral.sol
    - [x]  _liquidate does not need to check_isLoanOpen(...) as _getLoanAndAccrueInterest calls it.
    - [x]  The service fee for the collapsible shorts from the SIP spec does not appear to be implemented
    - [x]  comments:
        - [x]  L548, "let child handle it", appears to be incorrect as the child contract is not really handling anything.
        - [x]  L577 "Otherwise multiple" → multiply
- [x]  contracts/CollateralErc20.sol
    - [x]  There is an assumption of tokens have a max of 18 decimals
    - [x]  transferFrom → should use safeTransferFrom 
    - [x]  transfer → should use safeTransfer
- [ ]  contracts/CollateralManager.sol
    - [ ]  L241 skew limit max should be considered. It's unlikely to be a problem with the current shortable synths, but in low volume synths it could lead to undesirable outcomes.
    - [x]  might run into gas issues on L2 if too many synths, e.g.  exceedsDebtLimit iterates through both as long/short synth arrays.
    - [x]  L102, iSynth → "ISynth" (to avoid confusion with inverse synths)
    - [x]  Should add event for setUtilisationMultiplier
    - [x]  accrueInterest no revert message on L489
    - [x]  accrueInterest is based on approximation as before, works if it's actively used -- simply informational, should be better than L1 with higher usage. 
- [x]  contracts/CollateralShort.sol
    - [x]  closeWithCollateral can add check for if collateral > 0 to avoid 0 value transfer


contracts/CollateralManager.sol
```diff
@@ -99,7 +99,7 @@ contract CollateralManager is ICollateralManager, Owned, Pausable, MixinResolver
         staticAddresses[0] = CONTRACT_ISSUER;
         staticAddresses[1] = CONTRACT_EXRATES;
 
-        // we want to cache the name of the synth and the name of its corresponding iSynth
+        // we want to cache the name of the synth and the name of its corresponding ISynth
```
jacko125, line 102: we can remove the iSynth from this comment

PR #1504, opened 2021-09-13
update deployer to support EIP-1559

KillerByte:
updates `--gas-price` to be replaced with `--max-fee-per-gas` which specifies the maximum base fee paid on a transaction. Additionally,
deployer may also specify `--max-priority-fee-per-gas` to specify a mining tip (default: 1 gwei)

If the network does not support 1559, the `gasPrice` is automatically determined by ethers.

For EIP-1559 support, ethers.js needed to be upgraded to 5.4.6

### DIFF
--- a/publish/src/Deployer.js
+++ b/publish/src/Deployer.js
@@ -23,6 +23,8 @@ class Deployer {
 		deploymentFile,
 		dryRun,
 		gasPrice,
+		maxFeePerGas,
+		maxPriorityFeePerGas,
 		methodCallGasLimit,
 		network,
 		providerUrl,
@@ -39,6 +41,8 @@ class Deployer {
 		this.deploymentFile = deploymentFile;
 		this.dryRun = dryRun;
 		this.gasPrice = gasPrice;
+		this.maxFeePerGas = maxFeePerGas;
+		this.maxPriorityFeePerGas = maxPriorityFeePerGas;
 		this.methodCallGasLimit = methodCallGasLimit;
 		this.network = network;
 		this.contractDeploymentGasLimit = contractDeploymentGasLimit;
@@ -110,13 +114,28 @@ class Deployer {
 		return ethers.utils.defaultAbiCoder.encode(types, params);
 	}
 
+	async addGasOptions(tx) {
+		if (this.gasPrice) {
+			tx.gasPrice = ethers.utils.parseUnits(this.gasPrice.toString(), 'gwei');
+		} else if ((await this.provider.getFeeData()).maxFeePerGas) {
+			if (this.maxFeePerGas)
+				tx.maxFeePerGas = ethers.utils.parseUnits(this.maxFeePerGas.toString(), 'gwei');
+			if (this.maxPriorityFeePerGas)
+				tx.maxPriorityFeePerGas = ethers.utils.parseUnits(
+					this.maxPriorityFeePerGas.toString(),
+					'gwei'
+				);
+		}
+	}
+
 	async sendDummyTx() {
 		const tx = {
 			to: '0x0000000000000000000000000000000000000001',
 			data: '0x0000000000000000000000000000000000000000000000000000000000000000',
 			value: 0,
-			gasPrice: ethers.utils.parseUnits(this.gasPrice.toString(), 'gwei'),
 		};
+
+		this.addGasOptions(tx);
 
 		const response = await this.signer.sendTransaction(tx);
 		await response.wait();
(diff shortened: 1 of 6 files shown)


publish/src/Deployer.js
```diff
@@ -110,14 +114,29 @@ class Deployer {
 		return ethers.utils.defaultAbiCoder.encode(types, params);
 	}
 
+	async addGasOptions(tx) {
+		if (this.gasPrice) {
+			tx.gasPrice = ethers.utils.parseUnits(this.gasPrice.toString(), 'gwei');
+		} else if ((await this.provider.getFeeData()).maxFeePerGas) {
```
justinjmoses, line 120: this indicates EIP-1559 support?

publish/src/Deployer.js
```diff
@@ -133,10 +152,9 @@ class Deployer {
 			? this.methodCallGasLimit
 			: this.contractDeploymentGasLimit;
 
-		const params = {
-			gasLimit,
-			gasPrice: ethers.utils.parseUnits(this.gasPrice.toString(), 'gwei'),
-		};
+		const params = { gasLimit };
```
justinjmoses, line 155: I think we can also potentially remove `gasLimit` now that we're using ethers and the estimate is much more robust. Thoughts?

publish/src/Deployer.js
```diff
@@ -110,14 +114,29 @@ class Deployer {
 		return ethers.utils.defaultAbiCoder.encode(types, params);
 	}
 
+	async addGasOptions(tx) {
```
justinjmoses, line 117: I don't love this approach of mutating the given argument - better to return a version mixed in with `Object.assign`

publish/src/commands/deploy/index.js
```diff
@@ -85,6 +86,8 @@ const deploy = async ({
 	deploymentPath = deploymentPath || getDeploymentPathForNetwork({ network, useOvm });
 	ensureDeploymentPath(deploymentPath);
 
+	let gasPrice = null;
```
justinjmoses, line 89: instead of doing this, we should remove it entirely from here and all functions that it is passed into :pray:

publish/src/Deployer.js
justinjmoses: Maybe instead of using Deployer we just have a util that is `mixinGasOptions` inside `publish/src/utils`, and can be used in all the commands, taking the `maxFeePerGas` and `maxPriorityFeePerGas` and mixing it into a new object, returning that? We can omit `gasPrice` altogether now

justinjmoses: Also if the mixin takes `provider` then it can perform the `provider.getFeeData()` itself

publish/src/Deployer.js
dbeal-eth: probably true

publish/src/Deployer.js
dbeal-eth: good idea :+1:

publish/src/Deployer.js
dbeal-eth: yes they added this function after the EIP-1559 switch to help users detect if the network supports EIP-1559 (as well as what the recommended gas prices are). see https://github.com/ethers-io/ethers.js/issues/1610#issuecomment-874370333

I checked and the value does not return `type` or anything like that so I agree the check does look a little awkward

dbeal-eth:
@justinjmoses all comments addressed in https://github.com/Synthetixio/synthetix/pull/1504/commits/ec3d4a4a89409a8504d23fd4472acdd6deef79ec

dbeal-eth:
@justinjmoses requesting review to merge this PR as is since the PR appears to be a net-positive of testing situation in the `develop` branch right now, and the current integration test error does not appear to be related to changes in this PR.

contracts/CollateralShort.sol
```diff
@@ -72,6 +72,9 @@ contract CollateralShort is Collateral {
 
     // Needed for Lyra.
     function getShortAndCollateral(address borrower, uint id) external view returns (uint principal, uint collateral) {
+        // needed because lint fails here
+        require(borrower != address(0), "dummy");
```
justinjmoses, line 76: ah this is a warning right? can we move this into it's own PR?

hardhat/tasks/task-node.js
```diff
@@ -27,6 +27,8 @@ task('node', 'Run a node')
 			console.log(yellow(`Forking ${network}...`));
 		}
 
+		console.log(taskArguments);
```
justinjmoses, line 30: pls remove

PR #1507, opened 2021-09-16
CI improvements and maintenance

drptbl:
**Changes:**
- updated docker containers to use latest node version (14.17.6) and npm (7.23.0)
- use [codecov uploader](https://docs.codecov.com/docs/codecov-uploader) instead of bash script which is deprecated
- added `job-audit` before `job-prepare` with two additional steps:
    - `npm audit` which will fail the whole workflow only if there is critical vulnerability detected in dependencies (`npm install` will never be triggered in this case increasing security)
    - [`lockfile-lint`](https://www.npmjs.com/package/lockfile-lint) which makes sure that lockfile is up to date and has no malicious stuff inside
- updated base runner ubuntu image to use the latest
- improved caching of node dependencies
- updated codechecks dependency

**New workflow:**
Screenshot 2021-09-16 at 17 19 00

_side note:_
`job-fork-tests` is also failing on `develop` branch, so it's not related to my changes.


justinjmoses:
Thanks! Yep `fork-tests` is failing due to some reason when run without anything to fork.

justinjmoses (approved):
💯

dbeal-eth (approved):
thanks for these updates.

should we create a ticket ot migrate this pipeline to github actions at some point?

PR #1516, opened 2021-09-21
fix slither crashing in CI and local

artdgn:
Context: was working through the different script targets and tests and this one was broken, so thought it might be nice to fix it. 

Details: slither was not working in both CI and locally, [example run]( https://app.circleci.com/pipelines/github/Synthetixio/synthetix/8795/workflows/1e547102-ac94-4f7d-8558-6b4b2d6fdae1/jobs/87273/parallel-runs/0/steps/0-103). It seems it has been broken for a while (at least 4mo back, haven't check earlier).

changes:
  - pass the correct artifacts path in config
  - avoid using OZ Math import to prevent name collision that was causing slither to fail https://github.com/crytic/slither/wiki#keyerror-or-nonetype-error
  - modify local slither target to install slither in virtual env instead of globally
  - store slither report as CI stage artifact

testing: local runs of slither, unit tests (`npx hardhat test --grep "Rewards"`)


artdgn:
Slither CI output after the fix: https://app.circleci.com/pipelines/github/Synthetixio/synthetix/8810/workflows/0fbd5882-dba6-4ba4-aef5-ba6b84a85f35/jobs/87499/parallel-runs/0/steps/0-103

artdgn:
`ci/circleci: job-test-deploy-script ` [fails with](https://app.circleci.com/pipelines/github/Synthetixio/synthetix/8811/workflows/944a8ae5-c4da-4add-a6aa-6b63a07f8bfe/jobs/87509?invite=true#step-103-44):

```
Error: No compiled source for: StakingRewardssETHUniswapV1. The source file is set to StakingRewards.sol - is that correct?
```

I don't how that's related to my changes.

And fork-tests seems to have been broken before as well.

artdgn:
@justinjmoses this should be better now (it's the third approach: not importing the OZ Math library at all).

The added `internal` method is just for the `min` function, which is very minimal and is tested through the tests that already exist for the methods that were using it before.

artdgn:
Now the report is stored as a CI artifact, [example](https://circleci.com/gh/Synthetixio/synthetix/88088?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

artdgn:
Closed and replaced by https://github.com/Synthetixio/synthetix/pull/1519

PR #1517, opened 2021-09-21
ci: lavamoat integration

drptbl:
**Changes:**
- integrated [`@lavamoat/allow-scripts`](https://github.com/LavaMoat/LavaMoat/tree/main/packages/allow-scripts) dependency which adds an allowlist for dependencies that are allowed to run `postinstall` scripts. This is an additional layer of protection from for ex. [software supply chain attacks](https://github.blog/2020-09-02-secure-your-software-supply-chain-and-protect-against-supply-chain-threats-github-blog/) and a case when `postinstall` scripts are stealing env vars on execution.
- added [initial allowlist](https://github.com/Synthetixio/synthetix/pull/1517/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R145-R160) of `postinstall` scripts. None of them are mandatory for us, so it's safe to disallow them.
- thanks to the change related to getting rid of external reference to `github.com` for `solidifier` package, I was able to [improve](https://github.com/Synthetixio/synthetix/pull/1517/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R30) `lockfile-lint` command and make it more strict and secure. `lockfile-lint` will fail if there is a reference to dependency from any other source than official `npm` registry.

fixed using occasion:
- codecov was posting double PR messages and status checks on PR, it's now [fixed](https://github.com/Synthetixio/synthetix/pull/1517/files#diff-3e878d9bdc47f29a0ab909a7db939acf08d2d2f1aaba81e6f897b32361fa40dbR3).


README.md
```diff
@@ -35,6 +35,10 @@ For a guide from the community, see [synthetix.community](https://synthetix.comm
 
 ## Repo Guide
 
+### Setup
+
+Run `npm run setup` script after installing dependencies with `npm install`. It will trigger postinstall scripts for dependencies which are whitelited inside allowlist and apply our custom patches for external dependencies if needed.
```
justinjmoses, line 40: QU: would a `postinstall` be the best way to do `setup`? I'd rather this be automated were possible.

drptbl: It's not possible because all `postinstall` scripts (from dependencies and this repository) are disabled inside `.npmrc` file with `ignore-scripts=true` flag. This is done this way to make sure that no one runs a `postinstall` script from a dependency which is not in allowlist (it forces users to use `allow-scripts` command). Sadly, `npm` doesn't offer a way to disable `postinstall` scripts selectively for dependencies.

drptbl: @justinjmoses I have updated the description in the readme. Actually, you don't have to run `npm install` before `npm run setup` because it's already included in it. All you have to do to setup the repo (and install dependencies) is to run `npm run setup`.

drptbl: @justinjmoses As an alternative solution, we could switch to use `yarn@2` instead of `npm`.

`yarn` offers allowlist feature for `postinstall` scripts out-of-the-box:
https://next.yarnpkg.com/configuration/manifest#dependenciesMeta.built

Thanks to that, we could resign to use `@lavamoat/allow-scripts` with conjunction of `npm` and use pure `yarn@2` instead.

I'm not having any other ideas currently. My currently proposed solution is also what other repositories use, for ex. `metamask-extension`:
https://github.com/MetaMask/metamask-extension/blob/develop/package.json#L10

patches/solidifier+2.1.0.patch
```diff
@@ -0,0 +1,125 @@
+diff --git a/node_modules/solidifier/cli.js b/node_modules/solidifier/cli.js
+deleted file mode 100755
+index 38eb705..0000000
+--- a/node_modules/solidifier/cli.js
++++ /dev/null
+@@ -1,14 +0,0 @@
```
justinjmoses, line 6: Why doesn't this patch include everything in the fork? Or did you just remove the version info from the `package.json`?

README.md
justinjmoses: Fair enough, but I don't really want to bring in yarn just for this. OK, since it's (running setup) super unlikely to be critical unless you are deploying I can get behind it. 

I am going to talk with @thekevinbrown about getting this fork into the parent so we don't even need `patch-package`

patches/solidifier+2.1.0.patch
drptbl: Everything is included except version bump from `package.json`. I don't think that it would well taking in to an account that solidifer is referenced from `synthetix` repo as `solidifier@2.1.0`.

justinjmoses (changes requested):
HOLD. Requesting an update to solidifier from @thekevinbrown

justinjmoses (approved):

PR #1519, opened 2021-09-23
fix slither crashing in CI and local

artdgn:
New PR instead of https://github.com/Synthetixio/synthetix/pull/1516 as now I can push into repo without forking

--------

Details: slither was not working in both CI and locally, [example run]( https://app.circleci.com/pipelines/github/Synthetixio/synthetix/8795/workflows/1e547102-ac94-4f7d-8558-6b4b2d6fdae1/jobs/87273/parallel-runs/0/steps/0-103). It seems it has been broken for a while (at least 4mo back, haven't check earlier).

changes:
  - pass the correct artifacts path in config
  - avoid using OZ Math import to prevent name collision that was causing slither to fail https://github.com/crytic/slither/wiki#keyerror-or-nonetype-error
  - modify local slither target to install slither in virtual env instead of globally
  - store slither report as CI stage artifact

testing: local runs of slither, unit tests (`npx hardhat test --grep "Rewards"`)


.gitignore
```diff
@@ -42,3 +42,6 @@ process.yml
 .db
 
 optimism
+
+# python venv
+.venv/
```
justinjmoses, line 47: can you ensure newlines at end of file is turned on for your IDE?

artdgn: :+1: fixed thx

.circleci/config.yml
```diff
@@ -1,3 +1,5 @@
+# DO NOT EDIT MANUALLY!
```
justinjmoses, line 1: 🙏🏼

contracts/RewardsDistributionRecipient.sol
```diff
@@ -17,4 +17,15 @@ contract RewardsDistributionRecipient is Owned {
...
     }
+
+    /* ========== Math helpers ========== */
+
+    /*
+     * @dev copied from openzeppelin-solidity-2.3.0/contracts/math/Math.sol
+     *  to avoid using OZ Math import because of contract name collision (Math)
+     *  with the internal Math.sol contract which e.g. is confusing for slither-analyzer
+     */
+    function _min(uint256 a, uint256 b) internal pure returns (uint256) {
```
justinjmoses, line 28: So you're mixing logic so you can get it in the staking rewards contracts but it really has no business being in here. I think we can just rewrite the logic in the two contracts rather than using the function. Thoughts?

justinjmoses (changes requested):

contracts/RewardsDistributionRecipient.sol
artdgn: Yeah I thought to put it as a helper in both of the using contracts, but it was used in logic that is really common between the two and seemed related to the inherited parts of `RewardsDistributionRecipient` /  `notifyRewardAmount`. So it seemed ok to move it there as well, but I agree that looking at that file in isolation it's not clear what the `_min` function is doing there.

Alternatively. the whole method `lastTimeRewardApplicable` with its only used state variable `periodFinish` can be moved up into `RewardsDistributionRecipient` as well, but that also feels awkward. 

I think if the option of repeating the code is cleaner, than since the logic of this function is so simple I can just inline it in the two places.

justinjmoses (approved):

package.json
```diff
@@ -12,7 +12,7 @@
 		"lint": "solhint \"contracts/*.sol\" && solhint \"contracts/test-helpers/*.sol\" && solhint --config contracts/interfaces/.solhint.json \"contracts/interfaces/*.sol\" && eslint \"**/*.js\"",
 		"lint:fix": "eslint --fix \"**/*.js\"",
 		"build:ci": "node .circleci/pack.js",
-		"slither": "pip3 install --user slither-analyzer && slither .",
+		"slither": "python3 -m venv .venv && .venv/bin/python -m pip install slither-analyzer && .venv/bin/python -m slither .",
```
justinjmoses, line 15: 🙏🏼

PR #1523, opened 2021-09-24
add slither code check github integration

artdgn:
This adds code quality checks on PRs (visible in `Checks` tab of the PR) which can be configured to alert / block etc via the the repo's `Security` tab. 

It won't show anything interesting for this PR, because it has nothing to compare to on the base branch. But I've checked that it shows stuff when PRing from another branch into this one: [example checks view for that PR](https://github.com/Synthetixio/synthetix/pull/1525/checks?check_run_id=3695508441)

How useful / noisy / configurable it is - remains to be seen in future PRs.


justinjmoses (commented):
This is awesome :100:! Should we remove it from circleci then? I think we can

artdgn:
@justinjmoses 

> This is awesome 100! Should we remove it from circleci then? I think we can

maybe it's ok to keep both for a while, and then remove whatever looks less useful after a while? The fancy github integration may have drawbacks, and the simple report dump may have its own advantages as a simple file stored for all runs. Is there any cost / risk for keeping the simple one for now?

I can set a reminder for two weeks time to have another look at the integration thingy and remove the circle-ci step if it looks redundant then.

justinjmoses (approved):

PR #1528, opened 2021-09-24
Adds block tag parameter to interact task

noahlitvin:
This allows the interact task to run at a specified [block tag](https://docs.ethers.io/v5/api/providers/types/#providers-BlockTag) for easier troubleshooting. For example, `npx hardhat interact --block-tag 13289452`.

Happy to make improvements prior to merge. Thanks!


noahlitvin:
I tested this further and it does seem to be working as expected.

justinjmoses:
Looking good. Merging to a local branch and will do the merge from there 🙏🏼

PR #1529, opened 2021-09-25
Reduce gas usage for opening loans

barrasso:
The goal of this PR is to reduce the gas used when calling `CollateralShort.open` so it doesn't hit the limit on L2.

This requires a balance between the contract size of `CollateralShort` and the gas usage of `_open`. 

Ideally, we want to minimize any external contract calls since they are quite costly.


contracts/Collateral.sol
```diff
@@ -251,7 +271,7 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
         require(amount <= maxLoan(collateral, currency), "Exceed max borrow power");
 
         // 7. This fee is denominated in the currency of the loan.
-        uint issueFee = amount.multiplyDecimalRound(getIssueFeeRate(address(this)));
+        uint issueFee = amount.multiplyDecimalRound(issueFeeRate);
```
barrasso, line 274: I am seeing a reduction of ~1M *(!)* gas used by moving `issueFeeRate` back into the `Collateral` contract.

test/contracts/Collateral.js
```diff
@@ -15,7 +15,13 @@ contract('Collateral', async accounts => {
 		ensureOnlyExpectedMutativeFunctions({
 			abi: Collateral.abi,
 			ignoreParents: ['Owned', 'Pausable', 'MixinResolver', 'Proxy'],
-			expected: ['addRewardsContracts', 'addSynths'],
+			expected: [
+				'addRewardsContracts',
```
justinjmoses, line 19: at the very least please add tests to make sure they are `onlyOwner` using `onlyGivenAddressCanInvoke`

test/contracts/CollateralShort.js
```diff
@@ -219,6 +230,70 @@ contract('CollateralShort', async accounts => {
...
+				beforeEach(async () => {
+					await short.setCanOpenLoans(false, { from: owner });
+				});
+				it('should update the flag', async () => {
+					assert.isFalse(await short.canOpenLoans());
+				});
+			});
+		});
+
+		describe('setMinCollateral', async () => {
```
justinjmoses, line 253: these should be moved to the `Collateral` tests.

justinjmoses: Also look at using the `onlyGivenAddressCanInvoke` function to test `onlyOwner`

contracts/Collateral.sol
```diff
@@ -173,6 +176,23 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
         return true;
     }
 
+    /* ---------- SETTERS ---------- */
+
+    function setMinCollateral(uint _minCollateral) external onlyOwner {
```
dbeal-eth, line 181: why were these setters missing before, if they were never removed from the `Collateral` contract?

test/contracts/Collateral.js
```diff
@@ -15,7 +15,13 @@ contract('Collateral', async accounts => {
...
+				'addSynths',
+				'setCanOpenLoans',
+				'setIssueFeeRate',
+				'setMinCollateral',
+			],
 		});
```
dbeal-eth, line 25: do tests need to be added to this file to verify the correct funcitoning of the `set` functions?

test/contracts/SystemSettings.js
```diff
@@ -546,6 +506,9 @@ contract('SystemSettings', async accounts => {
 	});
 
 	describe('setLiquidationRatio()', () => {
+		before(async () => {
+			await setupSettings();
```
dbeal-eth, line 510: why is this added here?

test/integration/behaviors/short.behavior.js
```diff
@@ -90,7 +88,10 @@ function itCanOpenAndCloseShort({ ctx }) {
 				before('open the loan', async () => {
 					tx = await CollateralShort.open(amountToDeposit, amountToBorrow, toBytes32('sETH'));
 
-					const { events } = await tx.wait();
+					const receipt = await tx.wait();
+					console.log(`short.open gas used: ${receipt.gasUsed.toString()}`);
```
dbeal-eth, line 92: remove log

dbeal-eth (approved):
look before merging

PR #1534, opened 2021-09-28
Print deploy param object value

barrasso:
This PR introduces a minor addition to `get-deploy-parameter-factory.js` so it can output the value of an overridden parameter of type `Object`.


dbeal-eth (approved):
yay

PR #1539, opened 2021-09-29
Refactor exceedsDebtLimit check

barrasso:
This PR fixes an issue where opening a loan may hit the gas limit on L2 since `CollateralManager.exceedsDebtLimit` calls `totalLong` and `totalShort` which iterate through both long and short synth arrays.

- Added the `totalLongAndShort` function so we can do just one `SLOAD` per fnc call rather than doing one for each synth in a loop.
- Reporting ~1M gas savings

### DIFF
--- a/contracts/CollateralManager.sol
+++ b/contracts/CollateralManager.sol
@@ -43,6 +43,9 @@ contract CollateralManager is ICollateralManager, Owned, Pausable, MixinResolver
     // The set of all collateral contracts.
     AddressSetLib.AddressSet internal _collaterals;
 
+    // The set of all available currency keys.
+    Bytes32SetLib.Bytes32Set internal _currencyKeys;
+
     // The set of all synths issuable by the various collateral contracts
     Bytes32SetLib.Bytes32Set internal _synths;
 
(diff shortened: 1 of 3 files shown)


contracts/CollateralManager.sol
```diff
@@ -277,17 +297,12 @@ contract CollateralManager is ICollateralManager, Owned, Pausable, MixinResolver
         (entryRate, lastRate, lastUpdated, newIndex) = state.getShortRatesAndTime(currency, index);
     }
 
-    // This function might run into gas issues on L2 if there are too many synths
-    // since it iterates through both long and short synth arrays.
     function exceedsDebtLimit(uint amount, bytes32 currency) external view returns (bool canIssue, bool anyRateIsInvalid) {
         uint usdAmount = _exchangeRates().effectiveValue(currency, amount, sUSD);
 
-        (uint longValue, bool longInvalid) = totalLong();
```
jacko125, line 285: @barrasso  If you're storing the currencyKeys now in `_currencyKeys` , you could update `totalLong()` and `totalShort()` to skip the external call to load the key where it does this on L176:

`bytes32 synth = _synth(synths[i]).currencyKey();` 

Is totalLong() and totalShort() still required ?

jacko125 (commented):
Lgtm, only one suggestion

contracts/CollateralManager.sol
```diff
@@ -204,6 +207,23 @@ contract CollateralManager is ICollateralManager, Owned, Pausable, MixinResolver
...
     }
 
+    function totalLongAndShort() public view returns (uint susdValue, bool anyRateIsInvalid) {
+        bytes32[] memory currencyKeys = _currencyKeys.elements;
+
+        if (currencyKeys.length > 0) {
+            (uint[] memory rates, bool invalid) = _exchangeRates().ratesAndInvalidForCurrencies(currencyKeys);
+            for (uint i = 0; i < rates.length; i++) {
+                uint longAmount = state.long(currencyKeys[i]).multiplyDecimal(rates[i]);
+                uint shortAmount = state.short(currencyKeys[i]).multiplyDecimal(rates[i]);
```
jacko125, line 217: This will return 0 when the shortable synth is not there. I wonder if you can use the `_shortableSynths` to check whether to even make this external check here, but then you need to store the shortableSynth currencykeys too.

contracts/CollateralManager.sol
barrasso: > If you're storing the currencyKeys now in _currencyKeys , you could update totalLong() and totalShort() to skip the external call to load the key where it does this on L176:

Resolved: https://github.com/Synthetixio/synthetix/pull/1539/commits/7b50aa8a8b3756368f938019b7be085b610bb711

> Is totalLong() and totalShort() still required ?

They are both still used in `BaseDebtCache._totalNonSnxBackedDebt`. Eventually this can be updated, but I think it is out of scope for this SIP.

contracts/CollateralManager.sol
barrasso: I think returning 0 here is acceptable behavior. I'm not sure that adding another piece of contract storage for shortable currency keys is a good idea here since it's yet another array to manage, and the lengths of these arrays can differ which may lead to out of bounds issues.

jacko125: All good

jacko125 (approved):
lgtm this change

PR #1546, opened 2021-10-04
Fixing releases to also mark SIPs as released

justinjmoses:
- Fixes issue whereby a finalized release (in the release pipeline) does not actually mark the SIPs as released for the given `layer`
- Also fixes bug whereby releasing with version `-ovm` was being incorrectly skipped.

### DIFF
--- a/publish/src/commands/finalize-release.js
+++ b/publish/src/commands/finalize-release.js
@@ -26,12 +26,12 @@ const finalizeRelease = async ({ layer, release, versionTag }) => {
 		await versionsUpdate({ release, useOvm: true, versionTag });
 	}
 
-	const prerelease = semver.prerelease(versionTag) && semver.prerelease(versionTag) !== 'ovm';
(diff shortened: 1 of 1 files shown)


publish/src/commands/finalize-release.js
```diff
@@ -26,12 +26,12 @@ const finalizeRelease = async ({ layer, release, versionTag }) => {
 		await versionsUpdate({ release, useOvm: true, versionTag });
 	}
 
-	const prerelease = semver.prerelease(versionTag) && semver.prerelease(versionTag) !== 'ovm';
+	const prerelease = semver.prerelease(versionTag) && semver.prerelease(versionTag)[0] !== 'ovm';
```
dbeal-eth, line 29: are we certain that `semver.prerelease(versionTag)[0]` is not null? otherwise the functionality would not be exactly the same as before.

barrasso (approved):
LGTM, just have same concern that db raised about being certain the first element in the array is not null

PR #1547, opened 2021-10-05
Futures: merge develop

artdgn:
- [x] Resolved conflicts with help from @liamzebedee 
- [ ] CI should be green eventually
- [ ] Review and verify `releases.json` makes sense, since we both weren't sure about our conflict resolution there


artdgn:
@liamzebedee 

I'm having some trouble resolving the last two CI steps - `integration-tests` and `validate-deployments`.

---------------

`integration-tests` fails with on CI after deployments:
```
TypeError: Cannot read property 'numMarkets' of undefined
    at module.exports (/home/circleci/repo/publish/src/commands/deploy/deploy-futures.js:115:69)
``` 

But locally fails during deployment with:
```
- Attempting to deploy Synthetix (with source MintableSynthetix)
An unexpected error occurred:

Error: nonce has already been used (error={"reason":"processing response error","code":"SERVER_ERROR","body":"{\"jsonrpc\":\"2.0\",\"id\":489,\"error\":{\"code\":-32000,\"message\":\"invalid transaction: nonce too low\"}}\n","error":{"code":-32000},"requestBody":"{\"method\":\"
``` 

---------------

`validate-deployments` fails during tests, and seems to be related to deployments not hapening correctly either:
```
TypeError: Cannot read property 'Synthetix' of undefined
```

liamzebedee:
> TypeError: Cannot read property 'Synthetix' of undefined

Re: this error. The deployment script doesn't work at all for ovm yet. So kovan-ovm and kovan-ovm-futures networks should be removed actually. Since this is just merging in develop, I don't think making the script work belongs in this PR. @artdgn

PR #1554, opened 2021-10-08
Helper script to distribute SNX/sUSD to accounts on kovan-futures-ovm

liamzebedee:
This is a helper script to fund accounts with SNX on L2. This is necessary in order to test the futures keeper locally, as we need SNX to create new positions in the futures markets. It's also useful in the case of running another futures testnet, as the SNX needs to be distributed here too.

I considered making this an external tool but it belongs colocated with the synthetix repo. 

Example usage:

```
npx hardhat fund-local-accounts --provider-url https://kovan.optimism.io --target-network kovan-ovm --deployment-path ./publish/deployed/kovan-ovm-futures/ --use-ovm --private-key $KOVAN_OVM_FUTURES_PRIVATE_KEY
```


hardhat/tasks/task-local.js
```diff
@@ -0,0 +1,125 @@
+const { task } = require('hardhat/config');
```
justinjmoses, line 1: Can you rename this file to match the task? `task-fund-local-accounts.js` 🙏🏼

jacko125 (approved):
lgtm the faucet helper script once file renamed!

hardhat/tasks/task-local.js
liamzebedee: Sure

PR #1562, opened 2021-10-13
Upgrade deployer for OVM 2.0

liamzebedee:
The OVM 2.0 upgrade means that all contracts can now be run natively using EVM code. This removes edge case code we have to navigate around the old OVM 1.0 architecture, where contracts were transpiled to OVM bytecode.

 - removing edge cases from deployer
 - removing edge cases from verify script
 - fixes integration tests to use a newer version of the Optimism ops docker image. My reason for using this instead of just replacing it with a Hardhat chain to test against, is that Optimism is still a moving target, and it will save time to keep testing against their geth instance rather than assuming it's completely EVM compatible with no bugs.


justinjmoses:
Thanks mate! I do think we should tackle #1504 first though

artdgn (approved):

liamzebedee:
Closing as these changes were merged in #1589, and JJ's done the legwork on the rest of them in https://github.com/Synthetixio/synthetix/pull/1598

PR #1563, opened 2021-10-13
Feat(DynamicFee): SIP-184 dynamic fees

leckylao:
- [x] Params.json (L1)
- [x] first implementation
- [x] Performance - have `lastDynamicFee` and `lastRoundSeen` for caching
- [x] Merge develop once https://github.com/Synthetixio/synthetix/pull/1127 merged
- [x] Add more test scenarios. 
- [x] Unit test
- [x] Add Exchanger test with/without dynamic fee
- [x] Add Exchanger test sBTC -> sUSD and sUSD -> sBTC with sBTC dynamic fee. sETH -> sBTC with both currency dynamic fee
- [x] integration test
- [ ] Audit
- [ ] Testnet deploy


contracts/DynamicFee.sol
```diff
@@ -0,0 +1,59 @@
...
+    function getDynamicFee(
+        uint[] memory prices,
+        uint threshold,
+        uint weightDecay
+    ) public pure returns (uint dynamicFee) {
+        uint size = prices.length;
+        require(size >= 2, "Not enough prices");
+        for (uint i = 0; i < size - 1; i++) {
+            uint priceDifferential = getPriceDifferential(prices[i], prices[i + 1], threshold);
+            uint priceWeight = getPriceWeight(i, weightDecay);
```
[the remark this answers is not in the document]
artdgn: > Both ways complexity should be the same that both need to calculate the decay 10 times right?

Not exactly. The previous way there is added complexity due to `.pow()` being calculated in every iteration inside the `getPriceWeight()`. That calculation runs it's own loop (doing multiplications). So that adds some complexity which can be removed. It's logarithmic, but still no need to run those extra internal loops.

> As DynamicFee is a library so looks like best place to store lastDynamicFee and lastRoundSeen would be on Exchanger the place that it get call. 

I think something along these lines, and probably the library interface would change a little. It would be something like `updateDynamicFee(uint lastDynamicFee, uint lastRoundSeen, address exchangeRates) returns (uint dynamicFee, uint lastRound)` or something. And inside there it would query `ExchangeRates` for the last round, and get the price data or something.

@leckylao

contracts/DynamicFee.sol
```diff
@@ -0,0 +1,61 @@
...
+        uint price,
+        uint previousPrice,
+        uint threshold
+    ) public pure returns (uint) {
+        require(price > 0, "Price cannot be 0");
+        require(previousPrice > 0, "Previous price cannot be 0");
+
+        int abs = int(price.divideDecimal(previousPrice)) - int(SafeDecimalMath.unit());
+        abs = abs > 0 ? abs : -abs;
+        int priceDifferential = abs - int(threshold);
```
jacko125, line 30: @leckylao So here we subtract the `threshold` instead of only reporting it if it is above the threshold?

contracts/DynamicFee.sol
```diff
@@ -0,0 +1,61 @@
...
+    /// @param previousPrice Previous round price
+    /// @param threshold Threshold constant -
+    /// A system constant for the price differential default to 40 bps
+    /// @return uint price differential with 18 decimals
+    /// only return if non-zero value, otherwise return 0
+    function getPriceDifferential(
+        uint price,
+        uint previousPrice,
+        uint threshold
+    ) public pure returns (uint) {
```
jacko125, line 24: shouldn't this be external ?

contracts/DynamicFee.sol
```diff
@@ -0,0 +1,59 @@
...
+        return weightDecay.powDecimal(round);
+    }
+
+    /// @notice Calculate dynamic fee based on preceding 10 price differential
+    /// @param prices A list of prices from the current round to the previous rounds
+    /// @param threshold A threshold to determine the price differential
+    /// @param weightDecay A weight decay constant
+    /// @return uint dynamic fee
+    function getDynamicFee(
+        uint[] memory prices,
```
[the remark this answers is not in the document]
jacko125: Given that we need the last 10 prices from the oracle, could we not store the last 10 rounds in our own ExchangeRates contract for example (as an improvement over making 10x external calls to the oracle).

Then only fetch the latest prices if they are not stored in the ExchangeRates and persist them to storage as well? This might be more expensive gas wise given we are writing to storage. 

Also has CL looked into improving their interface to allow retrieving a list of roundID / prices with one call?

contracts/DynamicFee.sol
```diff
@@ -0,0 +1,61 @@
...
+        uint threshold,
+        uint weightDecay
+    ) public pure returns (uint dynamicFee) {
+        uint size = prices.length;
+        require(size >= 2, "Not enough prices");
+        for (uint i = prices.length - 1; i > 0; i--) {
+            uint priceDifferential = getPriceDifferential(prices[i - 1], prices[i], threshold);
+            uint roundDecay = getRoundDecay(i, weightDecay);
+            dynamicFee = (dynamicFee.multiplyDecimal(roundDecay)).add(priceDifferential);
```
jacko125, line 58: @leckylao Might be worth having you run through this PR and the logic here for calculating the `dynamicFee` to make sure we're on the right track. 

Would you apply the `roundDecay` to the priceDifferential by `priceDifferential.multiplyDecimal(roundDecay)` if you want to decay the priceDifferential based on the number of rounds it was before?

contracts/DynamicFee.sol
leckylao: It's also being used in `getDynamicFee`  https://github.com/Synthetixio/synthetix/pull/1563/files/a999fe36ae951fb9f55124dd057bd0f3a68402d3#diff-6019708811a67d9f46fdf302d9455ff6d5f2016232a8de6753a33ef1dc218405R56

contracts/DynamicFee.sol
leckylao: I break it down into two lines, essentially it's `return abs - int(threshold) > 0 ? uint(abs - int(threshold)) : uint(0);`

contracts/DynamicFee.sol
leckylao: Yes, I have that on the first place. @artdgn points out this way could store the `lastDynamicFee` and `lastRoundSeen` for improvement. And I've checked test case passed in both ways. Will implement the improvement once I have fixed all tests
https://github.com/Synthetixio/synthetix/pull/1563/commits/8aa8c5e5e0ff25504f9487a2fc0dbe5f9d47c1f9#file-contracts-dynamicfee-sol-L58

contracts/DynamicFee.sol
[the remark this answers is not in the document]
leckylao: Yes, I am thinking to have a separate PR to try reading the feed round data from cache first using `_rate` which is already there. That way the whole system can also benefits from it as when frequent exchange it would be just reading from cache. 
https://github.com/Synthetixio/synthetix/blob/develop/contracts/ExchangeRates.sol#L446

Not necessary need to be persistent in storage as it's just cacheing, if cacheing not found would just go fetch again as it's doing currently.

Here's the last update from CL. Still ongoing conversations
https://discord.com/channels/413890591840272394/809168416350928946/896036115570655323

contracts/DynamicFee.sol
leckylao: > I'm not sure about the calculation in this revision. It seems to apply the decay exponentially more than needed because it multiplies by `roundDecay` instead of just by `weightDecay`. This code would result for e.g. 4 iterations in this: Iteration 1: `dynamicFee = 0 * weightDecay ^ 9 + priceDiff9 = priceDiff9` Iteration 2: `dynamicFee = priceDiff9 * weightDecay ^ 8 + priceDiff8` Iteration 3: `dynamicFee = (priceDiff9 * weightDecay ^ 8 + priceDiff8) * weightDecay ^ 7 + priceDiff7 = priceDiff9 * weightDecay ^ 15 + priceDiff8 * weightDecay ^ 7 + priceDiff7` Iteration 4: `dynamicFee = priceDiff9 * weightDecay ^ 21 + priceDiff8 * weightDecay ^ 13 + priceDiff7 * weightDecay ^ 6 + priceDiff6` ... Which is probably incorrect.
> 
> Wouldn't the code in [#1563 (comment)](https://github.com/Synthetixio/synthetix/pull/1563#discussion_r736116702) be more correct and result in this ?: Iteration 1: `dynamicFee = 0 * weightDecay + priceDiff9 = priceDiff9` Iteration 2: `dynamicFee = priceDiff9 * weightDecay + priceDiff8` Iteration 3: `dynamicFee = (priceDiff9 * weightDecay + priceDiff8) * weightDecay + priceDiff7 = priceDiff9 * weightDecay ^ 2 + priceDiff8 * weightDecay + priceDiff7` Iteration 4: `dynamicFee = priceDiff9 * weightDecay ^ 3 + priceDiff8 * weightDecay ^ 2 + priceDiff7 * weightDecay + priceDiff6` Which is what we need?

Nah, it's round decay then it would match Afif's data. Can run you through on the Monday brief session.

contracts/DynamicFee.sol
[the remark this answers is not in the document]
leckylao: > > Both ways complexity should be the same that both need to calculate the decay 10 times right?
> 
> Not exactly. The previous way there is added complexity due to `.pow()` being calculated in every iteration inside the `getPriceWeight()`. That calculation runs it's own loop (doing multiplications). So that adds some complexity which can be removed. It's logarithmic, but still no need to run those extra internal loops.

Can run through the roundDecay with Afif on the Monday brief session. 

> > As DynamicFee is a library so looks like best place to store lastDynamicFee and lastRoundSeen would be on Exchanger the place that it get call.
> 
> I think something along these lines, and probably the library interface would change a little. It would be something like `updateDynamicFee(uint lastDynamicFee, uint lastRoundSeen, address exchangeRates) returns (uint dynamicFee, uint lastRound)` or something. And inside there it would query `ExchangeRates` for the last round, and get the price data or something.
> 
> @leckylao

Then the library would have dependancy on the ExchangeRates and need to get it from AddressResolver which is another dependancy. The idea of using it as library is less dependancy and can be import anywhere else to use it.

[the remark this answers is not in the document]
artdgn: > Then the library would have dependancy on the ExchangeRates and need to get it from AddressResolver which is another dependancy.

That's not what I meant. You pass in the address in `updateDynamicFee(uint lastDynamicFee, uint lastRoundSeen, address exchangeRates)` and then there is no dependency on resolver, just need to import the interface.

leckylao:
Gas cost for Synthetix.exchange:

Before implementation is 500666 Gwei:
image

After implementation is 740175 Gwei:
image

After optimisation is 5845863 Gwei:
image

test/contracts/BaseSynthetix.js
```diff
@@ -856,7 +856,7 @@ contract('BaseSynthetix', async accounts => {
 		it("should lock newly received synthetix if the user's collaterisation is too high", async () => {
 			// Set sEUR for purposes of this test
 			const timestamp1 = await currentTime();
-			await exchangeRates.updateRates([sEUR], [toUnit('0.75')], timestamp1, { from: oracle });
+			await exchangeRates.updateRates([sEUR], [toUnit('1')], timestamp1, { from: oracle });
```
artdgn, line 859: why were these changes needed? doesn't it make testing harder by not being able to tell if exchange rate is used correctly or not (when exchange rate is 1)

test/contracts/TestableDynamicFee.js
```diff
@@ -48,7 +48,41 @@ contract('TestableDynamicFee', () => {
 			toUnit('49933.34034209'),
 			toUnit('49871.92313713'),
 		];
-		const dynamicFee = await testableDynamicFee.testGetDynamicFee(prices);
+		const dynamicFee = await testableDynamicFee.testGetDynamicFee(prices, '0');
 		assert.bnEqual(dynamicFee, '2064427530203592');
 	});
+
+	it('Can get dynamic fee round 15-24', async () => {
```
artdgn, line 55: can you add some details on what part of the dynamic fee calculation this range is testing? can we upload the reference table to github gist, or as tests data so that it can be used for future test updates?

test/contracts/TestableDynamicFee.js
```diff
@@ -48,7 +48,41 @@ contract('TestableDynamicFee', () => {
...
+			toUnit('49722.83886705'),
+			toUnit('49838.87627216'),
+			toUnit('49842.74988613'),
+			toUnit('49933.34034209'),
+		];
+		const dynamicFee = await testableDynamicFee.testGetDynamicFee(prices, '0');
+		assert.bnEqual(dynamicFee, '1857984777183232');
+	});
+
+	it('Can get dynamic fee round 34-43', async () => {
```
artdgn, line 72: same thing - as a reader / maintainer I have no idea what this is testing or what to the values should be (both for setup and the expected result)

test/contracts/TestableDynamicFee.js
```diff
@@ -48,7 +48,41 @@ contract('TestableDynamicFee', () => {
...
+			toUnit('49009.46274509'),
+			toUnit('49054.03062906'),
+			toUnit('49093.89744338'),
+			toUnit('49095.24231598'),
+			toUnit('49101.41'),
+			toUnit('49208'),
+		];
+		const dynamicFee = await testableDynamicFee.testGetDynamicFee(prices, '0');
+		assert.bnEqual(dynamicFee, '8062531530836597');
+	});
```
artdgn, line 87: I think it would be good to have ranges and setups that test these parts of the dynamic fee calculation:
1. When price diffs are all below threshold.
2. When price diffs are all above threshold.
3. When some price diffs are above, and some below threshold.
4. When there was a caching gap longer than 10 rounds.
5. When there was a caching gap shorter than 10 rounds.
6. That with caching and without caching (e.g. gap) - there values are the same.
7. That if there was no caching gap longer than 10 rounds, dynamic fee is persisted correctly (doesn't start from 0, but starts from previous values from before 10 rounds)
8. Reverse trades BTC->USD have same dynamic fee as USD->BTC
9. ETH->BTC, BTC->ETH trades have correct dynamic fee (max of either side?)

test/contracts/Exchanger.spec.js
```diff
@@ -712,6 +683,8 @@ contract('Exchanger (spec tests)', async accounts => {
 							from: oracle,
 						}
 					);
+					// Disable Dynamic Fee by setting rounds to 0
+					await systemSettings.setExchangeDynamicFeeRounds('0', { from: owner });
```
artdgn, line 687: It looks like dynamic fee in actual exchanges is never tested - always disabled. If so, we should add test cases that test that dynamic fee is correctly used and deducted during exchanges. That it's correctly in test cases that probe:
1. Different kinds of directions - BTC->USD, USD->BTC, ETH->BTC, BTC->ETH - otherwise we're not covering the logic that relates to sUSD.
2. That dynamic fee is persisted correctly after exchanges and is used in next exchange.
3. That caching works correctly with gaps and without gaps and results in same results.

artdgn (changes requested):
Good stuff, but I think there some gaps (but maybe I'm missing something):
1. More testing for the dynamic fee calculation - I've pointed out below some test scenarios that I think are important.
2. Testing of dynamic fee being correctly used and persisted in actual `exchange()` functionality (it seems disabled in exchange tests). And more test cases to cover the pairs logic and the caching logic.
3. The refactor for gas optimisation in ExchangeRates for latest rate seems to possibly increase gas usage.
3. The refactor for gas optimisations in exchange() seems to not be needed - doesn't look to be reducing gas.
4. The gas usage measerement is probably too optimistic because in most `exchange()` tests dynamic fee is disabled.
5. There seems to be a an issue with dynamic fee and direction of trade (only destination is used).
6. There seems to be an issue with currentRoundId state in Rates that can be set to arbitrary user values and the incrementing looks prone to get out of sync with oracle.
7. Wasn't sure about reentrancy risks due to calling only internal contracts and not arbitrary external contracts.
8. The actual calculation of the dynimic fee still looks weird and not according to the spec. I think the conclusion of the discussion is that more test cases, and more clarity in the test cases as to what they're testing are needed.


There more details at specific line numbers in the review.

test/contracts/TestableDynamicFee.js
leckylao: Thanks for the test cases suggestion. After I added more cases turns out that current formula is not correct so I have to rollback to the previous formula. Which is good as now it align with the SIP. 

And I have to remove the `lastExchangeDynamicFeeRate` cache as it doesn't really work. Different dynamic fee using different rounds and they have different `roundDecay`. Additionally the the last dynamic fee contains the last 10 rounds. When using for the new calculations it will need to subtract the additional rounds. So not worth it.

leckylao:
> Gas cost for Synthetix.exchange:
> 
> Before implementation is 500666 Gwei: image
> 
> After implementation is 740175 Gwei: image
> 
> After optimisation is 572653 Gwei: image

Gas cost after adding dynamic fee for `sourceCurrencyKey`

image

barrasso (changes requested):

contracts/ExchangeRates.sol
```diff
@@ -483,8 +543,14 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
         }
     }
 
-    function _getRateAndTimestampAtRound(bytes32 currencyKey, uint roundId) internal view returns (uint rate, uint time) {
+    function _getRateAndUpdatedTimeAtRound(bytes32 currencyKey, uint roundId) internal view returns (uint rate, uint time) {
```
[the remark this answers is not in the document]
leckylao: The propose here is to make it consistent with the function `_getRateAndUpdatedTime` as the code are pretty much the same and with the `AtRound` to indicate accept `roundId`.

contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -142,8 +142,8 @@ contract ExchangeRatesWithDexPricing is ExchangeRates {
         // updates is a good proxy for price volatility.
         uint considerationWindowStart = block.timestamp.sub(considerationWindow);
         uint roundId = _getCurrentRoundId(currencyKey);
-        for (updateThreshold; updateThreshold > 0; updateThreshold--) {
-            (uint rate, uint time) = _getRateAndTimestampAtRound(currencyKey, roundId);
```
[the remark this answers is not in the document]
leckylao: Right, good point. Will rollback so don't need to deploy this contract

contracts/Exchanger.sol
```diff
@@ -443,8 +435,33 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
+        // SIP-65: Decentralized Circuit Breaker
+        // mutative call to suspend system if the rate is invalid
+        if (
+            _suspendIfRateInvalid(sourceCurrencyKey, entry.sourceRate) ||
+            _suspendIfRateInvalid(destinationCurrencyKey, entry.destinationRate)
+        ) {
+            return (0, 0, IVirtualSynth(0));
+        }
+
         uint sourceAmountAfterSettlement = _settleAndCalcSourceAmountRemaining(sourceAmount, from, sourceCurrencyKey);
```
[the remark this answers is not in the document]
leckylao: So that it can has less cost to suspend the system? As circuit breaker only need to check the rate.

publish/deployed/mainnet/params.json
```diff
@@ -1,4 +1,8 @@
 [
+	{
+		"name": "EXCHANGE_DYNAMIC_FEE_ROUNDS",
+		"value": "1"
```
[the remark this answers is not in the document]
leckylao: Yes, `EXCHANGE_DYNAMIC_FEE_ROUNDS` takes minimum value 2 to be effective, setting it to 1 on Mainnet in order to disable it on L1 and not being override back to default 10 after new `SystemSettings` deployment.

test/contracts/CollateralShort.js
```diff
@@ -54,17 +57,22 @@ contract('CollateralShort', async accounts => {
...
 	const updateRatesWithDefaults = async () => {
-		const timestamp = await currentTime();
-
-		await exchangeRates.updateRates([sETH], ['100'].map(toUnit), timestamp, {
-			from: oracle,
-		});
+		let timestamp;
+		for (let i = 0; i < EXCHANGE_DYNAMIC_FEE_ROUNDS; i++) {
+			timestamp = await currentTime();
+			await exchangeRates.updateRates([sETH], ['100'].map(toUnit), timestamp, {
```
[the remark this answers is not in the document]
leckylao: Dynamic Fee applies when minimal 2 rates available.

test/contracts/DebtCache.js
```diff
@@ -966,6 +968,9 @@ contract('DebtCache', async accounts => {
 			});
 
 			it('exchanging between synths updates sUSD debt total due to fees', async () => {
+				// Disable Dynamic fee so that we can neglect it.
```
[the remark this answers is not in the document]
leckylao: Through this way so that we know what impact it could brings in and only disable if it's not relevant and in intention.

test/contracts/Exchanger.spec.js
```diff
@@ -209,6 +214,7 @@ contract('Exchanger (spec tests)', async accounts => {
 									new web3.utils.BN(1),
 									new web3.utils.BN(2),
 								],
+								bnCloseVariance,
```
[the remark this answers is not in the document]
leckylao: Good point, should disable dynamic fee instead to test for `ExchangeEntryAppended` match.

test/contracts/TestableDynamicFee.js
```diff
@@ -0,0 +1,132 @@
...
+		assert.bnEqual(priceWeight0, toUnit('1'));
+
+		const priceWeight1 = await testableDynamicFee.testGetPriceWeight('1');
+		assert.bnEqual(priceWeight1, toUnit('0.9'));
+
+		const priceWeight2 = await testableDynamicFee.testGetPriceWeight('2');
+		assert.bnEqual(priceWeight2, toUnit('0.81'));
+	});
+
+	it('Can get dynamic fee from dynamic-fee-calc.csv round 13-22, all below threshold', async () => {
```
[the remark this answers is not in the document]
leckylao: Made sense, changing it to "according to SIP feasibility spreadsheet"

test/contracts/Exchanger.spec.js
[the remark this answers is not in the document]
leckylao: Oh, actually it's not caused by dynamic fee. It's actually because introducing of the cache rate so that the actual exchange rate might not be the same from the view function `getAmountsForExchange`. As it's not enquiring the same round data. `exchanger.getAmountsForExchange` is getting the current round rate but ExchangeEntry rate is the historical rate.

contracts/ExchangeRates.sol
```diff
@@ -372,13 +425,26 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
...
-        currentRoundForRate[currencyKey]++;
+        if (roundId > 0) {
+            currentRoundForRate[currencyKey] = roundId;
+        } else {
+            // Note: this will effectively start the rounds at 1, which matches Chainlink's Agggregators
+            currentRoundForRate[currencyKey]++;
+        }
+
+        // skip writing if the rate is the same
+        if (rate == _rates[currencyKey][currentRoundForRate[currencyKey]].rate) return;
```
artdgn, line 447: nit: inline returns don't seem to be the usual style in the solidity in this codebase, and can be confusing to keep track when reasoning about where the method returns

contracts/ExchangeRates.sol
```diff
@@ -372,13 +422,23 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
...
     function _setRate(
         bytes32 currencyKey,
+        uint256 roundId,
         uint256 rate,
         uint256 time
     ) internal {
-        // Note: this will effectively start the rounds at 1, which matches Chainlink's Agggregators
-        currentRoundForRate[currencyKey]++;
+        if (roundId > 0) {
+            currentRoundForRate[currencyKey] = roundId;
```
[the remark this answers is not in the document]
artdgn: what happens when user passes 0 as round id in `mutativeEffectiveValueAndRatesAtRound`, what will oracle return and what will be cached in `_setRate()`? What will be the effect of that causing an increment in `currentRoundForRate[currencyKey]` because of the `roundId == 0` flow?

contracts/ExchangeRates.sol
```diff
@@ -372,13 +422,23 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
...
+        } else {
+            // Note: this will effectively start the rounds at 1, which matches Chainlink's Agggregators
```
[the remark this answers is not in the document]
artdgn: Still not sure if this flow needs to be in this contract if it's only used for testing. Shouldn't it be extended in a `Testable...` contract instead adding complexity and surface area in this contract?

I'm reading this again and again and it still doesn't sit right with me.

It seems that maybe anyone can increment `currentRoundForRate[currencyKey]` at will as much as they want by passing a `0` as round id in the `mutativeEffectiveValueAndRatesAtRound`. Can that cause some cache poisoning or some accounting bugs?

contracts/Exchanger.sol
```diff
@@ -626,7 +632,7 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
         }
 
         // if no last exchange for this synth, then we need to look up last 3 rates (+1 for current rate)
-        (uint[] memory rates, ) = exchangeRates().ratesAndUpdatedTimeForCurrencyLastNRounds(currencyKey, 4);
+        (uint[] memory rates, ) = exchangeRates().ratesAndUpdatedTimeForCurrencyLastNRounds(currencyKey, 4, 0);
```
artdgn, line 635: nit: mysterious `4` magic number should probably be a named constant (I know it's not new code, but still might be nice to fix since the line is touched).

contracts/Exchanger.sol
```diff
@@ -755,21 +775,31 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
+        // No dynamic fee for sUSD
+        if (currencyKey == sUSD) {
+            return 0;
         }
-
-        return exchangeFeeRate;
+        (uint threshold, uint weightDecay, uint rounds) = getExchangeDynamicFeeData();
+        uint[] memory prices;
+        // Note: We are using cache round ID here for cheap read
+        uint currentRoundId = exchangeRates().currentRoundForRate(currencyKey);
```
artdgn, line 800: can this be manipulated externally through `mutativeEffectiveValueAndRatesAtRound` and the ability to pass `0` roundId to increment `currentRoundForRate`?

contracts/MixinSystemSettings.sol
```diff
@@ -132,6 +139,48 @@ contract MixinSystemSettings is MixinResolver {
...
+    function getExchangeDynamicFeeData()
+        internal
+        view
+        returns (
+            uint threshold,
+            uint weightDecay,
+            uint rounds
+        )
+    {
+        threshold = getExchangeDynamicFeeThreshold();
```
artdgn, line 171: gas: can be further improved by using `getUIntValues` to get multiple values from `flexibleStorage` in a single call.

contracts/SystemSettings.sol
```diff
@@ -358,6 +386,46 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
         }
     }
 
+    /// @notice Set exchange dynamic fee threshold constant default 40bps
```
artdgn, line 389: nit: can probably mention that value is not in bips but is decimal ratio

contracts/SystemSettings.sol
```diff
@@ -358,6 +386,46 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
...
+        flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_EXCHANGE_DYNAMIC_FEE_WEIGHT_DECAY, weightDecay);
+
+        emit ExchangeDynamicFeeWeightDecayUpdated(weightDecay);
+    }
+
+    /// @notice Set exchange dynamic fee last N rounds
+    /// @param rounds The exchange dynamic fee last N rounds
+    /// @return uint dynamic fee last N rounds
+    function setExchangeDynamicFeeRounds(uint rounds) external onlyOwner {
+        flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_EXCHANGE_DYNAMIC_FEE_ROUNDS, rounds);
```
artdgn, line 415: can be 0?

contracts/SystemSettings.sol
```diff
@@ -358,6 +386,46 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
...
+
+        emit ExchangeDynamicFeeRoundsUpdated(rounds);
+    }
+
+    /// @notice Set max exchange dynamic fee
+    /// @param maxFee The max exchange dynamic fee
+    /// @return uint dynamic fee last N rounds
+    function setExchangeMaxDynamicFee(uint maxFee) external onlyOwner {
+        flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_EXCHANGE_MAX_DYNAMIC_FEE, maxFee);
```
artdgn, line 424: can be 0?

test/contracts/CollateralShort.js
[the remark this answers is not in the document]
artdgn: Is there a test for fewer than 10 rounds available?

test/contracts/TestableDynamicFee.js
```diff
@@ -0,0 +1,132 @@
...
+		assert.bnEqual(priceWeight0, toUnit('1'));
+
+		const priceWeight1 = await testableDynamicFee.testGetPriceWeight('1');
+		assert.bnEqual(priceWeight1, toUnit('0.9'));
+
+		const priceWeight2 = await testableDynamicFee.testGetPriceWeight('2');
+		assert.bnEqual(priceWeight2, toUnit('0.81'));
+	});
+
+	it('Can get dynamic fee according to SIP feasibility spreadsheet round 13-22, all below threshold', async () => {
```
artdgn, line 41: can you please commit the spreadsheet as a csv with the formulas as text so that future maintainers can reference that if needed to update / check the tests?

artdgn (changes requested):
- Still have some concerns about setRate caching w.r.t to roundId == 0 case.
- Would prefer to see the reference spreadsheet comitted as without it the tests for dynamic fee aren't maintainable.
- Some other small issues.

contracts/ExchangeRates.sol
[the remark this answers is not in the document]
artdgn: Seems that anyone can poison the cache by calling it with 0 - and setting several future rates to the current rate. Which will result in incorrect rate being used when future rounds happen because poisoned cached values will be used.

It seems that this whole currentRoundId incrementing should be avoided.

contracts/ExchangeRates.sol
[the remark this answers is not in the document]
artdgn: Added the exploit scenario above.

contracts/ExchangeRates.sol
artdgn: It works but it's easy to miss so hurts readability IMO. Still a nit though.

contracts/SystemSettings.sol
artdgn: :+1: maybe worth to add that into the comment

contracts/SystemSettings.sol
artdgn: Is there a use case for setting this to 0? aren't you disabling dynamic fee by setting rounds to 0 or 1?

contracts/Exchanger.sol
artdgn: just a named constant in the contract that is set to 4.

contracts/SystemSettings.sol
leckylao: This is to cap max dynamic fee to 1 (100%), otherwise it could go negative/overflow.

publish/deployed/mainnet/params.json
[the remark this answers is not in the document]
justinjmoses: You should just remove this though - this is only meant for overrides @leckylao

artdgn:
work continued in https://github.com/Synthetixio/synthetix/pull/1642 , closing this one

PR #1579, opened 2021-10-22
fix some minor logical issues

artdgn:
Fixes a few minor issues found in https://github.com/Synthetixio/synthetix/pull/1576/ that can still be nice to fix:
- interface and implementation mismatch (interface defines address, implementation IERC20)
- contract overriding constant values (happens to work, but not sure if it's a defined behavior)
- reducing array size with `.length--` 
- using uint(-1)

Not sure if worth fixing (if not going to redeploy), but FWIW.

### DIFF
--- a/contracts/Exchanger.sol
+++ b/contracts/Exchanger.sol
@@ -75,8 +75,6 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
         uint timestamp;
     }
 
-    bytes32 public constant CONTRACT_NAME = "Exchanger";
-
     bytes32 private constant sUSD = "sUSD";
 
     // SIP-65: Decentralized circuit breaker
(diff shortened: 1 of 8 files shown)


contracts/Exchanger.sol
```diff
@@ -115,6 +113,11 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
         addresses = combineArrays(existingAddresses, newAddresses);
     }
 
+    // solhint-disable func-name-mixedcase
+    function CONTRACT_NAME() external pure returns (bytes32) {
```
liamzebedee, line 117: Do you have the specific error that necessitated this change? I tried it in Remix and didn't see any warnings - so I assume it's somewhere else.

liamzebedee (changes requested):

contracts/Exchanger.sol
artdgn: The error happens in solidity 0.8, but the problem this change (in 0.5) is trying to fix is the implicit overriding of a constant value `CONTRACT_NAME`. I think 0.5 is overriding it according to our expectations, but I'm not sure this is a well defined behaviour (couldn't find it in the docs e.g.). It might be undefined with out usage of multiple inheritance (which of the base contract's "constant" is being used in a child?), so perhaps being explicit about it not being a constant, but an overridable function may make more sense. @liamzebedee 

Definitely not a super important change, but just wanted to fix the minor issues found in that solidity upgrade exercise.

liamzebedee: Right yeah - I just read through the docs:

> While public state variables can override external functions, they themselves cannot be overridden.
Screen Shot 2021-11-01 at 4 39 28 pm


🤦 It seems like the best approach here would be to make this a method instead of a public variable.

liamzebedee (approved):

PR #1582, opened 2021-10-27
Futures: skew scaling logic update

artdgn:
Fixes https://github.com/Synthetixio/issues/issues/363

- Renames currency valued parameters to be clearly in USD terms
- Updates parameter values and usage for `skewScale` to be in USD terms (renamed to `skewScaleUSD` - to make this risk parameter easier to set, and reduce potential of mistake due to confusion or miscalculation of `baseAsset` <> USD values for different markets.
- Sets `skewScaleUSD` to $50M for all markets.
- Updates the skew scaling calculation to always use the the constant scaling parameter (previously would use market size when it was larger). Still clips to [-1, 1] when skew is over the scaling constant.

Main functional change is that `minSkewScale` is named and set differntly:
- **Before**: `"minSkewScale": 300,` for BTC, and `"minSkewScale": 2000,` for ETH.
- **After**: `"skewScaleUSD": 50000000,` in all markets.


artdgn:
Fixes https://github.com/Synthetixio/issues/issues/363

liamzebedee (approved):

jacko125 (approved):
Lgtm

PR #1583, opened 2021-10-27
WIP: Convert FuturesMarket to use OpenZeppelin UUPS proxy

intmainvoid:


contracts/open-zeppelin/Initializable.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: MIT
```
intmainvoid, line 1: Convention is to keep it flat in contracts folder

PR #1588, opened 2021-11-01
ci: use node 16 (LTS)

drptbl:
> This release marks the transition of Node.js 16.x into Long Term Support (LTS) with the codename 'Gallium'. The 16.x release line now moves into "Active LTS" and will remain so until October 2022. After that time, it will move into "Maintenance" until end of life in April 2024.

source: https://nodejs.org/en/blog/release/v16.13.0/

New docker containers also use `npm@8`. No changes required while updating from `npm@7`.


dbeal-eth (approved):
if CI processes are succeeding then it is probably OK

barrasso (approved):

PR #1592, opened 2021-11-02
Removing inverse synths

justinjmoses:
In preparation for SIP-120 removing `ExchangeRates` (#1127) it's prudent to simply remove inverse pricing altogether.


dbeal-eth (approved):
a lot is gone

barrasso (approved):

contracts/PurgeableSynth.sol
```diff
@@ -51,10 +51,7 @@ contract PurgeableSynth is Synth {
         uint maxSupplyToPurge = exRates.effectiveValue("sUSD", maxSupplyToPurgeInUSD, currencyKey);
 
         // Only allow purge when total supply is lte the max or the rate is frozen in ExchangeRates
```
artdgn, line 53: nit: does the comment and revert message need updating too?

contracts/ExchangeRates.sol
```diff
@@ -96,71 +87,6 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
...
-            inverse.frozenAtUpperLimit = false;
-            inverse.frozenAtLowerLimit = false;
-            // remove any tracking
-            roundFrozen[currencyKey] = 0;
-        }
-
-        // SIP-78
-        uint rate = _getRate(currencyKey);
-        if (rate > 0) {
-            exchanger().setLastExchangeRateForSynth(currencyKey, rate);
```
artdgn, line 145: Can `Exchanger.setLastExchangeRateForSynth` be removed now as well?

justinjmoses: good catch

contracts/PurgeableSynth.sol
justinjmoses: fixed both

artdgn (approved):

test/contracts/SynthUtil.js
```diff
@@ -2,10 +2,7 @@
 
 const { contract } = require('hardhat');
 const { assert, addSnapshotBeforeRestoreAfterEach } = require('./common');
-const {
-	toBytes32,
-	constants: { ZERO_BYTES32 },
-} = require('../..');
+const { toBytes32 } = require('../..');
 const { toUnit, currentTime } = require('../utils')();
 const { setExchangeFeeRateForSynths } = require('./helpers');
```
sohkai, line 7: Just wondering, is it worth removing the `iBTC` bits in this file as well? For example, l51 still sets the `iBTC` rate on ExchangeRates.

test/contracts/PurgeableSynth.js
```diff
@@ -290,7 +290,6 @@ contract('PurgeableSynth', accounts => {
 				});
 
 				describe('when the user holds 5000 USD worth of the purgeable synth iETH', () => {
```
sohkai, line 292: Similar to above, worth removing `iETH` bits altogether here?

This test also seems shallow now; will holders of isynths not be able to invoke a purge from here on?

publish/src/commands/deploy/deploy-core.js
```diff
@@ -57,7 +57,6 @@ module.exports = async ({
 
 	await deployer.deployContract({
 		name: 'ExchangeRates',
-		source: useOvm ? 'ExchangeRatesWithoutInvPricing' : 'ExchangeRates',
```
sohkai, line 60: For SIP-120 we might need to keep / re-add this during deployment to switch between the base ExchangeRates and the one with dex pricing

sohkai (commented):
🙌 not totally up to speed on the status of isynths and if any are still left in wallets, but I'm assuming there aren't any. Looks good!

Just noted a couple of tests that still had isynths (not an exhaustive search though).

jacko125 (approved):
changes LGTM

test/contracts/PurgeableSynth.js
justinjmoses: correct. iSynths will never be purged - we've used the `SynthRedeemer` for deprecated ones and no new iSynths will be created.

publish/src/commands/deploy/deploy-core.js
justinjmoses: Yep, going to update 120 with this change.

PR #1594, opened 2021-11-05
Futures liquidation fee update

artdgn:
Updates the futures liquidation related logic for the first two items of https://github.com/synthetixio/issues/issues/352:
1. Proportional liquidation fee - to incentivise liquidations of larger positions first. Tentatively set to 35 bps.
2. Proportional liquidation margin buffer - to prevent negative margin liquidation, and to allow proportional fee to be paid to the fee pool. Tentatively set at 25 bps. If this buffer will not be needed in practice for safety (or for fees), the value can be set to 0 in the future.

These two values now allow controlling the proportional liquidation margin, and the allocation of that margin upon liquidation to the liquidator and fee pool. 

Other important changes in this PR:
- Adding the parameters to `FuturesMarketSettings`: `liquidationFeeBPs` and `liquidationBufferBPs`
- Adding new views to `FuturesMarket`: `liquidaionFee` and `liquidationMargin`
- Changing the logic of `_liquidationPrice` to calculate unrecorded funding using the current price.
- Changing the event and debt adjustment during `_liquidatePosition` to use the current price instead of the `liquidationPrice` (which may be higher or lower than actual price at that time, depending on the parameters, buffer, and oracle price).
- Sending any positive remainder of the margin to Fee Pool after deducting the `liquidationFee` (that goes to liquidator).

### DIFF
--- a/contracts/FuturesMarket.sol
+++ b/contracts/FuturesMarket.sol
@@ -676,7 +676,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
      * @return lMargin liquidation margin to maintain in sUSD fixed point decimal units
      */
     function liquidationMargin(address account) external view returns (uint lMargin) {
-        require(positions[account].size > 0, "0 size position");
+        require(positions[account].size != 0, "0 size position");
         (uint price, ) = _assetPrice();
         return _liquidationMargin(positions[account].size, price);
     }
--- a/contracts/FuturesMarket.sol
+++ b/contracts/FuturesMarket.sol
@@ -516,7 +516,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
 
         uint uMargin = uint(newMargin);
         int positionSize = position.size;
-        if (positionSize != 0 && uMargin <= _liquidationFee()) {
+        if (positionSize != 0 && uMargin <= _minLiquidationFee()) {
             return (uMargin, Status.CanLiquidate);
         }
 
@@ -587,11 +587,11 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         uint currentPrice
     ) internal view returns (uint) {
         // A position can be liquidated whenever:
-        //     remainingMargin <= liquidationFee
+        //     remainingMargin <= minLiquidationFee
         // Hence, expanding the definition of remainingMargin the exact price
         // at which a position can first be liquidated is:
-        //     margin + profitLoss + funding  =  liquidationFee
-        //     price                          =  lastPrice + (liquidationFee - margin) / positionSize - netFundingPerUnit
+        //     margin + profitLoss + funding  =  minLiquidationFee
+        //     price                          =  lastPrice + (minLiquidationFee - margin) / positionSize - netFundingPerUnit
         // This is straightforward if we neglect the funding term.
 
         int positionSize = position.size;
@@ -601,11 +601,13 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         }
 
         int result =
-            int(position.lastPrice).add(int(_liquidationFee()).sub(int(position.margin)).divideDecimalRound(positionSize));
+            int(position.lastPrice).add(
+                int(_minLiquidationFee()).sub(int(position.margin)).divideDecimalRound(positionSize)
+            );
 
         if (includeFunding) {
             // If we pay attention to funding, we have to expanding netFundingPerUnit and solve again for the price:
-            //     price         =  (lastPrice + (liquidationFee - margin) / positionSize - netAccrued) / (1 + netUnrecorded)
+            //     price         =  (lastPrice + (minLiquidationFee - margin) / positionSize - netAccrued) / (1 + netUnrecorded)
             // Where, if fundingIndex == sequenceLength:
             //     netAccrued    =  fundingSequence[fundingSequenceLength - 1] - fundingSequence[position.fundingIndex]
             //     netUnrecorded =  currentFundingRate * (block.timestamp - fundingLastRecomputed)
@@ -643,7 +645,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
 
     function _canLiquidate(
         Position memory position,
-        uint liquidationFee,
+        uint minLiquidationFee,
         uint fundingIndex,
         uint price
     ) internal view returns (bool) {
@@ -652,7 +654,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
             return false;
         }
 
-        return _remainingMargin(position, fundingIndex, price) <= liquidationFee;
+        return _remainingMargin(position, fundingIndex, price) <= minLiquidationFee;
     }
 
     /*
@@ -660,7 +662,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
      */
     function canLiquidate(address account) external view returns (bool) {
         (uint price, bool invalid) = _assetPrice();
-        return !invalid && _canLiquidate(positions[account], _liquidationFee(), fundingSequence.length, price);
+        return !invalid && _canLiquidate(positions[account], _minLiquidationFee(), fundingSequence.length, price);
     }
 
     function _currentLeverage(
@@ -769,7 +771,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         }
 
         // The order is not submitted if the user's existing position needs to be liquidated.
-        if (_canLiquidate(oldPos, _liquidationFee(), fundingIndex, price)) {
+        if (_canLiquidate(oldPos, _minLiquidationFee(), fundingIndex, price)) {
             return (oldPos, 0, Status.CanLiquidate);
         }
 
@@ -1168,7 +1170,7 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         address liquidator,
         uint fundingIndex,
         uint price,
-        uint liquidationFee
+        uint minLiquidationFee
     ) internal {
         Position storage position = positions[account];
 
@@ -1191,10 +1193,10 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         delete positions[account];
 
         // Issue the reward to the liquidator.
-        _manager().issueSUSD(liquidator, liquidationFee);
+        _manager().issueSUSD(liquidator, minLiquidationFee);
 
         emitPositionModified(positionId, account, 0, 0, 0, price, fundingIndex, 0);
-        emitPositionLiquidated(positionId, account, liquidator, positionSize, lPrice, liquidationFee);
+        emitPositionLiquidated(positionId, account, liquidator, positionSize, lPrice, minLiquidationFee);
     }
 
     /*
@@ -1206,10 +1208,10 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
         uint price = _assetPriceRequireChecks();
         uint fundingIndex = _recomputeFunding(price);
 
-        uint liquidationFee = _liquidationFee();
-        _revertIfError(!_canLiquidate(positions[account], liquidationFee, fundingIndex, price), Status.CannotLiquidate);
+        uint minLiquidationFee = _minLiquidationFee();
+        _revertIfError(!_canLiquidate(positions[account], minLiquidationFee, fundingIndex, price), Status.CannotLiquidate);
 
-        _liquidatePosition(account, messageSender, fundingIndex, price, liquidationFee);
+        _liquidatePosition(account, messageSender, fundingIndex, price, minLiquidationFee);
     }
 
     /* ========== EVENTS ========== */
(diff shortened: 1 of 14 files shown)


contracts/FuturesMarket.sol
```diff
@@ -583,51 +585,102 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+
+        // A position can be liquidated whenever:
+        //     remainingMargin <= liquidationMargin
+        // Hence, expanding the definition of remainingMargin the exact price
+        // at which a position can first be liquidated is:
+        //     margin + profitLoss + funding =  liquidationMargin
+        //     profitLoss = (price - last-price) * positionSize
+        //     price  = lastPrice + (liquidationMargin - margin) / positionSize - netFundingPerUnit
+        int result =
+            int(position.lastPrice).add(int(liqMargin).sub(int(position.margin)).divideDecimalRound(positionSize)).sub(
```
liamzebedee, line 620: This may be a dumb question but I definitely don't claim to be an expert in this area, so please, educate me. Why use the `lastPrice` here when we calculate the liqMargin above using the `currentPrice`?

liamzebedee (changes requested):

contracts/FuturesMarket.sol
artdgn: According to the derivation in the comments above, `lastPrice` is used in order to calculate the profit / loss, which is ` (futurePrice - lastPrice) * positionSize` (which was the price the position was opened / updated at). This is because the formula is trying to solve for the `futurePrice` at which the position will be liquidatable. Using `currentPrice` would give incorrect P&L w.r.t. to position opening price (at which point the "the loan" from the debt pool was taken).

Liquidation margin is calculated using the current price because it's the best estimation we can do currently - because we can't really solve for the actual future liquidation margin because we don't know how funding in the future will affect it (because it depends on 1) time passage and future skew 2) at what times it's being recorded due to other transactions).

Previously liquidation margin was a constant parameter, but making it proportional to price & position introduces this imprecision in solving for liquidation price for the future.

The imprecision is not an issue for the actual liquidation transaction - because then `futurePrice` is `currentPrice`. It's also small when calculated close to the possible liquidation time.

The comments above the code are using trying to give more details about the derivation, but probably can be improved, any idea for what extra details needs to go there?

contracts/FuturesMarket.sol
```diff
@@ -638,12 +689,28 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+     * @param account address of the trader's account
+     * @return fee that will be paid for liquidating the account if it can be liquidated
+     *  in sUSD fixed point decimal units or 0 if account is not liquidatable.
+     */
+    function liquidationFee(address account) external view returns (uint) {
+        (uint price, bool invalid) = _assetPrice();
+        if (!invalid && _canLiquidate(positions[account], fundingSequence.length, price)) {
+            return _liquidationFee(positions[account].size, price);
+        } else {
+            return 0;
```
zyzek, line 708: Instead of returning zero, it is possible to return an estimated fee by returning `_liquidationFee(size, _liquidationPrice(...))` no?

contracts/FuturesMarket.sol
artdgn: I think you've left a similar comment in your review below, will respond there.

contracts/FuturesMarketSettings.sol
```diff
@@ -206,21 +221,33 @@ contract FuturesMarketSettings is Owned, MixinFuturesMarketSettings, IFuturesMar
...
-    function setLiquidationFee(uint _sUSD) external onlyOwner {
+    function setMinLiquidationFee(uint _sUSD) external onlyOwner {
         require(_sUSD <= _minInitialMargin(), "min margin < liquidation fee");
-        _flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_LIQUIDATION_FEE, _sUSD);
-        emit LiquidationFeeUpdated(_sUSD);
+        _flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_MIN_LIQUIDATION_FEE, _sUSD);
+        emit MinLiquidationFeeUpdated(_sUSD);
+    }
+
+    function setLiquidationFeeBPs(uint _bps) external onlyOwner {
```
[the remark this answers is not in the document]
artdgn: Fair point, will see how I can make it more consistent.

contracts/FuturesMarket.sol
zyzek: Oops, somehow managed to duplicate the comment. Sorry about that.

contracts/FuturesMarketSettings.sol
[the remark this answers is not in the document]
artdgn: Do you see any reason for such differences? If we see a reason it should be easy enough to add, otherwise, better to have fewer parameters probably.

contracts/FuturesMarket.sol
```diff
@@ -583,51 +585,100 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+     * @param price price of single baseAsset unit in sUSD fixed point decimal units
+     * @return lFee liquidation fee to be paid to liquidator in sUSD fixed point decimal units
+     */
+    function _liquidationFee(int positionSize, uint price) internal view returns (uint lFee) {
+        // size * price * fee-BPs / 10000
+        // the first multiplication is decimal because price is fixed point decimal, but BPs and 10000 are plain int
+        uint proportionalFee = _abs(positionSize).multiplyDecimalRound(price).mul(_liquidationFeeBPs()).div(10000);
+        uint minFee = _minLiquidationFee();
+        // max(proportionalFee, minFee) - to prevent not incentivising liquidations enough
+        return proportionalFee > minFee ? proportionalFee : minFee;
```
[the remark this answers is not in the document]
artdgn: The problem with that is that the fee is what needs to be `max()`ed so that the keeper isn't underincentivised. And the `_liquidationFee()` is used separately in during liquidation to determine the payment sent to the keeper.

This would mean that if we make the liquidation margin calculation more straight forward like you suggested, it would use a slightly different calculation to the actual fee, and so would be slightly inconsistent (and also require ensuring both get updated correctly in any future changes).

I think both approaches are close enough to not split hairs, but I prefer the current one for the internal consistency, and encapsulation of the liquidationFee calculation in one function.

--------------------

> The minimum initial margin requirement actually accounts properly for the buffer

Not sure about that one, as the minimum initial margin is not really tied to any position size I think, so there's no buffer that can be applied in all cases.

contracts/FuturesMarket.sol
artdgn: It's possible to calculate, but it seemed to me a strange value to return - a fee for a price at which the liquidation cannot be done, so won't ever be paid (because what will be paid - is for a different price, so a different fee). 

More specifically it seemed that this function is what actually should be used by liquidators (because `canLiquidate()` doesn't have enough information to decide how profitable this liquidation is). So it makes sense to have 0 fee for liquidation that cannot happen in the context of a liquidator using this function in a script.

I will add a comment to explain this reasoning in the code.

contracts/FuturesMarketSettings.sol
```diff
@@ -206,21 +221,33 @@ contract FuturesMarketSettings is Owned, MixinFuturesMarketSettings, IFuturesMar
...
+    }
+
+    function setLiquidationBufferBPs(uint _bps) external onlyOwner {
+        _flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_LIQUIDATION_BUFFER_BPS, _bps);
+        emit LiquidationBufferBPsUpdated(_bps);
     }
 
     function setMinInitialMargin(uint _minMargin) external onlyOwner {
-        require(_liquidationFee() <= _minMargin, "min margin < liquidation fee");
+        require(_minLiquidationFee() <= _minMargin, "min margin < liquidation fee");
```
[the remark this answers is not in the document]
artdgn: I think the initial margin enforces minimum margin even in the absence of a position, so the only thing we can check in that case is that it covers the `minLiqudiationFee`, because can't really assume a position size until they try to open it. 

Currently min-fee is set at `20` and min-margin is at `100`.

contracts/FuturesMarket.sol
artdgn: Thanks, will have a more thorough think through these scenarios in terms of debt and post back my conclusions here.

contracts/FuturesMarket.sol
```diff
@@ -583,51 +585,100 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+     * so that stakers would not leak value to liquidators through minting rewards that are not from the
+     * account's margin.
+     * @param positionSize size of position in fixed point decimal baseAsset units
+     * @param price price of single baseAsset unit in sUSD fixed point decimal units
+     * @return lBuffer liquidation buffer to be paid to liquidator in sUSD fixed point decimal units
+     */
+    function _liquidationBuffer(int positionSize, uint price) internal view returns (uint lBuffer) {
+        // size * price * buffer-BPs / 10000
+        // the first multiplication is decimal because price is fixed point decimal, but BPs and 10000 are plain int
+        return _abs(positionSize).multiplyDecimalRound(price).mul(_liquidationBufferBPs()).div(10000);
```
jacko125, line 657: @artdgn What's the reason to div by 10,000 ?

contracts/FuturesMarket.sol
zyzek: I'll add that other than this last point, I'm more or less satisfied that the PR looks pretty solid. As long as the debt considerations are being kept in mind, I'll leave it to full-time contributors for further review and approval (I haven't looked at test cases, config files and so on).

contracts/FuturesMarket.sol
artdgn: Hey, it's because it seemed simplest to set those fees in basis points instead of decimals. This is because they are always discussed and documented in basis points - so it seemed the most natural unit for them instead of decimals. I can make them decimals if it's too confusing.

contracts/FuturesMarket.sol
jacko125: Perhaps a view function of the possible liquidation fees based on position size regardless would be useful for the keeper bots to sort / organise which positions have the most fees if they become liquidatble? 

The price to pass in per unit of base asset would be the position.liquidationPrice ?

contracts/FuturesMarket.sol
```diff
@@ -583,51 +585,98 @@ contract FuturesMarket is Owned, Proxyable, MixinFuturesMarketSettings, IFutures
...
+     * @param positionSize size of position in fixed point decimal baseAsset units
+     * @param price price of single baseAsset unit in sUSD fixed point decimal units
+     * @return lFee liquidation fee to be paid to liquidator in sUSD fixed point decimal units
+     */
+    function _liquidationFee(int positionSize, uint price) internal view returns (uint lFee) {
+        // size * price * fee-ratio
+        uint proportionalFee = _abs(positionSize).multiplyDecimalRound(price).multiplyDecimalRound(_liquidationFeeRatio());
+        uint minFee = _minLiquidationFee();
+        // max(proportionalFee, minFee) - to prevent not incentivising liquidations enough
+        return proportionalFee > minFee ? proportionalFee : minFee; // not using _max() helper because it's for signed ints
```
jacko125, line 641: Feels like the comment here at the end of line isn't needed

jacko125 (approved):
Happy with the changes. What are the default liquidation fee ratio % and buffer % in mind that we will set for deployment ? And then these can be added to sip-80

contracts/FuturesMarket.sol
artdgn: Had a deeper look into the debt calculations:

1. It seems that for the subject of liquidations, the liquidation price actually makes no difference for the debt calculations. This is because the new position is of 0 size and 0 margin, so the price used doesn't matter. So the new position adds no debt and the old position's debt is subtracted at its last price.
```
// in _liquidatePosition
        _applyDebtCorrection(
            Position(0, 0, 0, price, fundingIndex),
            Position(0, position.margin, positionSize, position.lastPrice, position.fundingIndex)
        );

// _positionDebtCorrection

    function _positionDebtCorrection(Position memory position) internal view returns (int) {
        return
            int(position.margin).sub(
                position.size.multiplyDecimalRound(int(position.lastPrice).add(fundingSequence[position.fundingIndex]))
            );
    }

// _applyDebtCorrection

    function _applyDebtCorrection(Position memory newPosition, Position memory oldPosition) internal {
        int newCorrection = _positionDebtCorrection(newPosition);
        int oldCorrection = _positionDebtCorrection(oldPosition);
        _entryDebtCorrection = _entryDebtCorrection.add(newCorrection).sub(oldCorrection);
    }
```

2. Unrelated to the liquidation related debt calculation I don't really understand the role of the funding rate component of the debt correction calculation. It seems to me that the FR shouldn't be included in it. Because for a new position - there is no funding component in the debt added (the position is new), and for a modified position the funding is already counted in the updated margin along with the P&L. So I think it should probably be removed. I'll follow up on that out of band and if needed it will be a separate PR.

contracts/FuturesMarket.sol
artdgn: the comment is there because it was suggested to use `_max()` in the reivew, and without this comment someone might have the same idea in future - and the comment will spare them the trouble :)

artdgn:
> Happy with the changes. What are the default liquidation fee ratio % and buffer % in mind that we will set for deployment ? And then these can be added to sip-80

Yeah the sip will need to be updated, but perhaps worth to wait to make a single update with all the changes in this task.

The values I've set right now is:
- 35 bps for fee incentive (so $35 for each 10K in notional risk).
- 25 bps for buffer. The logic is that if actual liquidation price is on average 0.5% worse than the first liquidation price - the stakers will not be impacted.

PR #1596, opened 2021-11-10
Sip 120 minor issues

artdgn:
This PR addresses several minor issues in https://github.com/Synthetixio/synthetix/pull/1127 contract changes:
- Not using safemath for price and amount calculations in `ExchangeRatesWithDexPricing`
- Allowing setting dex oracle to zero address in `SystemSettings`
- Not checking dex oracle returned price to at least not be 0 in `effectiveAtomicValueAndRates`
- Not having comments about the the dex oracle `IDexPriceAggregator`: impelementation source, refernce interface source.

This PR also fixes a few issues with tests in that PR:
- Tests for `exchangeAtomically` not being correctly set up to trigger volatility checks (no configurtion for the check which causes the check to be skipped comletely). Which makes the test skip an important part of functionality it's testing, and also makes gas measurement inaccureate.
- Using `.sort()` for numeric times incorrectly (using alphabetic sort for numbers) - and so setting up the mock correctly only for some test values. 
- Setting minutes to 3600 seconds in tests instead of 60 seconds, which makes the test use numbers that are too different from the values used in real world config.

### DIFF
--- a/test/contracts/SystemSettings.js
+++ b/test/contracts/SystemSettings.js
@@ -1145,9 +1145,17 @@ contract('SystemSettings', async accounts => {
 				);
 			});
 
+			it('cannot be set to 0 address', async () => {
+				await assert.revert(
+					systemSettings.setAtomicEquivalentForDexPricing(sETH, ZERO_ADDRESS, { from: owner }),
+					'Atomic equivalent is 0 address'
+				);
+			});
+
 			it('allows to be reset', async () => {
-				await systemSettings.setAtomicEquivalentForDexPricing(sETH, ZERO_ADDRESS, { from: owner });
-				assert.equal(await systemSettings.atomicEquivalentForDexPricing(sETH), ZERO_ADDRESS);
+				// using account1 (although it's EOA) for simplicity
+				await systemSettings.setAtomicEquivalentForDexPricing(sETH, account1, { from: owner });
+				assert.equal(await systemSettings.atomicEquivalentForDexPricing(sETH), account1);
 			});
 		});
 	});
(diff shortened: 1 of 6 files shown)


justinjmoses (approved):
This PR is solid. Thank you. :pray:

test/contracts/SystemSettings.js
```diff
@@ -1145,9 +1145,17 @@ contract('SystemSettings', async accounts => {
...
 			});
 
+			it('cannot be set to 0 address', async () => {
+				await assert.revert(
+					systemSettings.setAtomicEquivalentForDexPricing(sETH, ZERO_ADDRESS, { from: owner }),
+					'Atomic equivalent is 0 address'
+				);
+			});
+
 			it('allows to be reset', async () => {
```
sohkai, line 1155: I found this a bit weird, now if you want to turn off atomic exchanges you'll have to set the equivalent to a random address like `0xdead`?

sohkai (commented):
🙏 Thanks, good catches!

test/contracts/SystemSettings.js
artdgn: It's worth to add an explicit way to turn it off through system settings if that was the previous usage of the zero-address. I don't think that setting to zero address or `0xdead` is a good choice in this case - it relies too much on the knowledge and (hope) of this parameter's usage implementation (that it's checked, and the call / result is checked correctly, etc).

sohkai: Fair points. I originally did it this way to avoid adding another system setting to enable/disable which assets are atomically exchangeable since it's unlikely we'll query from a synth-based AMM pool in future.

For clarity, [`ExchangeRatesWithDexPricing`](https://github.com/synthetixio/synthetix/blob/develop/contracts/ExchangeRatesWithDexPricing.sol#L78-L82) does check the parameter, failing price queries if the mapping doesn't exist, and at the moment the only way to disable a synth from being atomically exchanged once the mapping is set would be to reset it to a random address to fail the price query.

PR #1597, opened 2021-11-10
Skip shorts integration tests when cannot open loans

justinjmoses:
This causes fork and integration tests to fail now that [SCCP-146](https://sips.synthetix.io/sccp/sccp-146/) was implemented.


dbeal-eth (approved):
nice

PR #1604, opened 2021-11-22
liquidations update

dbeal-eth:


publish/releases.json
```diff
@@ -363,7 +363,7 @@
 			"sip": 167,
 			"layer": "both",
 			"sources": {
-				"base": ["SystemSettings", "OwnerRelayOnEthereum"],
+				"base": ["SystemSettings", "OwnerRelayOnEthereum", "Issuer"],
 				"ovm": ["OwnerRelayOnOptimism"]
```
barrasso, line 367: The `Issuer` should be added to the `ovm` sources, not `base`.

publish/src/commands/deploy/deploy-core.js
```diff
@@ -201,7 +201,7 @@ module.exports = async ({
 
 	await deployer.deployContract({
 		name: 'Issuer',
-		source: useOvm ? 'IssuerWithoutLiquidations' : 'Issuer',
+		source: useOvm ? 'Issuer',
```
barrasso, line 204: I think we can just remove the `source` line altogether here

barrasso (changes requested):
Couple of minor changes

PR #1607, opened 2021-11-24
fix test lints

dbeal-eth:
sometimes the dual tests fail due to race condition built into the tests

in addition, a better event is now being used for monitoring transaction relay status.

Withdrawal tests still don't work because the transaction is not relayed on the L1 side,
I'm guessing because the user has to call `finalize` transaction somehow. Not sure who to talk to if we want to get those tests working.


test/integration/utils/optimism-temp.js
```diff
@@ -205,7 +223,7 @@ function _printMessengerLog(log) {
 		const argValue = event.args[0];
 		console.log(chalk.gray(`> ${event.name}(${argName}:${argType} = ${argValue})`));
 	} catch (err) {
-		console.error('could not parse messenger log:', log);
+		// console.error('could not parse messenger log:', log);
```
barrasso, line 226: should this be commented out?

barrasso (approved):
LGTM, one minor comment

test/integration/utils/optimism-temp.js
dbeal-eth: https://github.com/Synthetixio/synthetix/pull/1607/commits/eb160b3ebfcb14a5085673dc1b11ec266e73fd30

justinjmoses (approved):

PR #1609, opened 2021-11-25
Futures next price mechanism

artdgn:
Implements https://github.com/Synthetixio/issues/issues/370

Refactors that were needed, but are not exactly part of the mechanism:
- Split futures market into base and next-price-mixin for ease of development / navigation.
- Add new test file for the next-price functionality (instead of adding to previous file).
- Introduce a new struct for `TradeParams` to pass between various methods to reduce signature sizes for internal methods and reduce potential for errors in passing arguments.

Tasks:
- [x] Implement latest mechanism
- [x] Add tests for all scenarios
- [x] Gas measurement & ~~optimisations~~

PR #1610, opened 2021-11-26
Futures remove closure fee and rounding

artdgn:
### Warning: This PR is on top of `futures-next-price` branch (https://github.com/Synthetixio/synthetix/pull/1609) so should be merged only **AFTER** that PR is merged. But can be reviewed already.

Implements https://github.com/Synthetixio/issues/issues/380

Also removes usage of rounding for decimal operations in futures contracts and tests.

The order fee calculation has been simplified to only consider the direction of the trade w.r.t. the skew before the trade. This makes the code simpler, more efficient, and easier to reason about and test. This means that in the edge-case of a trade that flips the skew - the fee will be charged as if it doesn't. This should be negligible both because this will a rare edge case, and because the difference in that case is very small (whether the full fee is all taker, all maker, or a combination).


contracts/FuturesMarketBase.sol
```diff
@@ -247,7 +247,7 @@ contract FuturesMarketBase is Owned, Proxyable, MixinFuturesMarketSettings, IFut
      */
     function _maxOrderSizes(uint price) internal view returns (uint, uint) {
         (uint long, uint short) = _marketSizes();
-        int sizeLimit = int(_maxMarketValueUSD(baseAsset)).divideDecimalRound(int(price));
+        int sizeLimit = int(_maxMarketValueUSD(baseAsset)).divideDecimal(int(price));
```
jacko125, line 250: This would reduce the `sizeLimit` capacity slightly (doesn't round up).

contracts/FuturesMarketBase.sol
```diff
@@ -648,7 +647,7 @@ contract FuturesMarketBase is Owned, Proxyable, MixinFuturesMarketSettings, IFut
      */
     function _liquidationFee(int positionSize, uint price) internal view returns (uint lFee) {
         // size * price * fee-ratio
-        uint proportionalFee = _abs(positionSize).multiplyDecimalRound(price).multiplyDecimalRound(_liquidationFeeRatio());
+        uint proportionalFee = _abs(positionSize).multiplyDecimal(price).multiplyDecimal(_liquidationFeeRatio());
```
jacko125, line 650: Same comment here it'll truncate instead of rounding the multiply / division

jacko125 (approved):
LGTM, only comment is no rounding in certain calculations for the market limits and liquidation calcs, accessible margin

PR #1612, opened 2021-11-29
use new `hardhat-interact` package instead of builtin

dbeal-eth:
the `interact` hardhat task has been refactored and moved
into its own package, `hardhat-interact`. This PR removes the old
code from the repository and uses the new package.

see the code for the new interact command here:
https://github.com/dbeal-eth/hardhat-interact

`synthetix` uses a custom system for managing deployments of its
contracts. To account for this, the `interact` command has been
extended to first regenerate the deployment manifests into a compliant
format.

Since `hardhat-interact` relies on hardhat configuration for network
and signer configuration, the `hardhat.config.js` file has been updated
to include our networks.

notes:
* `hardhat-interact` should be usable for both v2 and v3 synthetix @ajsantander want to implement on v3 repo?
* package-lock.json verification may fail because of new NPM 8 format.
how do we work around this?
* `hardhat-interact` was recently refactored and might have some bugs. Still need to write a couple tests for it as well over the coming week.


barrasso (approved):
💯 🔥

hardhat.config.js
```diff
@@ -35,6 +37,9 @@ module.exports = {
 		tests: './test/contracts',
 		artifacts: path.join(BUILD_FOLDER, 'artifacts'),
 		cache: path.join(BUILD_FOLDER, CACHE_FOLDER),
+
+		// required for interact
+		deployments: './deployments',
```
justinjmoses, line 42: shouldn't this be `publish/deployed`?

dbeal-eth: `hardhat-interact` expects a different, standardized format for the deployment artifacts, so this is the directory of the converted format

hardhat.config.js
```diff
@@ -54,6 +59,22 @@ module.exports = {
 			blockGasLimit: 12e6,
 			url: 'http://localhost:8545',
 		},
+		mainnet: {
```
justinjmoses, line 62: we should probably update the `loadConnections` util in `publish/src/util` to use these providers and chain ids.

justinjmoses (approved):

hardhat.config.js
dbeal-eth: maybe save that for another PR? I'm concerned this will impact existing functionality in subtle ways and its not really necessary to do it right this second

justinjmoses: I do follow a *now or never* approach to tech debt as often times these things are just missed. I do think we have enough test coverage to fix it now without anything crazy happening.

Tuneabq (approved):

Tuneabq (approved):

hardhat.config.js
dbeal-eth: after conversation we have determined that the interface for `loadConnections` is not well suited to be converted right now and so this change will not be completed in this PR.

PR #1614, opened 2021-12-01
reduce storage variables sizes for gas savings

artdgn:
This reduces the stored variables sizes for some gas savings. 

Main savings are for the most expensive methods - `modifyPosition` and `executeNextPriceOrder` of around 80K gas:
- `modifyPosition` : 420K gas -> 340K gas
- `executeNextPriceOrder`: 470K gas -> 390K gas

On OVM this will translate to minor savings at this time because L2 gas is not the main cost, but given possible future added complexity and rising prices, it's still wise to take advantage of the easy savings. 

In terms of savings on calldata (for the L1 part of cost) - it's already minimal with most methods accepting at most one argument.

-----------------
Full tables (third numeric:

### Before
```
|  Contract                     ·  Method                            ·  Min         ·  Max        ·  Avg              ·  # calls      ·  usd (avg)  │
|  TestableFuturesMarket        ·  transferMargin                    ·      135898  ·     314024  ·           296034  ·          307  ·          -  │
|  TestableFuturesMarket        ·  withdrawAllMargin                 ·      173682  ·     277950  ·           245639  ·           19  ·          -  │
|  TestableFuturesMarket        ·  closePosition                     ·      303863  ·     330149  ·           313949  ·           45  ·          -  │
|  TestableFuturesMarket        ·  liquidatePosition                 ·      258390  ·     323471  ·           283890  ·           15  ·          -  │
|  TestableFuturesMarket        ·  modifyPosition                    ·      304164  ·     453436  ·           422787  ·          256  ·          -  │
|  TestableFuturesMarket        ·  submitNextPriceOrder              ·      327384  ·     368257  ·           336695  ·           35  ·          -  │
|  TestableFuturesMarket        ·  executeNextPriceOrder             ·      439018  ·     506507  ·           471997  ·           16  ·          -  │
|  TestableFuturesMarket        ·  cancelNextPriceOrder              ·      199686  ·     339162  ·           275570  ·           12  ·          -  │
```

### After
```
|  Contract                     ·  Method                            ·  Min         ·  Max        ·  Avg              ·  # calls      ·  usd (avg)  │
|  TestableFuturesMarket        ·  transferMargin                    ·      136421  ·     296422  ·           280604  ·          307  ·      26.56  │
|  TestableFuturesMarket        ·  withdrawAllMargin                 ·      167044  ·     273247  ·           241681  ·           19  ·      22.88  │
|  TestableFuturesMarket        ·  closePosition                     ·      292804  ·     312440  ·           300439  ·           45  ·      28.44  │
|  TestableFuturesMarket        ·  liquidatePosition                 ·      234595  ·     305572  ·           269646  ·           15  ·      25.53  │
|  TestableFuturesMarket        ·  modifyPosition                    ·      304005  ·     364752  ·           344986  ·          256  ·      32.66  │
|  TestableFuturesMarket        ·  submitNextPriceOrder              ·      277397  ·     299843  ·           283778  ·           35  ·      26.86  │
|  TestableFuturesMarket        ·  executeNextPriceOrder             ·      368287  ·     418676  ·           392716  ·           16  ·      37.18  │
|  TestableFuturesMarket        ·  cancelNextPriceOrder              ·      200720  ·     319312  ·           270256  ·           12  ·      25.58  │
```

### DIFF
--- a/contracts/FuturesMarketBase.sol
+++ b/contracts/FuturesMarketBase.sol
@@ -98,13 +98,13 @@ contract FuturesMarketBase is Owned, Proxyable, MixinFuturesMarketSettings, IFut
     bytes32 public baseAsset;
 
     // The total number of base units in long and short positions.
-    uint public marketSize;
+    uint128 public marketSize;
 
     /*
      * The net position in base units of the whole market.
      * When this is positive, longs outweigh shorts. When it is negative, shorts outweigh longs.
      */
-    int public marketSkew;
+    int128 public marketSkew;
 
     /*
      * The funding sequence allows constant-time calculation of the funding owed to a given position.
@@ -115,8 +115,8 @@ contract FuturesMarketBase is Owned, Proxyable, MixinFuturesMarketSettings, IFut
      * the net funding on a given position, obtain from this sequence the net funding per base unit
      * since the position was confirmed and multiply it by the position size.
      */
-    uint public fundingLastRecomputed;
-    int[] public fundingSequence;
+    uint32 public fundingLastRecomputed;
+    int128[] public fundingSequence;
 
     /*
      * Each user's position. Multiple positions can always be merged, so each user has
@@ -129,10 +129,10 @@ contract FuturesMarketBase is Owned, Proxyable, MixinFuturesMarketSettings, IFut
      * Then marketSkew * (_assetPrice() + _nextFundingEntry()) + _entryDebtCorrection yields the total system debt,
      * which is equivalent to the sum of remaining margins in all positions.
      */
-    int internal _entryDebtCorrection;
+    int128 internal _entryDebtCorrection;
 
     // This increments for each position; zero reflects a position that does not exist.
-    uint internal _nextPositionId = 1;
+    uint64 internal _nextPositionId = 1;
 
     // Holds the revert message for each type of error.
     mapping(uint8 => string) internal _errorMessages;
@@ -271,7 +271,7 @@ contract FuturesMarketBase is Owned, Proxyable, MixinFuturesMarketSettings, IFut
     function _marketDebt(uint price) internal view returns (uint) {
         // see comment explaining this calculation in _positionDebtCorrection()
         int totalDebt =
-            marketSkew.multiplyDecimal(int(price).add(_nextFundingEntry(fundingSequence.length, price))).add(
+            int(marketSkew).multiplyDecimal(int(price).add(_nextFundingEntry(fundingSequence.length, price))).add(
                 _entryDebtCorrection
             );
 
@@ -302,7 +302,7 @@ contract FuturesMarketBase is Owned, Proxyable, MixinFuturesMarketSettings, IFut
             return 0;
         }
 
-        return marketSkew.divideDecimal(int(skewScaleBaseAsset));
+        return int(marketSkew).divideDecimal(int(skewScaleBaseAsset));
     }
 
     /*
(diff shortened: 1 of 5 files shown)


contracts/FuturesMarketBase.sol
```diff
@@ -98,13 +98,13 @@ contract FuturesMarketBase is Owned, Proxyable, MixinFuturesMarketSettings, IFut
     bytes32 public baseAsset;
 
     // The total number of base units in long and short positions.
-    uint public marketSize;
+    uint128 public marketSize;
```
jacko125, line 101: These would be better left as uint256

contracts/FuturesMarketBase.sol
```diff
@@ -98,13 +98,13 @@ contract FuturesMarketBase is Owned, Proxyable, MixinFuturesMarketSettings, IFut
...
 
     /*
      * The net position in base units of the whole market.
      * When this is positive, longs outweigh shorts. When it is negative, shorts outweigh longs.
      */
-    int public marketSkew;
+    int128 public marketSkew;
```
jacko125, line 107: These would be better left as int256 then you dont need to cast

contracts/FuturesMarketBase.sol
```diff
@@ -129,10 +129,10 @@ contract FuturesMarketBase is Owned, Proxyable, MixinFuturesMarketSettings, IFut
      * Then marketSkew * (_assetPrice() + _nextFundingEntry()) + _entryDebtCorrection yields the total system debt,
      * which is equivalent to the sum of remaining margins in all positions.
      */
-    int internal _entryDebtCorrection;
+    int128 internal _entryDebtCorrection;
```
jacko125, line 132: How much gas does this save? Feels like not worth the casting required for this change.

contracts/FuturesMarketBase.sol
artdgn: Should be pretty safe given that:
1. Uniswap V2 and V3 use 112 bits and 128 bits for token amounts storage:
  - v2 112 bits: https://github.com/Uniswap/v2-core/blob/4dd59067c76dea4a0e8e4bfdda41877a6b16dedc/contracts/UniswapV2Pair.sol#L22
  - v3 128 bits: https://github.com/Uniswap/v3-core/blob/62d65bf88c4fb23671104c28f5bcae566274cb15/contracts/UniswapV3Pool.sol#L54
2. We're unlikely to create perp markets for tokens that even Uniswap doesn't support.
3. SHIB and other meme tokens (which really stretch the token amounts) have 10**15 total supply, which with 18 decimal points need roughly 110 bits (which is likely because of the 112 bits uni limit)

contracts/FuturesMarketBase.sol
artdgn: Same reply as for market size above.

contracts/FuturesMarketBase.sol
artdgn: The casting is only used once (the line that stores the variable).

The savings are probably worth it:
- 4K gas for `modifyPosition` and `executeNextPriceOrder`
- 10K gas for `transferMargin`

contracts/FuturesMarketBase.sol
jacko125: sounds good if the convention is used in other major DeFi protocols.

jacko125 (approved):

PR #1615, opened 2021-12-01
refactor views into mixin to reduce clutter

artdgn:
Minor refactor to reduce clutter in main futures contract.


jacko125 (approved):
lgtm splitting out the view functions into MixinFuturesViews

PR #1616, opened 2021-12-01
Fix flaky unit test

barrasso:
It seems like taking a debt snapshot doesn't work properly when coupled with `before`, so we moved the `takeDebtSnapshot` invocation to a `beforeEach` block.

Attempt to fix:
image


dbeal-eth (approved):
as long as the tests are working better, though tbh I'm not sure this actually solves the problem

PR #1617, opened 2021-12-02
fix duplicate fee reporting on WrapperFactory

dbeal-eth:
`WrapperFactory` is submitting a second `recordFeePaid` call. We need to fund the fee pool the missing $257k, and update the fee pool afterwards to ensure correct accounting.

The error was covered by unit tests prior to the discovery of this bug, but it lead to diverging functionality on-chain because the Synth used for `sUSD` is a `MockSynth`, not the actual `Synth`. This lead to coming to the wrong conclusion that having the additional `recordFeePaid()` was actually necessary.

To correct the unit tests, `MockSynth` was updated to match the functieonality of the `Synth` class with regards to the Fee period record fee paid.

This bug affects both L1 and L2, but only L2 will require funding or fee pool correction.

### DIFF
--- a/contracts/test-helpers/MockSynth.sol
+++ b/contracts/test-helpers/MockSynth.sol
@@ -2,12 +2,17 @@ pragma solidity ^0.5.16;
 
 import "../ExternStateToken.sol";
 import "../interfaces/ISystemStatus.sol";
+import "../interfaces/IAddressResolver.sol";
+import "../interfaces/IFeePool.sol";
 
 // Mock synth that also adheres to system status
 
 contract MockSynth is ExternStateToken {
-    ISystemStatus private systemStatus;
+    IAddressResolver private addressResolver;
     bytes32 public currencyKey;
+
+    // Where fees are pooled in sUSD
+    address public constant FEE_ADDRESS = 0xfeEFEEfeefEeFeefEEFEEfEeFeefEEFeeFEEFEeF;
 
     constructor(
         address payable _proxy,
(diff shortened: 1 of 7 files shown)


test/contracts/WrapperFactory.js
```diff
@@ -206,6 +207,16 @@ contract('WrapperFactory', async accounts => {
 					.find(({ name }) => name === 'Transfer'),
 			});
 		});
+		
+		it('records fee paid', async () => {
```
dbeal-eth, line 211: this test is not strictly necessary because its covered in `FeePool` test, but added here for completeness

dbeal-eth:
@justinjmoses FYI, the unit tests are passing on my local machine, not sure why we are getting that one test failure

contracts/test-helpers/MockSynth.sol
```diff
@@ -21,18 +26,49 @@ contract MockSynth is ExternStateToken {
...
+            // for now, do nothing
+        }
+
+        // Notify feePool to record sUSD to distribute as fees
+        IFeePool(addressResolver.getAddress("FeePool")).recordFeePaid(amountInUSD);
+
+        return true;
+    }
+
     function transfer(address to, uint value) external optionalProxy returns (bool) {
```
justinjmoses, line 59: As discussed, I'm ok with this for an emergency fix, but we should upgrade `MockSynth` to extend from `Synth` and not dupe this logic - duping logic from code in tests is basically how we ended up here in the first place!

PR #1621, opened 2021-12-07
SIP-194 Fix Liquidations on L2

justinjmoses:
The `liquidateDelinquentAccount` logic from `Issuer.sol` was indeed added to OVM during the [Menkalinan release](https://blog.synthetix.io/the-menkalinan-release/) however it can only be invoked via `Synthetix` which was not updated during that release.

This addresses that by moving the logic to the base class which both layers use.


barrasso (approved):
LGTM

barrasso:
https://www.notion.so/Alhena-v2-56-e8d0504dee1c4e85b59365e045ecfad4

PR #1627, opened 2021-12-08
Feat(SystemSettings): SIP-193 Reduce size for SystemSettings

leckylao:
contract size has been reduced to 64%

Note: As events been moved from `SystemSettings` to `SystemSettingsLib`. All existing js code getting events from `SystemSettings` has impacted and needed to be updated. 
Ref: [test/contracts/SystemSettings.js](https://github.com/Synthetixio/synthetix/pull/1627/files?authenticity_token=K%2BwqkQ%2FKcZufVlKcJBqCwO929hEINJ1PRPvp%2Bc6l9Au0%2BPEvTG9q1f44iR7jCbAfKCCtLGIEMDS%2FUSPOTRIVUg%3D%3D&file-filters%5B%5D=.js#diff-b2ffd43db74622f1c3ae5ae10518bee563b0a6d3d9a77e0c3be2a8f5a0fb5e8e)


contracts/SystemSettingsLib.sol
```diff
@@ -0,0 +1,53 @@
...
+    // }
+
+    function setCrossDomainMessageGasLimit(
+        address flexibleStorage,
+        bytes32 settingContractName,
+        bytes32 gasLimitSettings,
+        uint crossDomainMessageGasLimit,
+        uint minCrossDomainGasLimit,
+        uint maxCrossDomainGasLimit
+    ) public {
```
justinjmoses, line 34: not `external`?

contracts/SystemSettingsLib.sol
```diff
@@ -0,0 +1,53 @@
...
+        require(
+            crossDomainMessageGasLimit >= minCrossDomainGasLimit && crossDomainMessageGasLimit <= maxCrossDomainGasLimit,
+            "Out of range xDomain gasLimit"
+        );
+        setUIntValue(flexibleStorage, settingContractName, gasLimitSettings, crossDomainMessageGasLimit);
+    }
+
+    function setIssuanceRatio(
```
justinjmoses, line 42: can you please ignore the re-entrancy errors in slither in this file?

contracts/MixinSystemSettings.sol
```diff
@@ -34,8 +34,6 @@ contract MixinSystemSettings is MixinResolver {
     bytes32 internal constant SETTING_WRAPPER_MAX_TOKEN_AMOUNT = "wrapperMaxTokens";
     bytes32 internal constant SETTING_WRAPPER_MINT_FEE_RATE = "wrapperMintFeeRate";
     bytes32 internal constant SETTING_WRAPPER_BURN_FEE_RATE = "wrapperBurnFeeRate";
-    bytes32 internal constant SETTING_MIN_CRATIO = "minCratio";
-    bytes32 internal constant SETTING_NEW_COLLATERAL_MANAGER = "newCollateralManager";
```
barrasso, line 38: Safe to remove these 👍 Following SIP-135: `minCratio` and `collateralManager` are now set in `Collateral.sol`.

contracts/SystemSettings.sol
```diff
@@ -244,338 +231,319 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
...
-
-        flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_FEE_PERIOD_DURATION, _feePeriodDuration);
-
-        emit FeePeriodDurationUpdated(_feePeriodDuration);
+        SystemSettingsLib.setFeePeriodDuration(
+            address(flexibleStorage()),
+            SETTING_CONTRACT_NAME,
+            SETTING_FEE_PERIOD_DURATION,
+            _feePeriodDuration,
+            MIN_FEE_PERIOD_DURATION,
```
justinjmoses, line 288: these are only used in `SystemSettingsLib` so they could be moved over there instead of passed in as args

contracts/SystemSettings.sol
```diff
@@ -244,338 +231,319 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
...
-            "Out of range xDomain gasLimit"
-        );
-        flexibleStorage().setUIntValue(
+        SystemSettingsLib.setCrossDomainMessageGasLimit(
+            address(flexibleStorage()),
             SETTING_CONTRACT_NAME,
             _getGasLimitSetting(_gasLimitType),
-            _crossDomainMessageGasLimit
+            _crossDomainMessageGasLimit,
+            MIN_CROSS_DOMAIN_GAS_LIMIT,
```
justinjmoses, line 239: these limit definitions can be moved into the lib

contracts/SystemSettings.sol
```diff
@@ -244,338 +231,319 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
...
         );
         emit CrossDomainMessageGasLimitChanged(_gasLimitType, _crossDomainMessageGasLimit);
     }
 
     function setTradingRewardsEnabled(bool _tradingRewardsEnabled) external onlyOwner {
-        flexibleStorage().setBoolValue(SETTING_CONTRACT_NAME, SETTING_TRADING_REWARDS_ENABLED, _tradingRewardsEnabled);
-        emit TradingRewardsEnabled(_tradingRewardsEnabled);
+        SystemSettingsLib.setTradingRewardsEnabled(
+            address(flexibleStorage()),
+            SETTING_CONTRACT_NAME,
```
justinjmoses, line 248: i say redefine this `SETTING_CONTRACT_NAME` in the lib to avoid passing it each and every time

test/contracts/SystemSettings.js
```diff
@@ -175,7 +177,17 @@ contract('SystemSettings', async accounts => {
 			const txn = await systemSettings.setWaitingPeriodSecs(newPeriod, { from: owner });
 			const actual = await systemSettings.waitingPeriodSecs();
 			assert.equal(actual, newPeriod, 'Configured waiting period is set correctly');
-			assert.eventEqual(txn, 'WaitingPeriodSecsUpdated', [newPeriod]);
+			const encodeEventSignature = AbiCoder.encodeEventSignature({
```
justinjmoses, line 180: This can be done without an extra package in `package.json`: see https://github.com/Synthetixio/synthetix/blob/develop/test/etherscan/spec.js#L113

justinjmoses (changes requested):
A few minor changes but overall looking good!

test/contracts/SystemSettings.js
leckylao: > This can be done without an extra package in `package.json`: see https://github.com/Synthetixio/synthetix/blob/develop/test/etherscan/spec.js#L113

I had thought of that initially but the problem is need to set up the web3 instance which is too verbose for this case. And the `web3-eth-abi` is actually an existing package on other dependancy https://raw.githubusercontent.com/Synthetixio/synthetix/develop/package-lock.json. So it's not a new package and putting it in will let the git hook pass on the require missing package.

leckylao:
After moving public constant it reduced size more down from 64% to 61%
image

contracts/MixinSystemSettings.sol
```diff
@@ -180,22 +178,6 @@ contract MixinSystemSettings is MixinResolver {
...
-
-    function getNewCollateralManager(address collateral) internal view returns (address) {
-        return
-            flexibleStorage().getAddressValue(
-                SETTING_CONTRACT_NAME,
-                keccak256(abi.encodePacked(SETTING_NEW_COLLATERAL_MANAGER, collateral))
-            );
-    }
-
     function getInteractionDelay(address collateral) internal view returns (uint) {
```
artdgn, line 181: What the reasoning for separating the setters from the getters, especially if they rely on this intimate knowledge of the storage mapping schema? 

Shouldn't the getters go to the lib as well? If the intention to split them is for contract size savings, than to me it makes more sense to split such that each setter is near it's getter (so e.g. some values are in the lib, and some are here, or have two libs or something). 

Maybe a "core" settings contract that also uses an "ExtraSettingsLib" that houses the extras? It seems that this getters and setters split makes things less readable, and breaks encapsulation of implementation details too much.

This would also reduce the long method signatures and passing of constants.

artdgn (changes requested):
Some readability and maintainability concerns with the contract separation and testing code.

contracts/MixinSystemSettings.sol
leckylao: Didn't touch much on `MixinSystemSettings.sol`, mainly removed the non-using functions. And the getters are mainly one line so not much can reduce.

test/contracts/setup.js
```diff
@@ -124,7 +124,15 @@ const setupContract = async ({
 
 	// if it needs library linking
 	if (Object.keys((await artifacts.readArtifact(contract)).linkReferences).length > 0) {
-		await artifact.link(await artifacts.require('SafeDecimalMath').new());
+		if (/^SystemSettings$/.test(artifact._json.contractName)) {
```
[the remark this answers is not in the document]
leckylao: Not sure how it could be refactored outside of the if. And it should be self explanatory that special linking library when SystemSettings otherwise normal linking. 

> Also, isn't this regex equivalent to just a comparison (===) ?
I think so, would leave it for now.

test/integration/behaviors/stake.behavior.js
```diff
@@ -7,7 +7,7 @@ const { skipFeePeriod, skipMinimumStakeTime } = require('../utils/skip');
 
 function itCanStake({ ctx }) {
 	describe('staking and claiming', () => {
-		const SNXAmount = ethers.utils.parseEther('100');
+		const SNXAmount = ethers.utils.parseEther('1000');
```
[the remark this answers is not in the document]
leckylao: It fixes this one. 
https://app.circleci.com/pipelines/github/Synthetixio/synthetix/9775/workflows/f2026f37-4e7f-4a23-9127-b1d31f20adac/jobs/101109?invite=true#step-109-1867

contracts/SystemSettings.sol
```diff
@@ -258,324 +200,286 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
...
+            SETTING_TRADING_REWARDS_ENABLED,
+            _tradingRewardsEnabled
+        );
     }
 
     function setWaitingPeriodSecs(uint _waitingPeriodSecs) external onlyOwner {
-        flexibleStorage().setUIntValue(SETTING_CONTRACT_NAME, SETTING_WAITING_PERIOD_SECS, _waitingPeriodSecs);
-        emit WaitingPeriodSecsUpdated(_waitingPeriodSecs);
+        SystemSettingsLib.setWaitingPeriodSecs(
```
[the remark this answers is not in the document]
leckylao: > why not use the usual library syntax by adding `using SystemSettingsLib for IFlexibleStorage` and then calling `flexibleStorage().setWaitingPeriodSecs(...)` instead?

Hmmm, the SystemSettingsLib looks more clear to me than `flexibleStorage().setWaitingPeriodSecs(...)`

> And also maybe avoid the casting to address here and to the interface the lib..

Example pls?

contracts/SystemSettingsLib.sol
```diff
@@ -0,0 +1,567 @@
...
+                crossDomainMessageGasLimit <= MAX_CROSS_DOMAIN_GAS_LIMIT,
+            "Out of range xDomain gasLimit"
+        );
+        setUIntValue(flexibleStorage, settingContractName, gasLimitSettings, crossDomainMessageGasLimit);
+    }
+
+    function setIssuanceRatio(
+        address flexibleStorage,
+        bytes32 settingContractName,
+        bytes32 settingName,
```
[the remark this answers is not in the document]
leckylao: The `settingName` is originally defined at `MixinSystemSettings.sol` and using there for the getters. So passing it in from there to make single source of variable definition.

artdgn (dismissed):

artdgn:
Contract sizes:
```
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤
│ SystemSettingsLib               │ 13158  │ 12.85 KB  │ 53.54%           │          │
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤
...
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤
│ SystemSettings                  │ 14440  │ 14.1 KB   │ 58.76%           │          │
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤
```

contracts/MixinSystemSettings.sol
justinjmoses: This is so contracts can mix in the getting of these without having to go externally to the `SystemSettings` contract - they can go directly to `FlexibleStorage`. (So, gas savings)

contracts/SystemSettings.sol
```diff
@@ -5,62 +5,28 @@ import "./Owned.sol";
...
-    // Minimum Stake time may not exceed 1 weeks.
-    uint public constant MAX_MINIMUM_STAKE_TIME = 1 weeks;
-
-    uint public constant MAX_CROSS_DOMAIN_GAS_LIMIT = 8e6;
-    uint public constant MIN_CROSS_DOMAIN_GAS_LIMIT = 3e6;
-
-    int public constant MAX_WRAPPER_MINT_FEE_RATE = 1e18;
-    int public constant MAX_WRAPPER_BURN_FEE_RATE = 1e18;
+    // SystemSettingsLib is a way to split out the setters to reduce contract size
+    using SystemSettingsLib for IFlexibleStorage;
```
justinjmoses, line 13: This is quite a cute solution to how to invoke this

contracts/SystemSettings.sol
```diff
@@ -244,338 +202,161 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
...
-                _crossDomainMessageGasLimit <= MAX_CROSS_DOMAIN_GAS_LIMIT,
-            "Out of range xDomain gasLimit"
-        );
-        flexibleStorage().setUIntValue(
-            SETTING_CONTRACT_NAME,
-            _getGasLimitSetting(_gasLimitType),
-            _crossDomainMessageGasLimit
-        );
+        flexibleStorage().setCrossDomainMessageGasLimit(_getGasLimitSetting(_gasLimitType), _crossDomainMessageGasLimit);
         emit CrossDomainMessageGasLimitChanged(_gasLimitType, _crossDomainMessageGasLimit);
```
justinjmoses, line 206: why not move this event as well?

contracts/SystemSettingsLib.sol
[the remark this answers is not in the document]
justinjmoses: We could technically have this library extend from `MixinSystemSettings` - and then not need to pass through the keys...

contracts/SystemSettingsLib.sol
```diff
@@ -0,0 +1,511 @@
...
+    }
+
+    function setIssuanceRatio(
+        IFlexibleStorage flexibleStorage,
+        bytes32 settingName,
+        uint _issuanceRatio
+    ) external {
+        require(_issuanceRatio <= MAX_ISSUANCE_RATIO, "New issuance ratio cannot exceed MAX_ISSUANCE_RATIO");
+        flexibleStorage.setUIntValue(SETTINGS_CONTRACT_NAME, settingName, _issuanceRatio);
+        // slither-disable-next-line reentrancy-events
```
justinjmoses, line 83: Want to do this for the whole file instead?

contracts/SystemSettingsLib.sol
```diff
@@ -0,0 +1,511 @@
...
+                crossDomainMessageGasLimit <= MAX_CROSS_DOMAIN_GAS_LIMIT,
+            "Out of range xDomain gasLimit"
+        );
+        flexibleStorage.setUIntValue(SETTINGS_CONTRACT_NAME, gasLimitSettings, crossDomainMessageGasLimit);
+    }
+
+    function setIssuanceRatio(
+        IFlexibleStorage flexibleStorage,
+        bytes32 settingName,
+        uint _issuanceRatio
```
justinjmoses, line 79: can you fix these slither complaints by removing the leading underscore?

package.json
```diff
@@ -139,7 +139,8 @@
 		"js-levenshtein": "^1.1.6",
 		"pretty-error": "^2.1.1",
 		"solidity-parser-antlr": "^0.4.11",
-		"web3-utils": "1.2.2"
+		"web3-utils": "1.2.2",
+		"web3-eth-abi": "1.6.0"
```
justinjmoses, line 143: why is this needed? you can grab it as we do here: https://github.com/Synthetixio/synthetix/blob/feat/reduce-system-settings-size/test/etherscan/spec.js#L102 - please remove this and update SystemSettings.js to use the aforementioned technique

test/contracts/setup.js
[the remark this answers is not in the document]
justinjmoses: So he means do the SafeDecimalMath linking to `artifact` regardless and then if SystemSettings, then also link `SystemSettingsLib`. I believe that would work the same way but not 100% sure...

justinjmoses (changes requested):
A few changes here and there to address, nothing major

contracts/SystemSettings.sol
artdgn: looks like the reason is the usage of the enum `CrossDomainMessageGasLimits` in the sig, which is used in the getter mixin

contracts/SystemSettingsLib.sol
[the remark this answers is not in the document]
artdgn: I've tried something similar and libraries can't inherit (at least in sol 0.5) from contracts of libraries. 

They can use (be linked), and than we can have a SettingsGettersLib and SettingsSettersLib (that uses the getters one), but it all seems over complicated at that point and the correct and simple solution IMO is to just reduce the contract size by splitting contracts into settings by domain (a bit similar to how it's done with e.g. futures settings). But that would be a very big refactor with questionable payoff at this point.

contracts/SystemSettingsLib.sol
artdgn: Slither doesn't really support that, we can exclude the file or the detector entirely. I think it's pretty safe to disable `reentrancy-events` detector entirely as it only relates to events https://github.com/crytic/slither/wiki/Detector-Documentation#reentrancy-vulnerabilities-3 and the usual pattern is to emit at the end of a method, which almost always triggers the detector.

package.json
artdgn: I've moved the events back into contract from the library, the contract size is 68%, but testing the events is much more readable and simple now

artdgn:
Contract sizes:
```
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤
│ SystemSettingsLib               │ 8189   │ 8 KB      │ 33.32%           │          │
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤

..
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤
│ SystemSettings                  │ 16967  │ 16.57 KB  │ 69.04%           │          │
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤

```

artdgn:
If we need to further reduce size later we can bundle all the atomic params getter into a single getter that returns a struct and use that instead of the individual getter calls. This will also be more gas efficient due to smaller amount of external calls probably.

justinjmoses (approved):

PR #1630, opened 2021-12-11
SIP-184: Remove currentRound auto incrementation from ExchangeRates

leckylao:
- setting sUSD on constructor;
- moved currentRound auto increment into internalUpdateRates

### DIFF
--- a/contracts/ExchangeRates.sol
+++ b/contracts/ExchangeRates.sol
@@ -22,6 +22,7 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
     using SafeDecimalMath for uint;
 
     bytes32 public constant CONTRACT_NAME = "ExchangeRates";
+    bytes32 public constant SUSD = "sUSD";
 
     // Exchange rates and update times stored by currency code, e.g. 'SNX', or 'sUSD'
     mapping(bytes32 => mapping(uint => RateAndUpdatedTime)) private _rates;
@@ -60,8 +61,8 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
         oracle = _oracle;
 
         // The sUSD rate is always 1 on round 1 and is never stale.
-        currentRoundForRate["sUSD"] = 1;
-        _rates["sUSD"][currentRoundForRate["sUSD"]] = RateAndUpdatedTime({
+        currentRoundForRate[SUSD] = 1;
+        _rates[SUSD][currentRoundForRate[SUSD]] = RateAndUpdatedTime({
             rate: uint216(SafeDecimalMath.unit()),
             time: uint40(now)
         });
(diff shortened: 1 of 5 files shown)


contracts/ExchangeRates.sol
```diff
@@ -436,12 +440,10 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
...
-        if (roundId > 0) {
-            currentRoundForRate[currencyKey] = roundId;
-        } else {
-            // Note: this will effectively start the rounds at 1, which matches Chainlink's Agggregators
-            currentRoundForRate[currencyKey]++;
-        }
+        require(currencyKey != "sUSD", "Rate of sUSD cannot be updated, it's always UNIT.");
+        require(roundId > 0, "Round id must be greater than 0.");
+
+        currentRoundForRate[currencyKey] = roundId;
```
artdgn, line 446: I was under the impression that we've agreed that `currentRoundForRate` should be removed because it introduces more risks, potential for bugs, and too much coupling (between rates and exchanger contract).

artdgn (commented):
I was sure we said that removing the curreRate state variable is the best course of action.

The only place it's used right now is the exchanger for getting the last cached rate for dynamic fee to get the correct set of prices. It can be removed there because  you already have the roundIds in the `_exchange` method that you can pass into the fees method to get the historic prices. 

This way, this whole issue with coupling, and incrementing, etc can be avoided.

contracts/ExchangeRates.sol
```diff
@@ -149,15 +153,15 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
         uint time;
         (sourceRate, time) = _getRateAndTimestampAtRound(sourceCurrencyKey, roundIdForSrc);
         // cacheing to save external call
-        _setRate(sourceCurrencyKey, roundIdForSrc, sourceRate, time);
+        if (sourceCurrencyKey != "sUSD") _setRate(sourceCurrencyKey, roundIdForSrc, sourceRate, time);
```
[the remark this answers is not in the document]
leckylao: It does has the != sUSD check in _setRate, here is when reading the sUSD rate don't try to set it. Will put a comment there.

contracts/ExchangeRates.sol
leckylao: This is the first iteration. And `currentRoundForRate` is more for improvement due to the tightly coupling and timeframe.

contracts/ExchangeRates.sol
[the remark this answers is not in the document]
artdgn: I meant to remove the `require` in _setRate, and instead just return if it's sUSD. 

So in `_setRate` instead of 
```
require(currencyKey != "sUSD", "Rate of sUSD cannot be updated, it's always UNIT.");
``` 
```
if (currencyKey == "sUSD") {
  // no caching is done for sUSD
  return;
}
``` 
This way, there is no need to have those ifs elsewhere and:
- code is easier to read (less ifs)
- whoever is calling setRate doesn't need to check the currency

contracts/Exchanger.sol
```diff
@@ -788,17 +805,15 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
-        if (currencyKey == sUSD) {
-            return 0;
-        }
+        if (currencyKey == sUSD) return 0;
         (uint threshold, uint weightDecay, uint rounds) = getExchangeDynamicFeeData();
         uint[] memory prices;
-        // Note: We are using cache round ID here for cheap read
-        uint currentRoundId = exchangeRates().currentRoundForRate(currencyKey);
-        (prices, ) = exchangeRates().ratesAndUpdatedTimeForCurrencyLastNRounds(currencyKey, rounds, currentRoundId);
+        if (roundId == 0) roundId = exchangeRates().getCurrentRoundId(currencyKey);
```
artdgn, line 815: I don't think this method should try to be smart about the roundId it's getting. It's better to avoid using 0 as a special case if we can. This will reduce coupling, edge cases, and surface area for bugs. This method should just work with the roundId it's getting IMO because it's a low level method.

contracts/Exchanger.sol
```diff
@@ -835,7 +850,7 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
             uint exchangeFeeRate
         )
     {
-        exchangeFeeRate = _feeRateForExchange(sourceCurrencyKey, destinationCurrencyKey);
+        exchangeFeeRate = _feeRateForExchange(sourceCurrencyKey, destinationCurrencyKey, 0, 0);
```
artdgn, line 853: same comment as above.

artdgn (changes requested):
That's great, now that `currentRoundForRate` is not used, I think it should be possible to remove it also from the `ExchangeRates` contract.

justinjmoses:
Closed in favor of #1649

PR #1636, opened 2021-12-16
Sip 196 remove internal oracle

artdgn:
Removes the contract methods for internal oracle, and the testing usage of it.

### DIFF
--- a/contracts/ExchangeRates.sol
+++ b/contracts/ExchangeRates.sol
@@ -22,6 +22,7 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
     using SafeDecimalMath for uint;
 
     bytes32 public constant CONTRACT_NAME = "ExchangeRates";
+    //slither-disable-next-line naming-convention
     bytes32 internal constant sUSD = "sUSD";
 
     // Decentralized oracle networks that feed into pricing aggregators
@@ -360,6 +361,7 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
                 // so let's call it low-level to suppress any reverts
                 bytes memory payload = abi.encodeWithSignature("latestRoundData()");
                 // solhint-disable avoid-low-level-calls
+                // slither-disable-next-line low-level-calls
                 (bool success, bytes memory returnData) = address(aggregator).staticcall(payload);
 
                 if (success) {
@@ -469,6 +471,7 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
     }
 
     function _notImplemented() internal pure {
+        // slither-disable-next-line dead-code
         revert("Cannot be run on this layer");
     }
 
--- a/contracts/ExchangeRates.sol
+++ b/contracts/ExchangeRates.sol
@@ -22,12 +22,7 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
     using SafeDecimalMath for uint;
 
     bytes32 public constant CONTRACT_NAME = "ExchangeRates";
-
-    // Exchange rates and update times stored by currency code, e.g. 'SNX', or 'sUSD'
-    mapping(bytes32 => mapping(uint => RateAndUpdatedTime)) private _rates;
-
-    // The address of the oracle which pushes rate updates to this contract
-    address public oracle;
+    bytes32 internal constant sUSD = "sUSD";
 
     // Decentralized oracle networks that feed into pricing aggregators
     mapping(bytes32 => AggregatorV2V3Interface) public aggregators;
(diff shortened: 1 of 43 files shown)


contracts/ExchangeRates.sol
```diff
@@ -447,60 +350,67 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
...
-                        rate: uint216(_formatAggregatorAnswer(currencyKey, answer)),
-                        time: uint40(updatedAt)
-                    });
-            }
+        // sUSD rate is 1.0
+        if (currencyKey == sUSD) {
+            return
+                RateAndUpdatedTime({
+                    rate: uint216(SafeDecimalMath.unit()),
+                    time: uint40(block.timestamp) // never "stale"
```
justinjmoses, line 358: one small note here is that the timestamp for sUSD will always be `now` instead of the time it was first set at construction. I can't think of anywhere this is an issue, but calling it out as a change.

contracts/ExchangeRates.sol
```diff
@@ -447,60 +350,67 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
...
-            if (success) {
-                (, int256 answer, , uint256 updatedAt, ) =
-                    abi.decode(returnData, (uint80, int256, uint256, uint256, uint80));
-                return (_formatAggregatorAnswer(currencyKey, answer), updatedAt);
-            }
+        // short circuit sUSD
+        if (currencyKey == sUSD) {
+            // sUSD has no rounds, and 0 time is preferrable for "volatility" heuristics
+            // which are used in atomic swaps and fee reclamation
+            return (SafeDecimalMath.unit(), 0);
```
justinjmoses, line 397: you should return `block.timestamp` to be consistent instead of `0`

artdgn: This screws up fee reclamation logic which looks at timestamps recency.

I will return 0 in both places for consistency.

contracts/ExchangeRates.sol
artdgn: Updated it to return 0. Which is still different from what it used to be, but more consistent with the previous value (being in the past and not changing). Also consistent with the other method below.

artdgn:
Updated gas measurements:

```
|  Synthetix                        ·  exchange                                ·      184283  ·     855876  ·           539572  ·          392  ·          -  │
····································|··········································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchangeAtomically                      ·      657641  ·     667903  ·           666560  ·           17  ·          -  │
····································|··········································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchangeOnBehalf                        ·      724943  ·     743529  ·           727266  ·            8  ·          -  │
····································|··········································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchangeOnBehalfWithTracking            ·      731558  ·     750156  ·           733883  ·            8  ·          -  │
····································|··········································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchangeWithTracking                    ·      713964  ·     732547  ·           723256  ·            2  ·          -  │
····································|··········································|··············|·············|···················|···············|··············
|  Synthetix                        ·  exchangeWithVirtual                     ·      920134  ·     926310  ·           925985  ·           38  ·          -  │
····································|··········································|··············|·············|···················|···············|··············
|  Synthetix                        ·  settle                                  ·       67192  ·     398781  ·           166307  ·           34  ·          -  │

```

leckylao (approved):

contracts/ExchangeRates.sol
```diff
@@ -22,12 +22,7 @@ contract ExchangeRates is Owned, MixinSystemSettings, IExchangeRates {
     using SafeDecimalMath for uint;
 
     bytes32 public constant CONTRACT_NAME = "ExchangeRates";
-
-    // Exchange rates and update times stored by currency code, e.g. 'SNX', or 'sUSD'
-    mapping(bytes32 => mapping(uint => RateAndUpdatedTime)) private _rates;
-
-    // The address of the oracle which pushes rate updates to this contract
-    address public oracle;
+    bytes32 internal constant sUSD = "sUSD";
```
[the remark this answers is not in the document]
artdgn: It's used this way in other places in the codebase, and to me seems more readable as `sUSD`. I'll add an ignore comment to silence the warning.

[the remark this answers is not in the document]
justinjmoses: please :pray:

contracts/ExchangeRates.sol
justinjmoses: Thanku

publish/src/commands/deploy/system-and-parameter-check.js
```diff
@@ -127,7 +121,7 @@ module.exports = async ({
 		}
 	}
 
-	for (const address of [account, oracleAddress]) {
+	for (const address of [account]) {
```
justinjmoses, line 124: remove the loop now?

test/contracts/helpers.js
```diff
@@ -11,6 +11,48 @@ const {
...
 } = require('../..');
 
+const MockAggregator = artifacts.require('MockAggregatorV2V3');
+
+/// utility function to setup price aggregators
+/// @param exchangeRates instance of ExchangeRates contract
+/// @param owner owner account of exchangeRates contract for adding an aggregator
+/// @param keys array of bytes32 currency keys
+/// @param decimalsArray optional array of ints for each key, defaults to 18 decimals
+async function setupPriceAggregators(exchangeRates, owner, keys, decimalsArray = []) {
```
justinjmoses, line 21: [np] prefer using named parameters rather than just ordered ones - so `{ exchangeRates, owner, keys, decimalsArray }` - it's pretty much standard practice now these days in JS.

test/contracts/ExchangeRates.js
```diff
@@ -1033,12 +420,26 @@ contract('Exchange Rates', async accounts => {
 					assert.bnEqual(await instance.effectiveValue(SNX, amountOfSynthetixs, sEUR), amountOfEur);
 				});
 
+				// it('should revert when relying on a non-existant dest exchange rate in effectiveValue()', async () => {
```
justinjmoses, line 423: do we need these in comments?

test/contracts/ExchangeRates.js
```diff
@@ -1410,45 +809,33 @@ contract('Exchange Rates', async accounts => {
...
-
-										it('then the rates are invalid again', () => {
-											assert.equal(response[1], true);
-										});
-
-										it('and JPY is 0 while the other is fine', () => {
-											assert.equal(response[0][0], '0');
-											assert.bnEqual(response[0][1], toUnit(newRateXTZ.toString()));
-										});
+									// it('when the ratesAndInvalidForCurrencies is queried it reverts', async () => {
```
justinjmoses, line 812: (ibid)

test/contracts/ExchangeRates.js
```diff
@@ -2990,20 +2016,28 @@ contract('Exchange Rates', async accounts => {
 		});
 	};
 
+	// utility function to setup price aggregators
+	async function setupAggregators(keys, decimalsArray = []) {
```
justinjmoses, line 2020: are these necessary? they seem redundant

test/contracts/setup.js
```diff
@@ -1132,6 +1122,13 @@ const setupAllContracts = async ({
 			.map(mock => mock.setAddressResolver(returnObj['AddressResolver'].address))
 	);
 
+	if (returnObj['ExchangeRates']) {
```
justinjmoses, line 1125: 👍🏼

justinjmoses (approved):
Just a few comments for cleanup and nitpicks

test/contracts/ExchangeRates.js
artdgn: yeah these two are used a lot in this file, so it's nicer to have the shorter versions.

test/contracts/ExchangeRates.js
artdgn: :+1:

test/contracts/helpers.js
artdgn: looked into it, and `keys` being an array makes it a bit awkward to use this way. It seems that `decimalsArray` is never actually used, so I'll remove it, and 3 params should be fine for ordered probably.

artdgn: Also the 4th arg is pretty much never used, so it's mostly used with just the 3 args, which doesn't look too bad.

If with time people start adding args we can make do something like `(arg1, arg2, arg3, args = {arg4: ... } )` to make it less error prone, but still short enough for the usual case

contracts/ExchangeRates.sol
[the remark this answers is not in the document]
leckylao: > please 🙏

Just saw merged with sUSD, I believe @justinjmoses please means please use SUSD not please use ignore 🙈

PR #1637, opened 2021-12-16
Update configuration of loan and system settings

barrasso:
Updated the predicate for changing values only if setting to non-zero from zero.

### DIFF
--- a/publish/src/commands/deploy/configure-loans.js
+++ b/publish/src/commands/deploy/configure-loans.js
@@ -80,13 +80,14 @@ module.exports = async ({
 			comment: 'Ensure the CollateralEth contract has all associated synths added',
 		});
 
+		const issueFeeRate = (await getDeployParameter('COLLATERAL_ETH'))['ISSUE_FEE_RATE'];
 		await runStep({
 			contract: 'CollateralEth',
 			target: CollateralEth,
 			read: 'issueFeeRate',
-			expected: input => input !== '0', // only change if zero,
+			expected: input => issueFeeRate === '0' || input !== '0',
 			write: 'setIssueFeeRate',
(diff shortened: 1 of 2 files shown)


justinjmoses (changes requested):
Please use a reusable predicate function for this rather than duplicating the function logic each time.

Something like 
```javascript
const allowZeroOrUpdateIfNonZero = param => input => param === '0' || input !== '0';

// then use it like
expected: allowZeroOrUpdateIfNonZero(issueFeeRate)
```

barrasso:
> Please use a reusable predicate function for this rather than duplicating the function logic each time.
> 
> Something like
> 
> ```js
> const allowZeroOrUpdateIfNonZero = param => input => param === '0' || input !== '0';
> 
> // then use it like
> expected: allowZeroOrUpdateIfNonZero(issueFeeRate)
> ```

Resolved in https://github.com/Synthetixio/synthetix/pull/1637/commits/f912390fcba9b38f52ea4591a323dfbf1806331f

index.js
```diff
@@ -203,6 +203,8 @@ const defaults = {
 const toBytes32 = key => w3utils.rightPad(w3utils.asciiToHex(key), 64);
 const fromBytes32 = key => w3utils.hexToAscii(key);
 
+const allowZeroOrUpdateIfNonZero = param => input => param === '0' || input !== '0';
```
justinjmoses, line 206: could you move this to `publish/src/util`?

barrasso: resolved https://github.com/Synthetixio/synthetix/pull/1637/commits/e258764340ef567f334461d1126dd6f37b365632

justinjmoses (approved):

PR #1640, opened 2021-12-25
Feat/sip 184 merge develop

leckylao:


leckylao:
After moving public constant it reduced size more down from 64% to 61%
image

PR #1647, opened 2022-01-11
SIP-148: Liquidatooor

barrasso:
Introduces the upgraded liquidation mechanism as described in SIP-148: https://sips.synthetix.io/sips/sip-148/

Tracking Issue: https://github.com/Synthetixio/issues/issues/397

### Release Notes

- The legacy `Liquidations` contract will be deprecated in favor of the new `Liquidator`
- Updated the `ISynthetix` interface to account for the new liquidation mechanism
- Added the `LiquidatorRewards` contract to distribute redeemed SNX amongst stakers

-------------

```js
/* SCCP Values */

LIQUIDATION_DELAY: (3600 * 24).toString(), // 24 hours

LIQUIDATION_RATIO: w3utils.toWei('0.5'), // 200% cratio

LIQUIDATION_PENALTY: w3utils.toWei('0.3'), // 30% penalty

LIQUIDATION_ESCROW_DURATION: (3600 * 24 * 365).toString(), // 1 year

SELF_LIQUIDATION_PENALTY: w3utils.toWei('0.20'), // 20% penalty

FLAG_REWARD: w3utils.toWei('10'), // 10 SNX
FLAG_REWARD (OVM): w3utils.toWei('1'), // 1 SNX


LIQUIDATE_REWARD: w3utils.toWei('20'), // 20 SNX
LIQUIDATE_REWARD (OVM): w3utils.toWei('2'), // 2 SNX
```


contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,182 @@
...
+
+    /* ========== RESTRICTED FUNCTIONS ========== */
+
+    /// @notice This is called only by the Issuer to update the rewards for when a given account's debt balance changes.
+    function notifyDebtChange(address account) external onlyIssuer updateReward(account) { }
+
+    // TODO: finish notifyRewardAmount implementation
+    /// @notice This is called only by Synthetix after an account is liquidated and the SNX rewards are sent to the escrow.
+    function notifyRewardAmount(uint256 reward) external onlySynthetix updateReward(address(0)) {
+        // Ensure the provided reward amount is not more than the balance in the contract.
```
noahlitvin, line 137: I think we'd at least need `rewardRate = reward` here?

barrasso: Agreed

artdgn (changes requested):
Had some concerns about the implementation both in terms of mechanism and some of the calculations.

contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,181 @@
...
+
+// Internal references
+import "./interfaces/ISynthetixDebtShare.sol";
+import "./interfaces/IIssuer.sol";
+import "./interfaces/IRewardEscrowV2.sol";
+import "./interfaces/ISynthetix.sol";
+
+/// @title Upgrade Liquidation Mechanism V2 (SIP-148)
+/// @notice This contract is a modification to the existing liquidation mechanism defined in SIP-15.
+contract LiquidatorRewards is Owned, MixinSystemSettings, ILiquidatorRewards, ReentrancyGuard {
```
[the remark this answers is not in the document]
barrasso: > I think the StakingRewards mechanism relies on some assumptions about time and rewards rate, that make it fair, but that that don't hold here.

Totally agree. I've taken a first stab at removing the time-based components, but I think it still needs some tweaking.

> The core issue being that the rewards are being rewarded in a non-aligned way (for instaneous debt, instead of time-aware debt).

Assuming the 7-day burn lock isn't a strong enough of a disincentive, we'll need to think of an alternative method where we could "take a snapshot" of everyone's debt share balance each time a liquidation occurs and distribute the rewards instantaneously.

contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,181 @@
...
+        return balanceOf(account).mul(rewardPerToken(amount).sub(userRewardPerTokenPaid[account])).div(1e18).add(rewards[account]);
+    }
+
+    /* ========== MUTATIVE FUNCTIONS ========== */
+
+    function getRewards() public nonReentrant updateReward(msg.sender, 0) {
+        uint256 reward = rewards[msg.sender];
+        if (reward > 0) {
+            rewards[msg.sender] = 0;
+            rewardsToken.approve(address(rewardEscrowV2()), reward);
```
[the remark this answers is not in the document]
barrasso: Resolved https://github.com/Synthetixio/synthetix/pull/1647/commits/fdefc8f311f941bb74f3ce42897bfb3d3273fb17

contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,181 @@
...
+
+        if (amount < 0) {
+            return
+                rewardPerTokenStored.add(
+                    (accumulatedRewards).mul(1e18).div(totalSupply().sub(uint(amount)))
+                );
+        } else {
+            return
+                rewardPerTokenStored.add(
+                    (accumulatedRewards).mul(1e18).div(totalSupply().add(uint(amount)))
```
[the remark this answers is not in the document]
barrasso: Resolved https://github.com/Synthetixio/synthetix/pull/1647/commits/fdefc8f311f941bb74f3ce42897bfb3d3273fb17

contracts/Issuer.sol
```diff
@@ -608,16 +614,61 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+
+        // if total SNX to redeem is greater than account's collateral
+        // account is under collateralised, liquidate all collateral and reduce sUSD to burn
+        if (totalRedeemed > collateralForAccount) {
+            // set totalRedeemed to all transferable collateral
+            totalRedeemed = collateralForAccount;
+
+            // whats the equivalent sUSD to burn for all collateral less penalty
+            amountToLiquidate = _snxToUSD(
+                collateralForAccount.divideDecimal(SafeDecimalMath.unit().add(selfLiquidationPenalty)),
```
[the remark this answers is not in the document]
barrasso: Resolved https://github.com/Synthetixio/synthetix/pull/1647/commits/fdefc8f311f941bb74f3ce42897bfb3d3273fb17

contracts/Issuer.sol
```diff
@@ -608,16 +614,61 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
             );
         }
 
-        // burn sUSD from messageSender (liquidator) and reduce account's debt
-        _burnSynths(account, liquidator, amountToLiquidate, debtBalance, totalDebtIssued);
+        // Reduce debt share of delinquent account by the amount to liquidate.
+        _removeFromDebtRegister(delinquentAccount, amountToLiquidate, debtBalance, totalDebtIssued);
 
         // Remove liquidation flag if amount liquidated fixes ratio
         if (amountToLiquidate == amountToFixRatio) {
```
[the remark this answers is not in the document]
barrasso: Resolved https://github.com/Synthetixio/synthetix/pull/1647/commits/fdefc8f311f941bb74f3ce42897bfb3d3273fb17

contracts/Issuer.sol
```diff
@@ -608,16 +614,61 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
             );
         }
 
-        // burn sUSD from messageSender (liquidator) and reduce account's debt
-        _burnSynths(account, liquidator, amountToLiquidate, debtBalance, totalDebtIssued);
+        // Reduce debt share of delinquent account by the amount to liquidate.
```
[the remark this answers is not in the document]
barrasso: https://github.com/Synthetixio/synthetix/pull/1647/commits/fdefc8f311f941bb74f3ce42897bfb3d3273fb17

contracts/Issuer.sol
```diff
@@ -615,21 +620,66 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+        address account
+    ) external onlySynthetix returns (uint totalRedeemed, uint amountToLiquidate) {        
+        // Ensure waitingPeriod and sUSD balance is settled as burning impacts the size of debt pool
+        require(!exchanger().hasWaitingPeriodOrSettlementOwing(account, sUSD), "sUSD needs to be settled");
+
+        // Check if account is eligible for self liquidation
+        require(liquidator().isSelfLiquidationOpen(account), "Account not open for self liquidation");
+
+        // Get the penalty for self liquidation
+        uint selfLiquidationPenalty = liquidator().selfLiquidationPenalty();
```
dbeal-eth, line 648: is there a reason for indirection here?

contracts/Issuer.sol
```diff
@@ -615,21 +620,66 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
         // Remove liquidation flag if amount liquidated fixes ratio
-        if (amountToLiquidate == amountToFixRatio) {
+        if (amountToLiquidate >= amountToFixRatio) {
             // Remove liquidation
-            liquidations().removeAccountInLiquidation(account);
+            liquidator().removeAccountInLiquidation(delinquentAccount);
         }
     }
 
+    function selfLiquidateAccount(
```
dbeal-eth, line 638: np it seems wierd that this function is mutative, but it doesn't  do everything needed for liquidation. it implies that there could be more than one invoker of this function, and the invoker could do different things depending. But that would basically never be the case...

contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,179 @@
...
+/// @title Upgrade Liquidation Mechanism V2 (SIP-148)
+/// @notice This contract is a modification to the existing liquidation mechanism defined in SIP-15.
+contract LiquidatorRewards is ILiquidatorRewards, Owned, MixinSystemSettings, ReentrancyGuard {
+    using SafeMath for uint256;
+    using SafeERC20 for IERC20;
+
+    /* ========== STATE VARIABLES ========== */
+
+    IERC20 public snx; // Synthetix Token
+    IERC20 public sds; // SynthetixDebtShare
```
dbeal-eth, line 31: if we redeploy debt shares, does that mean we have to redeploy this contract too?

contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,179 @@
...
+        // console.log("accumulatedRewards: %s", accumulatedRewards);
+
+        return
+            rewardPerTokenStored.add(
+                (lastUpdateTime).mul(accumulatedRewards).mul(1e18).div(supply)
+            );
+    }
+
+    function earned(address account) public view returns (uint256) {
+        return synthetixDebtShare().balanceOf(account).mul(rewardPerToken().sub(userRewardPerTokenPaid[account])).div(1e18).add(rewards[account]);
```
dbeal-eth, line 108: use `multiplyDecimal` instead of `mul` and then `div(1e18)`

contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /* ========== MUTATIVE FUNCTIONS ========== */
+
+    function getReward() external nonReentrant updateReward(msg.sender) {
+        uint256 reward = rewards[msg.sender];
+        if (reward > 0) {
+            rewards[msg.sender] = 0;
+            snx.approve(address(rewardEscrowV2()), reward);
```
dbeal-eth, line 117: probably best to do an infinite approval whenever address resolver addresses are refreshed, but we can confirm what the auditors think

contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,179 @@
...
+
+    // TODO: May need two functions for mint and burn.
+
+    /// @notice This is called only by the Issuer to update when a given account's debt balance changes.
+    function notifyDebtChange(address account) external onlyIssuer updateReward(account) { }
+
+    /// @notice This is called only by Synthetix after an account is liquidated and the SNX rewards are sent to this contract.
+    function notifyRewardAmount(uint256 reward) external onlySynthetix {        
+        // Ensure the provided reward amount is not more than the balance in the contract.
+        // console.log("reward: %s", reward);
```
dbeal-eth, line 133: remove comments

contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,179 @@
...
+        if (supply == 0) {
+            // console.log("supply is zero: %s", supply);
+            return rewardPerTokenStored;
+        }
+        // console.log("lastUpdateTime: %s", lastUpdateTime);
+        // console.log("accumulatedRewards: %s", accumulatedRewards);
+
+        return
+            rewardPerTokenStored.add(
+                (lastUpdateTime).mul(accumulatedRewards).mul(1e18).div(supply)
```
dbeal-eth, line 103: divide decimal instead of `mul(1e18).div(supply)`

dbeal-eth: thsi parenthesis around `lastUpdateTime) is confusing with the `.add` from before and unnecessary

contracts/SystemSettingsLib.sol
```diff
@@ -28,10 +28,10 @@ library SystemSettingsLib {
     uint public constant MAX_LIQUIDATION_RATIO = 1e18; // 100% issuance ratio
     uint public constant RATIO_FROM_TARGET_BUFFER = 2e18; // 200% - mininimum buffer between issuance ratio and liquidation ratio
 
-    uint public constant MAX_LIQUIDATION_PENALTY = 1e18 / 4; // Max 25% liquidation penalty / bonus
+    uint public constant MAX_LIQUIDATION_PENALTY = 3e18 / 10; // Max 30% liquidation penalty / bonus
 
     uint public constant MAX_LIQUIDATION_DELAY = 30 days;
-    uint public constant MIN_LIQUIDATION_DELAY = 1 days;
+    uint public constant MIN_LIQUIDATION_DELAY = 43200; // 12 hours
```
dbeal-eth, line 34: does solidity not support `12 hours;`?

publish/src/commands/deploy/configure-legacy-settings.js
```diff
@@ -69,15 +69,15 @@ module.exports = async ({
 		});
 	}
 
-	if (Liquidations && EternalStorageLiquidations) {
+	if (Liquidator && EternalStorageLiquidator) {
 		await runStep({
-			contract: 'EternalStorageLiquidations',
-			target: EternalStorageLiquidations,
+			contract: 'EternalStorageLiquidator',
```
dbeal-eth, line 74: random question I just thought of: are we going to need to redeploy the existing `EternalStorageLiquidations` because of this rename, or are we just reusing

test/contracts/CollateralErc20.js
```diff
@@ -890,8 +890,8 @@ contract('CollateralErc20', async accounts => {
 		});
 
 		describe('should allow liquidations on an undercollateralised sUSD loan', async () => {
-			const renAmount = new BN('19642857');
-			const internalAmount = new BN('196428571428571428');
+			const renAmount = new BN('46428571');
```
dbeal-eth, line 893: how did you come out with these numbers? just trying to validate the change of the test

test/contracts/CollateralEth.js
```diff
@@ -177,7 +177,7 @@ contract('CollateralEth', async accounts => {
 		assert.equal(await ceth.collateralKey(), sETH);
 		assert.equal(await ceth.synths(0), toBytes32('SynthsUSD'));
 		assert.equal(await ceth.synths(1), toBytes32('SynthsETH'));
-		assert.bnEqual(await ceth.minCratio(), toUnit('1.3'));
+		assert.bnEqual(await ceth.minCratio(), toUnit('1.35'));
```
dbeal-eth, line 180: uh?

dbeal-eth (changes requested):
I think this needs more testing, and some of the existing tests changed for reasons that I think should be explained for clarity.

also we need to consider storage values a little more, such as whether or not we should continue using a new eternal stroage contract, or use flexible storage?

test/contracts/CollateralEth.js
barrasso: Since the `liquidationPenalty` has increased to 30%, we had to increase the `minCratio` of `CollateralEth` and `CollateralShort` to be greater than 130% otherwise `CollateralUtil.liquidationAmount` will try to divide by zero at this line: 

`uint sUSDamount = dividend.divideDecimal(divisor);`

publish/src/commands/deploy/configure-legacy-settings.js
barrasso: Good point, maybe it makes sense to leave this alone and revert the name change so we don't have to redeploy.

contracts/BaseSynthetix.sol
```diff
@@ -308,9 +324,50 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
 
         emitAccountLiquidated(account, totalRedeemed, amountLiquidated, messageSender);
 
-        // Transfer SNX redeemed to messageSender
+        // Transfer the flag reward to the account who flagged this account for liquidation.
+        address flagger = liquidator().getLiquidationCallerForAccount(account);
+        uint flagReward = liquidator().flagReward();
```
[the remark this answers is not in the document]
barrasso: out of scope

contracts/BaseSynthetix.sol
```diff
@@ -308,9 +324,50 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+        bool flagRewardTransferSucceeded = _transferByProxy(account, flagger, flagReward);
+        require(flagRewardTransferSucceeded, "Flag reward transfer did not succeed");
+
+        // Transfer the liquidate reward to message sender.
+        uint liquidateReward = liquidator().liquidateReward();
```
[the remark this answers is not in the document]
barrasso: also out of scope

contracts/Issuer.sol
```diff
@@ -615,21 +620,66 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
         }
 
-        // burn sUSD from messageSender (liquidator) and reduce account's debt
-        _burnSynths(account, liquidator, amountToLiquidate, debtBalance, totalDebtIssued);
+        // Reduce debt share of delinquent account by the amount to liquidate.
+        _removeFromDebtRegister(delinquentAccount, amountToLiquidate, debtBalance, totalDebtIssued);
 
         // Remove liquidation flag if amount liquidated fixes ratio
-        if (amountToLiquidate == amountToFixRatio) {
+        if (amountToLiquidate >= amountToFixRatio) {
```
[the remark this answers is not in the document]
barrasso: Yeah it's a good point. Arthur proposed this change I suppose to cover the case in which there may be a rounding error or something where these values are not exactly equal.

publish/src/commands/deploy/configure-legacy-settings.js
barrasso: What are the implications if we do continue on with this name change? Can we reuse the existing `EternalStorageLiquidations` and copy over the state to this new contract? 

I tried to revert the name change and was getting a undefined error at:
```js
await cache['EternalStorageLiquidations'].setAssociatedContract(instance.address, {...}
```

Does the name after `EternalStorage{NAME}` actually have to match the naming convention of the contract it is associated with? ie. `Liquidator`

contracts/Issuer.sol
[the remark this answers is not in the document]
barrasso: https://github.com/Synthetixio/synthetix/commit/86b55b9213e2864763ff0df0242859f7df9b972f

contracts/Issuer.sol
```diff
@@ -615,21 +620,66 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
-            liquidations().removeAccountInLiquidation(account);
+            liquidator().removeAccountInLiquidation(delinquentAccount);
         }
     }
 
+    function selfLiquidateAccount(
+        address account
+    ) external onlySynthetix returns (uint totalRedeemed, uint amountToLiquidate) {        
+        // Ensure waitingPeriod and sUSD balance is settled as burning impacts the size of debt pool
+        require(!exchanger().hasWaitingPeriodOrSettlementOwing(account, sUSD), "sUSD needs to be settled");
```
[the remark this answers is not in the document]
barrasso: https://github.com/Synthetixio/synthetix/commit/86b55b9213e2864763ff0df0242859f7df9b972f

contracts/Issuer.sol
barrasso: https://github.com/Synthetixio/synthetix/commit/86b55b9213e2864763ff0df0242859f7df9b972f

contracts/SystemSettingsLib.sol
barrasso: https://github.com/Synthetixio/synthetix/commit/86b55b9213e2864763ff0df0242859f7df9b972f

contracts/BaseSynthetix.sol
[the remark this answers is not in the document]
dbeal-eth: what does out of scope mean in this case? this should reduce scope?

contracts/Issuer.sol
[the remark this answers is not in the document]
dbeal-eth: yea sorry this was np wording, just trying to make the error as obvious as possible

contracts/BaseSynthetix.sol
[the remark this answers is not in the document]
barrasso: sorry meant `undeclared identifier`. I can't access `getFlagReward()` here because this base class only inherits `MixinResolver` and not `MixinSystemSettings`.

contracts/LiquidatorRewards.sol
barrasso: now that I think about it... yes.

publish/src/commands/deploy/configure-legacy-settings.js
justinjmoses: So there's this:

We wrote `FlexibleStorage` to solve the EternalStorage problem. There's no need to deploy a new EternalStorage contract for each contract - you use the existing `FlexibleStorage` and read/write to it using your contract name via the AddressResolver. Look at how we do it in `Issuer` here: https://github.com/Synthetixio/synthetix/blob/master/contracts/Issuer.sol#L807-L811. We can rewrite Liquidations to do this for reading and writing to storage. And IIRC, we don't need to port the existing liqudations storage data over unless there are any open liquidations.

justinjmoses: As for the naming convention - it's only really important in the unit tests `setup.js` - but it is a nice convention.

contracts/Issuer.sol
```diff
@@ -615,61 +627,58 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
 
+        // Get the total amount of SNX collateral (including escrows and rewards)
         uint collateralForAccount = _collateral(account);
-        uint amountToFixRatio =
-            liquidations().calculateAmountToFixCollateral(debtBalance, _snxToUSD(collateralForAccount, snxRate));
 
-        // Cap amount to liquidate to repair collateral ratio based on issuance ratio
-        amountToLiquidate = amountToFixRatio < susdAmount ? amountToFixRatio : susdAmount;
+        // Calculate the amount of debt to liquidate to fix c-ratio
+        amountToLiquidate = liquidator().calculateAmountToFixCollateral(
```
dbeal-eth, line 652: i might be confused, but if we are using `amountToLiquidate` to determine amount for liquidating, wouldn't this be a partial liquidation (assuming a "full" liquidation erases all the user's sUSD and any SNX required to cover the amount erased)

contracts/Issuer.sol
```diff
@@ -615,61 +627,58 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
-            // whats the equivalent sUSD to burn for all collateral less penalty
-            amountToLiquidate = _snxToUSD(
-                collateralForAccount.divideDecimal(SafeDecimalMath.unit().add(liquidationPenalty)),
-                snxRate
-            );
+            // Set totalRedeemed to all transferable collateral (SNX not in escrow)
+            totalRedeemed = IERC20(address(synthetix())).balanceOf(account);
+
+            // Get the equivalent debt amount to burn for all collateral (less penalty).
+            amountToLiquidate = _snxToUSD(collateralForAccount.divideDecimal(SafeDecimalMath.unit().sub(penalty)), snxRate);
```
dbeal-eth, line 670: shouldn't it be the other way around, where `amountToLiquidate` is the total debt shares for the user, and `totalRedeemed` is the corresponding value of the SNX + the liquidation penalty?

contracts/Issuer.sol
```diff
@@ -615,61 +627,58 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+        _removeFromDebtRegister(account, amountToLiquidate, debtBalance);
+
+        // Account for the burnt debt in the cache.
+        debtCache().updateCachedsUSDDebt(-SafeCast.toInt256(amountToLiquidate));
 
-        // Remove liquidation flag if amount liquidated fixes ratio
-        if (amountToLiquidate == amountToFixRatio) {
-            // Remove liquidation
-            liquidations().removeAccountInLiquidation(account);
+        if (!isSelfLiquidation) {
```
dbeal-eth, line 679: what if the account is self liquidated and they have fixed their c-ratio, but they have the liquidation flag on their account?

contracts/Liquidator.sol
```diff
@@ -129,30 +151,46 @@ contract Liquidations is Owned, MixinSystemSettings, ILiquidations {
...
+
     /**
      * r = target issuance ratio
      * D = debt balance
      * V = Collateral
      * P = liquidation penalty
      * Calculates amount of synths = (D - V * r) / (1 - (1 + P) * r)
      */
-    function calculateAmountToFixCollateral(uint debtBalance, uint collateral) external view returns (uint) {
+    function calculateAmountToFixCollateral(
```
dbeal-eth, line 166: since `debtBalance` and `collateral` are already passed into this function, wouldn't it just be better to pass in the resulting `collateral` by removing the penalty that way?

contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,125 @@
...
+        if (sharesSupply == 0) {
+            return 0;
+        }
+
+        return (IERC20(address(synthetix())).balanceOf(address(this))).div(sharesSupply);
+    }
+
+    function earned(address account) public view returns (uint256) {
+        return
+            synthetixDebtShare().balanceOf(account).mul(rewardPerShare().sub(userRewardPerSharePaid[account])).div(1e18).add(
```
dbeal-eth, line 94:
```
synthetixDebtShare().balanceOf(account).mul(rewardPerShare().sub(userRewardPerSharePaid[account])).div(1e18)
```

->
```
synthetixDebtShare().balanceOf(account).multiplyDecimal(rewardPerShare().sub(userRewardPerSharePaid[account]))
```

contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,125 @@
...
+            );
+    }
+
+    /* ========== MUTATIVE FUNCTIONS ========== */
+
+    function getReward() external nonReentrant updateReward(msg.sender) {
+        uint256 reward = rewards[msg.sender];
+        if (reward > 0) {
+            rewards[msg.sender] = 0;
+            IERC20(address(synthetix())).approve(address(rewardEscrowV2()), reward);
```
dbeal-eth, line 105: significant gas benefit would be to put this approval as infinite in the constructor. All that doing the approval here accomplishes is preventing funds from being sent from another function, which can be verified is not the case in this file

contracts/LiquidatorRewards.sol
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    function rewardPerShare() public view returns (uint256) {
+        (, uint sharesSupply, ) = issuer().allNetworksDebtInfo();
+
+        if (sharesSupply == 0) {
+            return 0;
+        }
+
+        return (IERC20(address(synthetix())).balanceOf(address(this))).div(sharesSupply);
```
dbeal-eth, line 89: i just realized, shouldn't this function record the total synthetix accumulated in the contract, not the current balance of SNX?

otherwise, as user's start redeeming their SNX, the balance will go down and so will the corresponding `rewardPerShare` even though that value should only be able to increase

test/contracts/LiquidatorRewards.js
```diff
@@ -0,0 +1,159 @@
...
+			await setupStaker();
+
+			const rewardValue = toUnit('100');
+			await synthetix.transfer(liquidatorRewards.address, rewardValue, { from: owner });
+
+			const earned = await liquidatorRewards.earned(stakingAccount1);
+			assert.bnGt(earned, ZERO_BN);
+		});
+
+		it('rewards balance should increase if new rewards come in', async () => {
```
dbeal-eth, line 125: where is the call to `earned`?

test/contracts/LiquidatorRewards.js
```diff
@@ -0,0 +1,159 @@
...
+			await setupStaker();
+
+			const rewardValue = toUnit('100');
+			await synthetix.transfer(liquidatorRewards.address, rewardValue, { from: owner });
+
+			const initialEarnedBal = await liquidatorRewards.earned(stakingAccount1);
+			await liquidatorRewards.getReward({ from: stakingAccount1 });
+			const postEarnedBal = await liquidatorRewards.earned(stakingAccount1);
+
+			assert.bnLt(postEarnedBal, initialEarnedBal);
```
dbeal-eth, line 153: if you call `getRewards()`, then you claim all your rewards from the contract, so `postEarnedBal` should be `0`

test/contracts/LiquidatorRewards.js
```diff
@@ -0,0 +1,159 @@
...
+			await synthetix.transfer(liquidatorRewards.address, newRewards, { from: owner });
+
+			const rewardsBalanceAfter = await synthetix.balanceOf(liquidatorRewards.address);
+
+			assert.bnEqual(rewardsBalanceBefore, ZERO_BN);
+			assert.bnGt(rewardsBalanceAfter, rewardsBalanceBefore);
+		});
+	});
+
+	describe('getReward()', () => {
```
dbeal-eth, line 140: need to add more tests for this function:

* if rewards are claimed twice
* if rewards are claimed when there are no rewards
* if two accounts are staking `earned()` remains the same for the account which did not call `getRewards`

test/contracts/LiquidatorRewards.js
```diff
@@ -0,0 +1,159 @@
...
+			const initialEarnedBal = await liquidatorRewards.earned(stakingAccount1);
+			await liquidatorRewards.getReward({ from: stakingAccount1 });
+			const postEarnedBal = await liquidatorRewards.earned(stakingAccount1);
+
+			assert.bnLt(postEarnedBal, initialEarnedBal);
+
+			const rewardsBalanceAfter = await synthetix.balanceOf(liquidatorRewards.address);
+			assert.bnGt(rewardsBalanceAfter, rewardsBalanceBefore);
+		});
+	});
```
dbeal-eth, line 158: need to add general purpose test to verify if debt shares balance changes, either by user changing their debt balance, or by total debt shares changing. The user's rewards should change in accordance with either of those values with respect to time, even if the user does not touch the `liquidatorRewards` contract

test/integration/behaviors/liquidations.behavior.js
```diff
@@ -96,41 +106,57 @@ function itCanLiquidate({ ctx }) {
...
+						const remainingReward = beforeDebttedSnx
+							.sub(await Synthetix.balanceOf(user5.address))
+							.sub(await Synthetix.balanceOf(user6.address));
+						assert.bnNotEqual(remainingReward, '0');
 						assert.bnEqual(
-							await Synthetix.balanceOf(otherUser.address),
-							beforeCredittedSnx.add(amountSent)
+							await Synthetix.balanceOf(LiquidatorRewards.address),
+							beforeRemainingRewardCredittedSnx.add(remainingReward)
 						);
```
dbeal-eth, line 160: might want to add a quick test to verify claiming of liquidator rewards after the fact

dbeal-eth (changes requested):
sorry, but I don't think we should be doing kovan deployment.

contracts/BaseSynthetix.sol
```diff
@@ -297,20 +311,45 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+
+        // Transfer the remaining SNX to the LiquidatorRewards contract.
+        uint remainingReward = totalRedeemed.sub(liquidateReward.add(flagReward));
+        bool remainingRewardTransferSucceedeed = _transferByProxy(account, address(liquidatorRewards()), remainingReward);
+        require(remainingRewardTransferSucceedeed, "Remaining rewards transfer did not succeed");
+
+        return remainingRewardTransferSucceedeed;
+    }
+
+    function liquidateSelf(address account) external systemActive optionalProxy returns (bool) {
```
dbeal-eth, line 340: `address account` not needed here

contracts/Issuer.sol
```diff
@@ -350,6 +358,10 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
             balance = balance.add(rewardEscrowV2().balanceOf(account));
         }
 
+        if (address(liquidatorRewards()) != address(0)) {
+            balance = balance.add(liquidatorRewards().earned(account));
+        }
```
justinjmoses, line 363: This change could be quite gas intensive. Can you please run the fork tests with and without this to compare gas on mint, burn, transfer and claim? 🙏🏼

contracts/Liquidator.sol
```diff
@@ -129,30 +151,46 @@ contract Liquidations is Owned, MixinSystemSettings, ILiquidations {
         return deadline > 0 && now > deadline;
     }
 
+    function _hasEnoughCollateral(address account) internal view returns (bool) {
```
justinjmoses, line 154: maybe `hasEnoughSNXBalance` to clarify it's difference from `Synthetix.collateral` which includes escrow

contracts/BaseSynthetix.sol
```diff
@@ -297,20 +311,45 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+        bool flagRewardTransferSucceeded = _transferByProxy(account, flagger, flagReward);
+        require(flagRewardTransferSucceeded, "Flag reward transfer did not succeed");
+
+        // Transfer the liquidate reward to message sender.
+        uint liquidateReward = liquidator().liquidateReward();
+        bool liquidateRewardTransferSucceeded = _transferByProxy(account, messageSender, liquidateReward);
+        require(liquidateRewardTransferSucceeded, "Liquidate reward transfer did not succeed");
+
+        // Transfer the remaining SNX to the LiquidatorRewards contract.
+        uint remainingReward = totalRedeemed.sub(liquidateReward.add(flagReward));
```
barrasso, line 333: check if remainingReward > 0

barrasso: add tests for this

test/integration/behaviors/liquidations.behavior.js
```diff
@@ -8,39 +9,48 @@ const { skipLiquidationDelay } = require('../utils/skip');
...
+			({
+				Synthetix,
+				SynthetixDebtShare,
+				Liquidator,
+				LiquidatorRewards,
+				SystemSettings,
+			} = ctx.contracts);
 
-			({ owner, someUser, otherUser } = ctx.users);
+			({ owner, user4, user5, user6 } = ctx.users);
```
barrasso, line 27: change names to flagger, liquidator, liquidated

test/integration/behaviors/liquidations.behavior.js
```diff
@@ -96,41 +106,57 @@ function itCanLiquidate({ ctx }) {
...
+						await Synthetix.connect(user6).liquidateDelinquentAccount(user4.address);
 					});
 
-					it('deducts sUSD debt from the liquidated', async () => {
-						assert.bnLt(
-							await Synthetix.debtBalanceOf(someUser.address, toBytes32('sUSD')),
-							beforeDebt
-						);
+					it('reduces the debt share balance of the liquidated', async () => {
+						assert.bnLt(await SynthetixDebtShare.balanceOf(user4.address), beforeDebt);
```
barrasso, line 127: check actual values

test/contracts/Liquidator.js
```diff
@@ -308,40 +343,174 @@ contract('Liquidations', accounts => {
...
+							const debtAfter = aliceDebtValueBefore.sub(amountToFixRatio);
+							const collateralAfterMinusPenalty = aliceCollateralBefore.sub(
+								multiplyDecimal(amountToFixRatio, toUnit('1').add(penalty))
+							);
+
+							// c-ratio = debt / collateral
+							const collateralRatio = divideDecimal(debtAfter, collateralAfterMinusPenalty);
+
+							assert.bnEqual(collateralRatio, ratio);
+						});
```
barrasso, line 494: also check that the liquidation flag is cleared here

contracts/Issuer.sol
```diff
@@ -350,6 +358,10 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
             balance = balance.add(rewardEscrowV2().balanceOf(account));
         }
 
+        if (address(liquidatorRewards()) != address(0)) {
```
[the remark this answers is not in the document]
barrasso: We should also consider if we even want to add this check to `_collateral`. Is it worth the gas increase?

contracts/Issuer.sol
barrasso: Did some testing and measured about ~10-15% increases in gas usage with this change.

contracts/Issuer.sol
barrasso: Yeah the terms are a bit confusing, but we're using `amountToLiquidate` here to calculate the amount of debt to burn in order to reach the target c-ratio (this is considered a "full liquidation"). 

A "full liquidation" essentially means an account can only get liquidated **once**.  It'll either attempt to fix their cratio with their SNX collateral, or it will clear all collateral and debt.

A "partial liquidation" is when an account is liquidated for an arbitrary amount of sUSD (`susdAmount`).
 e.g. you could "partially" liquidate someone by calling the function multiple times in $5 intervals.

This SIP doesn't allow for partial liquidations anymore -- that's why `uint susdAmount` was removed from this function.

contracts/Issuer.sol
barrasso: In this case, the amount of SNX to be redeemed is greater than the amount of account's collateral balance.

So, we need to set `totalRedeemed` to the account's collateral balance. It shouldn't matter about the penalty here since all the SNX is being redeemed anyway.

And then `amountToLiquidate` is the amount of debt to burn (equivalent to the value of the redeemed SNX).

contracts/LiquidatorRewards.sol
barrasso: Good point. Although I was wondering if we put an infinite approval in the constructor, what happens when we deploy a new `Synthetix` contract? Wouldn't we need to do another approval somehow?

contracts/BaseSynthetix.sol
barrasso: added check in https://github.com/Synthetixio/synthetix/pull/1647/commits/44262613afb644a98a5637db95df83692c79c6d6


unit tests pending

contracts/BaseSynthetix.sol
barrasso: https://github.com/Synthetixio/synthetix/pull/1647/commits/44262613afb644a98a5637db95df83692c79c6d6

contracts/Issuer.sol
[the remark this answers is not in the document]
barrasso: removed this for now (https://github.com/Synthetixio/synthetix/pull/1647/commits/44262613afb644a98a5637db95df83692c79c6d6) as I don't think the marginal benefit to minters is worth the gas usage increase across the board

contracts/Issuer.sol
barrasso: https://github.com/Synthetixio/synthetix/pull/1647/commits/44262613afb644a98a5637db95df83692c79c6d6

contracts/Liquidator.sol
barrasso: https://github.com/Synthetixio/synthetix/pull/1647/commits/44262613afb644a98a5637db95df83692c79c6d6

contracts/Liquidator.sol
barrasso: Could you elaborate on this?

dbeal-eth: sorry, this was really poorly written

based on the name of the function and the alternatives available, i think its kind of out of scope to factor in the additional parameter `isSelfLiquidation` to figure out what penalty to put into the equation.

the problem is this function has so many args and nuances to the args that it might be better to just have a generic function `calculateAmountToFixCollateral(address account)` which does the whole shebang for a given account passed to `liquidateDelinquentAccount`

also, this function does not take into account the liquidation rewards--only the penalty. Therefore, it is necessary to pass in the `collateralForAccount.sub(liquidationReward)` or else their account will remain undercollateralized after liquidation is completed

test/integration/behaviors/liquidations.behavior.js
barrasso: https://github.com/Synthetixio/synthetix/commit/f5537535433620c4b0bfec8d672598de1da7b97f

contracts/LiquidatorRewards.sol
barrasso: https://github.com/Synthetixio/synthetix/pull/1647/commits/34afd173fdb4bbb33f7da07e9abd32aac9720fc6

contracts/LiquidatorRewards.sol
barrasso: https://github.com/Synthetixio/synthetix/pull/1647/commits/34afd173fdb4bbb33f7da07e9abd32aac9720fc6

test/contracts/LiquidatorRewards.js
barrasso: https://github.com/Synthetixio/synthetix/pull/1647/commits/34afd173fdb4bbb33f7da07e9abd32aac9720fc6

test/integration/behaviors/liquidations.behavior.js
barrasso: https://github.com/Synthetixio/synthetix/pull/1647/commits/34afd173fdb4bbb33f7da07e9abd32aac9720fc6

test/integration/behaviors/liquidations.behavior.js
barrasso: https://github.com/Synthetixio/synthetix/pull/1647/commits/7e9d1d23c861da261c7f103b75fc1aa4541a297e

test/integration/behaviors/liquidations.behavior.js
barrasso: https://github.com/Synthetixio/synthetix/pull/1647/commits/7e9d1d23c861da261c7f103b75fc1aa4541a297e

test/contracts/LiquidatorRewards.js
barrasso: https://github.com/Synthetixio/synthetix/commit/5d8824ac463f5ce92977f81a5b94ff095ead1edc

test/contracts/LiquidatorRewards.js
barrasso: https://github.com/Synthetixio/synthetix/commit/5d8824ac463f5ce92977f81a5b94ff095ead1edc

test/contracts/LiquidatorRewards.js
```diff
@@ -0,0 +1,388 @@
...
+
+				const afterEarnedValue = await liquidatorRewards.earned(stakingAccount1);
+				const afterDebtShareBalance = await synthetixDebtShare.balanceOf(stakingAccount2);
+				const afterDebtSharesSupply = await synthetixDebtShare.totalSupply();
+
+				assert.bnGt(afterEarnedValue, beforeEarnedValue);
+				assert.bnLt(afterDebtShareBalance, beforeDebtShareBalance);
+				assert.bnLt(afterDebtSharesSupply, beforeDebtSharesSupply);
+			});
+		});
```
dbeal-eth, line 304: can we add one more compound test somewhere in this file where (for example)

1. user A mints 100 debt shares
2. user B mints 100 debt shares
3. 10 sUSD of liquidation rewards are collected after someone is liquidated
4. user A claims
5. user C mints 100 debt shares

ensure user B still has same amount of rewards as what user A claimed

its sort a similar test to the `earned()` test added previously where we ensure that other users are not affected by one user's actions (in this case, a new user coming in and increasing the `sharesSupply` should not affect already earned rewards)

dbeal-eth (changes requested):

contracts/BaseSynthetix.sol
```diff
@@ -297,20 +311,66 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
 
-        // Transfer SNX redeemed to messageSender
-        // Reverts if amount to redeem is more than balanceOf account, ie due to escrowed balance
-        return _transferByProxy(account, messageSender, totalRedeemed);
+        if (totalRedeemed > 0) {
+            uint stakerRewards; // The amount of rewards to be sent to the LiquidatorRewards contract.
+            uint flagReward = liquidator().flagReward();
+            uint liquidateReward = liquidator().liquidateReward();
+            // Check if the total amount of redeemed SNX is enough to payout the liquidation rewards.
+            if (totalRedeemed > (flagReward.add(liquidateReward))) {
```
dbeal-eth, line 326: nit unnecessary extra parenthesis

contracts/SystemSettingsLib.sol
```diff
@@ -28,10 +28,10 @@ library SystemSettingsLib {
     uint public constant MAX_LIQUIDATION_RATIO = 1e18; // 100% issuance ratio
     uint public constant RATIO_FROM_TARGET_BUFFER = 2e18; // 200% - mininimum buffer between issuance ratio and liquidation ratio
 
-    uint public constant MAX_LIQUIDATION_PENALTY = 1e18 / 4; // Max 25% liquidation penalty / bonus
+    uint public constant MAX_LIQUIDATION_PENALTY = 3e18 / 10; // Max 30% liquidation penalty / bonus
 
     uint public constant MAX_LIQUIDATION_DELAY = 30 days;
-    uint public constant MIN_LIQUIDATION_DELAY = 1 days;
+    uint public constant MIN_LIQUIDATION_DELAY = 24 hours;
```
dbeal-eth, line 34: we might want to let the council set this to a lower minimum value. in severe market conditions, 6 hour liquidations seem like somethign taht could be a real option they could legit take. who are we to stop them?

test/contracts/LiquidatorRewards.js
```diff
@@ -0,0 +1,324 @@
...
+				address: synthetix.address,
+				skipPassCheck: true,
+				reason: 'Synthetix only',
+			});
+		});
+	});
+
+	describe('earned()', () => {
+		it('should be 0 when not staking', async () => {
+			assert.bnEqual(await liquidatorRewards.earned(stakingAccount1), ZERO_BN);
```
dbeal-eth, line 143: might want to add a call to `notifyRewardAmount` here to drive home that nobody can earn when there is no staking

dbeal-eth (changes requested):

contracts/SystemSettingsLib.sol
barrasso: consulted with spartan council in discord

contracts/BaseSynthetix.sol
```diff
@@ -314,9 +314,6 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
     /// @notice Force liquidate a delinquent account and distribute the redeemed SNX rewards amongst the appropriate recipients.
     /// @dev The SNX transfers will revert if the amount to send is more than balanceOf account (i.e. due to escrowed balance).
     function liquidateDelinquentAccount(address account) external systemActive optionalProxy returns (bool) {
-        // Get the unclaimed liquidator rewards for this account.
-        liquidatorRewards().getReward(account);
```
barrasso: removed this because the rewards get escrowed anyway and didn't add much value.

plus it reset `messageSender` (bc of the approval call from within LiquidatorRewards)

contracts/Issuer.sol
```diff
@@ -707,27 +707,23 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
         );
 
         // Get the equivalent amount of SNX for the amount to liquidate
-        uint snxRedeemed = _usdToSnx(amountToLiquidate, snxRate);
+        totalRedeemed = _usdToSnx(amountToLiquidate, snxRate);
 
-        // Factor in the liquidation penalty
-        totalRedeemed = snxRedeemed.multiplyDecimal(SafeDecimalMath.unit().add(penalty));
-
```
barrasso, line 697: removed because the penalty is already accounted for in `calculateAmountToLiquidate`

contracts/Issuer.sol
```diff
@@ -707,27 +707,23 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
             amountToLiquidate = debtBalance.multiplyDecimal(transferableBalance).divideDecimal(collateralForAccount);
         }
 
-        // Reduce debt by amount to liquidate.
+        // Reduce debt shares by amount to liquidate.
         _removeFromDebtRegister(account, amountToLiquidate, debtBalance);
 
-        // Account for the burnt debt in the cache.
-        debtCache().updateCachedsUSDDebt(-SafeCast.toInt256(amountToLiquidate));
-
```
barrasso, line 713: removed because we don't actually burn any debt (no sUSD is taken out of circulation), only debt share balance changes.

contracts/Liquidator.sol
```diff
@@ -129,30 +151,46 @@ contract Liquidations is Owned, MixinSystemSettings, ILiquidations {
...
         return dividend.divideDecimal(divisor);
     }
 
     // get liquidationEntry for account
     // returns deadline = 0 when not set
     function _getLiquidationEntryForAccount(address account) internal view returns (LiquidationEntry memory _liquidation) {
-        _liquidation.deadline = eternalStorageLiquidations().getUIntValue(_getKey(LIQUIDATION_DEADLINE, account));
+        _liquidation.deadline = flexibleStorage().getUIntValue(
+            CONTRACT_NAME,
+            keccak256(abi.encodePacked(_getKey(LIQUIDATION_DEADLINE, account), account))
```
dbeal-eth, line 186: we are double packing and hashing here (`_getKey` already doess `keccak256(abi.encodePacked())`). If we are trying to do this to get a rekey, its not necessary because `Liquidations` was renamed to `Liquidator` and that affects the final storage slot in `FlexibleStorage`.

since its unnecessary code complexity and increases gas, its recommended to revert to old code and just use flexible storage

contracts/Liquidator.sol
```diff
@@ -219,23 +256,33 @@ contract Liquidations is Owned, MixinSystemSettings, ILiquidations {
         uint _deadline,
         address _caller
     ) internal {
-        // record liquidation deadline
-        eternalStorageLiquidations().setUIntValue(_getKey(LIQUIDATION_DEADLINE, _account), _deadline);
-        eternalStorageLiquidations().setAddressValue(_getKey(LIQUIDATION_CALLER, _account), _caller);
+        // record liquidation deadline and caller
+        flexibleStorage().setUIntValue(
+            CONTRACT_NAME,
+            keccak256(abi.encodePacked(_getKey(LIQUIDATION_DEADLINE, _account), _account)),
```
dbeal-eth, line 262: same here

contracts/SystemSettings.sol
```diff
@@ -480,7 +528,11 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
     event TargetThresholdUpdated(uint newTargetThreshold);
     event LiquidationDelayUpdated(uint newDelay);
     event LiquidationRatioUpdated(uint newRatio);
+    event LiquidationEscrowDurationUpdated(uint newDuration);
     event LiquidationPenaltyUpdated(uint newPenalty);
+    event SelfLiquidationPenaltyUpdated(uint newPenalty);
+    event FlagRewardUpdated(uint newReward);
```
dbeal-eth, line 534: might be good to rename these something like `LiquidationFlagRewardUpdated`

dbeal-eth (approved):

PR #1648, opened 2022-01-11
SIP 193 handover changes

artdgn:
Addressing comments and small fixes for https://github.com/Synthetixio/synthetix/pull/1627/files

### DIFF
--- a/contracts/MixinSystemSettings.sol
+++ b/contracts/MixinSystemSettings.sol
@@ -7,6 +7,7 @@ import "./interfaces/IFlexibleStorage.sol";
 
 // https://docs.synthetix.io/contracts/source/contracts/mixinsystemsettings
 contract MixinSystemSettings is MixinResolver {
+    // must match the one defined SystemSettingsLib, defined in both places due to sol v0.5 limitations
     bytes32 internal constant SETTING_CONTRACT_NAME = "SystemSettings";
 
     bytes32 internal constant SETTING_WAITING_PERIOD_SECS = "waitingPeriodSecs";
--- a/contracts/SystemSettings.sol
+++ b/contracts/SystemSettings.sol
@@ -9,11 +9,18 @@ import "./SystemSettingsLib.sol";
 
 // https://docs.synthetix.io/contracts/source/contracts/systemsettings
 contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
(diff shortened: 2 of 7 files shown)


contracts/MixinSystemSettings.sol
```diff
@@ -7,6 +7,7 @@ import "./interfaces/IFlexibleStorage.sol";
 
 // https://docs.synthetix.io/contracts/source/contracts/mixinsystemsettings
 contract MixinSystemSettings is MixinResolver {
+    // must match the one defined SystemSettingsLib, defined in both places due to sol v0.5 limitations
     bytes32 internal constant SETTING_CONTRACT_NAME = "SystemSettings";
```
leckylao, line 11: NIT: This might create human error in the future, I do see there's test case to guard it a bit, but would still prefer to just pass the variable around in order for single source of truth.

leckylao (approved):
NIT for the SETTING_CONTRACT_NAME, others LGTM

contracts/MixinSystemSettings.sol
artdgn: This is not ideal, but I don't think there's potential for human error in this case, because not only the added test case would fail, but almost every read & write test in the SystemSettings.js would fail as well - because the write and read storage slots would be different for all settings. So I think there is no practical risk assuming the tests are run at all.

artdgn:
Current size:
```
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤
│ SystemSettingsLib               │ 13158  │ 12.85 KB  │ 53.54%           │          │
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤

....
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤
│ SystemSettings                  │ 14440  │ 14.1 KB   │ 58.76%           │          │
├─────────────────────────────────┼────────┼───────────┼──────────────────┼──────────┤

```

contracts/MixinSystemSettings.sol
artdgn: In any case, just to be extra safe, and not rely on tests, I've added a check for equality in the constructor:
```
require(SETTING_CONTRACT_NAME == SystemSettingsLib.contractName(), "read and write keys not equal");

PR #1651, opened 2022-01-13
Peacock release v2.57 to kovan

barrasso:


dbeal-eth (approved):
confirmed release is good

PR #1656, opened 2022-01-21
Optimism Forking with Hardhat

dbeal-eth:
This PR uses the newly released forking support on hardhat to run `fork` tests on optimism, as well as allowing the starting of a fork by just running `npm run fork:ovm` (with appropriate `PROVIDER_URL` or similar set)

Notes:
* I ran into an insane number of issues debugging owners for both the forked and the integration tests, so there are a lot of misc test chagnes in oredr to deal with that. Also,
* I think we should get rid of the `directRelay` test, now that we have locked the relay system and we don't need to redeploy it ever again, its just making the management of these tests much harder
* Had to change the synth for the redeemer test to `sBTC` since `sDEFI` does not exist on mainnet l2

Let me know if I missed anything else.


.circleci/src/jobs/job-fork-tests.yml
```diff
@@ -7,15 +7,30 @@ steps:
   - run:
       command: npm run fork:mainnet
       background: true
+  - run:
```
justinjmoses, line 10: Can we split out into two jobs - `job-fork-tests` and `job-fork-tests-ovm`? The reason `job-test-integration` isn't split out is due to the need to run the complex ops tool in the background (though frankly, now that it's using caching we should probably split these as well). This makes it easier to identify flakes and where the error is without having to open the CI job

justinjmoses (changes requested):

.circleci/src/jobs/job-fork-tests.yml
dbeal-eth: good idea

hardhat/tasks/task-test-integration.js
```diff
@@ -89,25 +96,47 @@ task('test:integration:l2', 'run isolated layer 2 production tests')
...
-			await prepareDeploy({ network, synthsToAdd, useOvm });
-			await deployInstance({
-				addNewSynths: true,
-				buildPath,
-				network,
-				providerPort: providerPortL2,
-				providerUrl,
-				useOvm,
-			});
+			if (taskArguments.useFork) {
```
justinjmoses, line 111: [np] we could probably dedupe a bunch of code here as the l1 and l2 tasks are looking pretty similar but I get you want to put this PR to bed so it's just a nitpick

publish/deployed/kovan-ovm/deployment.json
```diff
@@ -155,11 +155,11 @@
 		},
 		"FeePool": {
 			"name": "FeePool",
-			"address": "0x129fd2f3a799bD156e8c00599760AfC2f0f953dA",
+			"address": "0x2F737bf6a32bf3AcBef4d5148DA507569204Fb61",
```
justinjmoses, line 158: This is mistaken - `0x129fd2f3a799bD156e8c00599760AfC2f0f953dA` is what it should be - looks like you don't have the latest develop merged in

justinjmoses (changes requested):
Something is up with the merge - it's trying to undo the peacock changes to deployment.json

hardhat/tasks/task-test-integration.js
dbeal-eth: appreciate your understanding

hardhat/tasks/task-test-integration.js
```diff
@@ -127,27 +156,25 @@ task('test:integration:dual', 'run integrated layer 1 and layer 2 production tes
...
+			await compileInstance({ useOvm: false, buildPath: buildPath });
 		}
 
 		if (taskArguments.deploy) {
 			await deployInstance({
 				useOvm: false,
 				providerUrl,
 				providerPort: providerPortL1,
-				buildPath: buildPathEvm,
+				buildPath: buildPath,
```
justinjmoses, line 170: [np] can just use `buildPath` here as shorthand

index.js
```diff
@@ -499,6 +499,7 @@ const getUsers = ({ network = 'mainnet', user, useOvm = false } = {}) => {
 		}),
 		'mainnet-ovm': Object.assign({}, base, {
 			owner: '0x6d4a64C57612841c2C6745dB2a4E4db34F002D20',
+			deployer: '0x3C05B1239B223c969540FeFc0270227a2B00e047',
```
justinjmoses, line 502: we use the same deployer as mainnet

test/integration/l2/directRelay.l2.integration.js
```diff
@@ -87,7 +79,7 @@ describe('tempOwner directRelay integration tests (L2)', () => {
 			originalMinimumStakeTime = await SystemSettingsL2.minimumStakeTime();
 		});
 
-		describe('when changing an L2 system setting with directRelay', () => {
+		describe.skip('when changing an L2 system setting with directRelay', () => {
```
justinjmoses, line 82: can you just comment why you are skipping this?

test/integration/l2/directRelay.l2.integration.js
```diff
@@ -1,11 +1,8 @@
 const { assert } = require('../../contracts/common');
 const { bootstrapL2 } = require('../utils/bootstrap');
 
-const {
-	defaults: { TEMP_OWNER_DEFAULT_DURATION },
-} = require('../../..');
-
-describe('tempOwner directRelay integration tests (L2)', () => {
+// skipped because tempOwner no longer will work for fork tests
+describe.skip('tempOwner directRelay integration tests (L2)', () => {
```
justinjmoses, line 5: why not conditionally skip if fork? we do that elsewhere in a `before` block

justinjmoses: still ping on this.

test/integration/l2/directRelay.l2.integration.js
justinjmoses, line 82: why skip this specifically if the whole thing is skipped?

justinjmoses (approved):
Just a few comments left.

test/integration/l2/directRelay.l2.integration.js
dbeal-eth: for some reason I get this error when I try to use `this.skip()` from within, like we do in other tests:

```
TypeError: this.skip is not a function
```

Maybe because the before to skip is in a top level `describe` block?

I added `this.skip()` to nested tests, lets see how that works.

PR #1661, opened 2022-01-24
SIPS 196, 193, 184 - audit fixes, full version

artdgn:
The more comprehensive (and more complex) version of addressing the audit comments. The main issue difference from the minimal version is in addressing the `maxDynamiFee` / `maxExchangeFee` possibly being very large (possibly 100% but realistically up to 30% with current parameters).

This version caps the max dynamic fee at 5% and if dynamic fee would exceed it - no trade is performed (exchanger returns `(0,0)` as with other no-op scenarios). This also removes the usage of dynamic fees from atomic exchanges in the contract code (it was previously disabled by different L1 <> L2 configs).

This version also fixes the minor issues noted in the audits: some clarifying comments, and minor consistency issue.

Still to do: 
 - [x] Decide which fix (full or minimal) is preferrable.
 - [x] Auditors review of fixes.
 - [x] Find out the correct course of action for high dynamic fee in `getAmountsForExchange` which is used by `Collateral` to repay loans using collateral.
 - [x] Find out if it's acceptable to remove dynamic fee from the atomic exchanges in contracts (rather than just in config).

* A version that only addresses the minor issues **without the max fee issue** is https://github.com/Synthetixio/synthetix/pull/1662*


contracts/Exchanger.sol
```diff
@@ -945,27 +978,12 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
-        view
-        returns (
-            uint amountReceived,
-            uint fee,
-            uint exchangeFeeRate
-        )
-    {
-        exchangeFeeRate = _feeRateForExchange(sourceCurrencyKey, destinationCurrencyKey);
+        if (tooVolatile) {
+            // TODO: should this continue or return (0,0,0)?
```
artdgn, line 985: @barrasso That's the spot I'm not sure about:
1. If code continues if `tooVolatile` is true, max fee (5%) will be charged even if volatility would require higher fee to prevent possible oracle delay frontrunning.
2. If code returns no-op in this case - it might cause issues to the collateral repayments that are using it.

contracts/ExchangerWithFeeRecAlternatives.sol
```diff
@@ -272,14 +272,17 @@ contract ExchangerWithFeeRecAlternatives is MinimalProxyFactory, Exchanger {
         view
         returns (uint)
     {
+        // unused
+        sourceCurrencyKey;
```
artdgn, line 276: this is there just to keep it in the interface, to be consistent with previous interfaces and in case in future sourceCurrency is also used to determine the fee (it makes sense that it's important)

contracts/Exchanger.sol
barrasso: I'm more in favor of option 1 since it saves us from introducing issues to the existing collateral repayments. Though I guess it depends on how big of a problem the oracle frontrunning can be with the fee at 5%.

artdgn: After the clarifications on the usage of this method in `Collateral` it seems that this method is only used to `sUSD` to `sUSD` exchange, which means that it will never be volatile (dynamic fee will be 0). So for that purpose it's safe to return no-op `(0, 0, 0)`. 
For the other uses of this method it's also more correct to return no-op as this method should be consistent with what `exchange()` will do.

artdgn: Ended up adding a separate PR for the collateral checks and revert in case of 0 `amountReceived`: https://github.com/Synthetixio/synthetix/pull/1663

barrasso (approved):

barrasso (approved):
checks LGTM

PR #1663, opened 2022-01-25
refactor collateral repay into short, add checks

artdgn:
This adds adds a fix to an issue found in https://github.com/Synthetixio/synthetix/pull/1661 where:
- `CollateralShort` may try to repay a loan using collateral, but may receive 0 amount from the Exchanger for the collateral exchanged (e.g. due to invalid rate, or in https://github.com/Synthetixio/synthetix/pull/1661 due to high volatility). But will still consider the loan repayed. This PR adds a check for the returned amount - and a revert in case it's 0.
- Also the said loan repayment implicitly assumed collateral to be sUSD, and this PR adds an explicit check for that, and also refactors the code from `Collateral` into `CollaterlShort` (because the code only works for sUSD collateral).


barrasso (commented):
These changes look good to me. 

However, the main problem to solve here is how to facilitate the upgrade of the `CollateralShort` contract (if it's even possible). 

I believe this will require some sort of import/migrate functionality moving the old contract state to the new one (specifically the `loans` state + `sUSD` collateral). 

At the time of this writing, there is <$300k sUSD locked in the `CollateralShort` contract (which isn't a huge amount but still presents some risk).

PR #1667, opened 2022-01-27
SIP-198: Atomic Exchange At Pure Chainlink Price

noahlitvin:
### Breaking changes
This update includes changes to the `ISynthetix` and `IExchanger` interfaces, adding a parameter `minAmount` for slippage control.


contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -2,10 +2,11 @@ pragma solidity ^0.5.16;
 
 // Inheritance
 import "./ExchangeRates.sol";
+import "./MixinSystemSettings.sol";
 import "./interfaces/IDexPriceAggregator.sol";
 
 // https://docs.synthetix.io/contracts/source/contracts/exchangerateswithdexpricing
-contract ExchangeRatesWithDexPricing is ExchangeRates {
+contract ExchangeRatesWithDexPricing is MixinSystemSettings, ExchangeRates {
```
justinjmoses, line 9: `ExchangeRates` already has this mixin so not needed

contracts/MixinSystemSettings.sol
```diff
@@ -270,4 +271,12 @@ contract MixinSystemSettings is MixinResolver {
                 keccak256(abi.encodePacked(SETTING_ATOMIC_VOLATILITY_UPDATE_THRESHOLD, currencyKey))
             );
     }
+
+    function getPureChainlinkPriceForAtomicSwapsEnabled(bytes32 currencyKey) internal view returns (bool) {
```
justinjmoses, line 275: [np] perhaps take in an array of `bytes32` and use `IFlexibleStorage.getBoolValues()` to reducd gas?

contracts/MixinSystemSettings.sol
```diff
@@ -50,6 +50,7 @@ contract MixinSystemSettings is MixinResolver {
     bytes32 internal constant SETTING_ATOMIC_PRICE_BUFFER = "atomicPriceBuffer";
     bytes32 internal constant SETTING_ATOMIC_VOLATILITY_CONSIDERATION_WINDOW = "atomicVolConsiderationWindow";
     bytes32 internal constant SETTING_ATOMIC_VOLATILITY_UPDATE_THRESHOLD = "atomicVolUpdateThreshold";
+    bytes32 internal constant SETTING_PURE_CHAINLINK_PRICE_FOR_ATOMIC_SWAPS_ENABLED = "pureChainlinkForAtomicsEnabled";
```
artdgn, line 53: I've suggested it before, but haven't had the time to do it: it might be good to consolidate all the many atomic settings into a single struct `AtomicExchangeConfig`, which one of the top level methods in its exchanger requests once from settings and than passes into the internal ones. The same pattern as with `DynamicFeeConfig` params. 

So perhaps something like `SystemSettings.getAtomicExchangeConfig(bytes32 sourceKey, bytes32 destKey) returns (AtomicExchangeConfig memory)`

There are many advantages to this:
- Reduced gas costs: instead of 8-10 different calls to flexible storage contract, there is one. An array of keys is used to get an array of values - and they are passed in a struct.
- Reduced contract size for system settings (possibly): one external getter method instead of many.
- Reduced verbosity: the one getter method is shorter, and usage in code is shorter `config.value`.
- Testability: injecting the config this way can allow better unit testing of internal methods in a `Testable...` variant of a contract.

test/contracts/Exchanger.spec.js
```diff
@@ -2978,6 +2979,189 @@ contract('Exchanger (spec tests)', async accounts => {
...
 			});
+
+			describe('when we can use the pure Chainlink price', () => {
+				let amountIn;
+				let amountReceived;
+				let amountFee;
+
+				beforeEach(async () => {
+
+					// Set up Chainlink Prices
```
artdgn, line 2990: nit: there are now helper methods for these steps in `helpers.js` that might be useful / shorter to use here

contracts/MixinSystemSettings.sol
noahlitvin: See: https://github.com/Synthetixio/synthetix/pull/1709

artdgn (dismissed):

contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -112,18 +112,27 @@ contract ExchangeRatesWithDexPricing is ExchangeRates {
...
         require(address(inputEquivalent) != address(0), "No atomic equivalent for input");
+
         IERC20 susdEquivalent = IERC20(getAtomicEquivalentForDexPricing("sUSD"));
+        require(address(inputEquivalent) != address(0), "No atomic equivalent for sUSD");
 
-        return
-            _dexPriceDestinationValue(inputEquivalent, susdEquivalent, amount).div(amount).mul(
-                10**uint(inputEquivalent.decimals())
+        uint result =
+            _dexPriceDestinationValue(inputEquivalent, susdEquivalent, amount).mul(10**uint(inputEquivalent.decimals())).div(
```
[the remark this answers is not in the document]
noahlitvin: Great catch. I think this needs to be changed per: https://github.com/Synthetixio/synthetix/pull/1667/commits/ec7026c643c38c0c65f8d401e3a43c5c7158b30c

The updated testing confirms this, but I still need to think on this a little more.

justinjmoses (approved):

PR #1671, opened 2022-01-28
Alshephina release v2.58

artdgn:
https://www.notion.so/Alsephina-v2-58-baf4b481d64440928d9820b978b2c391


contracts/migrations/Migration_Alsephina.sol
```diff
@@ -0,0 +1,229 @@
+pragma solidity ^0.5.16;
```
justinjmoses, line 1: no need to check this in - it's kovan

contracts/migrations/Migration_Alkaid.sol
```diff
@@ -1,1038 +0,0 @@
-pragma solidity ^0.5.16;
```
justinjmoses, line 1: why delete these?

dbeal-eth: sorry, they shouldn't have been delteed. we made a small change to the `BaseMigration.sol` file and didn't realize it would affect the previous contracts (that change has since been reverted since it was just removing the `onlyDeployer` function since it won't be used in future migration contracts)

publish/deployed/mainnet/params.json
```diff
@@ -1,4 +1,8 @@
 [
+	{
+		"name": "EXCHANGE_DYNAMIC_FEE_ROUNDS",
+		"value": "0"
```
justinjmoses, line 4: It's meant to be 0 for L1? So disabled on mainnet?

barrasso: Yes - confirmed with Afif. 

> "That's correct. We don't want it to be active on L1, only L2."

justinjmoses (approved):

publish/src/commands/deploy/generate-solidity-output.js
```diff
@@ -206,7 +206,7 @@ contract Migration_${releaseName} is BaseMigration {
 			.join('\n\t\t')}
 	}
 
-	function migrate(address currentOwner) external onlyDeployer {
```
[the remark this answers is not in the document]
barrasso: cc @dbeal-eth [He can't keep getting away with it!](https://www.youtube.com/watch?v=a3_PPdjD6mg)

PR #1673, opened 2022-02-01
Futures dynamic fee

artdgn:
Use dynamic fee during modifying positions actions in futures:
- Fee is applied to trades (modify position, close position, next price orders)
- Fee is not applied to liquidations (different fee structure)

Dynamic fee over max reverts modifying positions, but does not revert margin actions and liquidations.

### DIFF
--- a/contracts/ExchangeCircuitBreaker.sol
+++ b/contracts/ExchangeCircuitBreaker.sol
@@ -201,13 +201,6 @@ contract ExchangeCircuitBreaker is Owned, MixinSystemSettings, IExchangeCircuitB
         return false;
     }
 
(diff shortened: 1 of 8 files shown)


contracts/ExchangeCircuitBreaker.sol
```diff
@@ -201,13 +201,6 @@ contract ExchangeCircuitBreaker is Owned, MixinSystemSettings, IExchangeCircuitB
         return false;
     }
 
-    // ========== MODIFIERS ==========
```
artdgn, line 204: unused

jacko125 (approved):
LGTM

PR #1691, opened 2022-02-08
address preliminary audit comments

artdgn:
Fixes and comment clarifications addressing the questions in the preliminary audit in https://github.com/iosiro/synthetix-audits/pull/7

### DIFF
--- a/contracts/FuturesMarketBase.sol
+++ b/contracts/FuturesMarketBase.sol
@@ -223,12 +223,7 @@ contract FuturesMarketBase is MixinFuturesMarketSettings, IFuturesMarketBaseType
         // marketSize is in baseAsset units so we need to convert from USD units
         require(price > 0, "price can't be zero");
         uint skewScaleBaseAsset = _skewScaleUSD(baseAsset).divideDecimal(price);
-
-        // parameters may not be set, don't divide by zero
-        if (skewScaleBaseAsset == 0) {
-            return 0;
-        }
-
+        require(skewScaleBaseAsset != 0, "skewScale is zero"); // don't divide by zero
         return int(marketSkew).divideDecimal(int(skewScaleBaseAsset));
     }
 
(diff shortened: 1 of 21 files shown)


contracts/FuturesMarketBase.sol
```diff
@@ -223,12 +223,7 @@ contract FuturesMarketBase is MixinFuturesMarketSettings, IFuturesMarketBaseType
         // marketSize is in baseAsset units so we need to convert from USD units
         require(price > 0, "price can't be zero");
         uint skewScaleBaseAsset = _skewScaleUSD(baseAsset).divideDecimal(price);
-
-        // parameters may not be set, don't divide by zero
-        if (skewScaleBaseAsset == 0) {
-            return 0;
-        }
-
+        require(skewScaleBaseAsset != 0, "skewScale is zero"); // don't divide by zero
```
jacko125, line 226: Can there be a scenario where this is 0? Like we actually don't want any proportional skew scale for the asset?

contracts/FuturesMarketBase.sol
```diff
@@ -600,9 +611,7 @@ contract FuturesMarketBase is MixinFuturesMarketSettings, IFuturesMarketBaseType
     // True if and only if two positions a and b are on the same side of the market;
     // that is, if they have the same sign, or either of them is zero.
     function _sameSide(int a, int b) internal pure returns (bool) {
-        // Since we only care about the sign of the product, we don't care about overflow and
-        // aren't using SignedSafeDecimalMath
-        return 0 <= a * b;
+        return (a >= 0) == (b >= 0);
```
jacko125, line 614: Why change the same side check here from a positive multiplication check?

jacko125 (changes requested):

contracts/FuturesMarketBase.sol
artdgn: It's not allowed to be zero when setting the parameter in settings contract, because it doesn't make sense to set a denominator param to 0. 

If we need the scale to be small we can always set it to some small non-zero value, like 1 wei, and it will effectively work like a 0 (to make funding rate 100% of max).

contracts/FuturesMarketBase.sol
artdgn: because the original could underflow and return the negative for two positives

jacko125: Waiting for us to move to solidity v0.8 this would've had Safemath inbuilt

jacko125 (approved):
Lgtm!

PR #1692, opened 2022-02-08
Futures pausing

artdgn:
Adding futures specific SystemStatus suspend and resume actions and views.

FuturesSettings will still be able to perform parameter changes when markets or system is suspended (but not if price is invalid).

### DIFF
--- a/contracts/FuturesMarketBase.sol
+++ b/contracts/FuturesMarketBase.sol
@@ -128,7 +128,7 @@ contract FuturesMarketBase is MixinFuturesMarketSettings, IFuturesMarketBaseType
 
     /*
      * This holds the value: sum_{p in positions}{p.margin - p.size * (p.lastPrice + fundingSequence[p.lastFundingIndex])}
-     * Then marketSkew * (_assetPrice() + _nextFundingEntry()) + _entryDebtCorrection yields the total system debt,
+     * Then marketSkew * (price + _nextFundingEntry()) + _entryDebtCorrection yields the total system debt,
      * which is equivalent to the sum of remaining margins in all positions.
      */
     int128 internal _entryDebtCorrection;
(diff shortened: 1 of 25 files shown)


contracts/FuturesMarketBase.sol
```diff
@@ -675,12 +688,26 @@ contract FuturesMarketBase is MixinFuturesMarketSettings, IFuturesMarketBaseType
...
+    /**
      * Pushes a new entry to the funding sequence at the current price and funding rate.
+     * @dev Admin only method accessible to FuturesMarketSettings. This is admin only because:
+     * - When system parameters change, funding should be recomputed, but system may be paused
+     *   during that time for any reason, so this method needs to work even if system is paused.
+     *   But in that case, it shouldn't be accessible to external accounts.
      */
     function recomputeFunding() external returns (uint lastIndex) {
+        // only FuturesMarketSettings is allowed to use this method
         _revertIfError(msg.sender != _settings(), Status.NotPermitted);
```
barrasso, line 700: should this be `address(_settings())`?

[np] Maybe consider using a modifier here instead to indicate it is a restricted external function:

```
    function recomputeFunding() external returns (uint lastIndex) onlySettings {...}

    function _onlySettings() internal view {
        require(msg.sender == address(_settings()), Status.NotPermitted);
    }

    modifier onlySettings() {
        _onlySettings(); // Use an internal function to save code size.
        _;
    }
```

barrasso (approved):
LGTM, couple of minor nitpicks

contracts/FuturesMarketBase.sol
artdgn: `_settings()` already returns and address.

Not sure about modifier: there's just one method that has this restriction now, and I'm not sure modifiers are more noticeable / readable / explicit than a check at the start of the method.

PR #1694, opened 2022-02-08
SIP-165 Debt Orcale

dbeal-eth:
In this PR you will find impl for [SIP-165](https://sips.synthetix.io/sips/sip-165/).

From a code perspective, this entails:
* `FeePool` and `Issuer` are changed to use the chainlink oracle instead of `DebtCache` and `SynthetixDebtShare` directly
* `FeePool` now closes its fee period across networks (using optimism relay call) to allow for synchronized sharing of close parameters between networks
* For testing and initial deployment, a dummy oracles `SingleNetworkAggregatorDebtRatio` and `SingleNetworkAggregatorIssuedSynths` are utilized to retrieve debt values for this network, meaning most unit tests can work exactly the same as before.

Notes:
* The SC has indicated that inflation should be divided evenly between networks based on amount of debt shares on each network, so this has been implemented. Also, fees will remain on the network they originate from for the time being
* Dual Integration test was added to verify fee pool closure
* Tests were removed from `DebtCache` because the functionality is no longer used within the system, but the actual code from solidity was not removed because there is no need to include `DebtCache` in an update. Doing so would require more migration complexity and it would be better if we could avoid that, so no changes have been made to `DebtCache` for the time being.

The release process for this SIP is 2 steps:
* First, we will release as usual with the included SingleNetworkAggregators, which will preserve current functionality while enabling for us to start reading from an oracle interface for debt info
* Second, we will use the pdao to change the AddressResolver setting for the two aggregators to be the chainlink provided ones, which will effectively complete the debt synthethsis and enable synth fungibility


test/integration/dual/staking.integration.js
```diff
@@ -0,0 +1,122 @@
...
+
+const { toBytes32 } = require('../../../index');
+
+const { exchangeSomething } = require('../utils/exchanging');
+const { ensureBalance } = require('../utils/balances');
+const { skipFeePeriod, skipMinimumStakeTime } = require('../utils/skip');
+
+const ethers = require('ethers');
+
+describe('stakingt & claiming integration tests (L1, L2)', () => {
```
barrasso, line 12: [np] typo: "stakingt"

barrasso (commented):
First pass LGTM. Some minor comments

contracts/SingleNetworkAggregators.sol
```diff
@@ -0,0 +1,142 @@
...
+import "./AddressResolver.sol";
+import "./interfaces/IDebtCache.sol";
+import "./interfaces/ISynthetixDebtShare.sol";
+import "./interfaces/AggregatorV2V3Interface.sol";
+
+import "./SafeDecimalMath.sol";
+
+// aggregator which reports the data from the system itself
+// useful for testing
+contract BaseSingleNetworkAggregator is Owned, AggregatorV2V3Interface {
```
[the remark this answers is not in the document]
dbeal-eth: its not a test helper, it will be deployed and used initially until chainlink oracles are ready, or as a backup

contracts/FeePool.sol
```diff
@@ -162,6 +177,26 @@ contract FeePool is Owned, Proxyable, LimitedSetup, MixinSystemSettings, IFeePoo
         return getTargetThreshold();
     }
 
+    function allNetworksSnxBackedDebt() internal view returns (uint256 debt, uint256 updatedAt) {
```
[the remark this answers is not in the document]
dbeal-eth: do you know why internal functions have underscores btw? I still have yet to figure this out...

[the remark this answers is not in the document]
barrasso: It's just a naming convention for functions and variables that are supposed to be internal or private to the contract. 
Sometimes it's also used to avoid shadowing variables. 

Technically it isn't needed (unless you got a linter like solint), but it does improve readability when scanning through contracts imo.
 
Like without reading all the keywords I can quickly say: "OK, this function is internal so it's not called by anyone else."

test/integration/dual/closeFeePeriod.integration.js
```diff
@@ -0,0 +1,70 @@
...
+
+			before('record current fee period', async () => {
+				prevFeePeriod = await FeePool.recentFeePeriods(0);
+			});
+
+			before('wait for close finalization', async () => {
+				await finalizationOnL2({ ctx, transactionHash: closeReceipt.transactionHash });
+			});
+
+			it('has new fee pool', async () => {
```
artdgn, line 62: nit: `new fee period`

test/integration/dual/closeFeePeriod.integration.js
```diff
@@ -0,0 +1,70 @@
...
+				const newFeePeriod = await FeePool.recentFeePeriods(0);
+				assert.bnNotEqual(newFeePeriod.feePeriodId, prevFeePeriod.feePeriodId);
+			});
+
+			it('has correct params', async () => {});
```
artdgn, line 67: would it make sense to add a sanity check that L2 debt has indeed effect on claims when closing from L1?

artdgn (approved):
Various small issues and questions. 

From futures point of view, as long as `takeDebtSnapshot()` is being respected by each chain for counting the debt (and it looks like it's being respected in `Issuer._totalIssuedSynths()` by calling `debtCache().cacheInfo()`), I see no issue.

contracts/FeePool.sol
```diff
@@ -162,6 +177,26 @@ contract FeePool is Owned, Proxyable, LimitedSetup, MixinSystemSettings, IFeePoo
...
+        (, int256 rawData, , uint timestamp, ) = AggregatorV2V3Interface(requireAndGetAddress(CONTRACT_EXT_AGGREGATOR_ISSUED_SYNTHS))
+            .latestRoundData();
+        
+        debt = uint(rawData);
+        updatedAt = timestamp;
+    }
+
+    function allNetworksDebtSharesSupply() internal view returns (uint256 sharesSupply, uint256 updatedAt) {
```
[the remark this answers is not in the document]
dbeal-eth: reccomended to just look directly at the aggregator provided by chainlink--shortly after release, we will provide additional subgraph functionality to record the data on the aggregator

test/contracts/BaseSynthetix.js
```diff
@@ -689,13 +692,17 @@ contract('BaseSynthetix', async accounts => {
 						baseSynthetix.issueSynths(toUnit('1'), { from: account2 }),
 					]);
 
+					// make aggregator debt info rate stale
```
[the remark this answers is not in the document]
dbeal-eth: its a 2 part operation, so forst we set timestamp to current time, and then it fast forwards past the stale period after that

test/integration/dual/closeFeePeriod.integration.js
dbeal-eth: 86b41d87

test/integration/dual/closeFeePeriod.integration.js
dbeal-eth: for now removed this empty test case. validation of fee closure routine is covered in a nother integration test

contracts/FeePool.sol
```diff
@@ -162,6 +177,26 @@ contract FeePool is Owned, Proxyable, LimitedSetup, MixinSystemSettings, IFeePoo
...
+        (, int256 rawIssuedSynths, , uint issuedSynthsUpdatedAt, ) = AggregatorV2V3Interface(requireAndGetAddress(CONTRACT_EXT_AGGREGATOR_ISSUED_SYNTHS))
```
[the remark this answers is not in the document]
dbeal-eth: will be impl

contracts/Issuer.sol
```diff
@@ -60,11 +61,16 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
 
     bytes32 public constant CONTRACT_NAME = "Issuer";
 
+    // SIP-65: Decentralized circuit breaker
```
artdgn, line 64: nit: 165

contracts/Issuer.sol
```diff
@@ -809,6 +836,37 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
         }
     }
 
+    function _verifyCircuitBreaker() internal returns (bool) {
+        (, int256 rawRatio, , uint ratioUpdatedAt, ) = AggregatorV2V3Interface(requireAndGetAddress(CONTRACT_EXT_AGGREGATOR_DEBT_RATIO))
```
artdgn, line 840: should the second oracle be checked as well, them being separate contracts and possibly failing not at the same time?

contracts/Issuer.sol
```diff
@@ -809,6 +836,37 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
         }
     }
 
+    function _verifyCircuitBreaker() internal returns (bool) {
```
artdgn, line 839: nit: return value is unclear from the name, maybe `_circuitBreakerOk`?

contracts/Issuer.sol
```diff
@@ -809,6 +836,37 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+        return true;
+    }
+
+    function _calculateDeviation(
+        uint last,
+        uint fresh
+    ) internal pure returns (uint deviation) {
+        if (last == 0) {
+            deviation = 1;
+        } else if (fresh == 0) {
```
artdgn, line 861: `fresh == 0` should be first condition to check, if both are 0 (fresh contract, and oracle misconfigured), deviation returned is 1.

test/contracts/Issuer.js
```diff
@@ -1269,9 +1269,55 @@ contract('Issuer (via Synthetix)', async accounts => {
...
 				});
+
+				it('circuit breaks when debt changes dramatically', async () => {
+					await synthetix.transfer(account1, toUnit('10000'), {
+						from: owner,
+					});
+
+					// They should now be able to issue sUSD
+					await synthetix.issueSynths(toUnit('100'), { from: account1 });
+					await synthetix.issueSynths(toUnit('1'), { from: account1 });
```
artdgn, line 1280: nit: is this line needed?

test/contracts/Issuer.js
```diff
@@ -1269,9 +1269,55 @@ contract('Issuer (via Synthetix)', async accounts => {
 					// And trying to issue the smallest possible unit of one should fail.
 					await assert.revert(synthetix.issueSynths('1', { from: account1 }), 'Amount too large');
 				});
+
+				it('circuit breaks when debt changes dramatically', async () => {
```
artdgn, line 1273: would be wise to check the edge cases as well
- last == 0
- fresh == 0
- fresh == 0 and last == 0 (which should fail with current code I believe)

artdgn (changes requested):
Two main points:
- Checking the values coming from the other oracle
- Logical error in condition check order

test/contracts/Issuer.js
dbeal-eth: yes

contracts/Issuer.sol
dbeal-eth: oracle may return `0` if there is no debt in the system, and that is not an error. if last is not set, then we have no data to go off of, so we shouldn't break the circuit.

whats important is that the value doesn't *become* 0 during normal operation, which the next condition will successfully catch.

so, in order for this case to occur, we have to be doing a release, and the oracle has to suddenly be misconfigured. 

I think we are ok.

contracts/Issuer.sol
dbeal-eth: the function also has the side effect of triggering issuance suspension so i thought calling it an `_is` function would have worse problem.

contracts/Issuer.sol
dbeal-eth: believe it or not, the second oracle (hereafter referred to as the `IssuedSynths` oracle) is not actually used to measurable extent in the `Issuer`, because if you look at the one function where `allNetworks

Also the `IssuedSynths` oracle, unlike the `DebtRatio` oracle or other price oracles, is sensitive to legitimate changes in value, where it can swing wildly in either direction if a large staker mints or burns.

Right now, the only thing the `IssuedSynths` oracle is really used for is calculating the amount of SNX to mint in `FeePool`, but that only happens once a week. If there is a mis-mint of debt, it can be corrected (since the minted supply will be sitting undistributed for most of a week)

Since both the `IssuedSynths` and `DebtRatio` are pushed by the same oracle code, if there was actually an issue that affected `IssuedSynths`, it would likely affect `DebtRatio` as well, and trigger the circuit breaker anyway, so because of all the reasons above, its been left out of the breaker check

test/contracts/Issuer.js
dbeal-eth: fresh == 0 and last == 0 is actually verified as part of this test, since it would fail or trigger issuance suspension on the first `issueSynths` call.

`fresh == 0` and `last != 0` would happen if either 1) all debt was burned, or 2) the oracle is borked. in either case, we would expect it to suspend. This test can be added.

`fresh != 0` and `last == 0` only happens if the contract is uninitialized, so it has no data to go off of and shouldn't break. to test this, we would have to deploy and hook up a new Issuer in the unit test I believe

contracts/Issuer.sol
dbeal-eth: anothre option is to call it `_operateCircuitBreaker`

contracts/Issuer.sol
dbeal-eth: c375a938

test/contracts/Issuer.js
dbeal-eth: tightened up the tests to verify that `totalIssuedSynths` is actually 0 at the beginning of the test, and to burn all synths in the burn circuit breaker test c375a938

contracts/Issuer.sol
artdgn: > so, in order for this case to occur, we have to be doing a release, and the oracle has to suddenly be misconfigured

Not suddenly, it might be misconfigured by the same release

> oracle may return 0 if there is no debt in the system

That's a fair point, but it only brings out a more important point - how do we tell between "oracle is broken" and "there is no debt".

contracts/Issuer.sol
artdgn: The deviation check might be the wrong check, but some validation still needs to be done. Surely the oracle can return bad values, and surely not every value can be a legitimate value. One possibility is to maybe to combine subtraction and division. If the diff is more that 1M and the deviation is more than 30% - break circuit. This way large values are possible when large values already exist, but large deviations for small values don't break the circuit.

artdgn: > Right now, the only thing the IssuedSynths oracle is really used for

and 

> Since both the IssuedSynths and DebtRatio are pushed by the same oracle code

Is IMO a combination of assumptions that might not hold true, and surely the code doesn't ensure this. This might still result in bad values returned and more trouble than adding basic validation on the numbers returned from oracle.

What is the risk / downside of adding those checks?

test/contracts/Issuer.js
artdgn: > Guess what? thats tested here also (thats why other comment from earlier "Is this really needed", yes it is needed or else it fails because last == 0 still)

might be good to add a comment there explaining this gotcha to make the test maintainer's life easier

test/contracts/Issuer.js
artdgn: please a code comment to explain it (the gotcha), as I still don't get it (and I would be puzzled if I were to touch this test)

contracts/Issuer.sol
barrasso: `_checkCircuitBreaker`?  😀

OR

call it `_circuitBreakerDidSuspendIssuance` and invert the boolean values

contracts/Issuer.sol
barrasso: Don't see any risk in adding these checks - I also think we should have moar validation. But I do see the downside of losing time testing on kovan. Let's consider which of those checks are absolutely necessary and whip up another PR in the meantime.

barrasso (approved):
I think this is good to merge into develop and test on Kovan for now. 
Also, the auditors haven't raised any major concerns. 

We can consider additional verification checks & circuit breakers in parallel, and then create a new PR later if needed.

PR #1704, opened 2022-02-18
Consolidating on ProxySynthetix and ProxysUSD

justinjmoses:


dbeal-eth (approved):
this is going to prevent so many bugs...

PR #1709, opened 2022-03-01
SIP-198: Atomic Configuration Getter Method

noahlitvin:
Implementation of the suggestion here, for discussion: https://github.com/Synthetixio/synthetix/pull/1667#discussion_r802255506

### DIFF
--- a/contracts/ExchangerWithFeeRecAlternatives.sol
+++ b/contracts/ExchangerWithFeeRecAlternatives.sol
@@ -274,7 +274,8 @@ contract ExchangerWithFeeRecAlternatives is MinimalProxyFactory, Exchanger {
         sourceCurrencyKey;
 
         // Get the exchange fee rate as per destination currencyKey
-        uint baseRate = getAtomicExchangeFeeRate(destinationCurrencyKey);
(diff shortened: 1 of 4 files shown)


noahlitvin:
@artdgn Makes sense! See the latest commit and let me know.

contracts/ExchangerWithFeeRecAlternatives.sol
```diff
@@ -274,7 +274,8 @@ contract ExchangerWithFeeRecAlternatives is MinimalProxyFactory, Exchanger {
         sourceCurrencyKey;
 
         // Get the exchange fee rate as per destination currencyKey
-        uint baseRate = getAtomicExchangeFeeRate(destinationCurrencyKey);
+        AtomicExchangeConfig memory destinationAtomicConfig = getAtomicExchangeConfig(destinationCurrencyKey);
```
artdgn, line 277: was passing the config into this method one from calling methods not a viable option?

artdgn (approved):

contracts/ExchangerWithFeeRecAlternatives.sol
noahlitvin: It looks like it's called from two places in the file where the rest of the config isn't needed, so I think this ends up being the cleanest.

PR #1710, opened 2022-03-06
SIP-204: Synth Teleporters

barrasso:
**WIP**: *This is an implementation of [SIP-204](https://sips.synthetix.io/sips/sip-204/).*

The `Issuer` uses the `SystemMessenger` to post a message to the target chain in order to "teleport" synths over.
Essentially this means the synths are burned on the source chain and then re-issued on the target chain (minus fees).

## TODO
- [x] deploy scripts
- [x] unit tests
- [ ] integration tests
- [x] fresh deploy to testnets
  - [x] `rinkeby`
  - [x] `goerli`
  - [x] `arbitrum-rinkeby`


contracts/BaseSynthetix.sol
```diff
@@ -297,6 +304,15 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         return issuer().burnSynthsToTargetOnBehalf(burnForAddress, messageSender);
     }
 
+    function teleportSynth(uint targetChainId, bytes32 currencyKey, address from, uint amount) external issuanceActive optionalProxy returns (bool) {
+        return issuer().teleportSynth(targetChainId, currencyKey, from, amount);
+    }
+
+    function receiveTeleportedSynth(uint targetChainId, bytes32 currencyKey, address from, uint amount) external issuanceActive optionalProxy returns (bool) {
```
dbeal-eth, line 311: rather than having the `systemMessenger` send a message to `BaseSynthetix`, it should just send the message directly to `issuer`. Reduces the chain of needed calls that way

contracts/Issuer.sol
```diff
@@ -597,6 +604,28 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+
+        bool successfullyTeleported = _teleportSynth(targetChainId, currencyKey, from, amount);
+        require(successfullyTeleported, "Issuer: Teleport failed");
+
+        emit SynthTeleported(currencyKey, from, amount);
+
+        return successfullyTeleported;
+    }
+
+    function receiveTeleportedSynth(uint targetChainId, bytes32 currencyKey, address from, uint amount) external onlySynthetix returns (bool) {
```
dbeal-eth, line 618: target chain ID shouldn't be needed here?

contracts/Issuer.sol
```diff
@@ -766,6 +795,64 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+        // What is their debt in sUSD?
+        (uint debtBalance, uint totalDebtIssued, bool anyRateIsInvalid) =
+            _debtBalanceOfAndTotalDebt(synthetixDebtShare().balanceOf(from), sUSD);
+        _requireRatesNotInvalid(anyRateIsInvalid);
+
+        // Calculate the teleportation fee
+        uint teleportFee = amount.multiplyDecimal(getTeleportFeeRate());
+
+        // Make sure they have enough synths available to teleport (plus fees)
+        require(IERC20(address(synths[currencyKey])).balanceOf(from) >= amount.add(teleportFee), "Issuer: not enough synths to teleport");
```
dbeal-eth, line 811: this brings up an interesting question of which is better: send the amount the user requested minus the fee, or the amount the user requested is what should be on destination chain, and then the fee is added on top of that?

for helping users to send their full account balance to another chain, it seems like the former would be better, and I don't see much of a downside either.

contracts/Issuer.sol
```diff
@@ -766,6 +795,64 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+
+        // Burn the synths first before teleporting
+        _burnSynths(from, from, amount.add(teleportFee), debtBalance, totalDebtIssued);
```
dbeal-eth, line 814: here, we actually don't call `_burnSynths`. Instead, you should just call `burn` on the Synth from the user's account with the requested amount.

this is because when a user sends debt to another chain, their debt is not forgiven (as `_burnDebt` will do), its simply being moved to another chain.

contracts/Issuer.sol
```diff
@@ -766,6 +795,64 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+            if (currencyKey != sUSD) {
+                teleportFee = exchangeRates().effectiveValue(currencyKey, teleportFee, sUSD);
+            }
+            synths[sUSD].issue(feePool().FEE_ADDRESS(), teleportFee);
+            feePool().recordFeePaid(teleportFee);
+        }
+
+        // Send a message to the target chain to validate the teleported synths
+        bytes memory encodedMessage = abi.encode("receiveTeleportedSynth", targetChainId, currencyKey, from, amount);
+        systemMessenger().post(targetChainId, CONTRACT_NAME, encodedMessage, 0);
```
dbeal-eth, line 827: here it appears the `SystemMessenger` is being instructed to send a message directly to `Issuer`, but it doesn't have permission for the cal lto succeed. With the current code impl, it has to go through `Synthetix` first (however, I would reccomend changing `Issuer` to accept calls from `SystemMessenger` directly, as I pointed out in another comment)

contracts/Issuer.sol
```diff
@@ -766,6 +795,64 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+
+        return true;
+    }
+
+    function _receiveTeleportedSynth(uint targetChainId, bytes32 currencyKey, address from, uint amount) internal returns (bool) {
+        // TODO: Verify message integrity?
+        require(targetChainId > 0, "Issuer: received invalid target chain id");
```
dbeal-eth, line 834: this check is unnecessary because the `SystemMessenger` validates chainID already. Instead, this function shouldn't really care about the chain ID. it shouldn't be a param

contracts/Issuer.sol
```diff
@@ -766,6 +795,64 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+        require(amount <= maxIssuable, "Issuer: amount too large to issue");
+
+        // Keep track of the debt they're about to create
+        _addToDebtRegister(from, amount, totalSystemDebt);
+
+        // Record issue timestamp
+        _setLastIssueEvent(from);
+
+        // Create their synths
+        synths[currencyKey].issue(from, amount);
```
dbeal-eth, line 848: you should be doing the same thing in the `_teleportSynth` function as well, `synths[currencyKey].burn(from, amount)`, as mentioned in previous comment

contracts/Issuer.sol
```diff
@@ -766,6 +795,64 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+        // TODO: Verify message integrity?
+        require(targetChainId > 0, "Issuer: received invalid target chain id");
+
+        // Make sure we can issue these synths
+        (uint maxIssuable, , uint totalSystemDebt, bool anyRateIsInvalid) = _remainingIssuableSynths(from);
+        _requireRatesNotInvalid(anyRateIsInvalid);
+        require(amount <= maxIssuable, "Issuer: amount too large to issue");
+
+        // Keep track of the debt they're about to create
+        _addToDebtRegister(from, amount, totalSystemDebt);
```
dbeal-eth, line 842: the teleporter is not sending their debt to another chain, the synths are simply being teleported. Therefore, this call is not needed.

contracts/Issuer.sol
```diff
@@ -833,4 +920,6 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
 
     event SynthAdded(bytes32 currencyKey, address synth);
     event SynthRemoved(bytes32 currencyKey, address synth);
+    event SynthTeleported(bytes32 currencyKey, address from, uint amount);
+    event SynthReceived(bytes32 currencyKey, address from, uint amount);
```
dbeal-eth, line 924: if there was one reason to have `targetChainId` in the functions, it is to be able to include it in the event here.

for record purposes, it is definitely a good idea to include the `chainID` within the events here since its a lot harder to track otherwise

contracts/Issuer.sol
```diff
@@ -766,6 +795,64 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+        if (teleportFee > 0) {
+            if (currencyKey != sUSD) {
+                teleportFee = exchangeRates().effectiveValue(currencyKey, teleportFee, sUSD);
+            }
+            synths[sUSD].issue(feePool().FEE_ADDRESS(), teleportFee);
+            feePool().recordFeePaid(teleportFee);
+        }
+
+        // Send a message to the target chain to validate the teleported synths
+        bytes memory encodedMessage = abi.encode("receiveTeleportedSynth", targetChainId, currencyKey, from, amount);
```
dbeal-eth, line 826: :ok_hand:

dbeal-eth (approved):
fix comments, and the unit tests should be very straightforward if you mock `SystemMessenger`

dbeal-eth (commented):
didnt mean to approve :laughing:

contracts/Issuer.sol
barrasso: resolved https://github.com/Synthetixio/synthetix/commit/a1c6f23a057ecc9739c536700b8cb6777b38b0db

contracts/Issuer.sol
barrasso: Former sounds good

resolved https://github.com/Synthetixio/synthetix/commit/a1c6f23a057ecc9739c536700b8cb6777b38b0db

contracts/Issuer.sol
```diff
@@ -766,6 +795,64 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
         }
     }
 
+    function _teleportSynth(uint targetChainId, bytes32 currencyKey, address from, uint amount) internal returns (bool) {
+        // Ensure waitingPeriod and sUSD balance is settled as burning impacts the size of debt pool
+        require(!exchanger().hasWaitingPeriodOrSettlementOwing(from, sUSD), "Issuer: sUSD needs to be settled");
+
+        // What is their debt in sUSD?
+        (uint debtBalance, uint totalDebtIssued, bool anyRateIsInvalid) =
```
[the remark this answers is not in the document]
barrasso: resolved https://github.com/Synthetixio/synthetix/commit/a1c6f23a057ecc9739c536700b8cb6777b38b0db

contracts/Issuer.sol
barrasso: resolved https://github.com/Synthetixio/synthetix/commit/a1c6f23a057ecc9739c536700b8cb6777b38b0db

contracts/Issuer.sol
barrasso: resolved https://github.com/Synthetixio/synthetix/commit/a1c6f23a057ecc9739c536700b8cb6777b38b0db

contracts/Issuer.sol
barrasso: resolved https://github.com/Synthetixio/synthetix/commit/a1c6f23a057ecc9739c536700b8cb6777b38b0db

contracts/Issuer.sol
barrasso: resolved https://github.com/Synthetixio/synthetix/commit/2f07a419955a20d1fcecf70dfb27fe80f345715e

PR #1711, opened 2022-03-07
Futures single market pausing

artdgn:
This adds the ability to pause a single futures market. Previously the whole futures system was pausable, this allows to pause a specific market. The following circumstances may require this:
- Some problem with a particular market that needs requires pausing.
- Pausing due to trading hours of the corresponding underlying asset.
- Migration related activities.

Also added a test to ensure that it's possible to pause funding rate for the duration of the pause by setting max funding rate to 0 before/after pausing and resetting it to correct value before/after resumption.

### DIFF
--- a/contracts/FuturesMarketBase.sol
+++ b/contracts/FuturesMarketBase.sol
@@ -627,8 +627,8 @@ contract FuturesMarketBase is MixinFuturesMarketSettings, IFuturesMarketBaseType
      * This is mutative because the circuit breaker stores the last price on every invocation.
      */
     function _assetPriceRequireSystemChecks() internal returns (uint) {
-        // check that futures markets aren't suspended, revert with appropriate message
-        _systemStatus().requireFuturesActive();
+        // check that futures market isn't suspended, revert with appropriate message
+        _systemStatus().requireFuturesMarketActive(baseAsset); // asset and market may be different
         // check that synth is active, and wasn't suspended, revert with appropriate message
         _systemStatus().requireSynthActive(baseAsset);
         // check if circuit breaker if price is within deviation tolerance and system & synth is active
--- a/contracts/SystemStatus.sol
+++ b/contracts/SystemStatus.sol
@@ -29,6 +29,8 @@ contract SystemStatus is Owned, ISystemStatus {
 
     mapping(bytes32 => Suspension) public synthSuspension;
 
+    mapping(bytes32 => Suspension) public futuresMarketSuspension;
+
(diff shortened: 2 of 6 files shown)


barrasso (approved):

contracts/FuturesMarketBase.sol
```diff
@@ -627,8 +627,8 @@ contract FuturesMarketBase is MixinFuturesMarketSettings, IFuturesMarketBaseType
      * This is mutative because the circuit breaker stores the last price on every invocation.
      */
     function _assetPriceRequireSystemChecks() internal returns (uint) {
-        // check that futures markets aren't suspended, revert with appropriate message
-        _systemStatus().requireFuturesActive();
+        // check that futures market isn't suspended, revert with appropriate message
+        _systemStatus().requireFuturesMarketActive(baseAsset); // asset and market may be different
```
jacko125, line 631: good

contracts/SystemStatus.sol
```diff
@@ -68,6 +70,14 @@ contract SystemStatus is Owned, ISystemStatus {
         _internalRequireFuturesActive();
     }
 
+    /// @notice marketKey doesn't necessarily correspond to asset key
+    function requireFuturesMarketActive(bytes32 marketKey) external view {
+        _internalRequireSystemActive();
+        _internalRequireExchangeActive(); // exchanging implicitely used
```
jacko125, line 76: @artdgn  If exchange is paused, should we assume that futures is paused as well?

artdgn: Yep. 

However individual synth exchanging can be paused separately (and futures will work), so that e.g. we can have futures for a synth working but not exchanger trades.

So if we need to pause both for a specific synth we need to pause the synth itself.

contracts/SystemStatus.sol
```diff
@@ -133,6 +143,21 @@ contract SystemStatus is Owned, ISystemStatus {
         }
     }
 
+    /// @notice marketKey doesn't necessarily correspond to asset key
+    function getFuturesMarketSuspensions(bytes32[] calldata marketKeys)
```
jacko125, line 147: marketKey is the address of the market ? or just the baseAsset?

artdgn: marketKey is just an identifier name for a market within the futures system (manager ans settings), so in this code it's just the same as `baseAsset`, but it can be stuff like `sBTC-perp`, or `sBTC-perp-v2` or something like that.

jacko125 (approved):
lgtm !

PR #1714, opened 2022-03-08
Futures volume source fee methods

artdgn:
Adds methods to emit tracking event for volume source fee integrations.

### DIFF
--- a/contracts/FuturesMarketBase.sol
+++ b/contracts/FuturesMarketBase.sol
@@ -96,6 +96,10 @@ contract FuturesMarketBase is MixinFuturesMarketSettings, IFuturesMarketBaseType
 
     /* ========== STATE VARIABLES ========== */
 
+    // The market identifier in the futures system (manager + settings). Multiple markets can co-exist
+    // for the same asset in order to allow migrations.
+    bytes32 public marketKey;
+
     // The asset being traded in this market. This should be a valid key into the ExchangeRates contract.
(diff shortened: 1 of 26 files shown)


barrasso (approved):
LGTM

dbeal-eth (approved):
:pray:

contracts/FuturesMarketBase.sol
```diff
@@ -901,25 +906,58 @@ contract FuturesMarketBase is MixinFuturesMarketSettings, IFuturesMarketBaseType
      * Reverts if the resulting position is too large, outside the max leverage, or is liquidating.
      */
     function modifyPosition(int sizeDelta) external {
```
jacko125, line 908: Why not remove `modifyPosition` and just keep `modifyPositionWithTracking` as the only function. Then you can leave the trackingCode as blank if not populated.

If it breaks alot of tests then I think leave it

jacko125 (approved):
LGTM

contracts/FuturesMarketBase.sol
artdgn: Yeah, it's mostly to still support the previous interface to reduce the need for urgent changes for tests, UI, scripts, keeper etc.

Also, the old methods are worth keeping anyway, because they're cheaper in terms of calldata (the main gas cost for optimism), so anyone not using the UI will have lower gas costs.

PR #1716, opened 2022-03-09
Debtcache import excluded debt entries

artdgn:
Adds the ability to import excluded debt entries which is needed for upgrading to new DebtCache instance.

### DIFF
--- a/contracts/BaseDebtCache.sol
+++ b/contracts/BaseDebtCache.sol
@@ -213,24 +213,30 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
     /// previous instance of the contract
     /// Also, in addition to this method it's possible to use recordExcludedDebtChange since
     /// it's accessible to owner in case additional adjustments are required
-    function importExcludedIssuedDebts(IDebtCache prevDebtCache) external onlyOwner {
+    function importExcludedIssuedDebts(IDebtCache prevDebtCache, IIssuer prevIssuer) external onlyOwner {
         // this can only be run once so that recorded debt deltas aren't accidentally
         // lost or double counted
-        require(excludedDebtImported == false, "import can only be run once");
+        require(!excludedDebtImported, "import can only be run once");
         excludedDebtImported = true;
 
-        bytes32[] memory currencyKeys = issuer().availableCurrencyKeys();
+        // get the currency keys from **previous** issuer, in case current issuer
+        // doesn't have all the synths at this point
+        // warning: if a synth won't be added to the current issuer before the next upgrade of this contract,
+        // its entry will be lost (because it won't be in the prevIssuer for next time).
+        // if for some reason this is a problem, it should be possible to use recordExcludedDebtChange() to amend
+        bytes32[] memory keys = prevIssuer.availableCurrencyKeys();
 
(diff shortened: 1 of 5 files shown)


contracts/BaseDebtCache.sol
```diff
@@ -31,6 +31,9 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
     uint internal _cacheTimestamp;
     bool internal _cacheInvalid = true;
 
+    // flag to ensure importing excluded debt is invoked only once
+    bool public excludedDebtImported = false; // public to avoid needing an event
```
dbeal-eth, line 35: np, i'd just call it `isInitialized` if you really want to be future proof lol

contracts/BaseDebtCache.sol
```diff
@@ -204,6 +207,34 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
+    /// used when migrating to new DebtCache instance in order to import the excluded debt records
+    /// If this method is not run after upgrading the contract, the debt will be
+    /// incorrect w.r.t to wrapper factory assets until the values are imported from
+    /// previous instance of the contract
+    /// Also, in addition to this method it's possible to use recordExcludedDebtChange since
+    /// it's accessible to owner in case additional adjustments are required
+    function importExcludedIssuedDebts(IDebtCache prevDebtCache) external onlyOwner {
+        // this can only be run once so that recorded debt deltas aren't accidentally
+        // lost or double counted
+        require(excludedDebtImported == false, "import can only be run once");
```
dbeal-eth, line 219: np `excludedDebtImported == false` should be replaced with `!excludedDebtImported`

js linters would throw a hissie fit about this

dbeal-eth (approved):
nice

contracts/BaseDebtCache.sol
jacko125: We do use this pattern elsewhere before of `isInitialized`

contracts/BaseDebtCache.sol
```diff
@@ -204,6 +207,34 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
...
+        excludedDebtImported = true;
+
+        bytes32[] memory currencyKeys = issuer().availableCurrencyKeys();
```
jacko125, line 222: So here we require the synths to be in the issuer() first before executing this function?  Ie, if new issuer is deployed we need to add synths into it first before invoking this function.

jacko125 (approved):

contracts/BaseDebtCache.sol
artdgn: fair point, but I wanted to be specific and save an event :)

contracts/BaseDebtCache.sol
artdgn: good point, will add a check for this in the script to ensure new issuer has synths.

artdgn: actually, better to just import using the keys from the previous issuer

dbeal-eth (approved):

jacko125 (approved):
LGTM!

PR #1721, opened 2022-03-15
futures integration and fork tests

artdgn:


test/integration/utils/bootstrap.js
```diff
@@ -73,12 +73,12 @@ function bootstrapL2({ ctx }) {
 			balance: ethers.utils.parseEther('1000000'),
 		});
 
-		if (!ctx.fork) {
-			startOpsHeartbeat({
-				l1Wallet: ctx.l1mock.users.user9,
-				l2Wallet: ctx.users.user9,
-			});
-		}
+		// if (!ctx.fork) {
```
justinjmoses, line 76: why are you commenting this out? can you add a reason if it's intentional?

justinjmoses (approved):
Just one note about commenting out code that should give a reason

test/integration/utils/bootstrap.js
artdgn: it's causing nonce issues, and has been already commented out for dual (since it was causing issues there), so had to comment it out for L2 as well because it was causing issues there as well.

artdgn: added a comment in the code in both places

barrasso (approved):

PR #1736, opened 2022-03-28
Sip 217 + 212

artdgn:
Changes:
- adds feeds and markets for [SIP 217](https://sips.synthetix.io/sips/sip-217/) on both kovan-ovm and mainnet-ovm
- updates `minInitialMargin` according to the already implemented [SCCP 172](https://sips.synthetix.io/sccp/sccp-172)
- :warning: this adds price feeds into exchange rates for previously existing synths - so if previously only `sBTC` would have a rate, not both `sBTC` but also, `BTC` are valid keys into exchange rates. This is to prevent inconsistencies due to changes of synths / futures markets with time (e.g. synths being added removed, futures markets being added removed).
- adds synths for [SIP 212](https://sips.synthetix.io/sips/sip-212/) but names them according to the source feed (XAU / XAG), other names can be done on the UI level to reduce potential for error and confusion in this repo


barrasso (approved):
approved pending 1 comment

publish/deployed/mainnet-ovm/synths.json
```diff
@@ -48,5 +48,15 @@
 		"name": "sUNI",
 		"asset": "UNI",
 		"subclass": "MultiCollateralSynth"
+	},
```
justinjmoses, line 51: so i believe the job-validation step in CI will fail if you add this and haven't added synths yet - better to just have this PR focus on kovan-ovm and then when planning mainnet release you can add the synths. However adding the feeds for both chains is totally fine and normal.

publish/src/commands/deploy/index.js
```diff
@@ -131,8 +131,6 @@ const deploy = async ({
 		useOvm,
 	});
 
-	const standaloneFeeds = Object.values(feeds).filter(({ standalone }) => standalone);
```
justinjmoses, line 134: so these are feeds that don't have synths - like `SNX` and `ETH` - and are useful to show

publish/deployed/kovan-ovm/synths.json
```diff
@@ -43,5 +43,10 @@
 		"name": "sMATIC",
 		"asset": "MATIC",
 		"subclass": "MultiCollateralSynth"
+	},
+	{
+		"name": "sWTI",
```
justinjmoses, line 48: wasn't the plan to call this `sOIL`? the same with `sGOLD` and `sSILVER`?

index.js
```diff
@@ -345,18 +345,9 @@ const getFeeds = ({ network, path, fs, deploymentPath, useOvm = false } = {}) =>
...
 
-	const synths = getSynths({ network, useOvm, path, fs, deploymentPath, skipPopulate: true });
-
 	// now mix in the asset data
 	return Object.entries(feeds).reduce((memo, [asset, entry]) => {
-		memo[asset] = Object.assign(
-			// standalone feeds are those without a synth using them
-			// Note: ETH still used as a rate for Depot, can remove the below once the Depot uses sETH rate or is
-			// removed from the system
-			{ standalone: !synths.find(synth => synth.asset === asset) || asset === 'ETH' },
```
justinjmoses, line 356: i don't think you should remove these. SNX is definitely still needed as a standalone rate.

publish/src/commands/deploy/index.js
justinjmoses: Have you checked that `SNX` will still be added as an aggregator?

publish/releases.json
```diff
@@ -543,11 +543,21 @@
 			"sources": ["Exchanger"],
 			"released": "both"
 		},
+		{
+			"sip": 212,
+			"layer": "ovm",
+			"sources": ["Synths.+"]
```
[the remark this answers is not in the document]
justinjmoses: > Unless we need to redeploy the synths because the FuturesMarkets are being redeployed?

I can't imagine why we would need to...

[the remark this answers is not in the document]
artdgn: removed the sources from releases json

publish/deployed/kovan-ovm/synths.json
artdgn: as with XAG and XAU, it's probably to keep the names consistent with the feeds throughout the codebase to reduce confusion and mistakes, and only rename them to "friendly" names as close as possible to the UI

PR #1737, opened 2022-03-28
deploy script changes for external use

DominicRomanowski:
Changes:

* Expose certain fields in deploy scripts to allow for external projects to deploy Synthetix within their own environments
* Filter out duplicate synths (probably unnecessary but caused problems in deploy scripts that tried adding synths)


barrasso (approved):

justinjmoses:
Going to merge this to a local branch `incoming` which is even with `develop` to try to fix the integration tests.

PR #1738, opened 2022-03-28
Fixing L1 forks where issuance circuit breaker was insitgating

justinjmoses:


test/integration/utils/issuance.js
```diff
@@ -0,0 +1,21 @@
...
+const { resumeIssuance } = require('./status');
+const { setSystemSetting } = require('./settings');
+
+async function ensureIssuance({ ctx }) {
+	if (ctx.fork) {
+		// Ensure issuance is not suspended for any reason
+		await resumeIssuance({ ctx });
+		// Note: if issuance has been suspended for some time, the circuit breaker could kick in for issuance,
+		// so up the ratio here
+		await setSystemSetting({
```
justinjmoses, line 11: a better solution is to read from the debt ratio aggregator and update the last ratio value from there. coming in a future PR.

PR #1740, opened 2022-03-28
Improving fork logic for issuance

justinjmoses:
* Improving fork logic for issuance to actually update the `setLastDebtRatio` owner function
* Cleanup logic in `Issuer` to only fetch answer not the whole round
* Change `setLastDebtRatio` to `updateDebtRatio` - fetching the latest on-chain from the aggregator (needs to be done after rebuildCache so can't be in constructor)
* Updating deploy script to invoke `updateDebtRatio` function as needed


justinjmoses:
Closing as no longer required.

PR #1743, opened 2022-03-31
Ensure connect-bridges handles approvals where necessary

justinjmoses:


justinjmoses:
Hmmm - looks like the dual int tests use the connect-bridge and it's failing trying to set the approval. Looking...

PR #1754, opened 2022-04-07
SIP-229 Synths Bridge

dbeal-eth:
Implementation for [SIP-229](https://sips.synthetix.io/sips/sip-229/)

Notes:
* chainlink has to do their end of the work to account for comparing the events on the send and receive side to ensure correct accounting of debt
* any synth can theoretically transferred from L2 to L1. However, in reality, we should probably just start with sUSD right now because other synths could have problems (ex. if synth only exists on L2 but not L1, and chainlink would have to account for the expected value of the synths in transit which makes a complicated problem even more complicated)
* Updated `BaseSynthetixBridge` so that we could keep things more symmetric and reduce code duplication.
* withdraw tests are not currently working due to them never being impl to work (this is the same with the withdraw tests that we have had the whole time)


contracts/BaseSynthetixBridge.sol
```diff
@@ -83,9 +104,44 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+
+        emit InitiateSynthTransfer(currencyKey, destination, amount);
+    }
+
+    function finalizeSynthTransfer(
+        bytes32 currencyKey,
+        address destination,
+        uint amount
+    ) external onlyCounterpart {
+        issuer().issueFreeSynths(currencyKey, destination, amount);
```
artdgn, line 134: require currencyKey sUSD?

contracts/BaseSynthetixBridge.sol
```diff
@@ -83,9 +104,44 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+        emit FinalizeSynthTransfer(currencyKey, destination, amount);
+    }
+
     // ========== EVENTS ==========
 
     event InitiationSuspended();
 
     event InitiationResumed();
+
+    event InitiateSynthTransfer(bytes32 indexed currencyKey, address indexed destination, uint256 amount);
```
artdgn, line 145: nit: uint256 here, just uint above

contracts/Issuer.sol
```diff
@@ -870,12 +896,17 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
 
     /* ========== MODIFIERS ========== */
 
-    function _onlySynthetix() internal view {
-        require(msg.sender == address(synthetix()), "Issuer: Only the synthetix contract can perform this action");
+    function _onlyBrokers() internal view {
+        require(
+            msg.sender == address(synthetix()) ||
+                msg.sender == resolver.getAddress(CONTRACT_SYNTHETIXBRIDGETOOPTIMISM) ||
```
artdgn, line 902: nit: one of these addresses is not set (so will be 0?), so we should check which one it is and use the other one for this condition. msg.sender cannot be 0 of course, but `require(msg.sender == address(0))` is a weird thing to have.

contracts/Issuer.sol
```diff
@@ -561,48 +563,72 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
         }
     }
 
-    function issueSynths(address from, uint amount) external onlySynthetix {
+    function issueFreeSynths(
```
artdgn, line 566: would be good to add documentation to these new methods. in which context they can be used and what does `Free` mean (memory, will, speech, beer, puppies?)

perhaps `issueBridgedSynths` and `burnBridgedSynths`?

contracts/Issuer.sol
```diff
@@ -870,12 +896,17 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
 
     /* ========== MODIFIERS ========== */
 
-    function _onlySynthetix() internal view {
-        require(msg.sender == address(synthetix()), "Issuer: Only the synthetix contract can perform this action");
+    function _onlyBrokers() internal view {
```
artdgn, line 899: I think it will be safer to create a second modifier instead of broadening this one's permissiveness.

With this code, now every one of the 3 brokers can call any one of the modifier protected methods. 

But in reality 1 broker needs to call one set of methods, and the other 2 brokers need to call another set - so separate modifiers will be more strict.

contracts/Issuer.sol
```diff
@@ -561,48 +563,72 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
         }
     }
 
-    function issueSynths(address from, uint amount) external onlySynthetix {
+    function issueFreeSynths(
+        bytes32 currencyKey,
+        address to,
+        uint amount
+    ) external onlyBrokers {
+        require(address(synths[currencyKey]) != address(0), "Issuer: synth doesn't exist");
```
artdgn, line 571: require sUSD?

contracts/Issuer.sol
```diff
@@ -561,48 +563,72 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+    function burnFreeSynths(
+        bytes32 currencyKey,
+        address from,
+        uint amount
+    ) external onlyBrokers {
+        require(address(synths[currencyKey]) != address(0), "Issuer: synth doesn't exist");
+        require(amount > 0, "Issuer: cannot issue 0 synths");
+
+        // Burn some synths
+        synths[currencyKey].burn(from, amount);
```
artdgn, line 587: shoudn't this use `_burnSynths` and update all the related accounting? (circuit breaker, debt register, cached susd debt)? Should nont of the accounting be updated, even debtCache? The supply does change, so next snapshot will update it?

If this doesn't need to happen, explaining this in comments would probably be important.

contracts/Issuer.sol
```diff
@@ -561,48 +563,72 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+        // Create their synths
+        synths[currencyKey].issue(to, amount);
+    }
+
+    function burnFreeSynths(
+        bytes32 currencyKey,
+        address from,
+        uint amount
+    ) external onlyBrokers {
+        require(address(synths[currencyKey]) != address(0), "Issuer: synth doesn't exist");
```
artdgn, line 583: require sUSD?

contracts/SynthetixBridgeToOptimism.sol
```diff
@@ -123,11 +127,7 @@ contract SynthetixBridgeToOptimism is BaseSynthetixBridge, ISynthetixBridgeToOpt
...
 
     // invoked by Messenger on L1 after L2 waiting period elapses
-    function finalizeWithdrawal(address to, uint256 amount) external {
-        // ensure function only callable from L2 Bridge via messenger (aka relayer)
-        require(msg.sender == address(messenger()), "Only the relayer can call this");
-        require(messenger().xDomainMessageSender() == synthetixBridgeToBase(), "Only the L2 bridge can invoke");
-
+    function finalizeWithdrawal(address to, uint256 amount) external onlyCounterpart {
         // transfer amount back to user
         synthetixERC20().transferFrom(synthetixBridgeEscrow(), to, amount);
```
artdgn, line 132: nit: slither complains about not checking return value, which is true. AFAIS Synthetix.transferFrom does return value, so we might as well check it here (otherwise this method assumes `transferFrom` is `safeTransferFrom` but it has no reason to assume or know that implementation detail)

contracts/BaseSynthetixBridge.sol
```diff
@@ -83,9 +104,44 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+        bytes memory messageData =
+            abi.encodeWithSelector(this.finalizeSynthTransfer.selector, currencyKey, destination, amount);
+
+        // relay the message to Bridge on L1 via L2 Messenger
+        messenger().sendMessage(
+            counterpart(),
+            messageData,
+            //uint32(getCrossDomainMessageGasLimit(CrossDomainMessageGasLimits.Withdrawal))
+            3000000
+        );
```
artdgn, line 124: so what happens if the issue on the other side fails for some reason? (circuit breaking, pausing)

artdgn (changes requested):
I think permissiveness should be tightened (sUSD only, modifiers). Some more documentation might be nice.

Have not looked at the tests yet (due to requesting changes).

contracts/BaseSynthetixBridge.sol
```diff
@@ -50,15 +56,30 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
         require(initiationActive, "Initiation deactivated");
     }
 
+    function counterpart() internal view returns (address);
```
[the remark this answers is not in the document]
dbeal-eth: if its not implemented than the compiler will fail, or if you try to construct at runtime it will fail because its an abstract contract

contracts/BaseSynthetixBridge.sol
```diff
@@ -83,9 +104,44 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+
+        // create message payload
+        bytes memory messageData =
+            abi.encodeWithSelector(this.finalizeSynthTransfer.selector, currencyKey, destination, amount);
+
+        // relay the message to Bridge on L1 via L2 Messenger
+        messenger().sendMessage(
+            counterpart(),
+            messageData,
+            //uint32(getCrossDomainMessageGasLimit(CrossDomainMessageGasLimits.Withdrawal))
```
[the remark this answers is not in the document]
dbeal-eth: whoops, was debugging

contracts/BaseSynthetixBridge.sol
```diff
@@ -83,9 +104,44 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
         emit InitiationResumed();
     }
 
+    function initiateSynthTransfer(
+        bytes32 currencyKey,
+        address destination,
+        uint amount
+    ) external {
+        issuer().burnFreeSynths(currencyKey, msg.sender, amount);
```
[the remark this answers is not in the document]
dbeal-eth: not certain yet, will be determined in discussion later today. most likely it will check a systemSetting

contracts/BaseSynthetixBridge.sol
```diff
@@ -83,9 +104,44 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+
+        // create message payload
+        bytes memory messageData =
```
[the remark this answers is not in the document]
dbeal-eth: probably

contracts/BaseSynthetixBridge.sol
dbeal-eth: this function has no benefit in checking the currency key because the corresponding initiate function should be doing that check. It would only serve to increase the risk of undelivered funds.

note that the `finalize` function must have the same parameters as the `initialize` function after crossing the bridge

contracts/Issuer.sol
dbeal-eth: definitely shouldn't add it here, this function is intended to allow issuance without the corresponding debt. There would be no benefit in restricting it here because its not a multi-chain aware function

contracts/Issuer.sol
dbeal-eth: * circuit breaker doesnot need to be updated because it covers an unrelated. If anything, the circuit breaker should be 
* debt register doesn't need to be updated because, by definition of the function of this function, . For example, someone with 0 staking/debt (AKA, just a "user" of the trading part of the protocol) should be able to transfer synths
* the debtCache should be updated and this will be added

contracts/Issuer.sol
dbeal-eth: no

contracts/SynthetixBridgeToOptimism.sol
dbeal-eth: we can talk about this in the afternoon, but I would like to refrain from making changes to sensitive parts of the system like withdrawals unless we need to. Sure, its a slither warning, but we know what we have now works and even if there is a failure here it wouldn't change the overall result for the user (they still would'nt receive their funds and we could have a major problem)

contracts/BaseSynthetixBridge.sol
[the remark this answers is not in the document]
dbeal-eth: c621ce8b

contracts/BaseSynthetixBridge.sol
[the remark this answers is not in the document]
dbeal-eth: c621ce8b

contracts/Issuer.sol
dbeal-eth: c621ce8b

contracts/BaseSynthetixBridge.sol
dbeal-eth: c621ce8b

contracts/BaseSynthetixBridge.sol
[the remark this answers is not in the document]
artdgn: I thought only sUSD was in spec for this feature for V2X because of the added possible complications of moving volatile synths.

contracts/BaseSynthetixBridge.sol
artdgn: > hould be doing that check

It's on another chain, this contract shouldn't trust that a message coming from some contract on another chain has done some checks. If it should only be minting sUSD, it should check that here as well.

contracts/Issuer.sol
artdgn: If Issuer is intended to only be minting sUSD in this method, it should check that it's only sUSD, 

Even if the bridge checked it as well, these are two different contracts, and if somehow the bridge is wrong, the Issuer should do its own input validation if it can, and that check is trivial.

What is the benefit of skipping that check here?

contracts/Issuer.sol
artdgn: why?

contracts/BaseSynthetixBridge.sol
artdgn: I think that if the system is paused it should be paused, it shouldn't be minting due to unchecked messages from another chain. 

What if the bridge contract had a bug, and now you know it was exploited. So now when it lends on L1 - it's can't even be paused?

contracts/BaseSynthetixBridge.sol
```diff
@@ -83,9 +104,45 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
         emit InitiationResumed();
     }
 
+    function initiateSynthTransfer(
```
artdgn, line 107: I think the wait period for L2->L1 creates some accounting problems that make immediate burning problematic in this case.

What if e.g. half of debt is being sent to L1 (one tx, or multiple). The synths are burned, so the debt on that chain is decreased. Which now allows more minting. Which now can be sent to the bridge and burned again. And again. 

During that time, the debt is not an accurate represenation of the debt due to the burning of the synths, so allows minting more than needed for the c-ratio that was set. This possibly opens an inflation attack. Which in combination with the other chain not being able to pause the issuance (in current code), can be quite destructive.

**I think the bridge needs to hold the synths, and only burn them upon receiving an issuance confirming message from the other chain.** This way the accounting problem is for temporarily having more then debt in the system than needed, than vice versa, which is safer (due to allowing to mint less rather and more).

Maybe L1 -> L2 can be done with burning immediately (due to the speed), but even then - how do we know there isn't going to be an issue due to incorrectly accounting the debt during that short window?

Even if it's not an attack, a big enough burn will affect the accounting for the rest of the stakers: their debt will be reduced for a week, they might mint, and then the debt will jump back up when it's minted on the other side, no?

artdgn (changes requested):

contracts/BaseSynthetixBridge.sol
artdgn: Also, a lot of my comments from previous review weren't addressed yet, so these points remain unaddressed in my view.

contracts/BaseSynthetixBridge.sol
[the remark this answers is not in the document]
artdgn: I don't think it's an abstract contract, we're in 0.5 still.

contracts/Issuer.sol
artdgn: not a nit.

contracts/BaseSynthetixBridge.sol
dbeal-eth: > What if e.g. half of debt is being sent to L1 (one tx, or multiple). The synths are burned, so the debt on that chain is decreased (e.g. by a snapshot). Which now allows more minting. Which now can be sent to the bridge and burned again. And again.

Thats why the chainlink oracle is being updated to account for synths in transit. Read the PR description please!

> I think the bridge needs to hold the synths, and only burn them upon receiving an issuance confirming message from the other chain. This way the accounting problem is for temporarily having more then debt in the system than needed, than vice versa, which is safer (due to allowing to mint less rather and more).

Suppose we have the bridge hold the synths while they are in transit. What happens when the synth arrives on the other end and is minted? Do we have to ? Note that there remains a 15-20 minute delay for sending messages L1->L2, meaning that we haven't really accomplished anything regarding the issue you mentioned for going the other way

contracts/BaseSynthetixBridge.sol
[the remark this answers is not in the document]
dbeal-eth: > I don't think it's an abstract contract, we're in 0.5 still.

https://docs.soliditylang.org/en/v0.5.16/contracts.html?highlight=abstract#abstract-contracts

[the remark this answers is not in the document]
justinjmoses: Yeah it won't compile if we don't implement it.

dbeal-eth:
as discussed in meeting last week, the bridge has been updated to not accept parameter for  `currencyKey`. Instead, it will assume `sUSD`.

contracts/BaseSynthetixBridge.sol
justinjmoses: I think this function should use the modifier `requireInitiationActive` - this is what we use to disable deposits and withdrawals of SNX when we are upgrading the bridge or destination contracts, so I think it's reasonable to also have it here. That being said, whenever we suspend initiation on the bridge, it will mean that both SNX and sUSD will be disabled. I think that's just cost of doing business, so to say.

contracts/Issuer.sol
justinjmoses: Agreed on a better name but not mandatory to add comments IMO - they are effectively protected internal

contracts/Issuer.sol
justinjmoses: I feel this is a nitpick. The function here is protected internal - if the public function in the bridge is limited to sUSD and this function is limited to the bridge, I think it's reasonably protected. If we added requires everywhere to double and triple check inputs our code would be much bigger. I think using unit tests to check this is sufficient.

contracts/Issuer.sol
artdgn: the comments are for developers looking into that code outside of the context of this PR and trying to figure out what accounting needs and doesn't need to happen here and **why** when compared to other issuing in the contract.

contracts/Issuer.sol
artdgn: It is protected internal in this PR, but it's not future proof for subsequent changes. It is easy to imagine future changes that open a way for a currencyKey to be passed in not through the bridge, especially now the modifier is written in this PR (not limiting to only the bridge).

I disagree that it's a nitpick, if the Issuer from its point of view "knows" that it only mints sUSD in this method, but accepts `currencyKey` just future interface changes - it should check its inputs.

contracts/BaseSynthetixBridge.sol
```diff
@@ -83,9 +106,36 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+        bytes memory messageData = abi.encodeWithSelector(this.finalizeSynthTransfer.selector, destination, amount);
+
+        // relay the message to Bridge on L1 via L2 Messenger
+        messenger().sendMessage(
+            counterpart(),
+            messageData,
+            uint32(getCrossDomainMessageGasLimit(CrossDomainMessageGasLimits.Withdrawal))
+        );
+
+        emit InitiateSynthTransfer(sUSD, destination, amount);
```
justinjmoses, line 124: Could you also add the `uint` ID in here to help Chainlink to figure out which transfers have been finalized?

contracts/BaseSynthetixBridge.sol
artdgn: Regarding the proposal for CL to track the burned / minted events. I have a variation on that, that's less dependant on harder to verify invents processing, and instead uses more easily verifiable (by external observer) contract views:

Each bridge will track and expose a view of total synths burned and minted, or even just the difference between those two sums. CL will use those views to count report the total debt in a way that will be trivial to check for someone who observes the rest of the debt values reported on both chains.

So e.g. when someone bridges L2->L1. They burn X. L2 side reports -X debt initially, L1 side reports 0. CL adds X to the rest of debt from both chains. After L1 finalises, it reports +X, L2 still reports -X - so there is no offset and CL adds 0 to the total debt.

We can even call that number something like `bridgeDebtOffset` and roll it into the number CL already uses, and this way no CL changes are required.

During migrations the offsets state should be migrated as well (just as external debt is migrated currently).

contracts/BaseSynthetixBridge.sol
```diff
@@ -121,12 +124,16 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
             uint32(getCrossDomainMessageGasLimit(CrossDomainMessageGasLimits.Withdrawal))
         );
 
+        synthTransferSent += amount;
```
artdgn, line 127: should probably be done before calling an external contract to follow check-effects-interactions. will likely be flagged by slither and the audit

contracts/BaseSynthetixBridge.sol
```diff
@@ -121,12 +124,16 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+
         emit InitiateSynthTransfer(sUSD, destination, amount);
     }
 
     function finalizeSynthTransfer(address destination, uint amount) external onlyCounterpart {
         issuer().issueFreeSynths(sUSD, destination, amount);
 
+        synthTransferReceived += amount;
```
artdgn, line 135: should probably be done before calling an external contract to follow check-effects-interactions

contracts/BaseSynthetixBridge.sol
dbeal-eth: no longer needed for chainlink, though lmk if you think we should still have it

contracts/BaseSynthetixBridge.sol
dbeal-eth: will be moved

contracts/BaseSynthetixBridge.sol
dbeal-eth: will be moved

contracts/Issuer.sol
dbeal-eth: will be changed

contracts/Issuer.sol
dbeal-eth: the function need-not be associated only with the bridge (even if the bridge is currently the only caller), which is why it wasn't named as such. as for the adding of comments, I can add comments if you think its needed, but in this file only the most complicated functions or noteworthy lines have added comments. Considering the brevity and simplicity of the function, we can probably go without

dbeal-eth: for naming, how does `issueSynthsWithoutDebt` sound? (and same for the corresponding `burn` function?)

contracts/Issuer.sol
dbeal-eth: > It is protected internal in this PR, but it's not future proof for subsequent changes. It is easy to imagine future changes that open a way for a currencyKey to be passed in not through the bridge, especially now the modifier is written in this PR (not limiting to only the bridge).

Can you give a specific hypothetical example of a subsequent change that could lead to unexpected complications because of the acceptance of this parameter?

> I disagree that it's a nitpick, if the Issuer from its point of view "knows" that it only mints sUSD in this method, but accepts currencyKey just for future proofing interface changes - it should check its inputs in current implementation.

I'm confused, all synths allow for `Issuer` to issue its tokens.

> Logically, the Issuer and the Bridge are different entities, and although internal (and have internal auth), they should still check their inputs without relying on implementation details knowledge of each other. Because when someone will go and change e.g. the bridge logic, they will also need to go and change the other contracts that implicitly rely on that logic, and they may forget to (and the reviewers of their PR might also forget to, and the auditors might forget to). That's why we should reduce this implicit reliance on internal knowledge of other contract's implementation if we can.

If we are creating a restritcion to only mint `sUSD` in order to match the impl of the bridge, then we are literally making an assumption within `Issuer` about how the bridge works and thereby increasing implementation complexity and hoops required if we have a case for this function to be used elsewhere

TBH, the synth contracts shouldn't be using `onlyInternalContracts` for auth. They should let the `Issuer` be the only caller for the `issue` function, and then everything calls the issuer which is much easier to upgrade and fits the implied purpose of this file by its name. But thats an issue from long ago.

contracts/Issuer.sol
dbeal-eth: henceforth, please assume all comments for the identically named `require sUSD?` comment earlier in this file now apply to this thread. apologies for my short response

contracts/BaseSynthetixBridge.sol
dbeal-eth: > Regarding the proposal for CL to track the burned / minted events. I have a variation on that, that's less dependant on harder to verify invents processing, and instead uses more easily verifiable (by external observer) contract views. It will even potentially require no changes from CL, so will be faster and easier to ship.
> 
> Each bridge will track and expose a view of total synths burned and minted, or even just the difference between those two sums. CL will use those views to report the total debt in a way that will be trivial to check for someone who observes the rest of the debt values reported on both chains.
> 
> So e.g. when someone bridges L2->L1. They burn X. L2 side reports -X debt offset initially, L1 side reports 0. CL adds X to the rest of debt from both chains. After L1 finalises, it reports +X, L2 still reports -X - so there is no offset and CL adds 0 to the total debt. When another bridging tx happens - it will update the offsets on each side at the right times when it lends on chain.
> 
> We can even call that number something like `bridgeDebtOffset` and roll it into the number CL already uses, and this way no CL changes are required.
> 
> During migrations the offsets state should be migrated as well (just as external debt is migrated currently).

this is being done now

contracts/BaseSynthetixBridge.sol
dbeal-eth: > I think that if the system is paused it should be paused, it shouldn't be minting due to unchecked messages from another chain.

after reviewing the implementation of the Synths file (https://github.com/Synthetixio/synthetix/blob/develop/contracts/Synth.sol#L167), it appears that it will allow minting or burning without checking to see if the system is paused, which is surprising. to prevent this, the `in

however, it doesn't solve the problem if one network is paused. If for example, L2 is paused, but L1 is not, how does L1 know that it can't send to L2? its not like a message could be sent to inform it. Therefore, I can think of only 2 things we can do:

1. always send the synths to the user in `finalizeSynthTransfer` even if the system is paused, which is probably the best thing to do tbh since the system being paused has little/nothing to do with the health of the bridge, and we have a separate lever to control that if need be
2. deposit any funds sent while the system was paused to a "lost funds" wallet of some sort, and manually distribute it after the system is unpaused, if necessary

however, we can't just have inaction, otherwise the synths would be lost in transit, unable to be recovered. we would probably have to refund.

contracts/BaseSynthetixBridge.sol
```diff
@@ -7,27 +7,38 @@ import "./MixinResolver.sol";
...
+    bytes32 private constant CONTRACT_FLEXIBLESTORAGE = "FlexibleStorage";
+    bytes32 private constant CONTRACT_EXCHANGERATES = "ExchangeRates";
 
     bytes32 private constant sUSD = "sUSD";
 
     bool public initiationActive;
 
-    uint public synthTransferSent;
-    uint public synthTransferReceived;
+    bytes32 private constant SYNTH_TRANSFER_NAMESPACE = "SynthTransfer";
```
artdgn, line 39: perhaps refactoring the SynthTransfer storage handling into a mixin will be nicer. It has a limited interface, and can be easier to read in a separate file

contracts/BaseSynthetixBridge.sol
```diff
@@ -124,17 +156,52 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
-        synthTransferSent += amount;
-
         emit InitiateSynthTransfer(sUSD, destination, amount);
     }
 
-    function finalizeSynthTransfer(address destination, uint amount) external onlyCounterpart {
-        issuer().issueFreeSynths(sUSD, destination, amount);
+    function finalizeSynthTransfer(bytes32 currencyKey, address destination, uint amount) external onlyCounterpart {
+        
+        _incrementSynthsTransferCounter(SYNTH_TRANSFER_RECV, currencyKey, amount);
```
artdgn, line 164: this doesn't do any validation, and can mint any synth with any amount according to a message from another chain (or from another contract, that makes it think it's got it from another chain)

contracts/BaseSynthetixBridge.sol
```diff
@@ -124,17 +156,52 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+
+    function _incrementSynthsTransferCounter(bytes32 group, bytes32 currencyKey, uint amount) internal {
+        bytes32 key = keccak256(abi.encodePacked(group, currencyKey));
+        
+        uint currentSynths = flexibleStorage().getUIntValue(SYNTH_TRANSFER_NAMESPACE,key);
+
+        flexibleStorage().setUIntValue(
+            SYNTH_TRANSFER_NAMESPACE,
+            key,
+            currentSynths + amount
```
artdgn, line 181: nit? safemath

contracts/BaseSynthetixBridge.sol
```diff
@@ -124,17 +156,52 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+        );
+    }
+
+    function _sumTransferAmounts(bytes32 group) internal view returns (uint sum) {
+        // get list of synths from issuer
+        bytes32[] memory currencyKeys = issuer().availableCurrencyKeys();
```
artdgn, line 187: what if a synth was deprecated / not added, its values should still be summed in perpetuity for correct debt accounting of the two sides, no?

contracts/SystemSettingsLib.sol
```diff
@@ -422,6 +422,19 @@ library SystemSettingsLib {
         );
     }
 
+    function setCrossSynthTransferEnabled(
```
artdgn, line 425: nit: `setCross`**Synth**`TransferEnabled` did you mean to have `Chain` instead of `Synth`?

contracts/BaseSynthetixBridge.sol
```diff
@@ -124,17 +156,52 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+
+        // get all synth rates
+        (uint[] memory rates, bool isInvalid) = exchangeRates().ratesAndInvalidForCurrencies(currencyKeys);
+
+        require(!isInvalid, "Rates are not accurate");
```
artdgn, line 192: nit: "invalid"

contracts/BaseSynthetixBridge.sol
```diff
@@ -124,17 +156,52 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+
+        // get all values
+        bytes32[] memory transferAmountKeys = new bytes32[](currencyKeys.length);
+        for (uint i = 0;i < currencyKeys.length;i++) {
+            transferAmountKeys[i] = keccak256(abi.encodePacked(group, currencyKeys));
```
artdgn, line 197: currencyKeys -> currencyKeys[i] ?

test/contracts/BaseSynthetixBridge.unit.js
```diff
@@ -185,25 +189,53 @@ contract('BaseSynthetixBridge (unit tests)', accounts => {
...
+						assert.bnEqual(issuer.smocked.burnSynthsWithoutDebt.calls[0].amount, toUnit('100'));
 					});
 
 					it('calls messenger', () => {
 						assert.bnEqual(messenger.smocked.sendMessage.calls[0]._target, issuer.address);
 					});
 
 					it('increments synthTransferSent', async () => {
-						assert.bnEqual(await instance.synthTransferSent(), toUnit('150'));
+						assert.bnEqual(flexibleStorage.smocked.setUIntValue.calls[0].value, toUnit('150'));
```
artdgn, line 238: key should be validated as well

contracts/SystemSettings.sol
```diff
@@ -225,7 +225,7 @@ contract SystemSettings is Owned, MixinSystemSettings, ISystemSettings {
 
     // SIP-229 Atomic exchanges
     // enable/disable sending of synths cross chain
-    function crossChainTransferEnabled(bytes32 currencyKey) external view returns (uint) {
+    function crossSynthTransferEnabled(bytes32 currencyKey) external view returns (uint) {
```
artdgn, line 228: nit: crossSynth?

artdgn (changes requested):

contracts/BaseSynthetixBridge.sol
dbeal-eth: the mixin is only expected to be implemented by `BaseSynthetixBridge`. Do you really think its worth the effort?

contracts/BaseSynthetixBridge.sol
dbeal-eth: what validation do you recomend? validation for message to only come from the other bridge contract is done with `onlyCounterpart`

contracts/BaseSynthetixBridge.sol
dbeal-eth: ty

contracts/BaseSynthetixBridge.sol
dbeal-eth: if a synth is deprecated/removed from the issuer, then we could solve this problem easily by just disabling bridging and waiting for all messages to be forwarded.

contracts/SystemSettingsLib.sol
dbeal-eth: I meant to have `Synth` as is in the code

contracts/BaseSynthetixBridge.sol
dbeal-eth: will fix

contracts/BaseSynthetixBridge.sol
dbeal-eth: yes

test/contracts/BaseSynthetixBridge.unit.js
dbeal-eth: not quite there yet but yes needs to be done

contracts/SystemSettings.sol
dbeal-eth: this is intended

contracts/BaseSynthetixBridge.sol
artdgn: it should be pretty straightforward, just splitting it into another file. not insisting on it, but it seemed a nice way to declutter the file from the storage related activities

contracts/BaseSynthetixBridge.sol
artdgn: It would be safer to control minting on the receiving side as well, at least in terms of supported currencyKeys, and later in amounts as well (rate limiting). 

So that the receiving side would have control over pausing / limiting damage in case of an issue. Right now it just does whatever the message tells it to. Feels too permissive, and no recourse for damage control / mitigation.

contracts/SystemSettings.sol
artdgn: in that case the naming doesn't seem right to me.. what is a "cross synth transfer"? is it cross-chain-synth-transfer? previous `crossChainTransferEnabled(bytes32 key)` seems more clear to me

contracts/BaseSynthetixBridge.sol
artdgn: should this be added into some checklist / doc that's relevant to migrations / deprecations?

contracts/SystemSettings.sol
dbeal-eth: it doesn't clarify that this function only applies to synths being transferred cross chain (and ex. not SNX being transferred cross chain).

perhaps `crossChainSynthTransferEnabled`? seems like its crossing a bad threshold of wordyness

test/contracts/BaseSynthetixBridge.unit.js
dbeal-eth: FYI this test coverage has been added in the integration tests instead because its difficult to add the required tests to unit tests since the exact hash used to store the data could be built many different ways, and it is not the role of tests to verify that. only that the value has been stored and could conceivably be recovered later.

contracts/BaseSynthetixBridge.sol
dbeal-eth: @justinjmoses how would you recommend considering this?

contracts/BaseSynthetixBridge.sol
dbeal-eth: e41306a2

contracts/BaseSynthetixBridge.sol
dbeal-eth: 55953254

contracts/BaseSynthetixBridge.sol
dbeal-eth: 5b81e07c332649c2d715fb4f6bc0feb59b7de911

test/contracts/BaseSynthetixBridge.unit.js
dbeal-eth: 5b81e07c332649c2d715fb4f6bc0feb59b7de911

contracts/BaseSynthetixBridge.sol
dbeal-eth: > So that the receiving side would have control over pausing / limiting damage in case of an issue. Right now it just does whatever the message tells it to. Feels too permissive, and no recourse for damage control / mitigation.

ok, what specific steps/measures can we take, besides implementing the rate limiting PR that has been under discussion? Should we keep rate limiting separate, or try and bundle it in this PR?

contracts/BaseSynthetixBridge.sol
kylriley: Worth mentioning that Optimism has [replay](https://github.com/ethereum-optimism/optimism/blob/2493e925ecff69ec352aa1a43471033ceb03a1a6/packages/contracts/contracts/L1/messaging/L1CrossDomainMessenger.sol#L231) / [relay](https://github.com/ethereum-optimism/optimism/blob/2493e925ecff69ec352aa1a43471033ceb03a1a6/packages/contracts/contracts/L1/messaging/L1CrossDomainMessenger.sol#L165) functions that can be used to resubmit failed cross-chain txs. 

If L1 => L2 fails on L2 `l1.replay()` can be used to resubmit the tx through to L2.
If L2 => L1 fails on L1 `l1.relay()` can be used to manually submit the L2 tx on L1.

Would need to have some condition on the front-end that determines when this is necessary, but it should be able to provide the user a mechanism for retrieving any lost synths.

kylriley: @dbeal-eth my concern with (1.) would be more around automated triggers (i.e. debt / price circuit breaker), where you don't have time to trigger `initiationActive` beforehand. If funds are moving around it could make triaging / correcting the problem harder.

contracts/BaseSynthetixBridge.sol
barrasso: some potential ideas for additional validation checks here:

- limit this functionality to only work with a single synth (i.e. `sUSD`) to reduce attack surface area
- add another modifier (e.g. `requireFinalizationActive`) that could be invoked on the receive side if some issue arises
  - this is kinda mentioned in your comment above (2.) where the in-flight funds that fail on the receiving side due to this pausing mechanism are recorded somehow and can be manually reconciled afterwards
- add a configurable variable that limits the amount of synths that can be sent/received
  - e.g. `maxSynthTransferAmount` (which can be tapered off and eventually set to `0` once the general purpose rate limiters are implemented)
  - or some `cooldown` variable (similar to burn lock) that allows for transfers only once per configurable time period ... this can also be tapered off and eventually set to `0` once the rate limiters are implemented
  - These setter variables can be internal to the bridge and set by owner (they don't have to live in SystemSettings so we can save some gas)

contracts/BaseSynthetixBridge.sol
barrasso: Won't this get exponentially more costly as we continue to add more and more synths? And what if we end up only using this bridge for a couple synths? Do we really have demand for enabling ALL synths to be transferable?

imo, we can significantly cut down the complexity here and make it only work with `sUSD` 

otherwise, we should only enable synths that have a high probability of never getting removed/deprecated (e.g. `sETH`, `sBTC`)

contracts/SystemSettings.sol
barrasso: nit: gonna have to agree with Arthur on naming convention here... 

Using `crossChainTransferEnabled(bytes32 key)` or `crossChainSynthTransferEnabled` seems more intuitive than the current `crossSynthTransferEnabled`

hardhat.config.js
```diff
@@ -92,6 +92,6 @@ module.exports = {
 
 	mocha: {
 		timeout: 120e3, // 120s
-		retries: 3,
```
barrasso, line 95: np: unsure if this is supposed to be here

test/contracts/BaseSynthetixBridge.unit.js
```diff
@@ -5,25 +5,37 @@ const { onlyGivenAddressCanInvoke, ensureOnlyExpectedMutativeFunctions } = requi
...
 
 	it('ensure only known functions are mutative', () => {
 		ensureOnlyExpectedMutativeFunctions({
 			abi: BaseSynthetixBridge.abi,
 			ignoreParents: ['Owned', 'MixinResolver'],
-			expected: ['resumeInitiation', 'suspendInitiation'],
+			expected: [
+				'resumeInitiation',
+				'suspendInitiation',
+				'initiateSynthTransfer',
```
barrasso, line 24: np: related to naming convention for the setting

If the setting is indeed changed to something `crossChainSynthTransferEnabled`, maybe it makes sense to also rename these to be more explicit and match accordingly:
- `initiateCrossChainSynthTransfer`
- `finalizeCrossChainSynthTransfer`

Just looking at the function names as they are now-- there's no indication that these synth transfers are "cross chain"

test/contracts/BaseSynthetixBridge.unit.js
```diff
@@ -160,6 +187,138 @@ contract('BaseSynthetixBridge (unit tests)', accounts => {
...
+				it('can only be called by messenger and registered counterpart', async () => {
+					await onlyGivenAddressCanInvoke({
+						fnc: instance.finalizeSynthTransfer,
+						accounts,
+						address: smockedMessenger,
+						args: [sUSD, owner, '100'],
+						reason: 'Only the relayer can call this',
+					});
+				});
+
```
barrasso, line 259: could also add some tests here to verify that if a circuit breaker is tripped or the system is suspended in some way while synths are in-flight to L2, they are still finalized or handled as expected

barrasso (commented):
added comments to existing discussions and some small nits

test/contracts/BaseSynthetixBridge.unit.js
dbeal-eth: good idea

test/contracts/BaseSynthetixBridge.unit.js
dbeal-eth: its on the bridge contract... shouldn't the name be corresponding with existing functions? other cross-transfer functions on the bridges aren't called "depositCrossChain" or "migrateEscrowCrossChain" so it seems reasonable for `initiateSynthTransfer` to be the name here as well

contracts/BaseSynthetixBridge.sol
dbeal-eth: cost is not a concern here because the function is only expected to be called by the oracle (and there really is any value to call on chain because the data is useless in isolation)

"exponentially more costly" no, only linearly. This function could probably easily support 100 synths before hitting the gas limit and failing read-only call

in v3 similar calculations will be segmented by the investments of a fund, so we will not have to worry about similar functions becoming

test/contracts/BaseSynthetixBridge.unit.js
dbeal-eth: actually, we can't test that in this file because the unit test doesn't have the real Issuer or SystemStatus.

will do it in the integration

contracts/BaseSynthetixBridge.sol
```diff
@@ -42,33 +61,70 @@ contract BaseSynthetixBridge is Owned, MixinSystemSettings, IBaseSynthetixBridge
...
+        newAddresses[5] = CONTRACT_FLEXIBLESTORAGE;
+        newAddresses[6] = CONTRACT_EXCHANGERATES;
         addresses = combineArrays(existingAddresses, newAddresses);
     }
 
+    function synthTransferSent() public view returns (uint) {
+        return _sumTransferAmounts(SYNTH_TRANSFER_SENT);
+    }
+
+    function synthTransferReceived() public view returns (uint) {
```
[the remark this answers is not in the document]
dbeal-eth: 0e97a0b0

hardhat.config.js
dbeal-eth: 0e97a0b0

contracts/SystemSettings.sol
dbeal-eth: 0e97a0b0

contracts/BaseSynthetixBridge.sol
dbeal-eth: we have to be really careful adding checks to the finalization function that could cause the transfer to fail in legitimate cases because recovery of lost funds is be difficult.

alternatively, we could set up some sort of a "impound box" which stores synths which have been rejected but can be unlocked by owner, but that raises other questions about, ex. what to do about debt (does it get included or not, both of which could be an attack vector on their own?), 

> limit this functionality to only work with a single synth (i.e. sUSD) to reduce attack surface area

SIP officially says to support any synth that is enabled

> add another modifier (e.g. requireFinalizationActive) that could be invoked on the receive side if some issue arises 

we could do that, but its redundant when we could just suspend synth transfers using the existing system status facility. Also this could be majorly impacted by the rejected transfer issue listed above

> add a configurable variable that limits the amount of synths that can be sent/received 

we have another SIP for that (or at least its a planned SIP). also, an attacker could still likely exploit this by just minting SNX on L2 instead, or unlocking SNX on L1, which is currently unlimited

also
* `maxTransferAmount` alone wouldn't stop an attacker from sending multiple large transfers at the same time
* `cooldown` would be the ideal way to go, but then we are basically implementing the rate limiting SIP. also, the cooldown/rate limiting would have to be enforced on the `initiate` side, because enforcement on the receiving side could result in failed receive transfers if ex. the finalize calls are delayed and then submitted in a batch (same issue listed first) 
* also `cooldown` has a problem where a large synth holder could prevent other, synth holders from sending. the user could continuously send transfers from L1 -> L2 and L2 -> L1 and almost completely block the pipe for everyone else. we could make this limit on a per address basis... oh wait.

barrasso (approved):

PR #1760, opened 2022-04-19
SIP-203: Volume Source Fee

noahlitvin:


contracts/FuturesMarketManager.sol
```diff
@@ -108,6 +108,13 @@ contract FuturesMarketManager is Owned, MixinResolver, IFuturesMarketManager {
         return results;
     }
 
+    /*
+     * Returns whether a given address is a market.
+     */
+    function isMarket(address market) external view returns (bool) {
```
artdgn, line 114: no need to add this, the market manager is the contract pays the fee in `payFee()` (and not spefic markets)

contracts/VolumePartner.sol
```diff
@@ -0,0 +1,198 @@
...
+
+    /* ========== Modifiers ========== */
+
+    modifier notFeeAddress(address account) {
+        require(account != FEE_ADDRESS, "Fee address not allowed");
+        _;
+    }
+
+    function _isInternalContract(address account) internal view returns (bool) {
+        return account == address(exchanger()) || futuresMarketManager().isMarket(account); // just change this to get futuresmarketmanager and then have an accrue fees proxy call like issueSUSD in the uturemarketsmanager contract
```
artdgn, line 178: the call will be form `futuresMarketManager()` and not from a specific market

noahlitvin: Got it, thanks!

PR #1761, opened 2022-04-22
Clean up futures and fix npm audit issue

artdgn:
- ~~run `npm run clean-install` because there was a big diff for `npm i`, and it seems that some update was forgotten along the way~~ package-lock.json is not used
- remove kovan-ovm-futures deployment artifacts network
- removes yaml-validator because it was failing npm audit and couldn't be fixed by npm fix
- fixes test failures that were masked by the audit issues


dbeal-eth (approved):
just remove lockfile

PR #1763, opened 2022-04-28
Perps v2 - split contracts

artdgn:
This PR only clones the futures contracts to allow working on v2 while v1 remains unchanged. This is needed to prevent a long living branch, and to keep the work on `develop`.

## The cloned contracts:
 - Are independent from the previous versions.
 - Only common contract still is the `FuturesMarketManager` which will be a common contract for now (the common touch point with rest of Synthetix)
 - Renamed to use `PerpsV2` names for contracts. This is because `Perps` is the more accurate and accepted name for this product, and to accommodate for future versions.
 - The market keys will use `pASSET` names in the new contracts.
 
## Testing and deployment considerations:
- There were no changes to deployment scripts, configuration scripts, and integration / fork tests. This is because it will be better to update these once the contracts approach their more final form and architecture.
- Unit tests were clones as well, with the exception of the common `FuturesMarketManager`.

## Changes to `FuturesMarketsManager`:
- Basic summary methods (in V1 handled by `FuturesMarketData`) were added into the manager, and no clone of `FuturesMarketData` was created. Tests were added into the manager test file.
- The scope of these summary methods was reduced compared to previous, to rely on smaller common interface (common to V1 and V2)

## Minor changes to functionality in new contracts:
- a fix for incorrectly checking for synth existence when using circuit breaker was done (not impacting circuit breaker, but providing a more correct revert reason)
- tracking event was renamed to `Tracking` from `FuturesTracking`


dbeal-eth (approved):
I think you mentioned this sometime before, but why are we we not putting perps v2 into a separate subdir rather just adding a prefix?

artdgn:
> I think you mentioned this sometime before, but why are we we not putting perps v2 into a separate subdir rather just adding a prefix?

Don't remember mentioning that, but that's a good idea probably.

PR #1770, opened 2022-05-06
Sccps 191 192

artdgn:


index.js
```diff
@@ -199,7 +199,7 @@ const defaults = {
 	ETHER_WRAPPER_MINT_FEE_RATE: w3utils.toWei('0.005'), // 5 bps
 	ETHER_WRAPPER_BURN_FEE_RATE: '0',
 
-	FUTURES_MIN_KEEPER_FEE: w3utils.toWei('20'), // 20 sUSD liquidation fee
+	FUTURES_MIN_KEEPER_FEE: w3utils.toWei('5'), // 5 sUSD min keeper fee
```
barrasso, line 202: why is this changing to `5 wei` if the override is set to `20 wei` below?

barrasso (changes requested):

index.js
barrasso: nvm I see it's set to `20` for test purposes

barrasso (approved):

PR #1778, opened 2022-05-11
SIP-236: fix exchange before reclaim risk

dbeal-eth:
`Exchanger.exchange` does not use the post-reclaim amount to calculate the amount of exchange for the
user trade. This leads to the user being left with more of their source token then they should
have had.

this PR resolves this problem by ensuring amounts used for exchange are
calculated for the amount after fee reclamation has occured. It also adds associated tests.

SIP to be assigned pending

### DIFF
--- a/contracts/Exchanger.sol
+++ b/contracts/Exchanger.sol
@@ -439,9 +439,17 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
         entry.roundIdForSrc = exchangeRates().getCurrentRoundId(sourceCurrencyKey);
         entry.roundIdForDest = exchangeRates().getCurrentRoundId(destinationCurrencyKey);
 
+        uint sourceAmountAfterSettlement = _settleAndCalcSourceAmountRemaining(sourceAmount, from, sourceCurrencyKey);
+
+        // If, after settlement the user has no balance left (highly unlikely), then return to prevent
+        // emitting events of 0 and don't revert so as to ensure the settlement queue is emptied
+        if (sourceAmountAfterSettlement == 0) {
+            return (0, 0, IVirtualSynth(0));
+        }
+
         (entry.destinationAmount, entry.sourceRate, entry.destinationRate) = exchangeRates().effectiveValueAndRatesAtRound(
             sourceCurrencyKey,
-            sourceAmount,
+            sourceAmountAfterSettlement,
             destinationCurrencyKey,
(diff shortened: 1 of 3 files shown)


contracts/Exchanger.sol
```diff
@@ -439,17 +439,25 @@ contract Exchanger is Owned, MixinSystemSettings, IExchanger {
...
+            sourceAmountAfterSettlement,
             destinationCurrencyKey,
             entry.roundIdForSrc,
             entry.roundIdForDest
         );
 
         _ensureCanExchangeAtRound(
             sourceCurrencyKey,
-            sourceAmount,
+            sourceAmountAfterSettlement,
```
justinjmoses, line 460: this is a change from what was in pre-184 but it's not critical as the check only ensures `sourceAmountAfterSettlement > 0` and that is already dealt with on L446 above.

test/contracts/Exchanger.spec.js
```diff
@@ -3258,6 +3330,94 @@ contract('Exchanger (spec tests)', async accounts => {
...
+								});
+
+								adjustedTransferBalance = balanceBefore.add(
+									expectedSettlement.reclaimAmount.gt(toUnit('0'))
+										? expectedSettlement.reclaimAmount.neg()
+										: expectedSettlement.rebateAmount
+								);
+							});
+
+							beforeEach('change', async () => {
```
justinjmoses, line 3379: `change` is pretty vague - is the price moving up or down? Is it cause a rebate or reclaim. - it will make this much more readable.

test/contracts/Exchanger.spec.js
```diff
@@ -3258,6 +3330,94 @@ contract('Exchanger (spec tests)', async accounts => {
...
+
+					beforeEach(async () => {
+						amountIn = toUnit('10000');
+						await sEURContract.issue(account1, amountIn);
+						await synthetix.exchange(sEUR, amountIn, sAUD, {
+							from: account1,
+						});
+					});
+
+					function settlementCase(newPrice) {
```
justinjmoses, line 3351: [np] why use `function` instead of `const settlementCase = (newPrice) => {};` as we do in this spec otherwise?

dbeal-eth: its just how I usually write functions. will be changed

barrasso (approved):

PR #1779, opened 2022-05-11
SIP-235: Update Issuer Modifier

barrasso:
https://sips.synthetix.io/sips/sip-235/


artdgn (approved):
Ideally we should have added a test for this as well, and maybe a fork test that ensures this condition doesn't occur due to misconfiguration.

contracts/Issuer.sol
```diff
@@ -960,6 +960,11 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
                 msg.sender == resolver.getAddress(CONTRACT_SYNTHETIXBRIDGETOBASE),
             "Issuer: Only trusted minters can perform this action"
         );
+        require(
```
artdgn, line 963: Perhaps using local variables for the addresses and shorter revert messages may reduce contract size @barrasso 
```solidity
address bridgeL1 = resolver.getAddress(CONTRACT_SYNTHETIXBRIDGETOOPTIMISM);
address bridgeL2 = resolver.getAddress(CONTRACT_SYNTHETIXBRIDGETOBASE);
require(msg.sender == bridgeL1 || msg.sender == bridgeL2, "Issuer: only trusted minters");
require(bridgeL1 == address(0) || bridgeL2 == address(0), "Issuer: one minter must be 0x0");
```

PR #1780, opened 2022-05-12
Perps changes batch 1

artdgn:
Some small changes to the contracts:
- better view for liquidation price and liquidation fee
- better views for iterating over position ids (number of positions, mapping of position to address)
- positions id persistance through closing / liquidting a position
- contract name getters
- remove the maker / taker fee distinction instead of a single fee
- rename mixins for easier navigation
- can add margin while market is paused (but not if whole system is paused)
- pass tracking code into manager for later integration with 203


artdgn:
btw, fork-tests are failing because AP & DYDX markets are paused although according to config they shouldn't be, and there's no automated choice for that situation in the deploy script (it asks for a decision).

contracts/FuturesMarketManager.sol
```diff
@@ -269,12 +269,22 @@ contract FuturesMarketManager is Owned, MixinResolver, IFuturesMarketManager {
         return amount;
     }
 
-    /*
+    /**
      * Allows markets to issue exchange fees into the fee pool and notify it that this occurred.
      * This function is not callable through the proxy, only underlying contracts interact;
      * it reverts if not called by a known market.
      */
+    function payFee(uint amount, bytes32 trackingCode) external onlyMarkets {
```
barrasso, line 277: perhaps rename this to something else in order to differentiate it from the futures v1 `payFee` function below.

e.g. `payFeeWithTracking`?

contracts/PerpsV2MarketBase.sol
```diff
@@ -573,22 +572,28 @@ contract PerpsV2MarketBase is MixinPerpsV2MarketSettings, IPerpsV2BaseTypes {
...
      * This is mutative because the circuit breaker stores the last price on every invocation.
+     * @param allowMarketPaused if true, checks everything except the specific market, if false
+     *  checks only top level checks (system, exchange, futures)
      */
-    function _assetPriceRequireSystemChecks() internal returns (uint) {
+    function _assetPriceRequireSystemChecks(bool allowMarketPaused) internal returns (uint) {
         // check that market isn't suspended, revert with appropriate message
-        _systemStatus().requireFuturesMarketActive(marketKey); // asset and market may be different
+        if (allowMarketPaused) {
+            // this will check system activbe, exchange active, futures active
```
barrasso, line 591: [np] typo `activbe` => `active`

barrasso (changes requested):

contracts/PerpsV2Settings.sol
```diff
@@ -82,8 +70,8 @@ contract PerpsV2Settings is Owned, MixinPerpsV2MarketSettings, IPerpsV2Settings
     /*
      * The maximum allowable notional value on each side of a market.
      */
-    function maxMarketValueUSD(bytes32 _marketKey) public view returns (uint) {
-        return _maxMarketValueUSD(_marketKey);
+    function maxSingleSideValueUSD(bytes32 _marketKey) public view returns (uint) {
```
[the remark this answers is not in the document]
artdgn: yeah good point, not sure why these views are public

contracts/PerpsV2Settings.sol
```diff
@@ -170,24 +154,14 @@ contract PerpsV2Settings is Owned, MixinPerpsV2MarketSettings, IPerpsV2Settings
...
-    function setMakerFee(bytes32 _marketKey, uint _makerFee) public onlyOwner {
-        require(_makerFee <= 1e18, "maker fee greater than 1");
-        _setParameter(_marketKey, PARAMETER_MAKER_FEE, _makerFee);
-    }
-
-    function setTakerFeeNextPrice(bytes32 _marketKey, uint _takerFeeNextPrice) public onlyOwner {
-        require(_takerFeeNextPrice <= 1e18, "taker fee greater than 1");
-        _setParameter(_marketKey, PARAMETER_TAKER_FEE_NEXT_PRICE, _takerFeeNextPrice);
+    function setBaseFee(bytes32 _marketKey, uint _baseFee) public onlyOwner {
+        require(_baseFee <= 1e18, "taker fee greater than 1");
```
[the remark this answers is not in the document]
artdgn: yep, missed when renaming stuff

contracts/FuturesMarketManager.sol
artdgn: I think the v1 method will be removed with v1 stuff after the migration, so I think it makes more sense to have the default method called just `payFee` and always expect a tracking code as part of its interface (and it's up to the caller to pass a 0).

barrasso (approved):

PR #1786, opened 2022-05-19
fix kovan futures: add axs feed, remove luna mkt

artdgn:
Futures summary views were broken because feeds were not added for AXS and LUNA.

Changes in config:
- add feed for AXS into feeds json
- remove LUNA from markets json

Changes on chain:
- add feed for AXS to rates contract
- remove LUNA market from futures manager

### DIFF
--- a/publish/deployed/kovan-ovm/futures-markets.json
+++ b/publish/deployed/kovan-ovm/futures-markets.json
@@ -182,20 +182,6 @@
 		"paused": false
 	},
 	{
-		"marketKey": "sLUNA",
-		"asset": "LUNA",
-		"takerFee": "0.0025",
-		"makerFee": "0.002",
-		"takerFeeNextPrice": "0.001",
-		"makerFeeNextPrice": "0.0",
-		"nextPriceConfirmWindow": "2",
-		"maxLeverage": "10",
-		"maxMarketValueUSD": "2000000",
-		"maxFundingRate": "0.1",
-		"skewScaleUSD": "300000000",
-		"paused": false
-	},
-	{
 		"marketKey": "sEUR",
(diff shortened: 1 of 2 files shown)


publish/deployed/kovan-ovm/futures-markets.json
```diff
@@ -181,20 +181,6 @@
 		"skewScaleUSD": "300000000",
 		"paused": false
 	},
-	{
-		"marketKey": "sLUNA",
-		"asset": "LUNA",
```
barrasso, line 186: Why are we removing LUNA from the `futures-markets.json` instead of adding the missing feed to `kovan-ovm/feeds.json`?

barrasso (changes requested):

publish/deployed/kovan-ovm/futures-markets.json
artdgn: I assumed that it might have invalid / paused feed again in the future (since it was delisted and CL paused feeds for it already previously), and might get forked etc etc...  so I thought that we probably don't want / need to really deal with that.

I guess on testnet we might as well have it and see how things break..  

So I can add it back. What would you prefer?

artdgn: Also CL removed it from mainnet-opti feeds (pretty sure it was there previously) and only left it on kovan, which is possibly a sign of future deprecation there as well (https://docs.chain.link/docs/optimism-price-feeds/)

barrasso: Let's keep LUNA removed 👍

barrasso (approved):

PR #1787, opened 2022-05-20
SIP-240: Update Liquidate Amount

barrasso:


contracts/Issuer.sol
```diff
@@ -717,8 +717,8 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
             // i.e. the value of the account's staking position relative to balanceOf will be unwound.
             totalRedeemed = transferableBalance;
 
-            // Liquidate their debt based on the ratio of their transferable collateral.
-            amountToLiquidate = debtBalance.multiplyDecimal(transferableBalance).divideDecimal(collateralForAccount);
+            // Liquidate the account's debt based on the liquidation penalty.
+            amountToLiquidate = debtBalance.multiplyDecimal(SafeDecimalMath.unit().add(penalty));
```
dbeal-eth, line 721: This line should be 

`amountToLiquidate = amountToLiquidate.multiplyDecimal(transferableBalance).divideDecimal(collateralForAccount)`

There are a couple other ways the calc could be done.

dbeal-eth (changes requested):

contracts/Issuer.sol
barrasso: resolved

artdgn (approved):

bachstatter:
When we calculate `totalRedeemed` on line 710 in `Issuer.sol` we add the penalty.
But doesn't `amountToLiquidate` from `liquidator().calculateAmountToFixCollateral` already have the penalty added?

dbeal-eth (approved):

PR #1792, opened 2022-05-26
SIP-243: Upgrade Collateral Short

barrasso:


contracts/Collateral.sol
```diff
@@ -178,6 +178,11 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
 
     /* ---------- SETTERS ---------- */
 
+    function setLoanClosed(uint id) external onlyOwner {
```
ajsantander, line 181: This will allow us to disable loans 722 and 771 for Lyra's hedger.

contracts/Collateral.sol
```diff
@@ -531,29 +540,38 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
...
-        // 2. Repay the accrued interest.
-        payment = payment.add(loan.accruedInterest);
+        // 2. Make sure they are not overpaying.
+        require(payment <= loan.amount, "Payment too high");
 
-        // 3. Make sure they are not overpaying.
-        require(payment <= loan.amount.add(loan.accruedInterest), "Payment too high");
+        // 3. Process the payment and pay fees if the loan has accrued interest.
+        payment = payment.add(loan.accruedInterest);
+        _processPayment(loan, payment);
```
ajsantander, line 548: _processPayment mints accrued interest and sends it to the fee pool.
So we account for that to be burnt later.

contracts/Collateral.sol
```diff
@@ -531,29 +540,38 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
...
-        loan.collateral = loan.collateral.sub(expectedAmount);
+        // 5. Reduce the collateral by the amount repaid.
+        // Note: expectedAmount is payment minus the exchange fees, so we need to include it.
+        uint collateralPaid = expectedAmount.add(exchangeFee);
+        loan.collateral = loan.collateral.sub(collateralPaid);
 
-        // 6. Process the payment and pay the exchange fees if needed.
-        _processPayment(loan, payment);
-        _payFees(fee, sUSD);
+        // 6. Pay the exchange fees (already converted to sUSD)
```
ajsantander, line 560: As above, _payFees mints the sUSD, so we need to burn it.

contracts/Issuer.sol
```diff
@@ -623,12 +623,9 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
      * @param short The address of the CollateralShort contract to be upgraded
      * @param amount The amount of sUSD collateral to be burnt
      */
-    bool public shortsUpgraded = false;
 
     function upgradeCollateralShort(address short, uint amount) external onlyOwner {
-        require(!shortsUpgraded, "Issuer: shorts already upgraded");
-        shortsUpgraded = true;
-        require(short == resolver.getAddress("CollateralShort"), "Issuer: wrong short address");
+        require(short == resolver.getAddress("CollateralShortLegacy"), "Issuer: wrong short address");
```
ajsantander, line 628: We need to register the CollateralShort contract that will be targeted by this purging function.

The function is now permanently open.

contracts/Collateral.sol
ajsantander: We would call this on the new contract but actually cause the disable in CollateralManager, which will affect the legacy CollateralShort.

contracts/Collateral.sol
leomassazza: _processPayment willl always pay `payment` whether it includes or not the accruedInterest (is added to payment at line 547).

contracts/Collateral.sol
```diff
@@ -382,9 +382,13 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
         (amount, collateral) = _repayWithCollateral(borrower, id, loan.amount);
 
         // 2. Pay the service fee for collapsing the loan.
-        uint serviceFee = amount.multiplyDecimalRound(getCollapseFeeRate(address(this)));
-        _payFees(serviceFee, sUSD);
-        collateral = collateral.sub(serviceFee);
+        if (collateral > 0) {
+            uint serviceFee = amount.multiplyDecimalRound(getCollapseFeeRate(address(this)));
+            _payFees(serviceFee, sUSD);
```
barrasso, line 387: The service fee should be paid in `loan.currency` instead of `sUSD` (since we are calculating the fee by multiplying by the `payment` amount).

It should then subtract the **sUSD equivalent value** of the `serviceFee` from `collateral`.

dbeal-eth (changes requested):
In general, to have full confidence in these fixes and the collateral shorts in the future, we need to rethink the tests and structure we have for. We should be breaking up the `Repaying shorts` section to have a `describe` clause that specifically covers every repayment function, and each describe clause should have `it`s that cover every state that changes.

By every state that changes, this includes:
* values for the entire `Loan` structure before and after calling the function
* sUSD balances for *all* affected accounts, including the fee address, collateral short, and the borrower
* globally registered `totalShorts` in the shorts manager
* any events which are emitted
* any other state changes that I may be not thinking of (check tenderly?)

contracts/Collateral.sol
```diff
@@ -382,9 +387,13 @@ contract Collateral is ICollateralLoan, Owned, MixinSystemSettings {
         (amount, collateral) = _repayWithCollateral(borrower, id, loan.amount);
 
         // 2. Pay the service fee for collapsing the loan.
-        uint serviceFee = amount.multiplyDecimalRound(getCollapseFeeRate(address(this)));
-        _payFees(serviceFee, sUSD);
-        collateral = collateral.sub(serviceFee);
+        if (collateral > 0) {
```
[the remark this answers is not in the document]
kylriley: What's the purpose of this `if` statement? Not sure why there should be a case for bypassing paying the service fee, unless it's to favor allowing users to close positions even if they can't afford it. In which case you may want to consider just removing the service fee altogether given the problem with it being %age based and being able to reduce your position before closing (effectively bypassing it).

ajsantander:
Deprecating in favor of https://github.com/Synthetixio/synthetix/pull/1793

PR #1782, opened 2022-05-13
Removing token transfer

justinjmoses:

PR #1794, opened 2022-05-30
SIP - 238, remove tokens double entry

artdgn:
Taking over from https://github.com/Synthetixio/synthetix/pull/1782


contracts/MultiCollateralSynth.sol
```diff
@@ -76,14 +76,13 @@ contract MultiCollateralSynth is Synth {
...
-    // Contracts directly interacting with multiCollateralSynth or wrapper to issue and burn
-    modifier onlyInternalContracts() {
-        bool isInternal = super._isInternalContract(msg.sender);
-        bool isEtherWrapper = msg.sender == address(etherWrapper());
-        bool isWrapper = wrapperFactory().isWrapper(msg.sender);
-        bool isMultiCollateral = collateralManager().hasCollateral(msg.sender);
-
-        require(isInternal || isWrapper || isMultiCollateral || isEtherWrapper, "Only internal contracts allowed");
-        _;
+    // overriding modifier from super to add more internal contracts and checks
```
artdgn, line 79: this is just removing code duplication between base and this

artdgn, line 79: just removing duplicate modifier

artdgn:
fork tests fail because of APE and DYDX futures being paused

contracts/BaseSynthetix.sol
```diff
@@ -475,6 +475,36 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+        // These entries are not required or cached in order to allow them to not exist (==address(0))
+        // e.g. due to not being available on L2 or at some future point in time.
+        return
+            caller == resolver.getAddress("RewardEscrowV2") ||
+            caller == resolver.getAddress("RewardEscrow") ||
+            caller == resolver.getAddress("SynthetixEscrow") ||
+            caller == resolver.getAddress("SynthetixBridgeToOptimism") ||
+            caller == resolver.getAddress("Depot");
+    }
+
```
ajsantander, line 507: Might be worth considering making this an external AddressResolver function `isInternalContract`. This would dedup code, but also improve performance a bit, since it would be 1 call + N reads vs N calls + N reads.

contracts/BaseSynthetix.sol
```diff
@@ -475,6 +475,36 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+    /// which isn't supported due to SIP-238 for other callers
+    function _isInternalTransferCaller(address caller) internal view returns (bool) {
+        // These entries are not required or cached in order to allow them to not exist (==address(0))
+        // e.g. due to not being available on L2 or at some future point in time.
+        return
+            caller == resolver.getAddress("RewardEscrowV2") ||
+            caller == resolver.getAddress("RewardEscrow") ||
+            caller == resolver.getAddress("SynthetixEscrow") ||
+            caller == resolver.getAddress("SynthetixBridgeToOptimism") ||
+            caller == resolver.getAddress("Depot");
```
ajsantander, line 505: Is this list missing TradingRewards?

ajsantander: Also ImportableRewardEscrowV2?

contracts/BaseSynthetix.sol
artdgn: I wanted to avoid having to upgrade resolver for this, and the resolver mixin as well (to avoid having to upgrade all inheriting contracts, or creating a divergence of source and deployed).

If I understand correctly, these calls are indeed inefficient for L1, but only happen in the exceptional case of these internal callers.

I'll look into optimizing the order of these checks so that short-circuiting reduces some gas costs for L1 users (e.g. putting more frequently called contracts first).

dbeal-eth: @artdgn is there a reason why the `resolverAddressesRequired` function cannot simply be updated to include all these contract addresses? Then you can just call `requireAndGetAddress()` from the resolver mixin, unless I am missing something

doesn't require update of resolver or its mixin

contracts/BaseSynthetix.sol
artdgn: Good catch, I missed it because it's using `safeTransfer` it seems. Interesting that no tests failed. 

It looks like in unit tests the rewards token is a mock, and no integration tests are done for it anymore.

Anyway, the contract isn't being used at all, but I've added it to the whitelist anyway.

artdgn:
> Question, do we need to also consider safeTransfer and safeTransferFrom for either SNX or synths?

Missed looking into those completely :+1:. 

Fortunately it seems nothing needs to be added for any of the contracts that use that library because it's mostly `StakingRewards*` / `ShortingRewards*` but they have the proxy SNX set as the reward token contract. (Fortunately because it would be difficult to add those to a whiltelist, since they have different name for each instance).

Also these contracts aren't active any longer (but the remaining ones should work still).

contracts/BaseSynthetix.sol
```diff
@@ -475,6 +475,36 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+        if (msg.sender == address(proxy)) {
+            // allow proxy through, messageSender should be already set correctly
+            return;
+        } else if (_isInternalTransferCaller(msg.sender)) {
+            // optionalProxy behaviour only for the internal legacy contracts
+            messageSender = msg.sender;
+        } else {
+            revert("Only the proxy can call");
+        }
+    }
```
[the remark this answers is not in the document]
ajsantander: Ok, to me dedup is always first, but ok. Feel free to close/resolve this comment.

ajsantander (approved):

contracts/BaseSynthetix.sol
artdgn: `resolverAddressesRequired` requires that addresses aren't 0, so if it will fail for those addresses:
- if they aren't present on one of the chains: some legacy contracts that aren't on L2, some L2 contracts that aren't on L1, the bridge from L1 when on L2... etc
- when those addresses are no longer exist on resolver (e.g. if resolver is upgraded but legacy contracts aren't added)

We can always have a workaround for having a dummy non-zero address for these cases, but it's just hacky, when we can just avoid checking 0 address, since msg.sender can't be zero anyway, and this whitelist is explicitly for the caller.

I don't think the gas is such a concern, because for regular transfers no extra calls or checks are done (because the first branch is taken), for for bridging only one check (call) is done which isn't too bad (after which the `||` short-circuits).

artdgn:
Integration test failure is the same as on `develop`

contracts/Synth.sol
```diff
@@ -270,15 +270,50 @@ contract Synth is Owned, IERC20, ExternStateToken, MixinResolver, ISynth {
...
+            caller == resolver.getAddress("SynthRedeemer") ||
+            caller == resolver.getAddress("WrapperFactory") || // transfer not used by users
+            // legacy
+            caller == resolver.getAddress("NativeEtherWrapper") ||
+            caller == resolver.getAddress("Depot");
+    }
+
     /* ========== EVENTS ========== */
     event Issued(address indexed account, uint value);
+
```
dbeal-eth, line 308: nit random lines added

dbeal-eth (approved):

contracts/Synth.sol
artdgn: auto formatting

PR #1797, opened 2022-06-01
move functions to collateralShort

leomassazza:


contracts/CollateralShort.sol
```diff
@@ -6,6 +6,10 @@ pragma experimental ABIEncoderV2;
 import "./Collateral.sol";
 
 contract CollateralShort is Collateral {
+    /* ========== CONSTANTS ========== */
+
+    bytes32 private constant sUSD = "sUSD";
```
ajsantander, line 11: nit: maybe this can be made internal in Collateral.sol instead of duplicated here?

ajsantander (approved):
Approved with just a nit.

contracts/CollateralShort.sol
leomassazza: fixed

PR #1825, opened 2022-06-20
SIP-252: Liquidation of SNX Escrow

artdgn:
Implements https://sips.synthetix.io/sips/sip-252/.

Changes other than what's covered in the SIP (and implementation details):
- Immutable storage now lives in `RewardEscrowV2Storage` (which talks to old `RewardEscrowV2`), so `RewardEscrowV2` is upgradable (and some storage fixes are possible if needed)
- `revokeFrom` is added to escrow (restricted to liqudiations)
- Contract size reductions in Issuer (since new code wasn't fitting in)
- Old `RewardEscrowV2` contracts are committed with`Frozen` suffix to be used in testing (and for reference)
- `RewardEscrow` (V1) migration functionality removed
- `Synthetix` has `migrateEscrowContractBalance` that is needed for migration of SNX balances from old to new escrow
- Liquidations of escrow can get expensive (4.5M gas for 100 entries in old contract)
- `liquidateDelinquentAccountEscrowIndex` is added to specify specific index to start looping from.
- liquidator collateral check considers escrow too.
- migration configuration script is added for deployment
- previous functionality for `RewardEscrowV2` extensions had missing tests (that were now added to validate the refactor)

--------- 
Projected runsheet actions:
- ~~Finish RewardEscrow (V1) migration~~ done by @dbeal-eth .
- Run the deployment as migration contract. If not running as migration contract, it's possible that FeePool might create an entry between two steps of the migration script, so if migration contract option is not used - issuance should be paused to prevent claiming for the duration of the migration.
- Test everything on old contract is bricked (also is tested in unit, integration & fork tests)
- Add `RewardEscrowV2Storage` to the `non-upgradable.json` list.

--------- 
Gas usage:
- `liquidateDelinquentAccount() with no escrow entries gas used: 684k` 
- `liquidateDelinquentAccount() with 100 escrow entries in **new** reward contract gas used: 1649k`
- `liquidateDelinquentAccount() with 100 escrow entries in **OLD** reward contract gas used: ~4200k` (not a result of a specific test, but estimated based on `setZeroAmountUntilTarget` tests below).
- `setZeroAmountUntilTarget() with 100 escrow entries on old contract gas used: 3683k`
- `setZeroAmountUntilTarget() with 100 escrow entries on new contract gas used: 1076k`

A more gas-efficient (but more complex alternative PR): https://github.com/Synthetixio/synthetix/pull/1839


contracts/Issuer.sol
```diff
@@ -697,7 +681,11 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
     function liquidateAccount(address account, bool isSelfLiquidation)
         external
         onlySynthetix
-        returns (uint totalRedeemed, uint amountToLiquidate)
+        returns (
+            uint totalRedeemed,
+            uint debtRemoved,
+            uint escrowToLiquidate
```
leomassazza, line 687: Update function documentation to reflect changes in return values

contracts/BaseSynthetix.sol
```diff
@@ -33,6 +34,7 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
     bytes32 private constant CONTRACT_REWARDSDISTRIBUTION = "RewardsDistribution";
     bytes32 private constant CONTRACT_LIQUIDATORREWARDS = "LiquidatorRewards";
     bytes32 private constant CONTRACT_LIQUIDATOR = "Liquidator";
+    bytes32 private constant CONTRACT_REWARDESCROWV2 = "RewardEscrowV2";
```
leomassazza, line 37: np: other contracts refers to this contract constant as CONTRACT_REWARDESCROW_V2

contracts/BaseSynthetix.sol
```diff
@@ -357,14 +408,23 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+        // incorrect accounting & transfers later
+        require(escrowToLiquidate == 0, "cannot self liquidate escrow");
 
-        emitAccountLiquidated(messageSender, totalRedeemed, amountLiquidated, messageSender);
+        emitAccountLiquidated(liquidatedAccount, totalRedeemed, debtRemoved, liquidatedAccount);
 
         // Transfer the redeemed SNX to the LiquidatorRewards contract.
         // Reverts if amount to redeem is more than balanceOf account (i.e. due to escrowed balance).
-        bool success = _transferByProxy(messageSender, address(liquidatorRewards()), totalRedeemed);
+        bool success = _transferByProxy(liquidatedAccount, address(liquidatorRewards()), totalRedeemed);
```
leomassazza, line 427: np: use same success variable name here and in line 384

artdgn: only one transfer in this scope, so naming it `liquidatorRewardTransferSucceeded` seems less readable to me

leomassazza: makes sense. Thanks

contracts/RewardEscrowV2Storage.sol
```diff
@@ -0,0 +1,293 @@
...
+
+        // append entryId to list of entries for account
+        _accountVestingEntryIds[account].push(entryId);
+
+        // Increment the next entry id.
+        nextEntryId++;
+
+        return entryId;
+    }
+
```
dbeal-eth, line 275: though this function would be unused, would be good to have a `setVestingEntry(address account, uint index, VestingEntries.VestingEntry calldata entry)` function as a escape hatch in case any future contract changes are conceived and the existing write functions that are provided are not sufficient for whatever reason.

I know technically its possible to do something like that with `addVestingEntry` and `setEntryZeroAmount`, but being able to update entries slots directly could be both more convenient and have gas benefits.

publish/releases.json
```diff
@@ -677,10 +677,21 @@
 			"released": "ovm"
 		},
 		{
-			"sip": 251,
+      "sip": 251,
```
dbeal-eth, line 680: np interesting tab?

dbeal-eth (commented):
looking good so far :+1: 

also, on the PR description could you post the latest example statistics/link for liquidation of one account with N number of escrow entries

this review is only half done. will provide other half of review tomorrow or something.

contracts/BaseRewardEscrowV2.sol
```diff
@@ -86,33 +74,57 @@ contract BaseRewardEscrowV2 is Owned, IRewardEscrowV2, LimitedSetup(8 weeks), Mi
...
+        return state().accountVestingEntryIDs(account, index);
     }
 
     /**
-     * @notice The number of vesting dates in an account's schedule.
+     * @notice A simple alias to totalEscrowedAccountBalance: provides ERC20 balance integration.
      */
-    function numVestingEntries(address account) external view returns (uint) {
-        return accountVestingEntryIDs[account].length;
+    function balanceOf(address account) public view returns (uint) {
```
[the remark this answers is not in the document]
artdgn: not new, it was part of previous contract too (and is used in tests for convenience)

contracts/BaseRewardEscrowV2.sol
```diff
@@ -128,16 +140,18 @@ contract BaseRewardEscrowV2 is Owned, IRewardEscrowV2, LimitedSetup(8 weeks), Mi
...
         // If the page extends past the end of the accountVestingEntryIDs, truncate it.
-        if (endIndex > accountVestingEntryIDs[account].length) {
-            endIndex = accountVestingEntryIDs[account].length;
+        if (endIndex > numVestingEntries(account)) {
+            endIndex = numVestingEntries(account);
         }
 
         uint256 n = endIndex - index;
+        uint256 entryID;
+        VestingEntries.VestingEntry memory entry;
```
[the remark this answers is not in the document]
artdgn: minute gas savings

contracts/BaseRewardEscrowV2.sol
```diff
@@ -166,14 +180,15 @@ contract BaseRewardEscrowV2 is Owned, IRewardEscrowV2, LimitedSetup(8 weeks), Mi
...
         uint256[] memory page = new uint256[](n);
         for (uint256 i; i < n; i++) {
-            page[i] = accountVestingEntryIDs[account][i + index];
+            page[i] = accountVestingEntryIDs(account, i + index);
         }
         return page;
     }
 
     function getVestingQuantity(address account, uint256[] calldata entryIDs) external view returns (uint total) {
+        VestingEntries.VestingEntry memory entry;
```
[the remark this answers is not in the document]
artdgn: same as above, no reason to redeclare variables every iteration

contracts/BaseRewardEscrowV2.sol
```diff
@@ -227,10 +246,60 @@ contract BaseRewardEscrowV2 is Owned, IRewardEscrowV2, LimitedSetup(8 weeks), Mi
...
+    function revokeFrom(
+        address account,
+        address recipient,
+        uint targetAmount,
+        uint startIndex
+    ) external onlySynthetix {
+        require(account != address(0), "account not set");
+        require(recipient != address(0), "recipient not set");
+
+        (uint total, uint endIndex, uint lastEntryTime) = state().setZerosUntilTarget(account, startIndex, targetAmount);
```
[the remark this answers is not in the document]
artdgn: I think the current name is narrow and specific, and something like `revokeEntries` implies that it also takes care of the aggregate accounting (total escrow amount) by handling all state related to revoking, which it doesn't. It only sets zeros until some target sum is reached, and everything else need to be handled outside of it.

[It can also be imagined to be used for more gas efficient vesting (if entry times are monotonic) or consolidating entries (e.g. before bridging / merging).]

contracts/BaseRewardEscrowV2.sol
```diff
@@ -227,10 +246,60 @@ contract BaseRewardEscrowV2 is Owned, IRewardEscrowV2, LimitedSetup(8 weeks), Mi
...
+            uint refund = total.sub(targetAmount);
+            state().addVestingEntry(
+                account,
+                VestingEntries.VestingEntry({endTime: uint64(lastEntryTime), escrowAmount: refund})
+            );
+        }
+
+        _subtractAndTransfer(account, recipient, targetAmount);
+
+        emit Revoked(account, recipient, targetAmount, startIndex, endIndex);
```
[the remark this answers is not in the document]
artdgn: `Vested` doesn't seem to be per entry though, only for the total amount, and the total amount here is not really vested. 

Do you think we should consider the revoked amount vested? Not sure myself really.

contracts/BaseRewardEscrowV2.sol
```diff
@@ -450,4 +488,5 @@ contract BaseRewardEscrowV2 is Owned, IRewardEscrowV2, LimitedSetup(8 weeks), Mi
         uint time
     );
     event NominateAccountToMerge(address indexed account, address destination);
+    event Revoked(address indexed account, address indexed recipient, uint targetAmount, uint startIndex, uint endIndex);
```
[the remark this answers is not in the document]
artdgn: Emitting one per entry would mean (other than extra gas) emitting the events in `setZerosUntilTarget`. That whole contract is not emitting any events (and if we start emitting there, we'd need to add events for all write methods for consistency).

The `targetAmount` and `account` make this event at least somewhat informative for accounting purposes.

Do you have any particular concerns or suggestions for this other then one event per entry?

contracts/BaseSynthetix.sol
```diff
@@ -279,6 +286,18 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
         return _transferFromByProxy(messageSender, from, to, value);
     }
 
+    // SIP-252: migration of SNX token balance from old to new escrow rewards contract
+    function migrateEscrowContractBalance() external onlyOwner {
```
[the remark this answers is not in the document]
artdgn: It either needs to remain there, or if it is removed, the migration script needs to be removed as well (and escrow will be assumed to not be upgradable). Right now the migration script is written allowing additional upgrades of RewardEscrowV2 (and it is removed from non-upgradable).

contracts/BaseSynthetix.sol
```diff
@@ -279,6 +286,18 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+        address from = resolver.requireAndGetAddress("RewardEscrowV2Frozen", "Old escrow address unset");
+        address to = resolver.requireAndGetAddress("RewardEscrowV2", "New escrow address unset");
```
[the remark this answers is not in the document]
artdgn: yes, but it will read from cache so another thing to possibly get wrong.

actually probably a good idea to require them being different addresses here.

contracts/RewardEscrowV2Storage.sol
artdgn: > I know technically its possible to do something like that with addVestingEntry and setEntryZeroAmount, but being able to update entries slots directly could be both more convenient and have gas benefits.

The reason entries were updated in place instead of being appended as in every other place in migrateAccount code is why ~~we can't have nice things~~ the gas costs are so high for current implementation. I think it's actually better to not allow this in the future, and setting an entry to zero, and append new ones should be sufficient for any scenario IMO. Can' think of a scenario that wouldn't be possible with this.

> in case any future contract changes are conceived and the existing write functions that are provided are not sufficient for whatever reason

I think this is a valid concern in general though, and I think it's worth to think through any other state changes that might be required but the current write methods don't allow (I think editing a specific entry is covered by zero & append though).

publish/releases.json
artdgn: used github UI for merge conflict resolution :disappointed:

contracts/BaseSynthetix.sol
```diff
@@ -314,9 +333,41 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
     /// @notice Force liquidate a delinquent account and distribute the redeemed SNX rewards amongst the appropriate recipients.
     /// @dev The SNX transfers will revert if the amount to send is more than balanceOf account (i.e. due to escrowed balance).
     function liquidateDelinquentAccount(address account) external systemActive optionalProxy returns (bool) {
-        (uint totalRedeemed, uint amountLiquidated) = issuer().liquidateAccount(account, false);
+        return _liquidateDelinquentAccount(account, 0);
+    }
+
+    /// @param escrowStartIndex: index into the account's vesting entries list to start iterating from
+    /// when liquidating from escrow in order to save gas (the default method uses 0 as default)
+    function liquidateDelinquentAccountEscrowIndex(address account, uint escrowStartIndex)
```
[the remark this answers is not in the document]
artdgn: nit, right?

contracts/Issuer.sol
```diff
@@ -128,8 +124,8 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
         return combineArrays(existingAddresses, newAddresses);
     }
 
-    function synthetix() internal view returns (ISynthetix) {
-        return ISynthetix(requireAndGetAddress(CONTRACT_SYNTHETIX));
+    function synthetixERC20() internal view returns (IERC20) {
```
artdgn, line 127: A lot of changes in `Issuer` are not related to the functionality, but were needed to reduce contract size (to allow space for the new changes)

contracts/BaseRewardEscrowV2.sol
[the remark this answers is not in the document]
dbeal-eth: `setZerosUntilTarget` makes the function sound more low level than it actually is. maybe instead call it `setEntryZeroAmountsUntilTarget` to make it clear this is sort of an range-capable wrapper around `setEntryZeroAmount` function. Alternately rename `setEntryZeroAmount` to be `setZeros`

contracts/BaseRewardEscrowV2.sol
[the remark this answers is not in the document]
dbeal-eth: i suspected that was the reason but most modern compilers are smart enough to optimize for declarations inside loops. Not sure if that means solidity compiler would actually optimize this, but even if it doesn't, the gas savings here would be almost negligible and it adds additional changes to a already long/detailed PR. The little things add up.

Since the change is already made its probably not worth it to change it back, but as I was reading through the PR it didn't seem to have any purpose related to the problem we are trying to solve so I figured I would verify.

contracts/BaseRewardEscrowV2.sol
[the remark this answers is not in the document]
dbeal-eth: in discord discussion it was decided that an additional `Vested` event is not necessary.

contracts/BaseRewardEscrowV2.sol
```diff
@@ -266,18 +335,35 @@ contract BaseRewardEscrowV2 is Owned, IRewardEscrowV2, LimitedSetup(8 weeks), Mi
...
+        /* No empty or already-passed vesting entries allowed. */
+        require(quantity != 0, "Quantity cannot be zero");
+        require(duration > 0 && duration <= max_duration, "Cannot escrow with 0 duration OR above max_duration");
 
-    function _reduceAccountEscrowBalances(address _account, uint256 _amount) internal {
-        // Reverts if amount being vested is greater than the account's existing totalEscrowedAccountBalance
-        totalEscrowedBalance = totalEscrowedBalance.sub(_amount);
-        totalEscrowedAccountBalance[_account] = totalEscrowedAccountBalance[_account].sub(_amount);
+        /* Add quantity to account's escrowed balance */
+        state().updateEscrowAccountBalance(account, int(quantity));
```
artdgn, line 348: this will use SafeCast

contracts/BaseSynthetix.sol
```diff
@@ -314,9 +336,41 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+    {
+        return _liquidateDelinquentAccount(account, escrowStartIndex);
+    }
 
-        emitAccountLiquidated(account, totalRedeemed, amountLiquidated, messageSender);
+    /// @notice Force liquidate a delinquent account and distribute the redeemed SNX rewards amongst the appropriate recipients.
+    /// @dev The SNX transfers will revert if the amount to send is more than balanceOf account (i.e. due to escrowed balance).
+    function _liquidateDelinquentAccount(address account, uint escrowStartIndex) internal returns (bool) {
+        // must store liquidator account address because below functions may attempt to transfer SNX which changes messageSender
+        address liquidatorAccount = messageSender;
```
artdgn, line 357: the usage of `messageSender` will be moved into the caller (which has the `optionalProxy` modifier)

publish/src/commands/deploy/deploy-core.js
```diff
@@ -93,11 +93,53 @@ module.exports = async ({
 		args: [account, addressOf(readProxyForResolver)],
 	});
 
+	const tokenStateSynthetix = await deployer.deployContract({
```
artdgn, line 96: this was moved but has no significance (was moved for a step that is no longer part of the PR)

artdgn:
To prevent confusion due to new commits I'm adding new commits in https://github.com/Synthetixio/synthetix/pull/1834 (so once you're done reviewing this PR, you can review the added changes there, and possible resolve some duplicate comments here if there are any)

contracts/BaseRewardEscrowV2.sol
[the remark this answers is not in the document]
artdgn: resolved in discord discussion

contracts/BaseRewardEscrowV2.sol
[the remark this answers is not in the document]
artdgn: will rename to `setZeroAmount` and `setZeroAmountUntilTarget` so that they are named similarly

test/contracts/Liquidator.js
```diff
@@ -984,6 +1059,235 @@ contract('Liquidator', accounts => {
...
+									);
+									assert.bnEqual(viewResult.escrowToLiquidate, escrowBefore.sub(escrowAfter));
+									assert.bnClose(
+										viewResult.debtToRemove,
+										toUnit('75').sub(debtAfter),
+										toUnit(0.01)
+									);
+								});
+							});
+							describe('last escrow entry remainder is added as new entry', () => {
```
dbeal-eth, line 1247: since this functionality is in `RewardEscrow` maybe move this test case over to `RewardEscrow` tests?

test/contracts/RewardEscrowV2Storage.js
```diff
@@ -0,0 +1,486 @@
...
+				assert.bnEqual(await instance.numVestingEntries(user1), 2);
+				// one unvested
+				assert.bnEqual(await instance.totalEscrowedAccountBalance(user1), entry1Amount);
+				// one vested
+				assert.bnEqual(await instance.totalVestedAccountBalance(user1), entry1Amount);
+			});
+		});
+	});
+
+	describe('public mutative methods on frozen contract', async () => {
```
dbeal-eth, line 136: unless this was changed later I believe we have these tests on integration as well, probably not necessary or useful to have the tests in unit because they don't get as accurate of a picture of the actual deployed chain state

test/contracts/Liquidator.js
```diff
@@ -984,6 +1059,235 @@ contract('Liquidator', accounts => {
...
+									await synthetix.liquidateDelinquentAccount(alice, { from: bob });
+									// no liquid balance added
+									assert.bnEqual(await synthetix.balanceOf(alice), 0);
+									// system debt is the same
+									assert.bnEqual(await synthetix.totalIssuedSynths(sUSD), totalDebt);
+									const escrowAfter = await rewardEscrowV2.balanceOf(alice);
+									const debtAfter = await synthetix.debtBalanceOf(alice, sUSD);
+									// escrow is mostly removed
+									assert.bnEqual(escrowAfter, 0);
+									assert.bnEqual(debtAfter, 0);
```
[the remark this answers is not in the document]
dbeal-eth: also cant remember as I was reading through, but do we have a test verifying that `liquidationAmounts` returns 0 after any of these liquidation operations is completed?

dbeal-eth:
also in `Liquidation.js` we should probably some specific tests for `liquidationAmounts` verifying it returns expected amounts in edge cases like no liquidatable amount (aka positive c-ratio)

contracts/Issuer.sol
```diff
@@ -716,38 +750,72 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+        (totalRedeemed, escrowToLiquidate) = _redeemableCollateralForTarget(account, redeemTarget, isSelfLiquidation);
+
+        // totalRedeemed == redeemTarget, all of the debt is removed, otherwise adjust it to redeemable collateral
+        if (totalRedeemed < redeemTarget) {
+            // Adjust debt amount to ratio of actual vs. target collateral amounts
+            debtToRemove = debtToRemove.multiplyDecimal(totalRedeemed).divideDecimal(redeemTarget);
+
+            // if all collateral is gone, erase the bad debt (some unliquidatable collateral can be in
+            // legacy SynthetixEscrow or LiquidatorRewards (if Synthetix liquidation method didn't call getReward() before this)
+            if (collateralForAccount == totalRedeemed) {
```
barrasso, line 775: Perhaps I'm being overly cautious here-- but is it worth considering the case where `totalRedeemed` is somehow not **exactly equal** to `collateralForAccount` and returns something slightly greater (i.e. by 1-2 wei)?  

e.g. due to a rounding error 

Or does this not matter since `collateralForAccount` will always be `>=` to `totalRedeemed` (when taking into account the unliquidatable collateral that's sitting in legacy `SynthetixEscrow` or `LiquidatorRewards`)?

contracts/Liquidator.sol
```diff
@@ -153,7 +174,7 @@ contract Liquidator is Owned, MixinSystemSettings, ILiquidator {
 
     /// @notice Checks if an account has enough SNX balance to be considered open for forced liquidation.
     function _hasEnoughSNX(address account) internal view returns (bool) {
-        uint balance = IERC20(address(synthetix())).balanceOf(account);
+        uint balance = issuer().collateral(account);
```
barrasso, line 177: Shouldn't this be checking only for *liquidateable* collateral, not *all* collateral?

e.g. `synthetix.balanceOf() + rewardEscrowV2().balanceOf()`

contracts/Liquidator.sol
```diff
@@ -153,7 +174,7 @@ contract Liquidator is Owned, MixinSystemSettings, ILiquidator {
...
         return balance > (getLiquidateReward().add(getFlagReward()));
```
barrasso, line 178: Side note:
I would consider removing this function `_hasEnoughSNX` altogether. 
This was added in SIP-148 mostly because it didn't make sense for liquidators to worry about insolvent accounts with such small amounts of SNX or escrow-only accounts. Now that escrow is liquidateable, it might not make sense to have this extra check.

Also, there's another problem with it:
  * you can flag an account via `flagAccountForLiquidation` even if `_hasEnoughSNX` returns `false`
  * after they are flagged, they cannot be liquidated since `isLiquidationOpen` will return `false`
  
 So flaggable, but unliquidateable accounts with small amounts of SNX will accumulate--  and this will become more of a problem if the reward values are raised in the future.

contracts/RewardEscrowV2.sol
```diff
@@ -237,9 +90,5 @@ contract RewardEscrowV2 is BaseRewardEscrowV2 {
     }
 
     /* ========== EVENTS ========== */
-    event MigratedAccountEscrow(address indexed account, uint escrowedAmount, uint vestedAmount, uint time);
```
barrasso, line 240: The `systemActive` modifier is unused. 
If this is intended, it could be removed-- along with `systemStatus()`.

test/integration/behaviors/liquidations.behavior.js
```diff
@@ -427,31 +479,33 @@ function itCanLiquidate({ ctx }) {
...
-				);
-				const expectedAmountToLiquidate = multiplyDecimal(
-					debtToLiquidate,
-					divideDecimal(beforeSnxBalance, totalRedeemed)
-				);
-
-				assert.bnEqual(snxRedeemed, beforeSnxBalance);
-				assert.bnClose(amountLiquidated.toString(), expectedAmountToLiquidate.toString(), '1000'); // the variance is due to a rounding error as a result of multiplication of the SNX rate
+				assert.bnEqual(snxRedeemed, beforeSnxBalance.add(beforeEscrowBalance));
+				assert.bnEqual(amountLiquidated.toString(), beforeDebtBalance.toString()); // the variance is due to a rounding error as a result of multiplication of the SNX rate
```
barrasso, line 508: can remove this comment

barrasso (changes requested):
overall contracts and tests look good - just had some minor feedback and nits

contracts/BaseRewardEscrowV2.sol
```diff
@@ -86,33 +74,57 @@ contract BaseRewardEscrowV2 is Owned, IRewardEscrowV2, LimitedSetup(8 weeks), Mi
...
+        return state().totalEscrowedAccountBalance(account);
+    }
+
+    function totalVestedAccountBalance(address account) public view returns (uint) {
+        return state().totalVestedAccountBalance(account);
+    }
+
+    function nextEntryId() public view returns (uint) {
+        return state().nextEntryId();
+    }
```
ajsantander, line 104: As slither is suggesting here, should we make these external? I see no internal reference to nextEntryId() from within this contract.

publish/src/commands/deploy/configure-reward-escrow.js
```diff
@@ -0,0 +1,93 @@
...
+		read: 'getAddress',
+		readArg: [toBytes32('RewardEscrowV2Frozen')],
+		expected: input => input === addressOf(frozenOrPreviousEscrow),
+		write: 'importAddresses',
+		writeArg: [[toBytes32('RewardEscrowV2Frozen')], [addressOf(frozenOrPreviousEscrow)]],
+		comment: 'Ensure that RewardEscrowV2Frozen is in the address resolver',
+	});
+
+	// move SNX balances
+	await runStep({
```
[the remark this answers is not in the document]
artdgn: I'll add the check, and this should address the gas consideration.

test/contracts/Liquidator.js
```diff
@@ -529,6 +564,37 @@ contract('Liquidator', accounts => {
...
+						// debt is fewer shares (but of higher debt per share), by (total - redeemed / total) more debt per share
+						const redeemed = toUnit('66.666666666666666667');
+						const shareMultiplier = divideDecimal(totalDebt, totalDebt.sub(redeemed));
+						assert.bnClose(
+							await synthetix.debtBalanceOf(alice, sUSD),
+							multiplyDecimal(debtBefore.sub(redeemed), shareMultiplier),
+							toUnit(0.001)
+						);
+					});
+				});
```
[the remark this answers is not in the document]
artdgn: The test checks that `"escrow balance is not used for self-liquidation"` by adding entries, triggering `liquidateSelf` and checking that escrow is untauched.

The revert is a precaution and can only be triggered if a bug is deployed in Liquidator somehow (in which case liquidator will ask Synthetix to liquidate from escrow, but Synthetix will revert, preventing it).

It's possible to trigger this by mocking a bad response from Liquidator, but I think it's not needed, as this precaution revert implementation is simple enough to verify, and cannot be triggered by any user input (I can think of).

test/contracts/Liquidator.js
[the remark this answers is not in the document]
artdgn: will add a post-action check as well

test/contracts/Liquidator.js
artdgn: the remainder functionality for `revokeFrom` is tested there as well (in `'should revoke and transfer SNX to recipient for partial amount'`)

so it might be a bit redundant, but it might be considered also a liquidation side effect, so might as well leave it here too.

test/contracts/RewardEscrowV2Storage.js
artdgn: This depends on functionality that should work in the unit tests as well (the inability to transfer using synthetix impl) because is part of the contract code (and not the contract state). It's one more place that might check a regression if this is changed in a way that breaks this assumtion.

contracts/Issuer.sol
artdgn: it shouldn't be possible to be in this condition, and I'm hesitant to change `==` to `<=` just because no rounding is taking place.

I think it would make sense to add a require to check for this if this was a value being passed from another contract. However, `totalRedeemed` is a result of the internal method `_redeemableCollateralForTarget` and it seems overkill to check the result of `<= collateralForAccount` in the same contract.

contracts/Liquidator.sol
artdgn: will add another view to Issuer for `liquidatableCollateral` (don't think `Liquidator` should have internal knowledge of that calculation)

contracts/Liquidator.sol
artdgn: I think the check does make some sense (if the reward is a constant amount), otherwise the liquidation tx cannot succeed. 

Wouldn't adding this check during `flagAccountForLiquidation` make this behavior consistent?

test/integration/behaviors/liquidations.behavior.js
artdgn: what do you mean? this is still a "close" comparison and not an exact comparison because of the `.toString()` (which truncates). There some issue with bnClose in integration tests (that's why it's used).

contracts/Liquidator.sol
barrasso: > otherwise the liquidation tx cannot succeed.

Where exactly would the tx fail? 
I was thinking that this case should be handled in the `else` block at L393 in `BaseSynthetix`:

```
  /* If the total amount of redeemed SNX is greater than zero 
  but is less than the sum of the flag & liquidate rewards,
  then just send all of the SNX to the LiquidatorRewards contract. */
  stakerRewards = totalRedeemed;
  ```
  
The liquidator doesn't get any flat reward for doing this, but it at least goes to stakers in the form of LiquidatorRewards.
It also cleans up small dust SNX that is stuck in inactive accounts. Let's say there are 10,000 accounts each with balanceOf 1-2 SNX... (which actually looks like is the case on Optimism already). That's already a decent amount of SNX that can't be liquidated and is just sitting until the reward threshold drops. And those rewards will likely never be <1 (*unless SNX moons to $2,192 and is altered by SCCP*).

Comment (artdgn) [reply contracts/Liquidator.sol:178]: Discussion continued in https://github.com/Synthetixio/synthetix/pull/1834#discussion_r917489876

File: contracts/BaseRewardEscrowV2.sol [hunk 30]
```diff
@@ -156,8 +170,8 @@ contract BaseRewardEscrowV2 is Owned, IRewardEscrowV2, LimitedSetup(8 weeks), Mi
         uint256 endIndex = index + pageSize;
 
         // If the page extends past the end of the accountVestingEntryIDs, truncate it.
-        if (endIndex > accountVestingEntryIDs[account].length) {
-            endIndex = accountVestingEntryIDs[account].length;
+        if (endIndex > numVestingEntries(account)) {
+            endIndex = numVestingEntries(account);
```
ajsantander, line 174: nit: avoid a call in the case the if statement body block is accessed by saving numVestingEntries to the stack

ajsantander (changes requested):
I see quite a bit of semantic overloading around the determination of wether a vesting entry is held in the old escrow contract or in the new one. Sometimes `entry.endTime == 0` is used, sometimes `entryId < fallbackCount`, and sometimes `amount = 0` (with the use of the `ZERO_PLACEHOLDER = -1`.

I call "semantic overloading" whenever you use a variable for more things that it's supposed to. E.g. `_totalEscrowedAccountBalance` should hold a balance, and only a balance, not an additional indicator (-1) pointing to where the value is stored.

As a developer, specially as a smart contract developer, one may be tempted to use semantic overloading for runtime optimization. My view regarding this is to always prioritize security over performance. Semantic overloading introduces points of fragility in the code, where one may forget assumptions in different parts of the code.

Even if the developer that introduces the overloads is smart enough to not make mistakes with them, code reviewers will find it harder to understand the code, and thus get lost in technicalities instead of detecting issues. Additionally, developers working in the code in the future are quite likely to make mistakes.

My suggestion for this PR then is to remove semantic overloading, and use explicit variables for understanding when values should be read from the old contract, and when from the new contract.

ajsantander (approved):

contracts/RewardEscrowV2Storage.sol
```diff
@@ -0,0 +1,345 @@
...
+    // id starting from which the new entries are stored in this contact only (and don't need to be read from fallback)
+    uint public firstNonFallbackId;
+
+    // -1 wei is a zero value placeholder in the read-through storage.
+    // needed to prevent writing zeros and reading stale values (0 is used to mean uninitialized)
+    // The alternative of explicit flags introduces its own set problems of ensuring they are written and read
+    // correctly (in addition to the values themselves). It adds code complexity, and gas costs, which when optimized
+    // lead to added coupling between different variables and even more complexity and potential for mistakenly
+    // invalidating or not invalidating the cache.
+    int internal constant ZERO_PLACEHOLDER = -1;
```
leomassazza, line 67: This is confusing and prone to errors. And the fact that there are two helper functions to read/write using the ZERO_PLACEHOLDER, those functions can be used to inspect an actual flag. 
Also, while looking at the helper function, the result is still an int (that can be negative), so -1 can be a valid value (from typing reading).

using a flag the complexity of the `_readWithZeroPlaceholder(int v)` is reduced to a transparent reading since the uninitialized flag is not artificially introduced in the value (the reading function it's replacing a -1 for a 0 and again, being the return type an int, -1 can be a valid number)

contracts/BaseSynthetix.sol
```diff
@@ -153,6 +161,23 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
 
+    /// the index of the first non zero RewardEscrowV2 entry for an account in order of iteration over accountVestingEntryIDs.
+    /// This is intended as a convenience off-chain view for liquidators to calculate the startIndex to pass
+    /// into liquidateDelinquentAccountEscrowIndex to save gas.
+    function getFirstNonZeroEscrowIndex(address account) external view returns (uint) {
+        uint numIds = rewardEscrowV2().numVestingEntries(account);
+        uint entryID;
+        VestingEntries.VestingEntry memory entry;
+        for (uint i = 0; i < numIds; i++) {
+            entryID = rewardEscrowV2().accountVestingEntryIDs(account, i);
```
leomassazza, line 172: I see those two functions are called together (same pattern) a couple of times. Here it implies two calls. 

You can add a single function that internally calls both to minimize the number of interactions.

contracts/Issuer.sol
```diff
@@ -354,21 +346,12 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
-        if (address(rewardEscrowV2()) != address(0)) {
-            balance = balance.add(rewardEscrowV2().balanceOf(account));
-        }
-
-        if (address(liquidatorRewards()) != address(0)) {
-            balance = balance.add(liquidatorRewards().earned(account));
-        }
-
-        return balance;
+        return
```
leomassazza, line 349: This now will revert if synthetixEscrow, rewardEscrowV2 or liquidatorRewards are not set, but I assume that will be seen easily at deployment time tests.

contracts/MintableSynthetix.sol
```diff
@@ -1,4 +1,5 @@
 pragma solidity ^0.5.16;
+pragma experimental ABIEncoderV2;
```
leomassazza, line 2: why do we have the pragma here?

artdgn: because of the use of `VestingEntries.VestingEntry` struct used in `BaseSynthetix` for `getFirstNonZeroEscrowIndex`

contracts/Issuer.sol
artdgn: This was previously the case as well because the checks in previous code would always be true. E.g. `address(synthetixEscrow()) != address(0) is always true because `synthetixEscrow()` uses `requireAndGetAddress` which requires it to not be 0.

This change is removing redundant code to reduce contract size (to be able to fit the other code).

contracts/BaseSynthetix.sol
artdgn: They are used in external calls only once, in this view method, which is expected to be called offline mostly.

Adding a convenience method doesn't really make sense since it adds contract code, and testing surface area, and won't really save much gas.

contracts/RewardEscrowV2Storage.sol
artdgn: >  uninitialized flag

Yes this issue was also raised by @ajsantander, I've explored a version with explicit flags (for all other cases of placeholders as well, which is almost all state variables) in https://github.com/Synthetixio/synthetix/pull/1838 and raised the issue to the auditors as well for their considerations.

In the end, that alternative code becomes prone to errors as well (because now you have to read and write two state variables for each state, and each value in an array), you have to both read and write them in sync, and handle and test possible mis-matches or misreads of the values. It adds gas costs, so leads to additional gas optimisations which lead to more complexity. It's also hard to read and reason about (but in a different way).

These value placeholders aren't super comfy in my view either, but even after implementing an alternative it didn't seem better. Also, no one was strongly in favour of that alternative version either (auditors or otherwise), so the more reviewed and simpler version remained.

> or change the helpers to return a uint and cast the result so

I'll move the casting around :+1:

test/integration/behaviors/liquidations.behavior.js
```diff
@@ -161,42 +142,55 @@ function itCanLiquidate({ ctx }) {
 
 				describe('getting liquidated', () => {
 					let tx;
-					let collateralBefore;
+					let cratioBefore;
```
leomassazza, line 145: rename to beforeCRatio as the other before*

test/integration/behaviors/liquidations.behavior.js
```diff
@@ -286,49 +293,54 @@ function itCanLiquidate({ ctx }) {
 				});
 
 				describe('getting liquidated', () => {
-					let tx;
-					let cratioBefore;
-					let beforeDebtShares, beforeSharesSupply;
+					let tx, viewResults;
+					let collateralBefore;
```
leomassazza, line 297: dito: beforeCRatio

test/contracts/RewardEscrowV2Storage.js
```diff
@@ -0,0 +1,566 @@
...
+
+const { ensureOnlyExpectedMutativeFunctions } = require('./helpers');
+const { toBytes32 } = require('../../index');
+
+const {
+	constants: { ZERO_ADDRESS },
+} = require('../..');
+
+contract('RewardEscrowV2Storage', async accounts => {
+	// const WEEK = 7 * 86400;
```
leomassazza, line 19: np: delete commented out line

test/contracts/RewardEscrowV2.js
```diff
@@ -73,221 +65,30 @@ contract('RewardEscrowV2', async accounts => {
 
 	describe('There are no escrow entries initially', async () => {
 		it('then numVestingEntries should return 0', async () => {
```
leomassazza, line 67: Comment says 0 and should be 2

test/contracts/RewardEscrowV2.js
```diff
@@ -73,221 +65,30 @@ contract('RewardEscrowV2', async accounts => {
 
 	describe('There are no escrow entries initially', async () => {
```
leomassazza, line 66: Fix comment. Expected 2 entries and not zero (no escrow entries)

leomassazza (approved):
Approved with some minor np in the tests. I marked as resolved the previous conversations you addressed.

test/integration/behaviors/liquidations.behavior.js
artdgn: this one is actually used as name implies it seems

barrasso (approved):

PR #1834, opened 2022-07-01
[draft]: internal review changes to #1825

artdgn:
Collecting changes added after internal review start (in order to stop changing reviewed PR) for https://github.com/Synthetixio/synthetix/pull/1825

### DIFF
--- a/contracts/BaseSynthetix.sol
+++ b/contracts/BaseSynthetix.sol
@@ -1,4 +1,5 @@
 pragma solidity ^0.5.16;
+pragma experimental ABIEncoderV2;
 
 // Inheritance
 import "./interfaces/IERC20.sol";
@@ -158,6 +159,23 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
 
     function transferableSynthetix(address account) external view returns (uint transferable) {
         (transferable, ) = issuer().transferableSynthetixAndAnyRateIsInvalid(account, tokenState.balanceOf(account));
+    }
+
+    /// the index of the first non zero RewardEscrowV2 entry for an account in order of iteration over accountVestingEntryIDs.
+    /// This is intended as a convenience off-chain view for liquidators to calculate the startIndex to pass
+    /// into liquidateDelinquentAccountEscrowIndex to save gas.
+    function getFirstNonZeroEscrowIndex(address account) external view returns (uint) {
+        uint numIds = rewardEscrowV2().numVestingEntries(account);
+        uint entryID;
+        VestingEntries.VestingEntry memory entry;
+        for (uint i = 0; i < numIds; i++) {
+            entryID = rewardEscrowV2().accountVestingEntryIDs(account, i);
+            entry = rewardEscrowV2().vestingSchedules(account, entryID);
+            if (entry.escrowAmount > 0) {
+                return i;
+            }
+        }
+        revert("all entries are zero");
     }
 
     function _canTransfer(address account, uint value) internal view returns (bool) {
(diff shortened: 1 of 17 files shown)


publish/src/commands/deploy/configure-legacy-settings.js
```diff
@@ -240,19 +239,6 @@ module.exports = async ({
 		});
 	}
 
-	// RewardEscrow on RewardsDistribution should be set to new RewardEscrowV2
```
artdgn, line 243: this step is done in the rewards escrow configuration script as well (and was a dupe here)

contracts/BaseSynthetix.sol
```diff
@@ -160,6 +161,23 @@ contract BaseSynthetix is IERC20, ExternStateToken, MixinResolver, ISynthetix {
...
+        uint entryID;
+        VestingEntries.VestingEntry memory entry;
+        for (uint i = 0; i < numIds; i++) {
+            entryID = rewardEscrowV2().accountVestingEntryIDs(account, i);
+            entry = rewardEscrowV2().vestingSchedules(account, entryID);
+            if (entry.escrowAmount > 0) {
+                return i;
+            }
+        }
+        revert("all entries are zero");
```
dbeal-eth, line 178: returning uint max would be better for bots than reverting here

contracts/Issuer.sol
```diff
@@ -735,6 +734,9 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
             uint debtBalance
         )
     {
+        // otherwise calculateAmountToFixCollateral reverts with unhelpful underflow error
+        require(liquidator().isLiquidationOpen(account, isSelfLiquidation), "Not open for liquidation");
```
dbeal-eth, line 738: would be better to do

```
if (!liquidator().isLiquidationOpen(account, isSelfLiquidation)) {
  return (0,0,0,0)
}
```

dbeal-eth (approved):

contracts/Liquidator.sol
```diff
@@ -173,8 +180,13 @@ contract Liquidator is Owned, MixinSystemSettings, ILiquidator {
...
     /// @notice Checks if an account has enough SNX balance to be considered open for forced liquidation.
-    function _hasEnoughSNX(address account) internal view returns (bool) {
-        uint balance = issuer().collateral(account);
+    function _hasEnoughSNXForRewards(address account) internal view returns (bool) {
+        /// SynthetixEscrow is unliquidatable so should be subtracted from SNX that can be used for rewards.
+        /// This has internal knowledge about collateral calculation (should only be in Issuer), and if SynthetixEscrow
+        /// is deprecated from Issuer's collateral balance - this needs to be updated too. The reason it's added here
+        /// is because it's an unlikely scenario and Issuer code size (and interface) are already too big.
+        uint unliquidatable = synthetixEscrow().balanceOf(account);
+        uint balance = issuer().collateral(account).sub(unliquidatable);
```
barrasso, line 189: This is better. Still think it can be removed altogether though: https://github.com/Synthetixio/synthetix/pull/1825#discussion_r917029785

barrasso (approved):

contracts/BaseSynthetix.sol
artdgn: The reason I revert here is that if there is not non-zero entry, than no number (be it last index, or maxUint) actually makes sense as an answer to `getFirstNonZeroEscrowIndex`. 

Why wouldn't a bot be able to handle a revert in a view?

contracts/Liquidator.sol
artdgn: I think removing it makes the rewards uncertain for the liquidators, which is probably a substantial mechanism change. I'm not sure it's a good change, now L1 flagger / liquidators know they are getting compensated for sure (the reward is small, but at least certain), if we remove this check - they now have to calculate themselves whether they'll be rewarded or not at all.

Also for the dust accounts: on L2 we can do a clean up by reducing the rewards, liquidating the dust, the raising rewards back up. It's an SCCP so is minimal effort. On L1 - probably not worth the gas anyway.

PR #1835, opened 2022-07-08
SIP-258 fix rate direction in ExchangeRatesWithDexPricing

dbeal-eth:
`ExchangeRatesWithDexPricing` was processing all dex exchange rate
queries in the same direction, creating unfavorable fills in unlikely circumstances on atomic swaps.

This change will ensure trade direction is taken into account while
computing rates.

Implements [SIP-258](https://sips.synthetix.io/sips/sip-258/)


contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -78,23 +78,21 @@ contract ExchangeRatesWithDexPricing is ExchangeRates {
...
-            // SRC is set to PureChainlinkPrice and DEST is NOT set to PureChainlinkPrice
-            sourceRate = systemSourceRate;
-            destRate = _getMaxValue(systemDestinationRate, _getPriceFromDexAggregator(destinationCurrencyKey, sourceAmount));
         } else {
-            // Otherwise, neither SRC nor DEST is set to PureChainlinkPrice.
-            sourceRate = _getMinValue(systemSourceRate, _getPriceFromDexAggregator(sourceCurrencyKey, sourceAmount));
-            destRate = _getMaxValue(systemDestinationRate, _getPriceFromDexAggregator(destinationCurrencyKey, sourceAmount));
+            destRate = _getMaxValue(
+                systemDestinationRate,
+                _getPriceFromDexAggregator(sUSD, destinationCurrencyKey, sourceAmount)
```
artdgn, line 92: Passing `sourceAmount` here makes not sense since this is a destination rate calculation, right?

I think this means that passing the amount into this method (and then `_dexPriceDestinationValue`) makes no sense in general, since it doesn't matter which amount is used, because it's just factored out anyway and the **rate** is returned. OR if external oracle somehow is using amount for liquidity based calculation (which I don't think it does) we need to pass a different amount here (because sourceAmount is incorrect if source currency is not sUSD).

Would probably be less confusing to remove the amount passing, and use an arbitrary amount for the oracle call.

contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -109,23 +107,30 @@ contract ExchangeRatesWithDexPricing is ExchangeRates {
     /// @notice Retrieve the TWAP (time-weighted average price) of an asset from its Uniswap V3-equivalent pool
     /// @dev By default, the TWAP oracle 'hops' through the wETH pool. This can be overridden. See DexPriceAggregator for more information.
     /// @dev The TWAP oracle doesn't take into account variable slippage due to trade amounts, as Uniswap's OracleLibary doesn't cross ticks based on their liquidity. See: https://docs.uniswap.org/protocol/concepts/V3-overview/oracle#deriving-price-from-a-tick
-    /// @param currencyKey The currency key of the synth we're retrieving the price for
+    /// @dev One of `sourceCurrencyKey` or `destCurrencyKey` should be sUSD. There are two parameters to indicate directionality. Because this function returns "price", if the source is sUSD, the result will be flipped.
```
artdgn, line 110: this seems confusing, I think it would be less confusing to pass the non-sUSD currencyKey, and a flag of whether it is source or destination.

contracts/ExchangeRatesWithDexPricing.sol
```diff
@@ -109,23 +107,30 @@ contract ExchangeRatesWithDexPricing is ExchangeRates {
...
+        IERC20 destEquivalent = IERC20(getAtomicEquivalentForDexPricing(destCurrencyKey));
+        require(address(destEquivalent) != address(0), "No atomic equivalent for dest");
 
         uint result =
-            _dexPriceDestinationValue(inputEquivalent, susdEquivalent, amount).mul(SafeDecimalMath.unit()).div(amount);
+            _dexPriceDestinationValue(sourceEquivalent, destEquivalent, amount).mul(SafeDecimalMath.unit()).div(amount);
         require(result != 0, "Result must be greater than 0");
 
-        return result;
+        return destCurrencyKey == "sUSD" ? result : SafeDecimalMath.unit().divideDecimalRound(result);
```
artdgn, line 133: nit: I think you already have `sUSD` constant in scope

artdgn (changes requested):
- I think there are tests only for sUSD trades (sUSD->sETH, and reverse). Should probably include a (sBTC -> sETH).
- using sourceAmount for destination rate calculation

contracts/ExchangeRatesWithDexPricing.sol
dbeal-eth: just trying to make the minimum changes to fix this issue which is why the `sourceAmount` argument was not removed

contracts/ExchangeRatesWithDexPricing.sol
dbeal-eth: booleans are opaque at the location of the function call as to what they indicate and don't follow the pattern used on other parts of the repo (almost every other function in Exchanger/ExchangeRates uses `sourceCurrencyKey, destCurrencyKey` for args). it is confusing that the function returns `price` instead of a `rate`, so I can understand ify ou wanted that changed, but would like more opinions on this because I would consider this exactly the opposite (ie, that using a boolean is more confusing/harder to track than argument ordering)

dbeal-eth:
> Looks like the SIP link is dead? Also, do we have a test that passes here and fails on `develop`?

yes I recall there are about 10 tests failing if the solidity code was checked into `develop`. The reason for the failure is because the mock dex aggregator has the wrong assumption about the behavior of aggregator when flipping input tokens (this can be verified by just using the on-chain aggregator and observing the result).

we should probably add another test to verify trade going the other direction as arthur mentioned, but the ticket is still in draft so give me a little more time :smile:

contracts/ExchangeRatesWithDexPricing.sol
dbeal-eth: whoops

contracts/ExchangeRatesWithDexPricing.sol
kaleb-keny: the fixes look fine to me, as long as `_getPriceFromDexAggregator` gets the min return amount, which i assume it does

kaleb-keny:
> Looks like the SIP link is dead? Also, do we have a test that passes here and fails on `develop`?

hey SIP-258 has now been merged, waiting for afif review, but seems to be in line with db's proposed here. Just one thing, need to redeploy `DexPriceAggregatorUniswapV3` with default fee of 5 bp and see if that works... In order to be able to hop over the 5 bp btc/eth & eth/usdc pools

PR #1838, opened 2022-07-12
[draft] Escrow liquidation explicit caching

artdgn:
Rewrote storage fallback access in `RewardEscrowV2Storage` https://github.com/Synthetixio/synthetix/pull/1825 to use explicit flags instead of implicit logic.


contracts/RewardEscrowV2Storage.sol
```diff
@@ -85,85 +96,79 @@ contract RewardEscrowV2Storage is IRewardEscrowV2Storage, State {
...
+                : fallbackRewardEscrow.numVestingEntries(account);
     }
 
     /* ========== MUTATIVE FUNCTIONS ========== */
 
     /// zeros out a single entry
-    function setZeroAmount(address account, uint entryId) public initialized onlyAssociatedContract {
+    /// note withCached(account) is not needed here since it operates on a specific entry, and non aggregate
+    /// views are used
+    function setZeroAmount(address account, uint entryId) public withFallback onlyAssociatedContract {
```
ajsantander, line 163: A bit of overloading here. I understand you are doing it to save a `CALL`. Perhaps be clear about it and call the function `setZeroAmountAndInitialize(...)`

ajsantander (approved):
IMO this makes the main PR much easier to read. Left a minor comment tho.

contracts/RewardEscrowV2Storage.sol
artdgn: not sure what you mean, any write has to also set the initialized flags if not set

PR #1844, opened 2022-07-21
SIP-257: update ExchangeRates to handle up to 27 decimals

artdgn:
https://sips.synthetix.io/sips/sip-257/

https://github.com/synthetixio/project-management/issues/362


publish/releases.json
```diff
@@ -681,6 +681,11 @@
 			"layer": "both",
 			"sources": ["Issuer", "SystemSettings", "SystemSettingsLib"],
 			"released": "both"
+		},
+		{
+			"sip": 257,
+			"layer": "both",
```
barrasso, line 687: Should consider releasing to `"layer": "ovm"` instead since:
* the SIP explicitly states that `Network: Optimism` 
* the intended purpose is to enable a futures market only available on L2

barrasso (changes requested):
LGTM, just one comment on deployment layer

publish/releases.json
artdgn: Functionally it is only needed for optimism, but releasing it only there will create a divergence between source and deployed for mainnet which is probably a bad idea.

Also, when the next change will happen in this contract (e.g. 230) it will be released to mainnet anyway, since we won't create two versions of it.

You're right about the SIP, although it is mainly about the new added market (which is only on optimism), but the Rates contracts change is technically (due to above reasons) only possible for both layers.

artdgn: Updating the SIP: https://github.com/Synthetixio/SIPs/pull/1143

barrasso (approved):

PR #1848, opened 2022-07-22
CI fixes

dbeal-eth:
fixes are:
* update cannon
* use locked version of anvil

this should fix:
* flakes
* misc issues when editing cannonfiles with unpredictable behavior
* better storage
* better UX


test/integration/utils/rates.js
```diff
@@ -122,6 +122,12 @@ async function _setMissingRates({ ctx }) {
 			await addAggregatorAndSetRate({ ctx, currencyKey, rate: ethers.utils.parseEther('1') });
 		}
 	}
+
```
artdgn, line 125: this shouldn't be done if in fork mode

artdgn (changes requested):
adding SNX rate shouldn't be done in fork tests

test/integration/utils/rates.js
dbeal-eth: ty https://github.com/Synthetixio/synthetix/pull/1848/commits/16ab4ba320857cec3e06151b11ecc7289a717dc1

artdgn (approved):

PR #1850, opened 2022-07-25
Goerli fixes

bachstatter:
Two small fixes for goerli.
 
We also have one other inconsistency:
On other networks the target name for the Synthetix proxy is `ProxyERC20`, on goerli its `ProxySynthetix`
And for sUSD we expect `ProxyERC20sUSD` but its called `ProxysUSD`. 
See: https://github.com/Synthetixio/js-monorepo/blob/master/packages/contracts-interface/src/index.ts#L111

Should I make updates to accomodate this change or do we plan to align the naming?


barrasso (approved):

barrasso:
LGTM - thx for finding these missing pieces @bachstatter 

> Should I make updates to accomodate this change or do we plan to align the naming?

I don't think you should make any changes just yet.

We'll likely want to align the naming convention to whatever we had previously used in `kovan` ... and more importantly, align it to whatever what we use for `mainnet` (which I assume is `ProxyERC20` and `ProxyERC20sUSD`).

If so, we'll make sure the target names are correctly updated to the ERC20 versions during the next testnet release.

bachstatter:
Cool yep I def prefer to have naming aligned!

PR #1854, opened 2022-07-30
impersonate the deployer using cannon

dbeal-eth:
this will allow for prototype testing on real forked network without
needing any keys or anything.

here is a command to fork mainnet. Any SIPs configured for an unreleased
release in `publish/releases.json` will be deployed

```
npx hardhat cannon:build --file cannonfile.release.toml --network mainnet --dry-run network=mainnet --impersonate --fund-signers --port 8545
```

once we feel more comfortable with the process we might want to add some
convenience `npm` commands to make bringin up these environments easier
so the command doesn't have to be as long


kaleb-keny (approved):

kaleb-keny (approved):

publish/deployed/local-ovm/deployment.json
```diff
@@ -5,7 +5,7 @@
 			"address": "0x08aE71038f091f18feE012de4Dba4365cAF80bCC",
 			"source": "SafeDecimalMath",
```
barrasso, line 6: why are there `local-ovm/deployment.json` changes? shouldn't this be empty?

publish/deployed/mainnet/deployment.json
```diff
@@ -16,16 +21,23 @@
...
 			"txn": "https://etherscan.io/tx/0xa582498dfb84e432ce0e8d0f1e61e4874169337b68ef38a35afd9bee23f00b84",
-			"network": "mainnet"
+			"network": "mainnet",
+			"constructorArgs": [
+				"0x971e78e0C92392A4E39099835cF7E6aB535b2227"
+			]
 		},
 		"ExchangeRates": {
 			"name": "ExchangeRates",
-			"address": "0xb4dc5ced63C2918c89E491D19BF1C0e92845de7C",
```
barrasso, line 23: I think these mainnet deployment files were accidentally committed

barrasso (changes requested):
need to revert deployment.json changes

publish/releases.json
```diff
@@ -1289,7 +1289,7 @@
 				"major": 2,
 				"minor": 75
 			},
-			"sips": []
```
barrasso, line 1292: also should revert this change as it is in `staging` and not yet in develop

barrasso (approved):

PR #1863, opened 2022-08-10
SIP 249, 269: Add BNB, OP perps

barrasso:


davidvuong (approved):
I don't see any obvious issues. LGTM 👍

PR #1870, opened 2022-08-17
Make very minor copy changes to require err msgs

davidvuong:
Most require error messages are in sentence case.

This also removes trailing white spaces on some files. Rather than adding 20 suggestion comments in PR #1859, I did it all in one easy to approve/merge commit.


leomassazza (approved):
LGTM. Just one comment: replacing `skewScale` with `Skew Scale` increases a bit the code size. I think we can spare those bytes, but just pointing out, since Arthur did a lot of work reducing the codesize.

PR #1873, opened 2022-08-19
Refactor and reduce stmts

davidvuong:
1. Remove the storage of numOfMarkets given it's only used in one
   location further below and it's already quite clear what it is.
2. Minor consistent comments changes / remove redundant comment
3. Use ++a instead of ++ then 2nd stmt to refetch on id incr


contracts/FuturesMarketManager.sol
```diff
@@ -165,8 +165,7 @@ contract FuturesMarketManager is Owned, MixinResolver, IFuturesMarketManager, IF
     function _totalDebtV1() internal view returns (uint debt, bool isInvalid) {
         uint total;
         bool anyIsInvalid;
-        uint numOfMarkets = _markets.elements.length;
-        for (uint i = 0; i < numOfMarkets; i++) {
+        for (uint i = 0; i < _markets.elements.length; i++) {
```
artdgn, line 168: Looking up the `.length` of a storage array is a storage lookup (SLOAD) which is a non-negligible amount of gas. Using it directly in the for loop stop condition repeats that for every iteration. So the pattern for these types of loops is to cache the length in a memory variable like `numOfMarkets` and use that instead.

It's a common pattern, that may not be important for L2 gas costs, but is kinda standard practice AFAIK.

contracts/PerpsStorageV2.sol
```diff
@@ -159,7 +157,7 @@ contract PerpsStorageV2 is IPerpsStorageV2External, IPerpsStorageV2Internal, IPe
         _position.lastPrice = price;
         // update funding entry to last entry
         _position.lastFundingEntry = lastFundingEntry[marketKey];
-        return _position; // returns memory
+        return _position;
```
artdgn, line 160: I think keeping this comment is good for clarity, since the method works with `Position storage` but in the end returns `Position memory`.

contracts/PerpsStorageV2.sol
```diff
@@ -108,11 +108,9 @@ contract PerpsStorageV2 is IPerpsStorageV2External, IPerpsStorageV2Internal, IPe
 
         // if position has no id, it wasn't initialized, initialize it:
         if (_position.id == 0) {
-            // id
-            marketScalars[marketKey].lastPositionId++; // increment position id
+            uint id = ++marketScalars[marketKey].lastPositionId; // increment position id and return
```
artdgn, line 111: I think I prefer keeping these two operations separated for clarity. These makes the code denser to read, and a bit more error prone to refactore (since the line now does two different important things).

artdgn (changes requested):
Some issues with a couple of the suggested changes.

artdgn (approved):

PR #1881, opened 2022-08-23
Liquidators QoL

leomassazza:
Ref: https://www.notion.so/Liquidations-QoL-077c9cedfb934ee8b167279ed45ffa02

### DIFF
--- a/contracts/PerpsEngineV2Base.sol
+++ b/contracts/PerpsEngineV2Base.sol
@@ -273,9 +273,9 @@ contract PerpsEngineV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2, IPerpsEn
     /// Upon liquidation, the positions will be closed, and the liquidation fees minted into the liquidator's account.
     function liquidatePositions(
         bytes32 marketKey,
-        address[] calldata accounts,
+        address[] memory accounts,
         address liquidator
-    ) external returns (bool[] memory liquidated) {
+    ) public returns (bool[] memory liquidated) {
         // check that market is supported by manager
         // this is needed in case market was removed, since this method is not guarded by
         // onlyOrdersRouter so doesn't check approvedRouterAndMarket on manager
@@ -293,6 +293,17 @@ contract PerpsEngineV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2, IPerpsEn
         }
 
         return liquidated;
+    }
+
+    /// backwards compatible liquidatePositon where a single position is liquidated by msg.sender
+    function liquidatePosition(bytes32 marketKey, address account) external {
+        address[] memory accounts = new address[](1);
+        accounts[0] = account;
+        bool[] memory result = liquidatePositions(marketKey, accounts, msg.sender);
+
+        if (!result[0]) {
+            revert(_errorMessages[uint8(Status.CannotLiquidate)]);
+        }
     }
 
     /* ========== INTERNAL TYPES ========== */
--- a/contracts/PerpsEngineV2Base.sol
+++ b/contracts/PerpsEngineV2Base.sol
@@ -268,14 +268,14 @@ contract PerpsEngineV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2, IPerpsEn
         _manager().issueSUSD(marketKey, to, amount);
     }
 
-    /// Liquidate a position if its remaining margin is below the liquidation margin.
-    /// Reverts if account should not be liquidated.
-    /// Upon liquidation, the position will be closed, and the liquidation fee minted into the liquidator's account.
-    function liquidatePosition(
+    /// Liquidate a group of position if for each, its remaining margin is below the liquidation margin.
+    /// Reverts if any account should not be liquidated.
+    /// Upon liquidation, the positions will be closed, and the liquidation fees minted into the liquidator's account.
+    function liquidatePositions(
         bytes32 marketKey,
-        address account,
+        address[] calldata accounts,
         address liquidator
-    ) external {
+    ) external returns (bool[] memory liquidated) {
         // check that market is supported by manager
         // this is needed in case market was removed, since this method is not guarded by
         // onlyOrdersRouter so doesn't check approvedRouterAndMarket on manager
@@ -285,7 +285,14 @@ contract PerpsEngineV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2, IPerpsEn
 
         uint price = _assetPriceRequireSystemChecks(marketKey);
         _recomputeFunding(marketKey, price);
-        _liquidatePosition(marketKey, account, price, liquidator);
+
+        liquidated = new bool[](accounts.length);
+
+        for (uint i = 0; i < accounts.length; i++) {
+            liquidated[i] = _liquidatePosition(marketKey, accounts[i], price, liquidator);
+        }
+
+        return liquidated;
     }
 
     /* ========== INTERNAL TYPES ========== */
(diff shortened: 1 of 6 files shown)


contracts/PerpsEngineV2ViewsMixin.sol
```diff
@@ -86,6 +86,22 @@ contract PerpsEngineV2ViewsMixin is PerpsEngineV2Base {
         return _stateViews();
     }
 
+    /// view that returns an array of bools indication for each account if can be liquidated
+    function canLiquidate(bytes32 marketKey, address[] calldata accounts)
```
bachstatter, line 90: 😍

contracts/PerpsEngineV2Base.sol
```diff
@@ -268,14 +268,14 @@ contract PerpsEngineV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2, IPerpsEn
         _manager().issueSUSD(marketKey, to, amount);
     }
 
-    /// Liquidate a position if its remaining margin is below the liquidation margin.
-    /// Reverts if account should not be liquidated.
-    /// Upon liquidation, the position will be closed, and the liquidation fee minted into the liquidator's account.
-    function liquidatePosition(
+    /// Liquidate a group of position if for each, its remaining margin is below the liquidation margin.
+    /// Reverts if any account should not be liquidated.
```
artdgn, line 272: this comment line is incorrect I think (it won't revert)

contracts/PerpsEngineV2Base.sol
```diff
@@ -285,7 +285,25 @@ contract PerpsEngineV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2, IPerpsEn
...
+
+        for (uint i = 0; i < accounts.length; i++) {
+            liquidated[i] = _liquidatePosition(marketKey, accounts[i], price, liquidator);
+        }
+
+        return liquidated;
+    }
+
+    /// backwards compatible liquidatePositon where a single position is liquidated by msg.sender
+    function liquidatePosition(bytes32 marketKey, address account) external {
```
artdgn, line 299: I think this method should be part of orders router, since 1. that's the contract that handles msg.sender auth for other orders, and 2. is technically just a convenience method, so shouldn't clutter the engine interface and contract.

I think the tests you've added for this can be just moved to the order tests as well.

contracts/PerpsEngineV2Base.sol
```diff
@@ -517,16 +535,19 @@ contract PerpsEngineV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2, IPerpsEn
...
     }
 
+    /// Returns false if position cannot be liquidated
     function _liquidatePosition(
         bytes32 marketKey,
         address account,
         uint price,
         address liquidator
-    ) internal {
+    ) internal returns (bool) {
```
artdgn, line 544: maybe this now needs a rename to `_tryLiquidatePosition` or something to be clear that it only may do the action, and the result should be checked

contracts/PerpsEngineV2ViewsMixin.sol
```diff
@@ -86,6 +86,22 @@ contract PerpsEngineV2ViewsMixin is PerpsEngineV2Base {
...
+        external
+        view
+        returns (bool[] memory liquidatables)
+    {
+        (uint price, bool isInvalid) = assetPrice(marketKey);
+        liquidatables = new bool[](accounts.length);
+
+        for (uint i = 0; i < accounts.length; i++) {
+            Position memory position = _stateViews().position(marketKey, accounts[i]);
+            liquidatables[i] = !isInvalid && _canLiquidate(position, price);
```
artdgn, line 100: can probably skip the loop entirely if `isInvalid` is true

contracts/PerpsEngineV2ViewsMixin.sol
```diff
@@ -86,6 +86,22 @@ contract PerpsEngineV2ViewsMixin is PerpsEngineV2Base {
         return _stateViews();
     }
 
+    /// view that returns an array of bools indication for each account if can be liquidated
```
artdgn, line 89: should mention that invalidity is rolled in to the resulting bools

contracts/PerpsOrdersV2Base.sol
```diff
@@ -126,6 +126,11 @@ contract PerpsOrdersV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2 {
         return engineContract().maxOrderSizes(marketKey);
     }
 
+    /// view returning true for each account if and only if the account can liquidate.
+    function canLiquidate(bytes32 marketKey, address[] calldata accounts) external view returns (bool[] memory) {
```
artdgn, line 130: coverage complains about this view. 

I think it's not needed really, because the single account `canLiquidate` is returned in `positionSummary` and for the batch view, they would use the engine anyway, because that's where the batch liquidation method is.

test/contracts/PerpsEngineV2.js
```diff
@@ -3277,7 +3311,441 @@ contract('PerpsEngineV2', accounts => {
 				assert.bnClose((await marketSummary()).marketDebt, toUnit('620'), toUnit('0.1'));
 				assert.bnClose((await marketSummary()).unrecordedFunding, toUnit('-8'), toUnit('0.01'));
 
-				await instance.liquidatePosition(marketKey, trader, liquidator);
+				await instance.liquidatePositions(marketKey, [trader], liquidator);
```
artdgn, line 3314: any idea what happened to the diff here, these test cases seem to already exist, but are shown as new here :thinking:

artdgn (changes requested):
Some small notes. Also if you can add some context into the PR description in addition to the notion link, since notion is not openly accessible, and it would be nice to provide some minimal context for PRs for anyone without that access.

contracts/PerpsEngineV2Base.sol
```diff
@@ -285,7 +285,25 @@ contract PerpsEngineV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2, IPerpsEn
...
+        liquidated = new bool[](accounts.length);
+
+        for (uint i = 0; i < accounts.length; i++) {
+            liquidated[i] = _liquidatePosition(marketKey, accounts[i], price, liquidator);
+        }
+
+        return liquidated;
+    }
+
+    /// backwards compatible liquidatePositon where a single position is liquidated by msg.sender
```
davidvuong, line 298: Minor: Typo in `liquidatePositon`.

contracts/PerpsEngineV2ViewsMixin.sol
davidvuong: Q & Nit: There's generally an inconsistent use of the `is` prefix to denote booleans throughout the project (see: `tooVolatile` but `isInvalid` and `invalid` mixed). What is the convention? `isXXX` is how I've always named my variables, web3 or otherwise.

contracts/PerpsEngineV2Base.sol
```diff
@@ -285,7 +285,25 @@ contract PerpsEngineV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2, IPerpsEn
 
         uint price = _assetPriceRequireSystemChecks(marketKey);
         _recomputeFunding(marketKey, price);
-        _liquidatePosition(marketKey, account, price, liquidator);
+
+        liquidated = new bool[](accounts.length);
```
davidvuong, line 289: Q: Is it generally accepted to simply noop on mutative calls when user input results in a pointless invocation (as opposed to reverting)? In our case, return an empty array when input `accounts.length == 0`?

artdgn: I think it makes sense to revert in this particular case (length 0). Although a no-op can still happen in this method if no account is actually liquidatable, but it seems reasonable.

leomassazza:
To be moved to PerpsV2 / V3

PR #1882, opened 2022-08-24
Feature/shortcut methods

davidvuong:
This PR adds additional convenience methods to the `contracts/PerpsOrdersV2Base.sol` contract. The intent is to reduce the number of transactions traders are required to sign and submit to improve usability when interacting with contracts through dApps like Kwenta.

The methods added are:

- `transferAndTrade`
- `tradeAndTransfer`
- `closeAndWithdraw`

You can find more details on specifics in comments within the codebase.

### DIFF
--- a/contracts/PerpsOrdersV2Base.sol
+++ b/contracts/PerpsOrdersV2Base.sol
@@ -282,6 +282,6 @@ contract PerpsOrdersV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2 {
 
     function _closePosition(bytes32 marketKey, bytes32 trackingCode) internal {
         int size = stateContract().position(marketKey, msg.sender).size;
-        _trade(marketKey, - size, trackingCode);
+        _trade(marketKey, -size, trackingCode);
     }
 }
--- a/contracts/PerpsOrdersV2Base.sol
+++ b/contracts/PerpsOrdersV2Base.sol
@@ -258,13 +258,23 @@ contract PerpsOrdersV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2 {
     }
 
     /// Shortcut method to transfer margin and trade in a single tx.
-    function transferAndTrade(bytes32 marketKey, int marginDelta, int sizeDelta, bytes32 trackingCode) external {
+    function transferAndTrade(
+        bytes32 marketKey,
+        int marginDelta,
+        int sizeDelta,
+        bytes32 trackingCode
+    ) external {
         _engineInternal().transferMargin(marketKey, msg.sender, marginDelta);
         _trade(marketKey, sizeDelta, trackingCode);
     }
 
     /// Shortcut method to trade and transfer margin in a single tx (inverse of transferAndTrade).
-    function tradeAndTransfer(bytes32 marketKey, int marginDelta, int sizeDelta, bytes32 trackingCode) external {
+    function tradeAndTransfer(
+        bytes32 marketKey,
+        int marginDelta,
+        int sizeDelta,
+        bytes32 trackingCode
+    ) external {
         _trade(marketKey, sizeDelta, trackingCode);
         _engineInternal().transferMargin(marketKey, msg.sender, marginDelta);
     }
--- a/contracts/PerpsOrdersV2Base.sol
+++ b/contracts/PerpsOrdersV2Base.sol
@@ -174,7 +174,7 @@ contract PerpsOrdersV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2 {
     /// helper for getting `int priceDelta` for the `engine.trade()` interface for making a trade at price different
     /// from current asset price (e.g. orders such as next price, limit, but also orders with slippage)
     function _priceDeltaFromCurrent(bytes32 marketKey, uint targetPrice) internal view returns (int) {
-        (uint currentPrice, ) = engineContract().assetPrice(marketKey);
+        (uint currentPrice,) = engineContract().assetPrice(marketKey);
         return int(targetPrice).sub(int(currentPrice));
     }
 
@@ -217,7 +217,7 @@ contract PerpsOrdersV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2 {
     function withdrawMaxMargin(bytes32 marketKey) external {
         address account = msg.sender;
         uint withdrawable = engineContract().withdrawableMargin(marketKey, account);
-        _engineInternal().transferMargin(marketKey, account, -int(withdrawable));
+        _engineInternal().transferMargin(marketKey, account, - int(withdrawable));
     }
 
     /*
@@ -252,6 +252,18 @@ contract PerpsOrdersV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2 {
         _closePosition(marketKey, trackingCode);
     }
 
+    /// Shortcut method to transfer margin and trade in a single tx.
+    function transferAndTrade(bytes32 marketKey, int marginDelta, int sizeDelta, bytes32 trackingCode) external {
+        _engineInternal().transferMargin(marketKey, msg.sender, marginDelta);
+        _trade(marketKey, sizeDelta, trackingCode);
+    }
+
+    /// Shortcut method to trade and transfer margin in a single tx (inverse of transferAndTrade).
+    function tradeAndTransfer(bytes32 marketKey, int marginDelta, int sizeDelta, bytes32 trackingCode) external {
+        _trade(marketKey, sizeDelta, trackingCode);
+        _engineInternal().transferMargin(marketKey, msg.sender, marginDelta);
+    }
+
     /* ========== INTERNAL MUTATIVE ========== */
 
     function _trade(
@@ -270,6 +282,6 @@ contract PerpsOrdersV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2 {
 
     function _closePosition(bytes32 marketKey, bytes32 trackingCode) internal {
         int size = stateContract().position(marketKey, msg.sender).size;
-        _trade(marketKey, -size, trackingCode);
+        _trade(marketKey, - size, trackingCode);
     }
 }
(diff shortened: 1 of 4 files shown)


contracts/PerpsOrdersV2Base.sol
```diff
@@ -252,6 +252,28 @@ contract PerpsOrdersV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2 {
...
     }
 
+    /// Shortcut method to transfer margin and trade in a single tx.
+    function transferAndTrade(
+        bytes32 marketKey,
+        int marginDelta,
+        int sizeDelta,
+        bytes32 trackingCode
+    ) external {
+        _engineInternal().transferMargin(marketKey, msg.sender, marginDelta);
```
artdgn, line 262: nit: another option was to make `transferMargin` public and call it instead.

contracts/PerpsOrdersV2Base.sol
```diff
@@ -252,6 +252,28 @@ contract PerpsOrdersV2Base is PerpsConfigGettersV2Mixin, IPerpsTypesV2 {
...
+        _trade(marketKey, sizeDelta, trackingCode);
+    }
+
+    /// Shortcut method to trade and transfer margin in a single tx (inverse of transferAndTrade).
+    function tradeAndTransfer(
```
artdgn, line 267: I think that `closeAndWithdraw` might be a good convenience method to add here as well, since `tradeAndTransfer` requires caller to know how much to withdraw, but `closeAndWithdraw` can just `closePosition` and then `withdrawMaxMargin` cleanly and simply closing everything.

test/contracts/PerpsOrdersV2.js
```diff
@@ -609,4 +612,189 @@ contract('PerpsOrdersV2', accounts => {
 			});
 		});
 	});
+
+	describe('trading using shortcut methods', () => {
+		describe('on transferAndTrade', () => {
+			const executeTransferAndTradeWithFee = async (margin, size) => {
+				const fee = (await perpsOrders.orderFee(marketKey, size, [baseFee, 0, toBytes32('')])).fee;
```
artdgn, line 619: that's weird, what's that third argument struct to `orderFee`?

test/contracts/PerpsOrdersV2.js
```diff
@@ -609,4 +612,189 @@ contract('PerpsOrdersV2', accounts => {
...
+				const size = toUnit('50');
+				const price = toUnit('200');
+
+				await setPrice(baseAsset, price);
+
+				const [fee] = await executeTransferAndTradeWithFee(margin, size);
+				const position = (await perpsOrders.positionSummary(marketKey, trader)).position;
+
+				assert.equal(position.id, '1');
+				assert.bnEqual(position.margin, margin.sub(fee), 'Margin does not match expected');
```
artdgn, line 637: good idea to check those views before the tx as well

test/contracts/PerpsOrdersV2.js
```diff
@@ -609,4 +612,189 @@ contract('PerpsOrdersV2', accounts => {
...
+			};
+
+			it('should succeed on valid invocation params', async () => {
+				const margin = toUnit('1000');
+				const size = toUnit('50');
+				const price = toUnit('200');
+
+				await setPrice(baseAsset, price);
+
+				const [fee] = await executeTransferAndTradeWithFee(margin, size);
```
artdgn, line 633: would be good to check the events are as expected (at least order of events) in at least one of those tests

test/contracts/PerpsOrdersV2.js
```diff
@@ -609,4 +612,189 @@ contract('PerpsOrdersV2', accounts => {
...
+				const position = (await perpsOrders.positionSummary(marketKey, trader)).position;
+
+				assert.equal(position.id, '1');
```
artdgn, line 636: nit: any particular reason this might be important?

test/contracts/PerpsOrdersV2.js
```diff
@@ -609,4 +612,189 @@ contract('PerpsOrdersV2', accounts => {
...
+			});
+
+			it('should revert when transferring with zero margin', async () => {
+				const margin = toUnit('0'); // zero margin.
+				const size = toUnit('100');
+				const price = toUnit('500');
+
+				await setPrice(baseAsset, price);
+
+				await assert.revert(
```
artdgn, line 685: revert checks should always specify the expected message (or a distinct part of it)

test/contracts/PerpsOrdersV2.js
```diff
@@ -609,4 +612,189 @@ contract('PerpsOrdersV2', accounts => {
...
+					perpsOrders.transferAndTrade(marketKey, margin, size, toBytes32('code'), {
+						from: trader,
+					})
+				);
+			});
+
+			it('should revert when size exceed max leverage on transferred margin', async () => {
```
artdgn, line 706: instead of this one I'd check with an existing position that withdrawing (negative marginDelta) and than trading reverts

test/contracts/PerpsOrdersV2.js
```diff
@@ -609,4 +612,189 @@ contract('PerpsOrdersV2', accounts => {
...
+				const price = toUnit('10');
+
+				await setPrice(baseAsset, price);
+
+				// Transfer margin and open position.
+				await perpsOrders.transferAndTrade(marketKey, margin, size, toBytes32(''), {
+					from: trader,
+				});
+
+				const withdrawableMargin = await perpsEngine.withdrawableMargin(marketKey, trader);
```
artdgn, line 734: this test can be stronger since withdrawable margin **after** the trade should be larger, I don't think what you're doing here withdraws all margin (because you call `withdrawableMargin` with an open position)

test/contracts/PerpsOrdersV2.js
```diff
@@ -609,4 +612,189 @@ contract('PerpsOrdersV2', accounts => {
...
+					withdrawableMargin.neg(),
+					size.neg(),
+					toBytes32(''),
+					{
+						from: trader,
+					}
+				);
+
+				const updatedPosition = (await perpsOrders.positionSummary(marketKey, trader)).position;
+				assert.equal(updatedPosition.size, '0');
```
artdgn, line 746: can check margin is close to 0 (or if you calculate it right with fees, exactly 0).

artdgn (changes requested):

test/contracts/PerpsOrdersV2.js
davidvuong: It just confirms the position was initialised correctly and since it's the first position, the id is 1. However, can be omitted if not necessary.

test/contracts/PerpsOrdersV2.js
```diff
@@ -630,12 +630,30 @@ contract('PerpsOrdersV2', accounts => {
 
 				await setPrice(baseAsset, price);
 
-				const [fee] = await executeTransferAndTradeWithFee(margin, size);
-				const position = (await perpsOrders.positionSummary(marketKey, trader)).position;
+				// Before executing a transfer and trade.
+				const position1 = (await perpsOrders.positionSummary(marketKey, trader)).position;
+				assert.equal(position1.id, '0');
+				assert.bnEqual(position1.margin, '0', '(before) Margin does not match expected');
```
artdgn, line 636: nit: the context messages are nice, but aren't really used much elsewhere (seem kinda cluttery as opposed to an inline comment about the assertions block)

test/contracts/PerpsOrdersV2.js
```diff
@@ -630,12 +630,30 @@ contract('PerpsOrdersV2', accounts => {
...
+				assert.bnEqual(position2.margin, margin.sub(fee), '(after) Margin does not match expected');
+				assert.bnEqual(position2.size, size, '(after) Size does not match expected');
+
+				// Verify order of events.
+				const decodedLogs = await getDecodedLogs({
+					hash: tx.tx,
+					contracts: [sUSD, perpsEngine],
+				});
+				assert.deepEqual(
+					decodedLogs.map(log => log?.name).filter(name => !!name),
```
artdgn, line 654: this filter is hiding events that also should be checked as they are part of the expected system flow (events from `perpsStorage` about position init, funding, etc)

test/contracts/PerpsOrdersV2.js
```diff
@@ -723,27 +759,48 @@ contract('PerpsOrdersV2', accounts => {
...
-				await perpsOrders.transferAndTrade(marketKey, margin, size, toBytes32(''), {
+				await perpsOrders.transferAndTrade(marketKey, margin, size, trackingCode, {
 					from: trader,
 				});
 
-				const withdrawableMargin = await perpsEngine.withdrawableMargin(marketKey, trader);
-				await perpsOrders.tradeAndTransfer(
+				// Close position (-size) and withdraw available margin (before) the final trade.
+				//
+				// There will be more margin to withdraw after the position is closed.
```
artdgn, line 773: you can use `simulateTrade` view on `perpsEngine` to get the exact expected trade results (sorry for realizing this now and not in the previous review when I was commenting on this)

artdgn (changes requested):

test/contracts/PerpsOrdersV2.js
davidvuong: I was going off the pseudo types provided in jsdoc comments. `context`, the 3rd argument of `assertBNEqual` is required but I can change that and remove the message here. I also agree this is cluttery.

test/contracts/PerpsOrdersV2.js
```diff
@@ -609,4 +612,311 @@ contract('PerpsOrdersV2', accounts => {
...
+
+				// Close position and withdraw all margin.
+				const tx = await perpsOrders.closeAndWithdraw(marketKey, toBytes32(''), {
+					from: trader,
+				});
+
+				// Position should be completely closed.
+				const position = (await perpsOrders.positionSummary(marketKey, trader)).position;
+
+				assert.bnEqual(position.margin, '0', 'Margin does not match expected');
```
artdgn, line 891: nit: some remaining context strings

davidvuong:
image

Merged.

PR #1891, opened 2022-09-02
Update Cannon to v.1.1.x

mjlescano:


dbeal-eth (changes requested):
looks like integration tests are failing. the reason for this is most likely that you just need to use the `cannon` network in the ci definition for the job.

dbeal-eth (approved):

PR #1896, opened 2022-09-09
SIP-237: Debt Migration

barrasso:


contracts/Issuer.sol
```diff
@@ -620,16 +622,25 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
...
+
     /**
      * Function used to migrate balances from the CollateralShort contract
      * @param short The address of the CollateralShort contract to be upgraded
      * @param amount The amount of sUSD collateral to be burnt
      */
     function upgradeCollateralShort(address short, uint amount) external onlyOwner {
-        require(short != address(0), "Issuer: invalid address");
         require(short == resolver.getAddress("CollateralShortLegacy"), "Issuer: wrong short address");
-        require(address(synths[sUSD]) != address(0), "Issuer: synth doesn't exist");
```
leomassazza, line 631: aren't those `require`s required?

contracts/DebtMigratorOnEthereum.sol
```diff
@@ -0,0 +1,126 @@
...
+        bytes32[] memory existingAddresses = MixinSystemSettings.resolverAddressesRequired();
+        bytes32[] memory newAddresses = new bytes32[](2);
+        newAddresses[0] = CONTRACT_EXT_MESSENGER;
+        newAddresses[1] = CONTRACT_OVM_DEBT_MIGRATOR_ON_OPTIMISM;
+        addresses = combineArrays(existingAddresses, newAddresses);
+    }
+
+    /* ========== RESTRICTED ========== */
+
+    function migrateEntireAccount(address account) external {
```
leomassazza, line 85: np this one is external, but not restricted (as the comment above suggests)

contracts/Issuer.sol
barrasso: Well technically they aren't *required* since these two simply prevent bad input by the owner: 

`require(short != address(0), "Issuer: invalid address");`
`require(amount > 0, "Issuer: cannot burn 0 synths");`

And this one is unlikely to ever be the case (when sUSD doesnt exist):
`require(address(synths[sUSD]) != address(0), "Issuer: synth doesn't exist");`
 -- if it ever were the case, it would just revert anyway later in the function.

I figured that removing these `require`s would be the most practical change to `Issuer` in order to reduce its size to fit in the new migration function. I'd definitely be interested in an alternative method for reducing size if one exists.

contracts/DebtMigratorOnEthereum.sol
barrasso: a688d4e1

leomassazza (approved):

PR #1897, opened 2022-09-09
SIP-299 Changes for V3 migration

noahlitvin:


cannonfile.toml
```diff
@@ -1,5 +1,6 @@
 name = "synthetix"
 description = "Infinite liquidity is here."
+version = "2.99.0"
```
barrasso, line 3: do we need `version` defined here? I thought we removed this before because it was leading to some weird deployment issues

contracts/Issuer.sol
```diff
@@ -714,6 +714,11 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
             uint escrowToLiquidate
         )
     {
+        // Consider triggering a migration of liquidateble accounts to V3
```
barrasso, line 717: is this supposed to be commented out?

barrasso (changes requested):

cannonfile.toml
dbeal-eth: yea I deliberately put it here now because we need a version lock on 2.99 for certain downstream v3 things to work.

we will continue using 2.99.0 for the rest of time on the cannon side, was my thought.

contracts/Issuer.sol
dbeal-eth: yes, the migration does not happen during liquidate

dbeal-eth: should probably be removed altogether tbh

dbeal-eth (approved):

barrasso (approved):

PR #1910, opened 2022-10-07
252 feedback

barrasso:


barrasso:
Already happens in BaseSynthetix.sol at the first line in the _liquidateDelinquentAccount function.

kaleb-keny (approved):

PR #1914, opened 2022-10-18
Feature/sip 279 new funding

davidvuong:
This PR is the implementation of both the premium/discount pricing function and the new funding rate velocity model described in SIP-279.

Main to-do list:

- [x] Funding rate velocity model
- [ ] Premium/discount pricing function

Smaller to-do list tasks:

- [x] Replace `skewScaleUsd` with `skewScale` (https://github.com/Synthetixio/synthetix/pull/1921)
- [x] Replace `maxMarketValueUSD` with `maxMarketValue`(https://github.com/Synthetixio/synthetix/pull/1921)
- [ ] Reason whether `fundingRateLastRecomputed` should be updated

Technical details:

- `PARAMETER_MAX_FUNDING_RATE` has been replaced with `PARAMETER_MAX_FUNDING_VELOCITY`
- `skewScaleUsd` replaced by just `skewScale` (not usd). Note that an intended byproduct of this is outside prices no longer impact the market skew
- `maxMarketValueUSD` replaced by just `maxMarketValue`
- `currentFundingRate` returns an inverted funding rate (when skewed long it returns negative but positive when skewed short). This is now changed and `currentFundingRate` no longer inverts. Funding entries in the funding sequence remain the same. The inversion happens during the `unrecordedFunding` call
- `FundingRecomputed` event now also includes the `fundingRate`
- A `fundingRateLastRecomputed` is now stored in the market state contract to track the floating funding rate
- A `currentFundingVelocity` is exposed to provide details on the current velocity
- Unlike `maxFundingRate` in V1, the `maxFundingVelocity` is capped. Although it is unlikely this will ever occur, there's code to ensure the proportional skew stays within [-1, 1] to avoid exceeding the max daily funding velocity

Resources:

- https://github.com/davidvuong/perpsv2-funding/blob/master/main.ipynb
- https://blog.synthetix.io/perps-v2-summary/
- https://sips.synthetix.io/sips/sip-279/

### DIFF
--- a/contracts/PerpsV2MarketBase.sol
+++ b/contracts/PerpsV2MarketBase.sol
@@ -143,11 +143,7 @@ contract PerpsV2MarketBase is Owned, MixinPerpsV2MarketSettings, IPerpsV2MarketB
         int pSkew = int(marketState.marketSkew()).divideDecimal(int(skewScaleBaseAsset));
 
         // Ensures the proportionalSkew is between -1 and 1.
-        //  - when pSkew = 1.1 then
-        //   min(1.1, 1) = 1
-        //  - when pSkew = -1.1 then
-        //   max(-0.02, -1) = -1
-        return pSkew > 0 ? _min(pSkew, _UNIT) : _max(pSkew, -_UNIT);
+        return _min(_max(-_UNIT, pSkew), _UNIT);
     }
 
     function _proportionalElapsed() internal view returns (int) {
--- a/contracts/PerpsV2MarketBase.sol
+++ b/contracts/PerpsV2MarketBase.sol
@@ -132,7 +132,6 @@ contract PerpsV2MarketBase is Owned, MixinPerpsV2MarketSettings, IPerpsV2MarketB
 
     /*
      * The size of the skew relative to the size of the market skew scaler.
-     * This value can be outside of [-1, 1] values.
      * Scaler used for skew is at skewScaleUSD to prevent extreme funding rates for small markets.
      *
      * TODO: Remove this. Replace skewScaleUsd with just skewScale, making `price` no longer necessary.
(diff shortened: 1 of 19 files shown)


contracts/PerpsV2MarketBase.sol
```diff
@@ -130,35 +130,86 @@ contract PerpsV2MarketBase is Owned, MixinPerpsV2MarketSettings, IPerpsV2MarketB
...
-        return int(marketState.marketSkew()).divideDecimal(int(skewScaleBaseAsset));
+
+        int pSkew = int(marketState.marketSkew()).divideDecimal(int(skewScaleBaseAsset));
+
+        // Ensures the proportionalSkew is between -1 and 1.
+        //  - when pSkew = 1.1 then
+        //   min(1.1, 1) = 1
+        //  - when pSkew = -1.1 then
+        //   max(-0.02, -1) = -1
+        return pSkew > 0 ? _min(pSkew, _UNIT) : _max(pSkew, -_UNIT);
```
leomassazza, line 150: you can get the same result with
_min(_max(-1 * _UNIT, pSkew), _UNIT)

but yours looks more clear. If size is a problem we should check the better option

test/contracts/PerpsV2Market.js
```diff
@@ -2046,7 +2063,12 @@ contract('PerpsV2Market', accounts => {
 					toUnit('1000').sub(fee),
 					toUnit('0.1')
 				);
-				assert.bnEqual((await futuresMarket.remainingMargin(trader2))[0], toUnit('5000').sub(fee2));
+
+				assert.bnClose(
```
leomassazza, line 2067: try to use the bnEqual instead bnClose (bnClose can hide a regression error later)

leomassazza (approved):
LGTM, just made a couple of comments (options). The more important is trying to avoid bnClose and use bnEqual to prevent regressions in the future

contracts/PerpsV2MarketBase.sol
davidvuong: I actually think yours is more clear. I'll make the switch.

test/contracts/PerpsV2Market.js
davidvuong: Yeah, much of the reason why `bnClose` is used is because funding rates and accrual is very sensitive to time so every calculation is always off by a very small amount. It's the same reason why you see `bnClose` is used in many other areas before these changes.

Thinking about this a bit more, this test is probably incorrect not needed. The remaining margin has definitely changed because funding has accrued. The 4 - 6 seconds between the first modifyPosition (with 50 size) and the subsequent transferMargin and modifyPosition with (-50) size has pushed the floating funding rate positive and longs will pay shorts (just a very very small amount).

PR #1920, opened 2022-10-28
Update Futures Market - SCCP-239

kaleb-keny:


barrasso (approved):

barrasso (changes requested):
need to update `maxMarketValueUSD`, not `skewScaleUSD`

barrasso (approved):

PR #1929, opened 2022-11-09
fix: npm audit critical vulnerability (override loader-utils)

drptbl:
Signed-off-by: Jakub Mucha <jakub.mucha@icloud.com>


leomassazza (approved):
Awesome!

PR #1934, opened 2022-11-15
Prepare PerpsV2 test deployment

leomassazza:


index.js
```diff
@@ -355,6 +356,29 @@ const getFeeds = ({ network, path, fs, deploymentPath, useOvm = false } = {}) =>
 	}, {});
 };
 
+const getOffchainFeeds = ({ network, path, fs, deploymentPath, useOvm = false } = {}) => {
+	let offchainFeeds;
```
davidvuong, line 360: Nit: Minor but you can probably avoid this var entirely and just return in the if/else.

publish/deployed/local-ovm/perpsv2-markets.json
```diff
@@ -17,7 +17,9 @@
 		"maxLeverage": "10",
 		"maxMarketValue": "20000000",
 		"maxFundingVelocity": "0.1",
-		"skewScale": "300000000"
+		"skewScale": "300000000",
```
davidvuong, line 20: I probably forgot to change these and they're just dummy values but do you mind lowering the `skewScale` and `maxMarketValue` something a bit smaller like `15000` and `1000` respectively? The values were previously in USD but now in native units. Same with pETH.

publish/releases.json
```diff
@@ -1420,6 +1435,25 @@
...
 			"sips": [267],
 			"released": true
+		},
+		{
+			"name": "Xxxxx",
+			"version": {
+				"major": 2,
+				"minor": 79
+			},
+			"sips": [279, 280, 281, 285],
```
davidvuong, line 1445: 🚀

publish/deployed/local-ovm/perpsv2-markets.json
leomassazza: Done (also removed data in sample mainnet-ovm/perpsv2-market.json, that should be empty and add markets as needed. Maintaining the file un-commited for testing release)

davidvuong:
Oh I forgot to mention before, I spoke with Afif and we have some realistic values BTC/ETH wrt `maxMarketValue` and `skewScale`:

```
ETH
skewScale - 1M ETH
maxMarketValue - 20k ETH

BTC
skewScale - 50k BTC
maxMarketValue - 1500 BTC
```

test/integration/dual/staking.integration.js
```diff
@@ -9,34 +9,34 @@ const { skipFeePeriod, skipMinimumStakeTime } = require('../utils/skip');
 
 const ethers = require('ethers');
 
-describe('staking & claiming integration tests (L1, L2)', () => {
+describe('staking & claiming integration tests (L1, L2)', async () => {
```
davidvuong, line 12: Oh how interesting. Is this why integration tests intermittently fail?

davidvuong (approved):
Couple of comments but LGTM

test/integration/dual/staking.integration.js
leomassazza: that was an attempt to fix it, but still not working and in fact IIRC async keyword doesn't impact describes so was a guesswork

PR #1940, opened 2022-11-21
Fix/pd calcs

davidvuong:
This PR includes:

1. Fixes the p/d-adjust price calculation to be multiplicative rather than additive `premium = price * (1 + adj)` before was `premium = price + adj`
2. Adds a `slippage` param to `closePosition` -> `closePosition(uint slippage)`
3. Fixes resolves broken `PerpsV2*` unit tests and added a few more for good measure


leomassazza (approved):
LGTM

PR #1943, opened 2022-11-23
SIP 279 (cont.) Premiums applied on Liquidation

davidvuong:
This PR introduces an additional premium to be paid by the trader when their position is liquidated. The period is _not_ paid to the keeper and given back to the stakers.

There are a few calculations that were changed, namely:

1. Liquidation price
2. Assertion in `canLiquidate`
3. Remaining margin on liquidation during liquidation

The liquidation premium is calculated as follows:

```
abs(size / skewScale) * 0.5
```

The adjustment to the liquidation price just subtracts the premium from their remaining margin:

```
int(position.lastPrice)
    .add(int(liqMargin).sub(int(position.margin).sub(int(liqPremium))).divideDecimal(positionSize))
    .sub(fundingPerUnit);
```

(2) and (3) use `_remainingLiquidatableMargin` to determine margin remaining after subtracting premium.

### DIFF
--- a/contracts/PerpsV2Market.sol
+++ b/contracts/PerpsV2Market.sol
@@ -174,8 +174,8 @@ contract PerpsV2Market is IPerpsV2Market, PerpsV2MarketProxyable {
      * Adjust the sender's position size.
      * Reverts if the resulting position is too large, outside the max leverage, or is liquidating.
      */
-    function modifyPosition(int sizeDelta, uint slippage) external {
-        _modifyPosition(sizeDelta, slippage, bytes32(0));
+    function modifyPosition(int sizeDelta, uint priceImpactDelta) external {
+        _modifyPosition(sizeDelta, priceImpactDelta, bytes32(0));
     }
 
     /*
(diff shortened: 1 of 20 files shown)


contracts/PerpsV2Market.sol
```diff
@@ -248,7 +248,7 @@ contract PerpsV2Market is IPerpsV2Market, PerpsV2MarketProxyable {
         Position memory position = marketState.positions(account);
 
         // get remaining margin for sending any leftover buffer to fee pool
-        uint remMargin = _remainingMargin(position, price);
+        uint remMargin = _remainingLiquidatableMargin(position, price);
```
davidvuong, line 251: I'm still not 100% on this. By using `_remainingLiquidatableMargin ` means the premium isn't sent to the `feePool`. If distributing the premium back to the stakers via the weekly claim is the mechanism to give back to stakers then this should just be `_remainingMargin`.

davidvuong:
CI box is super flakey but running integration tests locally renders:

```
npx hardhat test:integration:l2 --compile --deploy --provider-port 9545

  87 passing (6m)
  14 pending
``

leomassazza (approved):

PR #1948, opened 2022-11-28
Fix/improvements feedback

davidvuong:
A couple things in this PR:

1. Fixes a bug `_liquidatePosition`, ensuring we always pass the premium made on the liquidation to the fee pool
2. Updates the price impact calculation to be side aware (applying a min/max depending if you're short/long respectively)
3. No longer defaults the min/max price impact to maker/taker fees. Instead, trades _must_ be explicit with the price impact delta (on bps)
4. Finally, comments and more unit tests for coverage (more to be added later)


leomassazza (approved):
LGTM

PR #1951, opened 2022-12-02
Update goerli-ovm perpsv2 markets params

leomassazza:


davidvuong (approved):
This may break a few integration tests, particularly ones around testing max leverage but we can fix those later.

PR #1956, opened 2022-12-03
PerpsV2: Add helper interface and contract

leomassazza:
This PR introduces a consolidated interface and empty test contract that can be used as a base to connect to the proxied PerpsV2 markets. 

This can be used to connect any PerpsV2 market deployed on using the proxy address of the market address and the ABI of `TestablePerpsV2MarketEmpty` contract and use it as a simple contract.

i.e. if you want to connect to the ETH PerpsV2 market deployed in Goerli-optimism with ethers.js you can use:

```
const consolidatedPerpsV2MarketABI = [...] // the abi of the contract
const ethPerpsAddress = '0x111BAbcdd66b1B60A20152a2D3D06d36F8B5703c'

const contract = new ethers.Contract(ethPerpsAddress, consolidatedPerpsV2MarketABI, provider);
```

### DIFF
--- a/contracts/interfaces/IPerpsV2MarketConsolidated.sol
+++ b/contracts/interfaces/IPerpsV2MarketConsolidated.sol
@@ -1,0 +1,166 @@
+pragma solidity ^0.5.16;
+pragma experimental ABIEncoderV2;
+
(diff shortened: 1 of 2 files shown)


davidvuong (approved):
This looks good @leomassazza. I'm using the consolidated interface in our keepers interact with markets.

PR #1959, opened 2022-12-06
fix(chainMapping): fix typos in chain mappings

roninjin10:
Copy paste pasta broke chain mappings.  This is causing optimism bridge to fail because l1 contracts get loaded instead of l2

image


roninjin10:
Ah yes you are right.   Going to step through the code live now to debug really quick.

roninjin10:
So when I step in I'm in this minimized code.   Will look at the object next
image

roninjin10:
I see the issue now.  It expects an object ({id: 10}) not an id (10).  I think this is pretty confusing and got a way of being a bit safer here

index.js
```diff
@@ -51,7 +49,12 @@ const chainIdMapping = Object.entries({
 	return memo;
 }, {});
 
-const getNetworkFromId = ({ id }) => chainIdMapping[id];
+/** @type {(obj: {id: number} | number) => number} */
+const getNetworkFromId = (obj) => {
+	const id = typeof obj === 'number' ? obj : obj.id;
```
roninjin10, line 54: The name is pretty confusing and it fails silently if you pass in a number id (which based on the name makes sense).   The two solutions are this where I handle a number being passed in (without making a breaking change).   OR throwing an error (which is kinda a breaking change if people relied on it not throwing an error before).

roninjin10:
@bachstatter I updated pr based on more throrough debugging.  Let me know what you think

index.js
```diff
@@ -31,10 +31,8 @@ const chainIdMapping = Object.entries({
 		fork: true,
 	},
 
-	// OVM networks: see https://github.com/ethereum-optimism/regenesis/
 	10: {
 		network: 'mainnet',
```
roninjin10, line 35:
```suggestion
		network: 'mainnet',
		useOvm: true,
```

index.js
bachstatter: Yea this seems fine to me

index.js
```diff
@@ -51,7 +50,12 @@ const chainIdMapping = Object.entries({
 	return memo;
 }, {});
 
-const getNetworkFromId = ({ id }) => chainIdMapping[id];
+/** @type {(obj: {id: number} | number) => number} */
+const getNetworkFromId = (obj) => {
+	const id = typeof obj === 'number' ? obj : obj.id;
+	chainIdMapping[id];
```
bachstatter, line 56: Dont forget to return this please :)

roninjin10: lol I'm incapable of writing a single line of working code without typescript it appears
```suggestion
	return chainIdMapping[id];
```

bachstatter (approved):

PR #1970, opened 2022-12-14
Fix/audit bugs perpsv2

davidvuong:
In this PR:

1. Updates `marketSummaries` to include fundingVelocity
2. `orderFees` to adjust fee calc based on `OrderType`
3. `postTradeDetails` to accept an `OrderType` which also uses `orderFees`'s new fee calc based on `orderType`
4. `postTradeDetails` to consider the `liquidationPremium` when verifying if the trade can succeed
5. `updatePosition` considers the `liquidationPremium` to avoid allowing the user to enter into a position that can be instantly liquidated
6. Typos in comments
7. Updates fee calculation in both `DelayedOrders` and `DelayedOffchainOrders` to use their respective maker/taker fees during order submission. Previously it would _always_ look at spot (aka market) fees.


test/contracts/PerpsV2Market.js
```diff
@@ -24,19 +24,20 @@ const MockExchanger = artifacts.require('MockExchanger');
...
-	CanLiquidate: 3,
-	CannotLiquidate: 4,
-	MaxMarketSizeExceeded: 5,
-	MaxLeverageExceeded: 6,
-	InsufficientMargin: 7,
-	NotPermitted: 8,
-	NilOrder: 9,
-	NoPositionOpen: 10,
-	PriceTooVolatile: 11,
+	InvalidOrderType: 2,
```
leomassazza, line 27: I like the order by kind of error, but I don't know if is not preferable to keep the backwards compatibility.

contracts/PerpsV2MarketData.sol
```diff
@@ -187,7 +189,7 @@ contract PerpsV2MarketData {
     }
 
     function allMarketSummaries() external view returns (MarketSummary[] memory) {
-        return _marketSummaries(_futuresMarketManager().allMarkets());
+        return _marketSummaries(_futuresMarketManager().allMarkets(true));
```
leomassazza, line 192: I'd rename this function to allPerpsV2MarketSummaries or something like that, since is not returning ALL but just the proxied.

contracts/PerpsV2MarketViews.sol
```diff
@@ -278,6 +302,46 @@ contract PerpsV2MarketViews is PerpsV2MarketBase, IPerpsV2MarketViews {
         return (newPosition.margin, newPosition.size, newPosition.lastPrice, liqPrice, fee_, status_);
     }
 
+    function _simulationTradePrice(uint tradePrice) internal view returns (uint, bool) {
```
leomassazza, line 305: Like it!

contracts/PerpsV2MarketDelayedOrdersBase.sol
```diff
@@ -81,13 +81,16 @@ contract PerpsV2MarketDelayedOrdersBase is PerpsV2MarketProxyable {
         _revertIfError(status);
 
         // deduct fees from margin
-        uint commitDeposit = _delayedOrderCommitDeposit(params);
+        //
+        // commitDeposit is simply the maker/taker fee. note the dynamic fee rate is 0 since for the purposes of the
+        // commitment deposit it is not important since at the time of the order execution it will be refunded and the
+        // correct dynamic fee will be charged.
+        uint commitDeposit = _orderFee(params, 0);
```
leomassazza, line 88: see discussion in Discord

leomassazza: added here: https://github.com/Synthetixio/synthetix/pull/1972

leomassazza (approved):
LGTM, just two np on namings

davidvuong:
> Failing after 1s — 93.33% of diff hit (target 95.58%)

Patch failing on a param rename. Not relevant to the changes in this PR.

PR #1980, opened 2023-01-03
Set unused params in off-chain delayed orders to zero

leomassazza:
This PR sets the `targetRoundId` and `executableAtTime` to zero on off-chain delayed orders since those params are not used for those kind of orders to prevent wrong interpretations.


leomassazza:
Added to David's PR

PR #1981, opened 2023-01-03
Fix/perps v2 calcs

davidvuong:
This PR includes:
- Fixes on liqPremium calculation 
- Changes the .5 factor on liquidationPremium calculation to an SCCP parameter `liquidationPremiumMultiplier`)
- Add a cap on the keeper fee for liquidations (new SCCP parameter `maxKeeperFee`)
- Remove (set to zero) unused params from submit delayed order events when using off-chain mode
- Some fixes on deploy/config scripts


contracts/PerpsV2MarketBase.sol
```diff
@@ -545,6 +548,8 @@ contract PerpsV2MarketBase is Owned, MixinPerpsV2MarketSettings, IPerpsV2MarketB
         //
         // Liquidation margin is considered without a fee (but including premium), because it wouldn't make sense to allow
         // a trade that will make the position liquidatable.
+        //
+        // TODO: params.price is the fillPrice. `liquidationPremium` requires the oracle price.
```
davidvuong, line 552: Needs to be resolved first before merging.

leomassazza: does it affect also to _canLiquidate() that is called with params.price some lines above (line 501 at this time)?

contracts/PerpsV2MarketDelayedOrdersBase.sol
```diff
@@ -100,8 +100,8 @@ contract PerpsV2MarketDelayedOrdersBase is PerpsV2MarketProxyable {
                 priceImpactDelta: uint128(priceImpactDelta),
                 targetRoundId: uint128(targetRoundId),
                 commitDeposit: uint128(commitDeposit),
-                keeperDeposit: uint128(keeperDeposit),
-                executableAtTime: block.timestamp + desiredTimeDelta,
+                keeperDeposit: uint128(keeperDeposit), // offchain orders do _not_ have an executableAtTime as it's based on price age.
+                executableAtTime: isOffchain ? 0 : block.timestamp + desiredTimeDelta, // zero out - not used and minimise confusion.
```
leomassazza, line 104: targetRoundId also should be set to zero if isOffchain since is not used there either (I'll add a PR to this to fix that and tests that uses it)

test/contracts/PerpsV2Market.js
```diff
@@ -1777,7 +1777,7 @@ contract('PerpsV2Market PerpsV2MarketAtomic', accounts => {
 			assert.bnEqual(postDetails.margin, minInitialMargin.sub(fee));
 			assert.bnEqual(postDetails.size, size);
 			assert.bnEqual(postDetails.price, fillPrice);
-			assert.bnEqual(postDetails.liqPrice, toUnit('2.0555527525'));
+			assert.bnEqual(postDetails.liqPrice, toUnit('2.056002775'));
```
leomassazza, line 1780: can we add a calculation to get this number so that the formula can be updated if the calculation method changes? (if is not overkilling)

test/contracts/PerpsV2Market.js
```diff
@@ -3166,15 +3166,15 @@ contract('PerpsV2Market PerpsV2MarketAtomic', accounts => {
 					account: trader2,
 					fastForwardBy: 29000,
 					expectedRate: toUnit('0.00839120'),
-					expectedFunding: toUnit('-0.001408'), // neg because longs pay short
+					expectedFunding: toUnit('-0.14086'), // neg because longs pay short
```
leomassazza, line 3169: same here if not overkilling. Not a np, but maybe the effort is too much.

contracts/PerpsV2MarketDelayedOrdersBase.sol
leomassazza: https://github.com/Synthetixio/synthetix/pull/1982 added it here, if you agree zeroing this one, you can merge it to yours.

contracts/PerpsV2MarketBase.sol
davidvuong: Good catch.

test/contracts/PerpsV2Market.js
davidvuong: Yeah, we can. I wrote a `getExpectedLiquidationPrice` function but it's nestled inside the liquidation price unit tests. I've just hardcoded the numbers here because we actually test liqPrice in much more detail later.

test/contracts/setup.js
```diff
@@ -57,6 +57,7 @@ const constantsOverrides = {
 	EXCHANGE_DYNAMIC_FEE_THRESHOLD: toWei('0.004'),
 	EXCHANGE_MAX_DYNAMIC_FEE: toWei('0.05'),
 	FUTURES_MIN_KEEPER_FEE: toWei('20'),
+	FUTURES_MAX_KEEPER_FEE: toWei('1000'),
```
barrasso, line 60: same thing here about maxKeeperFee

barrasso (approved):
LGTM. just one minor comment

index.js
```diff
@@ -206,6 +206,7 @@ const defaults = {
 	ETHER_WRAPPER_BURN_FEE_RATE: '0',
 
 	FUTURES_MIN_KEEPER_FEE: w3utils.toWei('1'), // 1 sUSD min keeper fee
+	FUTURES_MAX_KEEPER_FEE: w3utils.toWei('1000'), // 1000 sUSD min keeper fee
```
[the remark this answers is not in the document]
leomassazza: yeah, the typo was in the SIP. It's fixed there.

PR #1985, opened 2023-01-04
Add Cap to keeper liquidation fee

leomassazza:
This PR introduces a new parameter `maxKeeperFee` and use it to cap keeper liquidation fee

### DIFF
--- a/contracts/MixinPerpsV2MarketSettings.sol
+++ b/contracts/MixinPerpsV2MarketSettings.sol
@@ -38,6 +38,8 @@ contract MixinPerpsV2MarketSettings is MixinResolver {
     // Global settings
     // minimum liquidation fee payable to liquidator
     bytes32 internal constant SETTING_MIN_KEEPER_FEE = "futuresMinKeeperFee";
+    // maximum liquidation fee payable to liquidator
+    bytes32 internal constant SETTING_MAX_KEEPER_FEE = "futuresMaxKeeperFee";
     // liquidation fee basis points payed to liquidator
     bytes32 internal constant SETTING_LIQUIDATION_FEE_RATIO = "futuresLiquidationFeeRatio";
     // liquidation buffer to prevent negative margin upon liquidation
@@ -151,6 +153,10 @@ contract MixinPerpsV2MarketSettings is MixinResolver {
 
     function _minKeeperFee() internal view returns (uint) {
         return _flexibleStorage().getUIntValue(SETTING_CONTRACT_NAME, SETTING_MIN_KEEPER_FEE);
+    }
+
+    function _maxKeeperFee() internal view returns (uint) {
+        return _flexibleStorage().getUIntValue(SETTING_CONTRACT_NAME, SETTING_MAX_KEEPER_FEE);
     }
 
     function _liquidationFeeRatio() internal view returns (uint) {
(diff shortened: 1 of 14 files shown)


contracts/MixinPerpsV2MarketSettings.sol
```diff
@@ -38,6 +38,8 @@ contract MixinPerpsV2MarketSettings is MixinResolver {
     // Global settings
     // minimum liquidation fee payable to liquidator
     bytes32 internal constant SETTING_MIN_KEEPER_FEE = "futuresMinKeeperFee";
+    // maximum liquidation fee payable to liquidator
+    bytes32 internal constant SETTING_MAX_KEEPER_FEE = "futuresMaxKeeperFee";
```
davidvuong, line 42: Dang, we forgot to rename `futures` to `perps` here. How difficult would it makes deploys if we were to rename both to be `perps{Min,Max}KeeperFee`?

contracts/PerpsV2MarketBase.sol
```diff
@@ -359,18 +359,21 @@ contract PerpsV2MarketBase is Owned, MixinPerpsV2MarketSettings, IPerpsV2MarketB
...
+     * liquidations of small positions, or overpaying.
      * @param positionSize size of position in fixed point decimal baseAsset units
      * @param price price of single baseAsset unit in sUSD fixed point decimal units
      * @return lFee liquidation fee to be paid to liquidator in sUSD fixed point decimal units
      */
     function _liquidationFee(int positionSize, uint price) internal view returns (uint lFee) {
         // size * price * fee-ratio
         uint proportionalFee = _abs(positionSize).multiplyDecimal(price).multiplyDecimal(_liquidationFeeRatio());
+        uint maxFee = _maxKeeperFee();
+        uint topedProportionalFee = proportionalFee > maxFee ? maxFee : proportionalFee;
```
davidvuong, line 372: Do you mean `topped` here instead of `toped`?

test/contracts/PerpsV2Market.js
```diff
@@ -5108,6 +5109,42 @@ contract('PerpsV2Market PerpsV2MarketAtomic', accounts => {
 				await perpsV2MarketSettings.setLiquidationFeeRatio(toUnit(0.02), { from: owner });
 				assert.bnEqual(await perpsV2Market.liquidationFee(trader2), toUnit(60));
 			});
+
+			it('top up liquidation fee', async () => {
```
davidvuong, line 5113: nit: Maybe rename this test to be 'max liquidation fee' for clarity but test looks good 👍

davidvuong (approved):
LGTM. Just a typo and some questions.

davidvuong:
Minor: Could you also write a test that checks if the fee pool grows by `remainingMargin - liqFees`? Unless I'm mistaken, looking at the `liquidatePosition`, only the `order.commitFee` is sent to the fee pool

contracts/MixinPerpsV2MarketSettings.sol
leomassazza: not that difficult, but we need to be sure we set all params to the right value when deploying the new contract

davidvuong: Cool. Might be worth doing this after this goes out.

davidvuong (approved):

PR #1990, opened 2023-01-06
SIP-255: Burn fees

barrasso:
> This SIP proposes burning fees collected by the fee pool on instead of allowing them to claim the fees and have it transferred to them. This will have the effect of reducing user's debt overall, and in theory they will be able to mint more at a later date.

See https://sips.synthetix.io/sips/sip-255/


contracts/FeePool.sol
```diff
@@ -570,6 +547,34 @@ contract FeePool is Owned, Proxyable, LimitedSetup, MixinSystemSettings, IFeePoo
...
+     */
+    function feesBurned(address account) public view returns (uint) {
+        // Add up the fees
+        uint[2][FEE_PERIOD_LENGTH] memory userFees = feesByPeriod(account);
+
+        uint totalFees = 0;
+
+        // Fees & Rewards in fee period [0] are not yet burned
+        for (uint i = 1; i < FEE_PERIOD_LENGTH; i++) {
+            totalFees = totalFees.add(userFees[i][0]);
```
leomassazza, line 561: Is this right? userFees is declared as `uint[2][N]`
but here is used as `uint[i][0]` (as if it were an `uint[N][2]`)

dbeal-eth (approved):

PR #1996, opened 2023-01-20
Feature/submit exec close order

davidvuong:
This PR sits on top of both `perps-v2-calcs` and `perpsv2-refactor-close`. In short, we allow traders to submit close position orders even if they are above maxLeverage. Currently, they must transfer more margin due to pnl, funding, and fees affecting remaining margin causing a revert.

This introduces two key components:

1. Adds `submitCloseDelayedOrderWithTracking` and `submitCloseOffchainDelayedOrderWithTracking`
2. Amends `_updatePositionMargin` to check maxLeverage _iff_ not closing a position

Tasks:

- [x] Implement `closeXXX` functions
- [x] Changes in `updatePositionMargin`
- [x] Add more unit tests


test/contracts/PerpsV2Market.delayedOrder.js
```diff
@@ -184,6 +184,51 @@ contract('PerpsV2Market PerpsV2MarketDelayedOrders', accounts => {
...
+
+			// (3) Fast forward time to allow this to be executable then execute.
+			await fastForward(minDelayTimeDelta + 1); // ff min + 1s buffer.
+			await perpsV2Market.executeDelayedOrder(trader, { from: trader });
+			const position = await perpsV2Market.positions(trader);
+
+			// (4) Price moves in the opposite direction -0.5% (50bps)
+			await setPrice(baseAsset, multiplyDecimal(price, toUnit('0.995')));
+
+			// (5) Attempt to close the position - expected to revert with not enough margin.
```
leomassazza, line 217: The comment is missleading since when this code goes to the branch, the issue is solved (and is not expected to revert)

davidvuong: Oops. Yes. The comments are a bit off.

davidvuong:
> LGTM. There's a np in a test comment

Nice. So I think this will work for any submitXXXDelayedOrder but I'll also update this to update the `closeDelayedOrder` stub but also to add more tests.

PR #2021, opened 2023-02-14
Remove commitment fee charged during delayed order submissions

davidvuong:
This PR implements the SIP defined here: https://github.com/Synthetixio/SIPs/pull/1452 (https://sips.synthetix.io/sips/sip-2004/)

In summary, it removes the commitment fee charged upon submission and removes the ability to cancel an order during the confirmation window. This change also updates a variety of tests to reflect the new change and removes the overrideCommitFee SCCP param as it's now redundant.

### DIFF
--- a/contracts/MixinPerpsV2MarketSettings.sol
+++ b/contracts/MixinPerpsV2MarketSettings.sol
@@ -85,6 +85,7 @@ contract MixinPerpsV2MarketSettings is MixinResolver {
         return _parameter(_marketKey, PARAMETER_MAKER_FEE);
     }
 
+    // TODO: Remove me
     function _overrideCommitFee(bytes32 _marketKey) internal view returns (uint) {
         return _parameter(_marketKey, PARAMETER_OVERRIDE_COMMIT_FEE);
     }
--- a/contracts/MixinPerpsV2MarketSettings.sol
+++ b/contracts/MixinPerpsV2MarketSettings.sol
@@ -142,10 +142,10 @@ contract MixinPerpsV2MarketSettings is MixinResolver {
 
     function _offchainMarketKey(bytes32 _marketKey) internal view returns (bytes32) {
         return
-        _flexibleStorage().getBytes32Value(
-            SETTING_CONTRACT_NAME,
-            keccak256(abi.encodePacked(_marketKey, PARAMETER_OFFCHAIN_MARKET_KEY))
-        );
+            _flexibleStorage().getBytes32Value(
+                SETTING_CONTRACT_NAME,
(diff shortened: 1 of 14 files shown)


test/contracts/PerpsV2Market.offchainDelayedOrder.js
```diff
@@ -726,13 +716,12 @@ contract('PerpsV2Market PerpsV2MarketOffchainOrders', accounts => {
 		});
 
 		describe('when an order exists', () => {
-			let commitFee, keeperFee, updateFeedData;
+			let keeperFee, updateFeedData;
 
 			beforeEach(async () => {
 				// keeperFee is the minimum keeperFee for the system
 				keeperFee = await perpsV2MarketSettings.minKeeperFee();
 				// commitFee is the fee that would be charged for a trade when order is submitted
```
leomassazza, line 724: remove also the orphan comment

contracts/PerpsV2MarketDelayedExecution.sol
```diff
@@ -265,9 +260,6 @@ contract PerpsV2MarketDelayedExecution is IPerpsV2MarketDelayedExecution, PerpsV
             _manager().issueSUSD(messageSender, order.keeperDeposit);
         }
```
leomassazza, line 262: I'd either 
- keep it to and add a comment as in line 278 saying commitDeposit should be zero, or
- set to zero the initialized toRefund in line 279 (and maybe assert that commitDeposit _is_ zero)

I'm inclined to 2nd option, but the only risk is ongoing txs when we do the upgrade since some txs can have a value different to zero and will get unaccounted.

contracts/MixinPerpsV2MarketSettings.sol
```diff
@@ -85,6 +85,7 @@ contract MixinPerpsV2MarketSettings is MixinResolver {
         return _parameter(_marketKey, PARAMETER_MAKER_FEE);
     }
 
+    // TODO: Remove me
```
leomassazza, line 88: let's do it :)

contracts/PerpsV2MarketDelayedExecution.sol
davidvuong: Good catch... 🤔 

Another option is to set it to 0 (like what I've done and what you've suggested) but rather than asserting (and presumably reverting?), just a simple if check if `commitDeposit` is actually zero. If not, we call `payFee`. Just thinking through what might happen:

1. During the upgrade: There may be a few orders pending execution but is cancelled for whatever reason. These orders will have a `commitFee`, which should be paid to the debt pool. These will be refunded.
2. Post upgrade and no additional orders with commitFee: the `if` branch is never executed as submissions no longer have a commit fee.

leomassazza (approved):
LGTM

PR #2032, opened 2023-02-27
Feature/market level liq buffer

davidvuong:
This PR adjusts the `liquidationBufferRatio` to be configurable per market rather than globally (i.e. across all markets). This change is part of SIP-2004 (amended https://github.com/Synthetixio/SIPs/pull/1480).


contracts/PerpsV2MarketState.sol
```diff
@@ -224,7 +224,7 @@ contract PerpsV2MarketState is Owned, StateShared, IPerpsV2MarketBaseTypes {
         address account,
         bool isOffchain,
         int128 sizeDelta,
-        uint128 priceImpactDelta,
+        uint128 desiredFillPrice,
```
leomassazza, line 227: Mmmm... that's not going to be that easy. 
The current state contract is immutable, this change will be picked in the migration (I'll pick this change to add there after merging). 
But there's still one issue: the legacy orders will have the value set as impact delta instead of desired price.

leomassazza (approved):
LGTM. Just need to add a couple of TODOs on deployment and migration (post-PRs)

PR #2033, opened 2023-03-02
Feature/desired fill price

davidvuong:
Implements `desiredPrice` changes defined in SIP-2004 (https://sips.synthetix.io/sips/sip-2004/)


test/contracts/PerpsV2Market.js
```diff
@@ -39,7 +39,7 @@ const Status = {
 	PriceTooVolatile: 12,
 };
 
-contract('PerpsV2Market PerpsV2MarketAtomic', accounts => {
+contract.only('PerpsV2Market PerpsV2MarketAtomic', accounts => {
```
leomassazza, line 42: rogue .only

test/contracts/PerpsV2Market.js
```diff
@@ -789,7 +814,7 @@ contract('PerpsV2Market PerpsV2MarketAtomic', accounts => {
 					await transferMarginAndModifyPosition({
 						market: perpsV2Market,
 						account: trader2,
-						fillPrice: price,
+						price: price,
```
leomassazza, line 817: if variable name is same you can ommit the assignation i.e. 
`price,`
but super np.

leomassazza (approved):
LGTM. Just one np (no need to update now) and one rogue `.only` might be hiding other errors :(

test/contracts/PerpsV2Market.js
kaleb-keny: Hey i don't know if it's possible, but keeping it as fill price would be preferable. I mean i often get confused going through the code on whether it's fillPrice or oraclePrice. this helps a bit

davidvuong: Yeah no prob I can revert the name here back to `fillPrice`.

PR #2044, opened 2023-03-10
Correct perpsV2 Interfaces

leomassazza:


contracts/interfaces/IPerpsV2MarketConsolidated.sol
```diff
@@ -75,6 +75,8 @@ interface IPerpsV2MarketConsolidated {
 
     function assetPrice() external view returns (uint price, bool invalid);
 
+    function fillPrice(int sizeDelta) external view returns (uint price, bool invalid);
```
davidvuong, line 78: Oh wow I'm surprised we didn't include this earlier. Good catch.

davidvuong (approved):
LGTM 🔥

PR #2046, opened 2023-03-12
[DRAFT] Split up migrated escrow amount over one year

barrasso:
RE: [SIP-237: Debt Migration](https://sips.synthetix.io/sips/sip-237/) 

For the migrated escrow amount, create twelve distinct escrow entries on L2 that vest over a year instead of packing it all into a single entry.


leomassazza (approved):
LGTM

PR #2055, opened 2023-03-25
Audit Fixes

leomassazza:


contracts/PerpsV2Market.sol
```diff
@@ -39,34 +39,50 @@ import "./interfaces/IPerpsV2Market.sol";
  *                                  the base asset, these settings determine the behaviour of each market.
  *                                  See that contract for descriptions of the meanings of each setting.
  *
+ * Proxy/Implementations/State Contracts architecture diagram
+ *
+ *                 ┌-> PerpsV2Market                 ─┐
+ *                 ├-> PerpsV2MarketViews            -┤
+ *   ProxyPerpsV2 ─┼-> PerpsV2MarketLiquidate        ─┼─> PerpsV2MarketState -> PerpsV2MarketStateLegacyR1
```
davidvuong, line 46: ❤️ I love this diagram

contracts/FuturesMarketManager.sol
```diff
@@ -441,10 +441,8 @@ contract FuturesMarketManager is Owned, MixinResolver, IFuturesMarketManager {
      */
     function addEndorsedAddresses(address[] calldata addresses) external onlyOwner {
         for (uint i = 0; i < addresses.length; i++) {
-            if (!_endorsedAddresses.contains(addresses[i])) {
-                _endorsedAddresses.add(addresses[i]);
-                emit EndorsedAddressAdded(addresses[i]);
-            }
+            _endorsedAddresses.add(addresses[i]);
```
davidvuong, line 444: Oh good catch. So every call just wipes it from index 0. I guess the problem here is if `addresses` is shorter than `_endorsedAddresses`, we can end up with lingering prior addresses.

contracts/PerpsV2MarketDelayedExecution.sol
```diff
@@ -308,19 +308,21 @@ contract PerpsV2MarketDelayedExecution is IPerpsV2MarketDelayedExecution, PerpsV
...
-            account,
-            position.margin,
-            position.size,
-            0,
-            fillPrice,
-            fundingIndex,
-            0,
-            marketState.marketSkew()
-        );
+        if (toRefund > 0) {
```
davidvuong, line 311: Strange I thought I had added this before but I could be mistaken 🤔

davidvuong (approved):
LGTM 👍

contracts/FuturesMarketManager.sol
leomassazza: is not a replacement, it adds all the addresses in the input array

contracts/PerpsV2MarketDelayedExecution.sol
leomassazza: me too... maybe a merge conflict?

PR #2066, opened 2023-04-17
Fix PerpsV2MarketData for legacy markets

leomassazza:


davidvuong (approved):
Looks good. However, two things:

- ~~A test which has both legacy and new markets would solidify whether this fixes the issue or not. I think it does for `currentFundingVelocity` but maybe `FeeRates` might reveal itself as another error.~~
- Given that `marketSummaries` accepts any markets, passing in a legacy `marketKey` into `_parameters` may cause an error when `{maker,taker}FeeOffChainDelayedOrder` is invoked so you might also need to check for `isLegacy` within `FeeRates`.

Edit: On (2), it might be fine since `_parameters` should just give default int 0 values if it can't find anything for that `marketKey`.

PR #2068, opened 2023-04-19
Fix flagging 

leomassazza:
Allow to flag when pending order and not enough margin


bachstatter (approved):
David asked me to have a look to.. Make sense to me!

davidvuong (approved):

PR #2078, opened 2023-05-02
SCCP-2002

kaleb-keny:
As per [SCCP-2002](https://sips.synthetix.io/sccp/sccp-2002) all max levergage numbers are bumped by 10% on crypto perps


leomassazza (approved):
LGTM

PR #2097, opened 2023-06-14
Merak-release-v2.90.0-alpha

leomassazza:


publish/src/command-utils/perps-v2-utils.js
```diff
@@ -85,7 +85,7 @@ const filteredLists = (originalList, newList) => {
 };
 
 const isNewMarket = ({ existingMarkets, marketKey }) =>
-	existingMarkets.includes(toBytes32(marketKey));
+	!existingMarkets.includes(toBytes32(marketKey));
```
davidvuong, line 88: 😱

leomassazza: that was a fix in the previous release that was already in develop but didn't get to staging since latest releases were only mainnet-ovm and not goerli-ovm :(

davidvuong (approved):

PR #2160, opened 2023-10-13
SIP-2036 fix reward distributor to bubble failures

dbeal-eth:
in the simplest possible way, fix a bug which prevents reward distribution from getting sent to L2 due to internal OOG gas logic in optimism contract (normally if you are OOG then the remaining code in RewardsDistribution would also fail, but optimism deliberately consumes an excessive amount of gas in a way that causes only the internal call to fail)

in accordance with [SIP-2036](https://sips.synthetix.io/sips/sip-2036/)

note: if you are wondering why I didnt just call the rewards distribution receiver contract normally and continue to be using hte low level call, the reason is im trying to keep changes to a minimal and this change basically ended up being a 1 line changes.

### DIFF
--- a/contracts/RewardsDistribution.sol
+++ b/contracts/RewardsDistribution.sol
@@ -166,10 +166,11 @@ contract RewardsDistribution is Owned, IRewardsDistribution {
                 bytes memory payload = abi.encodeWithSignature("notifyRewardAmount(uint256)", distributions[i].amount);
 
                 // solhint-disable avoid-low-level-calls
-                (bool success, bytes result) = distributions[i].destination.call(payload);
+                (bool success, bytes memory result) = distributions[i].destination.call(payload);
 
                 if (!success) {
                     // if the error was emitted by the destination contract, bubble
+                    uint len = result.length;
                     assembly {
                         revert(add(result, 0x20), len)
                     }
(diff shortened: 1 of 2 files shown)


contracts/RewardsDistribution.sol
```diff
@@ -166,10 +166,14 @@ contract RewardsDistribution is Owned, IRewardsDistribution {
                 bytes memory payload = abi.encodeWithSignature("notifyRewardAmount(uint256)", distributions[i].amount);
 
                 // solhint-disable avoid-low-level-calls
-                (bool success, ) = distributions[i].destination.call(payload);
+                (bool success, bytes memory result) = distributions[i].destination.call(payload);
 
                 if (!success) {
-                    // Note: we're ignoring the return value as it will fail for contracts that do not implement RewardsDistributionRecipient.sol
+                    // if the error was emitted by the destination contract, bubble
```
leomassazza, line 172: as discussed in a chat. 
This will require any future recipient of rewards distribution to be a contract and implement the `notifyRewardAmount(uint256)` function in order to not revert the whole distribution.
Is not an issue for this solution, but a requirement for future recipients.

leomassazza (approved):
LGTM

PR #2216, opened 2024-03-07
Aljanah release v2.100 to mainnet

barrasso:


kaleb-keny (approved):
code readability can be improved a bit by not having multiple variables pointing to the same contract, also having them old contracts, in the old section, new contracts in the new section. But overall the logic lgtm.

Issue #2214, opened 2024-03-05
Ops tool dependency failure


barrasso:
The `ops` tool fails to build due to a deprecated node version:

<img width="1120" alt="image" src="https://github.com/Synthetixio/synthetix/assets/4982406/19338294-5e8e-4fb6-9660-2fb86b0e2033">

PR #2222, opened 2024-03-18
support for latest version of cannon

dbeal-eth:


barrasso (approved):

dbeal-eth:
@barrasso are the CI tests supposed to succeed here? seems to be the same failures as develop.

barrasso:
> @barrasso are the CI tests supposed to succeed here? seems to be the same failures as develop.

Yes the integration job failing is expected at the moment: https://github.com/Synthetixio/synthetix/issues/2214

PR #2227, opened 2024-03-26
SIP-2059: Legacy Spot Synth Migration

barrasso:
Introduces a new contract, `DynamicSynthRedeemer`, which quotes synth redemption prices based on real-time chainlink price data * a discount rate (as opposed to the `SynthRedeemer`, which does so by removing the synth from the issuer and redeems at a fixed rate).

https://sips.synthetix.io/sips/sip-2059/


contracts/DynamicSynthRedeemer.sol
```diff
@@ -0,0 +1,114 @@
...
+
+    function redeemPartial(IERC20 synthProxy, uint amountOfSynth) external {
+        // technically this check isn't necessary - Synth.burn would fail due to safe sub,
+        // but this is a useful error message to the user
+        require(synthProxy.balanceOf(msg.sender) >= amountOfSynth, "Insufficient balance");
+        _redeem(synthProxy, amountOfSynth);
+    }
+
+    function _redeem(IERC20 synthProxy, uint amountOfSynth) internal {
+        uint rateToRedeem = redemptions[address(synthProxy)];
```
kaleb-keny, line 82: This needs to be updated to using `exchanger.rateForCurrency` ?, instead of a redemption rate

contracts/DynamicSynthRedeemer.sol
```diff
@@ -0,0 +1,114 @@
...
+        require(rateToRedeem > 0, "Synth not redeemable");
+        require(amountOfSynth > 0, "No balance of synth to redeem");
+        issuer().burnForRedemption(address(synthProxy), msg.sender, amountOfSynth);
+        uint amountInsUSD = amountOfSynth.multiplyDecimal(rateToRedeem);
+        sUSD().transfer(msg.sender, amountInsUSD);
+        emit SynthRedeemed(address(synthProxy), msg.sender, amountOfSynth, amountInsUSD);
+    }
+
+    function deprecate(IERC20 synthProxy, uint rateToRedeem) external onlyIssuer {
```
kaleb-keny, line 91: Yeah i think we need to kill this? Supply will not change, right?

barrasso: removing the `deprecate` function as it is not needed for the DynamicSynthRedeemer. It will stay in the regular synth redeemer

barrasso: we do, however, need to add a method to `Issuer` that is similar to `_removeSynth`, but specific to the dynamic redeemer instead

contracts/DynamicSynthRedeemer.sol
barrasso: https://github.com/Synthetixio/synthetix/pull/2227/commits/a0ffdf891c8ec8584e181a3b548292a7271fdb1e

contracts/DynamicSynthRedeemer.sol
barrasso: https://github.com/Synthetixio/synthetix/pull/2227/commits/a0ffdf891c8ec8584e181a3b548292a7271fdb1e

contracts/DynamicSynthRedeemer.sol
kaleb-keny: ohhh ok, must have misunderstood that it was ready for review or not

contracts/DynamicSynthRedeemer.sol
```diff
@@ -0,0 +1,91 @@
...
+
+    function redeemPartial(address synthProxy, uint amountOfSynth) external {
+        // technically this check isn't necessary - Synth.burn would fail due to safe sub,
+        // but this is a useful error message to the user
+        require(IERC20(synthProxy).balanceOf(msg.sender) >= amountOfSynth, "Insufficient balance");
+        _redeem(synthProxy, amountOfSynth);
+    }
+
+    function _redeem(address synthProxy, uint amountOfSynth) internal {
+        bytes32 currencyKey = ISynth(IProxy(synthProxy).target()).currencyKey();
```
kaleb-keny, line 78: Do you think it's worthwhile adding a allowForRedemption map, set by governance, in order to protect against situations of things being allowed for redemption, while that not being the intention.... Like SNX

contracts/DynamicSynthRedeemer.sol
```diff
@@ -0,0 +1,91 @@
...
+// Internal references
+import "./interfaces/IIssuer.sol";
+import "./interfaces/IExchangeRates.sol";
+import "./interfaces/ISynth.sol";
+
+interface IProxy {
+    function target() external view returns (address);
+}
+
+contract DynamicSynthRedeemer is Owned, IDynamicSynthRedeemer, MixinResolver {
```
kaleb-keny, line 20: Will disabling exchanges / atomic exchanges on all synths be done in this PR?

contracts/BaseDebtCache.sol
```diff
@@ -144,7 +151,8 @@ contract BaseDebtCache is Owned, MixinSystemSettings, IDebtCache {
             address synthAddress = address(synths[i]);
             require(synthAddress != address(0), "Synth does not exist");
             uint supply = IERC20(synthAddress).totalSupply();
-            values[i] = supply.multiplyDecimalRound(rates[i]);
+            uint value = supply.multiplyDecimalRound(rates[i]);
+            values[i] = value.multiplyDecimalRound(dynamicSynthRedeemer().getDiscountRate());
```
kaleb-keny, line 155: I believe sUSD needs to always have a `discountRate` of 1. The `discountRate` is only applied on synths that are open for dynamic redemption... I think another reason why we may need `allowedForRedemption` list. In addition, the old redemption tokens `redeemer`, are hopefully not affected?

contracts/DynamicSynthRedeemer.sol
barrasso: Don't need a code change for that. It can be done with an owner action: `SystemStatus.suspendSynthsExchange[]`

contracts/DynamicSynthRedeemer.sol
barrasso: That is one way to do it. Another way would be to simply require the `currencyKey` to not be `sUSD`. This way all synths except for `sUSD` can be redeemed.

Also, since `SNX` is not an `ISynth`, this function will revert for any non-synth tokens.

barrasso: Can also add a `setActive` flag to enable/disable dynamic redemptions.

contracts/BaseDebtCache.sol
barrasso: Instead of maintaining a list, we can also add an exception here for `sUSD` only.

contracts/DynamicSynthRedeemer.sol
kaleb-keny: works for killing atomic exchanges I assume?

contracts/BaseDebtCache.sol
kaleb-keny: sure thing 🙏🏼  and for old tokens like sLINK, i am assuming they are no used here, because we removed from the exchangeRates and being synths, right?

contracts/DynamicSynthRedeemer.sol
kaleb-keny: makes sense 🙏🏼

contracts/DynamicSynthRedeemer.sol
kaleb-keny: How do synths that are pending fee reclamation go into play here? Should we run a script that eliminates all settles all pending fee reclamations before executing this or should we tackle this in a PR, to kill pending fee reclamations?

barrasso: Yessir. `exchangeAtomically` also requires `exchangeActive` to be true

barrasso: > How do synths that are pending fee reclamation go into play here? Should we run a script that eliminates all settles all pending fee reclamations before executing this or should we tackle this in a PR, to kill pending fee reclamations?

This can be done as part of the release:

* 1 - suspend exchanges
* 2- `settle()` all pending fee reclamations (set waiting period to 0 if needed)
* 3 - deploy dynamic redeemer
* 4 - optional: for redeemer, setActive = true

contracts/BaseDebtCache.sol
barrasso: > sure thing 🙏🏼 and for old tokens like sLINK, i am assuming they are not captured here (rather the sUSD in redeemer is used), because we removed from the exchangeRates and being considered as synths, right?

Correct. Old synths like `sLINK` have been removed from the Issuer and are no longer considered to be synths in the system. They are only available for redemption through the legacy `SynthRedeemer`.

contracts/DynamicSynthRedeemer.sol
barrasso: Addressed here https://github.com/Synthetixio/synthetix/pull/2227/commits/5b0ec9935702fb37eb923881575f85a53f861361

contracts/BaseDebtCache.sol
barrasso: > I believe sUSD needs to always have a discountRate of 1

Addressed here https://github.com/Synthetixio/synthetix/pull/2227/commits/0e4eff7360724b7f84947fea0980340a7e368746

kaleb-keny (approved):

PR #2251, opened 2024-11-07
ability for cachedDebt to become negative

dbeal-eth:
to avoid subtraction overflow


dbeal-eth:
Decided not to continue with this

PR #2260, opened 2025-02-10
minor change to allow v3 to mint sUSD backed by v3

dbeal-eth:
also update cannon

do not merge until audit complete


contracts/Issuer.sol
```diff
@@ -343,6 +344,9 @@ contract Issuer is Owned, MixinSystemSettings, IIssuer {
     }
 
     function _maxIssuableSynths(address _issuer) internal view returns (uint, bool) {
+        if (_issuer == resolver.getAddress(CONTRACT_V3_LEGACYMARKET)) {
+            return (uint(-1), false);
```
noisekit, line 348: what is the story behind `uint(-1)`? It does not make much sense on the first look.

dbeal-eth: in these older versions of solidity, there was no `type(uint256).max`, so this is the closest equivalent.

noisekit: Thanks. If you plan to do any changes to the pr code - would be great to add this as a comment to the code.

noisekit (approved):

noisekit (commented):
I suspect that `publish/deployed/local/deployment.json` should not be committed

noisekit (approved):

barrasso (approved):

